PR #4, opened 2018-03-29
Interpreter

TristonianJones:
Initial implementation of a CEL interpreter.

The interpreter is a register machine-like implementation of CEL which
first walks an expressions abstract syntax tree and converts it to a linear
set of instructions. All built-in functions declared in the CEL language
definition, including all macros, are supported by this implementation.

Usage:

```go
import cel "github.com/google/cel-go/interpreter"
interpreter   := cel.StandardInterpreter(<proto.Message(s)>)
interpretable := interpreter.NewInterpretable(
   cel.NewProgram("optional.package.name",
        <expr.Expr>,
        <expr.SourceInfo>))
result, state := 
   interpretable.Eval(
      cel.NewActivation(map[string]interface{}{<identifier_bindings>}))
```

The interpreter supports protobuf and primitive types, but lacks an 
abstract type registry. This will be forthcoming. Additionally, the internal
type representations between the parser, checker, and interpreter have
yet to be unified. This too will be coming in future pull requests, along
with support for the canonical forms of evaluation unknowns and errors.


interpreter/activation.go
```diff
@@ -0,0 +1,115 @@
...
+//
+// Named bindings may lazily supply values by providing a function which
+// accepts no arguments and produces an interface value.
+// TODO: consider passing the current activation to the supplier.
+type MapActivation struct {
+	references map[int64]interface{}
+	bindings   map[string]interface{}
+}
+
+var _ Activation = &MapActivation{}
```
eobrain: This is basically a way of doing "implements" in Go.  However, I think it is contrary to the spirit of Go. See [Why doesn't Go have "implements" declarations?][1]

Suggest deleting this line.

[1]: https://golang.org/doc/faq#implements_interface

interpreter/activation.go
```diff
@@ -0,0 +1,115 @@
...
+
+	// Parent returns the parent of the current activation, may be nil.
+	// If non-nil, the parent will be searched during resolve calls.
+	Parent() Activation
+}
+
+// NewActivation returns an activation based on a map-based binding where the
+// map keys are expected to be qualified names used with ResolveName calls.
+// TODO: supply references from checked.proto.
+func NewActivation(bindings map[string]interface{}) *MapActivation {
```
eobrain: Suggest not exporting `MapActivation` (i.e. renaming it `mapActivation`) and returning `Activation` from this function. (Or is there some reason that the implementation needs to be visible outside this package?)

interpreter/activation.go
```diff
@@ -0,0 +1,115 @@
...
+}
+
+// HierarchicalActivation which implements Activation and contains a parent and
+// child activation.
+type HierarchicalActivation struct {
+	parent Activation
+	child  Activation
+}
+
+var _ Activation = &HierarchicalActivation{}
```
eobrain: Ditto. Suggest removing this eager type check.

interpreter/activation.go
```diff
@@ -0,0 +1,115 @@
...
+		default:
+			return object, true
+		}
+	}
+	return nil, false
+}
+
+// HierarchicalActivation which implements Activation and contains a parent and
+// child activation.
+type HierarchicalActivation struct {
```
eobrain: Suggest not exporting this type.

interpreter/activation.go
```diff
@@ -0,0 +1,115 @@
...
+func (a *HierarchicalActivation) ResolveName(name string) (interface{}, bool) {
+	if object, found := a.child.ResolveName(name); found {
+		return object, found
+	}
+	return a.parent.ResolveName(name)
+}
+
+// ExtendActivation takes two activations and produces a new one which prioritizes
+// resolution in the child first and parent(s) second.
+func ExtendActivation(parent Activation, child Activation) Activation {
```
eobrain: Suggest renaming this NewHierarchicalActivation

interpreter/astwalker.go
```diff
@@ -0,0 +1,498 @@
...
+			fieldValues[entry.GetFieldKey()] = valueId
+		case *expr.Expr_CreateStruct_Entry_MapKey:
+			keyValues[w.getId(entry.GetMapKey())] = valueId
+			entrySteps = append(entrySteps, w.walk(entry.GetMapKey())...)
+		}
+		entrySteps = append(entrySteps, w.walk(entry.GetValue())...)
+	}
+	if len(structExpr.MessageName) == 0 {
+		return append(entrySteps, NewMap(node.Id, keyValues))
+	} else {
```
eobrain: It would be more idiomatic to drop the `else`.

interpreter/types/types.go
```diff
@@ -0,0 +1,122 @@
...
+	aspects.Equaler
+	Name() string
+	IsDyn() bool
+}
+
+// MessageType produces a new Type instance for a qualified message name.
+func MessageType(name string) Type {
+	if name == TimestampType.Name() {
+		return TimestampType
+	} else if name == DurationType.Name() {
```
eobrain: It would be more idiomatic to drop the `else`.

interpreter/dispatcher.go
```diff
@@ -0,0 +1,207 @@
...
+			byFunction[argCount] = append(byArgCount, overloadRef)
+		}
+	}
+	return nil
+}
+
+func (d *defaultDispatcher) Dispatch(ctx *CallContext) (interface{}, error) {
+	if overload, err := d.findOverload(ctx); err == nil {
+		return overload.impl(ctx.args...)
+	} else {
```
eobrain, line 141: Ditto. Suggest dropping `else` statements after `return`.

interpreter/dispatcher.go
```diff
@@ -0,0 +1,207 @@
...
+}
+
+// Find the overload that matches the call context. There should be only one,
+// otherwise the call is ambiguous and a runtime error produced.
+func (d *defaultDispatcher) findOverload(ctx *CallContext) (*overload, error) {
+	// TODO: Add location metadata to overloads.
+	if overloadId, found := ctx.Overload(); found {
+		if match, found := d.overloads[overloadId]; found {
+			return match, nil
+		} else {
```
eobrain: Ditto. Suggest dropping `else` statements after `return`.  Similarly, many more below in this PR.

eobrain (commented):
First set of review comments, mostly just about style.  More coming.

interpreter/activation.go
TristonianJones: I tend to find the eager type check to be helpful during development and refactors, but I can remove these.

interpreter/activation.go
TristonianJones: Done

interpreter/activation.go
TristonianJones: Done

interpreter/activation.go
TristonianJones: Done

interpreter/dispatcher.go
TristonianJones: Done

interpreter/dispatcher.go
TristonianJones: Done

interpreter/types/types.go
TristonianJones: Done

interpreter/astwalker.go
TristonianJones: Done

interpreter/activation.go
TristonianJones: Done

interpreter/dispatcher.go
```diff
@@ -0,0 +1,204 @@
...
+	// Add one or more overloads, returning an error if any Overload has the
+	// same Overload#Name.
+	Add(overloads ...*functions.Overload) error
+
+	// Dispatch a call to its appropriate Overload and return the result or
+	// error.
+	Dispatch(ctx *CallContext) (interface{}, error)
+}
+
+// CallContext description of a function call invocation.
```
eobrain: nit: not a sentence

interpreter/dispatcher.go
```diff
@@ -0,0 +1,204 @@
...
+		for _, candidate := range candidates {
+			if candidate.handlesArgs(args) {
+				matches = append(matches, candidate)
+			}
+		}
+		if len(matches) == 1 {
+			return matches[0], nil
+		}
+		return nil, fmt.Errorf("ambiguous overloads for function '%s'. "+
+				"candidates: ['%s']",
```
eobrain: %s is probably not what you want -- perhaps %v

interpreter/activation.go
```diff
@@ -0,0 +1,111 @@
...
+
+func (a *mapActivation) Parent() Activation {
+	return nil
+}
+
+func (a *mapActivation) ResolveReference(exprId int64) (interface{}, bool) {
+	if object, found := a.references[exprId]; found {
+		return object, true
+	}
+	return nil, false
```
eobrain: Or ...

```go
	object, found := a.references[exprId]
	return object, found
```

interpreter/evalstate.go
```diff
@@ -0,0 +1,69 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// The interpreter package provides functions to evaluate CEL programs against
+// a series of inputs and functions supplied at runtime.
+package interpreter
+
+// EvalState tracks the values associated with expression ids during execution.
+type EvalState interface {
+	// Get the value of the given expression id, false if not found.
```
eobrain: Comment should start with "Value"

interpreter/evalstate.go
```diff
@@ -0,0 +1,69 @@
...
+
+// MutableEvalState permits the mutation of evaluation state for a given
+// expression id.
+type MutableEvalState interface {
+	EvalState
+	SetValue(int64, interface{})
+}
+
+var _ EvalState = &defaultEvalState{}
+var _ MutableEvalState = &defaultEvalState{}
```
eobrain: Suggest removing wager type checks.

interpreter/evalstate.go
```diff
@@ -0,0 +1,69 @@
...
+
+// NewEvalState returns a MutableEvalState.
+func NewEvalState() *defaultEvalState {
```
eobrain: Suggest returning interface:

```go
func NewEvalState() MutableEvalState {
```

interpreter/evalstate.go
```diff
@@ -0,0 +1,69 @@
...
+func (o *UnknownValue) Equal(other interface{}) bool {
+	return false
+}
+
+// TODO: replace this with the value.proto equivalents.
+type ErrorValue struct {
+	ErrorSet []error
+}
+
+func (o *ErrorValue) Equal(other interface{}) bool {
```
eobrain: Missing comment

interpreter/functions/standard.go
```diff
@@ -0,0 +1,1039 @@
...
+				rhs, rhsIsBool := values[1].(bool)
+				if lhsIsBool && rhsIsBool {
+					return lhs || rhs, nil
+				}
+				if lhsIsBool && lhs ||
+					rhsIsBool && rhs {
+					return true, nil
+				}
+				if !rhsIsBool {
+					return rhs, nil
```
eobrain: Ditto. `rhs` will always have its zero value of `false`.

interpreter/functions/standard.go
```diff
@@ -0,0 +1,1039 @@
...
+				rhs, rhsIsBool := values[1].(bool)
+				if lhsIsBool && rhsIsBool {
+					return lhs && rhs, nil
+				}
+				if lhsIsBool && !lhs ||
+					rhsIsBool && !rhs {
+					return false, nil
+				}
+				if !rhsIsBool {
+					return rhs, nil
```
eobrain: Is this correct? In this case `rhs` will always have its zero value of `false`.  If this is what you intend maybe it should be `return false, nil`.

interpreter/functions/standard.go
```diff
@@ -0,0 +1,1039 @@
...
+				cond, isBool := values[0].(bool)
+				trueVal := values[1]
+				falseVal := values[2]
+				if isBool {
+					if cond {
+						return trueVal, nil
+					}
+					return falseVal, nil
+				}
+				return values[0], nil
```
eobrain: So any non-Boolean cond will be returned, not just errors?  Is that the intent?

So 
```c
999 ? "AAA" : "BBB"
```
will evaluate to `999` rather than an error.

If that is the intent, I suggest adding a comment to explain.

interpreter/functions/standard.go
```diff
@@ -0,0 +1,1039 @@
...
+				}
+				return lhs, nil
+			}},
+		// Logical or (a || b)
+		{
+			operators.LogicalOr, operators.LogicalOr,
+			func(value1, value2 bool) {},
+			func(values ...interface{}) (interface{}, error) {
+				lhs, lhsIsBool := values[0].(bool)
+				rhs, rhsIsBool := values[1].(bool)
```
eobrain, line 96: Can you be guarantee that `len(values)` is at least 2?  Could a bad AST cause a panic here?

interpreter/functions/standard.go
```diff
@@ -0,0 +1,1039 @@
...
+			func(values ...interface{}) (interface{}, error) {
+				value1 := values[0].([]byte)
+				value2 := values[1].([]byte)
+				return bytes.Equal(value1, value2), nil
+			}},
+
+		{operators.Equals, operators.Equals,
+			func(value1, value2 interface{}) {},
+			func(values ...interface{}) (interface{}, error) {
+				if value1, ok := values[0].(aspects.Equaler); ok {
```
eobrain, line 141: It's a bit confusing to be mixing zero-base numbering and one-base numbering.  Maybe use `value0` and `value1` for the Equalar versions of the values?

interpreter/functions/standard.go
```diff
@@ -0,0 +1,1039 @@
...
+					return !value2.Equal(values[0]), nil
+				}
+				return values[0] != values[1], nil
+			}},
+
+		// Less than operator
+		{operators.Less, LessBool,
+			func(value1, value2 bool) {},
+			func(values ...interface{}) (interface{}, error) {
+				return !values[0].(bool) && values[1].(bool), nil
```
eobrain, line 172: Is there any way these might not be `bool`?  If so you would get a panic.

interpreter/functions/standard.go
```diff
@@ -0,0 +1,1039 @@
...
+		// Divide operator
+		{operators.Divide, DivideInt64,
+			func(value1 int64, value2 int64) {},
+			func(values ...interface{}) (interface{}, error) {
+				return values[0].(int64) / values[1].(int64), nil
+			}},
+		{operators.Divide, DivideUint64,
+			func(value1 uint64, value2 uint64) {},
+			func(values ...interface{}) (interface{}, error) {
+				return values[0].(uint64) / values[1].(uint64), nil
```
eobrain, line 512: How is division by zero handled?  Should this be returning a non-nil error?

interpreter/instructions.go
```diff
@@ -0,0 +1,226 @@
...
+var _ Instruction = &IdentExpr{}
+var _ Instruction = &SelectExpr{}
+var _ Instruction = &CallExpr{}
+var _ Instruction = &CreateListExpr{}
+var _ Instruction = &CreateMapExpr{}
+var _ Instruction = &CreateObjectExpr{}
+var _ Instruction = &MovInst{}
+var _ Instruction = &JumpInst{}
+var _ Instruction = &PushScopeInst{}
+var _ Instruction = &PopScopeInst{}
```
eobrain: Ditto. Suggest removing these.

interpreter/instructions.go
```diff
@@ -0,0 +1,226 @@
...
+type ConstExpr struct {
+	*baseInstruction
+	Value interface{}
+}
+
+func (e *ConstExpr) String() string {
+	return fmt.Sprintf("const %v, r%d", e.Value, e.GetId())
+}
+
+func NewConst(exprId int64, value interface{}) *ConstExpr {
```
eobrain, line 48: Suggest return `Instruction`.

interpreter/instructions.go
```diff
@@ -0,0 +1,226 @@
...
+type IdentExpr struct {
+	*baseInstruction
+	Name string
+}
+
+func (e *IdentExpr) String() string {
+	return fmt.Sprintf("local '%s', r%d", e.Name, e.GetId())
+}
+
+func NewIdent(exprId int64, name string) *IdentExpr {
```
eobrain, line 62: Suggest return `Instruction`.

interpreter/instructions.go
```diff
@@ -0,0 +1,226 @@
...
+	Operand int64
+	Field   string
+}
+
+func (e *SelectExpr) String() string {
+	return fmt.Sprintf("call  select(%d, '%s'), r%d",
+		e.Operand, e.Field, e.GetId())
+}
+
+func NewSelect(exprId int64, operandId int64, field string) *SelectExpr {
```
eobrain, line 105: Suggest return `Instruction`.

interpreter/instructions.go
```diff
@@ -0,0 +1,226 @@
...
+type CreateListExpr struct {
+	*baseInstruction
+	Elements []int64
+}
+
+func (e *CreateListExpr) String() string {
+	return fmt.Sprintf("mov   list(%v), r%d", e.Elements, e.GetId())
+}
+
+func NewList(exprId int64, elements []int64) *CreateListExpr {
```
eobrain, line 119: Suggest return `Instruction`.

interpreter/instructions.go
```diff
@@ -0,0 +1,226 @@
...
+type CreateMapExpr struct {
+	*baseInstruction
+	KeyValues map[int64]int64
+}
+
+func (e *CreateMapExpr) String() string {
+	return fmt.Sprintf("mov   map(%v), r%d", e.KeyValues, e.GetId())
+}
+
+func NewMap(exprId int64, keyValues map[int64]int64) *CreateMapExpr {
```
eobrain, line 134: Suggest return `Instruction`.

interpreter/instructions.go
```diff
@@ -0,0 +1,226 @@
...
+	*baseInstruction
+	Name        string
+	FieldValues map[string]int64
+}
+
+func (e *CreateObjectExpr) String() string {
+	return fmt.Sprintf("mov   type(%s%v), r%d", e.Name, e.FieldValues, e.GetId())
+}
+
+func NewObject(exprId int64, name string,
```
eobrain, line 150: Suggest return `Instruction`.

interpreter/instructions.go
```diff
@@ -0,0 +1,226 @@
...
+}
+
+func (e *JumpInst) String() string {
+	if e.OnValue != nil {
+		return fmt.Sprintf("jump  %d if r%d == %v", e.Count, e.GetId(), e.OnValue)
+	}
+	return fmt.Sprintf("jump  %d", e.Count)
+}
+
+func NewJump(exprId int64, instructionCount int, jumpOnValue interface{}) *JumpInst {
```
eobrain, line 169: Suggest return `Instruction`.

interpreter/instructions.go
```diff
@@ -0,0 +1,226 @@
...
+type MovInst struct {
+	*baseInstruction
+	ToExprId int64
+}
+
+func (e *MovInst) String() string {
+	return fmt.Sprintf("mov   r%d, r%d", e.GetId(), e.ToExprId)
+}
+
+func NewMov(exprId int64, toExprId int64) *MovInst {
```
eobrain, line 183: Suggest return `Instruction`.

interpreter/instructions.go
```diff
@@ -0,0 +1,226 @@
...
+type PushScopeInst struct {
+	*baseInstruction
+	Declarations map[string]int64
+}
+
+func (e *PushScopeInst) String() string {
+	return fmt.Sprintf("block  %v", e.Declarations)
+}
+
+func NewPushScope(exprId int64, declarations map[string]int64) *PushScopeInst {
```
eobrain, line 198: Suggest return `Instruction`.

interpreter/instructions.go
```diff
@@ -0,0 +1,226 @@
...
+// previous activation.
+type PopScopeInst struct {
+	*baseInstruction
+}
+
+func (e *PopScopeInst) String() string {
+	return fmt.Sprintf("end")
+}
+
+func NewPopScope(exprId int64) *PopScopeInst {
```
eobrain, line 212: Suggest return `Instruction`.

interpreter/interpreter.go
```diff
@@ -0,0 +1,286 @@
...
+			// Special instruction for modifying the program cursor
+		case *JumpInst:
+			jmpExpr := step.(*JumpInst)
+			// TODO: Add test for whether the jump should be made based
+			// on nil as the jump value, equality as the value, or unknown/err
+			if jmpExpr.OnValue == nil || i.value(resultId) == jmpExpr.OnValue {
+				if !stepper.JumpCount(jmpExpr.Count) {
+					// TODO: Error, the jump count should never exceed the
+					// program length.
+					panic("jumped too far")
```
eobrain, line 110: Should not panic. Can we add an error return to the function? Or can this never really happen?

interpreter/interpreter.go
```diff
@@ -0,0 +1,286 @@
...
+		state:       NewEvalState()}
+}
+
+type exprInterpretable struct {
+	interpreter *exprInterpreter
+	program     Program
+	state       MutableEvalState
+}
+
+var _ Interpretable = &exprInterpretable{}
```
eobrain: Suggest removing eager type checking.

interpreter/interpreter.go
```diff
@@ -0,0 +1,286 @@
...
+// NewInterpreter builds an Interpreter from a Dispatcher and TypeProvider
+// which will be used throughout the Eval of all Interpretable instances
+// gerenated from it.
+func NewInterpreter(dispatcher Dispatcher, typeProvider providers.TypeProvider) *exprInterpreter {
+	return &exprInterpreter{dispatcher, typeProvider}
+}
+
+// StandardInterpreter builds a Dispatcher and TypeProvider with support
+// for all of the CEL builtins defined in the language definition.
+func StandardIntepreter(types ...proto.Message) *exprInterpreter {
```
eobrain: Suggest return `Interpreter`.

interpreter/interpreter.go
```diff
@@ -0,0 +1,286 @@
...
+	dispatcher   Dispatcher
+	typeProvider providers.TypeProvider
+}
+
+var _ Interpreter = &exprInterpreter{}
+
+// NewInterpreter builds an Interpreter from a Dispatcher and TypeProvider
+// which will be used throughout the Eval of all Interpretable instances
+// gerenated from it.
+func NewInterpreter(dispatcher Dispatcher, typeProvider providers.TypeProvider) *exprInterpreter {
```
eobrain: Suggest return `Interpreter`.

interpreter/interpreter.go
```diff
@@ -0,0 +1,286 @@
...
+	// Eval an Activation to produce an output and EvalState.
+	Eval(activation Activation) (interface{}, EvalState)
+}
+
+type exprInterpreter struct {
+	dispatcher   Dispatcher
+	typeProvider providers.TypeProvider
+}
+
+var _ Interpreter = &exprInterpreter{}
```
eobrain: Suggest remove eager type checking.

interpreter/interpreter.go
```diff
@@ -0,0 +1,286 @@
...
+// the expression was not type-checked and the select, in fact, refers to a
+// qualified identifier name instead of a series of field selections.
+func (i *exprInterpretable) resolveUnknown(unknown *UnknownValue,
+	selExpr *SelectExpr,
+	currActivation Activation) {
+	if object, found := currActivation.ResolveReference(selExpr.Id); found {
+		i.setValue(selExpr.Id, object)
+	} else {
+		var validIdent = true
+		var identifier = selExpr.Field
```
eobrain: Suggest

```go
		validIdent := true
		identifier := selExpr.Field
```

interpreter/interpreter.go
```diff
@@ -0,0 +1,286 @@
...
+			} else {
+				i.setValue(selExpr.Id,
+					&UnknownValue{append([]Instruction{selExpr}, unknown.Args...)})
+			}
+		}
+	}
+}
+
+func (i *exprInterpretable) evalCall(step Instruction, currActivation Activation) {
+	callExpr := step.(*CallExpr)
```
eobrain: Suggest avoiding type assertion here.

```go
func (i *exprInterpretable) evalCall(callExpr *CallExpr, currActivation Activation) {
	argVals := make([]interface{}, len(callExpr.Args), len(callExpr.Args))
	for idx, argId := range callExpr.Args {
		argVals[idx] = i.value(argId)
	}
	ctx := &CallContext{
		call:       callExpr,
		activation: currActivation,
		args:       argVals,
		metadata:   i.program.Metadata()}
	if result, err := i.interpreter.dispatcher.Dispatch(ctx); err == nil {
		i.setValue(callExpr.GetId(), result)
	} else {
		i.setValue(callExpr.GetId(), err)
	}
}
```

interpreter/interpreter.go
```diff
@@ -0,0 +1,286 @@
...
+		resultId = step.GetId()
+		switch step.(type) {
+		case *ConstExpr:
+			i.evalConst(step)
+		case *IdentExpr:
+			i.evalIdent(step, currActivation)
+		case *SelectExpr:
+			i.evalSelect(step, currActivation)
+		case *CallExpr:
+			i.evalCall(step, currActivation)
```
eobrain: Suggest doing the type assertion here where we can be sure it won't panic, rather than in the function.

```go
		case *CallExpr:
			i.evalCall(step.(*CallExpr), currActivation)
```

interpreter/interpreter.go
```diff
@@ -0,0 +1,286 @@
...
+		case *CreateListExpr:
+			i.evalCreateList(step)
+		case *CreateMapExpr:
+			i.evalCreateMap(step)
```
eobrain: Suggest doing type assertion here rather than in function.

interpreter/interpreter.go
```diff
@@ -0,0 +1,286 @@
...
+			currActivation = NewHierarchicalActivation(currActivation, NewActivation(childActivaton))
+		case *PopScopeInst:
+			currActivation = currActivation.Parent()
+		}
+	}
+	return i.value(resultId), i.state
+}
+
+func (i *exprInterpretable) evalConst(step Instruction) {
+	i.setValue(step.GetId(), step.(*ConstExpr).Value)
```
eobrain: Ditto on moving type assertion out to the type switch.

interpreter/interpreter.go
```diff
@@ -0,0 +1,286 @@
...
+}
+
+func (i *exprInterpretable) evalIdent(step Instruction, currActivation Activation) {
+	idExpr := step.(*IdentExpr)
```
eobrain: Ditto on moving type assertion out to the type switch.

interpreter/interpreter.go
```diff
@@ -0,0 +1,286 @@
...
+		i.setValue(step.GetId(), result)
+	} else if enum, found := i.interpreter.typeProvider.EnumValue(idExpr.Name); found {
+		i.setValue(step.GetId(), enum)
+	} else {
+		i.setValue(step.GetId(), &UnknownValue{[]Instruction{step}})
+	}
+}
+
+func (i *exprInterpretable) evalSelect(step Instruction, currActivation Activation) {
+	selExpr := step.(*SelectExpr)
```
eobrain: ditto

interpreter/interpreter.go
```diff
@@ -0,0 +1,286 @@
...
+		metadata:   i.program.Metadata()}
+	if result, err := i.interpreter.dispatcher.Dispatch(ctx); err == nil {
+		i.setValue(step.GetId(), result)
+	} else {
+		i.setValue(step.GetId(), err)
+	}
+}
+
+func (i *exprInterpretable) evalCreateList(step Instruction) {
+	listExpr := step.(*CreateListExpr)
```
eobrain: ditto on suggesting moving type assertion out to calling location

interpreter/interpreter.go
```diff
@@ -0,0 +1,286 @@
...
+		entries[i.value(keyId)] = i.value(valueId)
+	}
+	// TODO: Add an error state if any key is repeated and any element in the
+	// map (key or value) is an error.
+	adaptingMap := types.NewMapValue(entries)
+	i.setValue(step.GetId(), adaptingMap)
+}
+
+func (i *exprInterpretable) evalCreateType(step Instruction) {
+	typeExpr := step.(*CreateObjectExpr)
```
eobrain: ditto

interpreter/interpreter.go
```diff
@@ -0,0 +1,286 @@
...
+	}
+	if value, err := i.newValue(typeExpr.Name, fields); err == nil {
+		i.setValue(step.GetId(), value)
+	} else {
+		i.setValue(step.GetId(), err)
+	}
+}
+
+func (i *exprInterpretable) evalMov(step Instruction) {
+	movExpr := step.(*MovInst)
```
eobrain: ditto

interpreter/program.go
```diff
@@ -0,0 +1,160 @@
...
+// to jump a certain number of instructions forward or back.
+type InstructionStepper interface {
+	// Next returns the next instruction, or false if the end of the program
+	// has been reached.
+	Next() (Instruction, bool)
+
+	// JumpCount moves a relative count of instructions forward or back in the
+	// program and returns whether the jump was successful.
+	//
+	// A jump may be unsuccessful if the number of instructions to jump exceeds
```
eobrain, line 51: Comment should start with "JumpCount".

interpreter/program.go
```diff
@@ -0,0 +1,160 @@
...
+}
+
+type exprProgram struct {
+	expression   *expr.Expr
+	container    string
+	instructions []Instruction
+	metadata     *exprMetadata
+}
+
+var _ Program = &exprProgram{}
```
eobrain: SUggest remove eager type checking.

interpreter/program.go
```diff
@@ -0,0 +1,160 @@
...
+
+// NewProgram creates a Program from a CEL expression and source information
+// within the specified container
+func NewProgram(container string, expression *expr.Expr,
+	sourceInfo *expr.SourceInfo) *exprProgram {
```
eobrain: Suggest return `Program`

interpreter/program.go
```diff
@@ -0,0 +1,160 @@
...
+}
+
+// The exprMetadata type provides helper functions for retrieving source
+// locations in a human readable manner based on the data contained within
+// the expr.SourceInfo message.
+type exprMetadata struct {
+	info *expr.SourceInfo
+}
+
+var _ Metadata = &exprMetadata{}
```
eobrain: Suggest remove eager type checking.

interpreter/program.go
```diff
@@ -0,0 +1,160 @@
...
+
+func newExprMetadata(info *expr.SourceInfo) *exprMetadata {
```
eobrain: Suggest return `Metadata`/

interpreter/program.go
```diff
@@ -0,0 +1,160 @@
...
+	// A jump may be unsuccessful if the number of instructions to jump exceeds
+	// the beginning or end of the program.
+	JumpCount(count int) bool
+}
+
+type exprProgram struct {
+	expression   *expr.Expr
+	container    string
+	instructions []Instruction
+	metadata     *exprMetadata
```
eobrain: Suggest

```go
	metadata     Metadata
```

interpreter/types/adapters_test.go
```diff
@@ -0,0 +1,137 @@
...
+func TestUnsupportedConversion(t *testing.T) {
+	if val, err := ProtoToExpr(NonConvertiable{}); err == nil {
+		t.Error("Expected error when converting non-proto struct to proto", val)
+	}
+	if val, err := ExprToProto(reflect.TypeOf(NonConvertiable{}), NonConvertiable{}); err == nil {
+		t.Error("Expected error when converting non-proto expr to proto", val)
+	}
+}
+
+func expectExprToProto(t *testing.T, in interface{}, out interface{}) {
```
eobrain, line 84: The Google Go style is to avoid assert libraries like these two expect*() functions. (Internally see http://go/gocomments#TOC-Assert)

interpreter/types/list.go
```diff
@@ -0,0 +1,265 @@
...
+	Contains(value interface{}) bool
+
+	// Concat two lists together into an adapted list.
+	Concat(other ListValue) ListValue
+
+	// Len returns the length of the list.
+	Len() int64
+}
+
+var _ ListValue = &listValue{}
```
eobrain: Ditto. Suggest removing.

interpreter/types/list.go
```diff
@@ -0,0 +1,265 @@
...
+	value := l.refValue.Index(int(i)).Interface()
+	return ProtoToExpr(value)
+}
+
+func (l *listValue) ToProto(refType reflect.Type) (interface{}, error) {
+	protoElem := refType.Elem()
+	protoElemKind := protoElem.Kind()
+	if protoElemKind == l.elemKind {
+		return l.value, nil
+	} else if protoElem.ConvertibleTo(l.listType.Elem()) {
```
eobrain: Suggest removing redundant `else`

interpreter/types/object.go
```diff
@@ -0,0 +1,165 @@
...
+}
+
+func NewProtoValue(value interface{}) ObjectValue {
+	// Unwrap any.Any values on construction.
+	_, isAny := value.(*any.Any)
+	if isAny {
+		// TODO: check whether the any being null is valid.
+		dynAny := &ptypes.DynamicAny{}
+		if err := ptypes.UnmarshalAny(value.(*any.Any), dynAny); err != nil {
+			panic(err)
```
eobrain: Generally should not panic.  Suggest adding error return parameter to this function instead.

interpreter/types/object.go
```diff
@@ -0,0 +1,165 @@
...
+		refMethod := it.msgType.Method(int(index))
+		if refMethod.Name[:3] == "Get" {
+			refValue := *it.msgRefValue
+			fieldRefValue := refMethod.Func.Call([]reflect.Value{refValue})[0]
+			if fieldRefValue.IsValid() {
+				fieldRefKind := fieldRefValue.Kind()
+				if fieldRefKind != reflect.Ptr && fieldRefKind != reflect.Interface {
+					it.cursorField = refMethod.Name[3:]
+					return true
+				} else if !fieldRefValue.IsNil() {
```
eobrain: Can remove redundant `else`

interpreter/types/providers/providers.go
```diff
@@ -0,0 +1,154 @@
...
+	// name could not be resolved.
+	EnumValue(enumName string) (int64, bool)
+}
+
+type defaultTypeProvider struct {
+	typesByName map[string]reflect.Type
+	enumsByName map[string]int64
+}
+
+var _ TypeProvider = &defaultTypeProvider{}
```
eobrain: ditto, suggest removing

interpreter/types/providers/providers.go
```diff
@@ -0,0 +1,154 @@
...
+
+// NewTypeProvider accepts a list of proto message instances and returns a type
+// provider which can create new instances of the provided message or any
+// message that proto depends upon in its FileDescriptor.
+func NewTypeProvider(types ...proto.Message) *defaultTypeProvider {
```
eobrain: Suggest returning `TypeProvider`

interpreter/types/providers/providers.go
```diff
@@ -0,0 +1,154 @@
...
+	}
+}
+
+func buildDescriptorSet(fileDesc *protobuf.FileDescriptorProto,
+	descriptorSet map[string]*protobuf.FileDescriptorProto) {
+	descriptorSet[fileDesc.GetName()] = fileDesc
+	for _, protoFileName := range fileDesc.Dependency {
+		if _, found := descriptorSet[protoFileName]; !found {
+			if fd, err := fileDescriptor(protoFileName); err != nil {
+				panic(err)
```
eobrain, line 129: Should avoid `panic`. Suggest adding `error` return parameter.

eobrain (changes requested):

interpreter/types/adapters_test.go
TristonianJones: The helper (marked as such with t.Helper()) seems to clarify the nature of the test. It primarily exists as a comparison function which is one area where the assert discussion seems to talk about needing to handle comparison of got versus wanted via libraries.

interpreter/dispatcher.go
TristonianJones: Actually, this one had to stay.

interpreter/dispatcher.go
TristonianJones: Done.

interpreter/dispatcher.go
TristonianJones: Thanks, updated.

interpreter/activation.go
TristonianJones: Done.

interpreter/evalstate.go
TristonianJones: Done.

interpreter/evalstate.go
TristonianJones: Done.

interpreter/evalstate.go
TristonianJones: Done.

interpreter/evalstate.go
TristonianJones: Removed, not used.

interpreter/functions/standard.go
TristonianJones: This is a bug. I've updated it. The original code was forgetting to check the case where the value was not bool, error, or unknown.

interpreter/functions/standard.go
TristonianJones: Same as below.

interpreter/functions/standard.go
TristonianJones: Bug. Fixed.

interpreter/functions/standard.go
TristonianJones: The function guard above ensures the the implementation is called with the appropriate argument count and types.

interpreter/functions/standard.go
TristonianJones: Acknowledged.

interpreter/functions/standard.go
TristonianJones: See the comment on the 'Signature' and 'Impl' fields of an Overload for more details about how this works.

interpreter/functions/standard.go
TristonianJones: None of the numeric edge cases are handled at the moment. I've added a TODO.

interpreter/interpreter.go
TristonianJones: This represents a bad logic bug in the CEL interpreter. It should never happen as it means expressions are not being evaluated properly.

interpreter/interpreter.go
TristonianJones: Thanks, done.

interpreter/interpreter.go
TristonianJones: SGTM, done.

interpreter/program.go
TristonianJones: It does, this is the second line of comments about when a jump may not be successful.

interpreter/types/object.go
TristonianJones: I've added a TODO. Agree that this should be handled differently, but wasn't sure how a missing type should have been represented until very recently.

interpreter/types/providers/providers.go
TristonianJones: This is also related to how missing types should be handled. The expectation is this should never happen unless you attempt to refer to a proto that is not compiled into your go library.

interpreter/instructions.go
TristonianJones: The types returned in this file are exported have their own inspectable properties, so I have left them as their types rather than as instructions.

TristonianJones:
PTAL

eobrain (approved):

eobrain:
LGTM

Issue #7, opened 2018-03-29
Use protobuf generated structs directly


TristonianJones:
The internal type system used within the parser and type-checker
is convenient for formatting debug strings, but using the protobuf
representations directly will simplify the implementation, provide
valuable feedback about the usability of the abstract CEL representation,
and reduce the likelihood of mismatches between the specification
and the implementation.

PR #8, opened 2018-03-30
Migrate from internal 'ast' to syntax.proto representation.

TristonianJones:
#7 The syntax.proto types are simple Go structs and should be
used in favor of any other sort of type representation, even at
the expense of having to update the parser / type-checker if
and when the syntax.proto changes.

The type-checker still uses its own flavor of semantic types, but
this will be removed in favor of checked.proto types in a future
pull request.


checker/checker.go
```diff
@@ -487,3 +507,25 @@ func newResolution(ref *semantics.FunctionReference, t types.Type) *overloadReso
 		Type:      t,
 	}
 }
+
+func (c *checker) location(e *expr.Expr) common.Location {
+	return c.locationById(e.Id)
+}
+
+func (c *checker) locationById(id int64) common.Location {
+	positions := c.sourceInfo.GetPositions()
```
JimLarson, line 516: If this just returns a reference to an already-constructed map, then ignore this comment.  Otherwise, if this does some work to construct this map, do it once at checker initialization time instead of on each call to get a location.

checker/checker.go
```diff
@@ -487,3 +507,25 @@ func newResolution(ref *semantics.FunctionReference, t types.Type) *overloadReso
...
+		for _, lineOffset := range c.sourceInfo.LineOffsets {
+			if lineOffset < offset {
+				line += 1
+				col = int(offset - lineOffset)
+			} else {
+				break;
+			}
+		}
+	}
+	return common.NewLocation(line, col)
```
JimLarson: Consider line/col -1/-1 for a not found location, since 1/0 is in-band.

debug/debug.go
```diff
@@ -0,0 +1,296 @@
...
+
+func newDebugWriter(a DebugAdorner) *debugWriter {
+	return &debugWriter{
+		adorner:   a,
+		indent:    0,
+		lineStart: true,
+	}
+}
+
+func (w *debugWriter) Append(e *expr.Expr) {
```
JimLarson: Is it a Go idiom to have two methods distinguished only by capitalization?

parser/macro.go
```diff
@@ -124,71 +119,69 @@ const (
...
-		zeroExpr := ast.NewInt64Constant(p.id(), loc, 0)
-		oneExpr := ast.NewInt64Constant(p.id(), loc, 1)
+		zeroExpr := p.newConstInt(ctx, 0)
+		oneExpr := p.newConstInt(ctx, 1)
 		init = zeroExpr
-		condition = ast.NewCallFunction(p.id(), loc, operators.LessEquals, accuIdent(), oneExpr)
-		step = ast.NewCallFunction(p.id(), loc, operators.Conditional, args[1],
-			ast.NewCallFunction(p.id(), loc, operators.Add, accuIdent(), oneExpr), accuIdent())
-		result = ast.NewCallFunction(p.id(), loc, operators.Equals, accuIdent(), oneExpr)
+		condition = p.newGlobalCall(ctx, operators.LessEquals, accuIdent(), oneExpr)
```
JimLarson, line 165: I thought existsOne() didn't terminate early - that it would evaluation each position even if it already saw 2 or more so that it noticed errors.

TristonianJones: It can terminate early on false, but if it encounters an error along the way it can't continue execution. It doesn't absorb errors like the exists / all.

debug/debug.go
TristonianJones: Not really an idiom, so much as that's how Go does visibility (Caps is public, lower is package private). I can rename the method to something else if it would make things clearer.

checker/checker.go
TristonianJones: GetPositions returns a reference to an already constructed map. The `locationById` method just resolves the raw offset to something line / col view of the data.

checker/checker.go
TristonianJones: Good suggestion. Thanks

debug/debug.go
JimLarson: Yeah, unless it's a well-established idiom, I'd prefer a rename.

parser/macro.go
JimLarson: Right, it doesn't ignore errors, but based on our earlier discussions I got the impression that it had to do a full count of all true values, then check that the count is equal to one, and an error anywhere in the list/map caused it to error out.

If the code here is what's intended, then the semantics need to be updated.

debug/debug.go
TristonianJones: Renamed to Buffer()

parser/macro.go
TristonianJones: I've added a TODO to make this consistent with CEL semantics.

TristonianJones:
PTAL

JimLarson (approved):
LGTM

PR #10, opened 2018-04-05
Migrate from internal 'semantic' types to checked.proto representations.

TristonianJones:
#7 This change also unifies the standard checked declarations with the standard
interpreter built-in names.

A future change will introduce a type registry to ensure that standard type
identifiers are published in a manner consistent with how these idents are
expected to be resolved within the interpreter.


checker/env.go
```diff
@@ -15,21 +15,22 @@
 package checker
 
 import (
+	"github.com/google/cel-go/checker/decls"
+	"github.com/google/cel-go/checker/types"
 	"github.com/google/cel-go/common"
-	"github.com/google/cel-go/semantics"
-	"github.com/google/cel-go/semantics/types"
+	"github.com/google/cel-spec/proto/checked/v1/checked"
 	expr "github.com/google/cel-spec/proto/v1/syntax"
```
eobrain, line 22: Its conventional to use a short name ending in "pb" when importing protos.  Suggest:

```go
	cpb "github.com/google/cel-spec/proto/checked/v1/checked"
￼  	spb "github.com/google/cel-spec/proto/v1/syntax"
```

Similarly in other files.

checker/env.go
```diff
@@ -39,25 +40,42 @@ func NewEnv(errors *common.Errors, typeProvider TypeProvider) *Env {
 	}
 }
 
-func (env *Env) AddFunction(function *semantics.Function) {
-	current := env.declarations.FindFunction(function.Name())
+func (env *Env) Add(decls ...*checked.Decl) {
```
eobrain: Optional. It would be more in keeping with Go conventions to use a single-letter receiver name here and elsewhere in this file.

```go
func (e *Env) Add(
```

checker/errors.go
```diff
@@ -32,7 +33,7 @@ func (errors *TypeErrors) expressionDoesNotSelectField(l common.Location) {
 	errors.ReportError(l, "expression does not select a field")
 }
 
-func (errors *TypeErrors) typeDoesNotSupportFieldSelection(l common.Location, t types.Type) {
+func (errors *TypeErrors) typeDoesNotSupportFieldSelection(l common.Location, t *checked.Type) {
```
eobrain: Optional: Suggest one- or two-letter receiver name, maybe `e`, `te`, or `es`.

checker/errors.go
```diff
@@ -44,41 +45,68 @@ func (errors *TypeErrors) fieldDoesNotSupportPresenceCheck(l common.Location, fi
...
+		}
+		result += types.FormatType(arg)
+	}
+	result += ")"
+	if resultType != nil {
+		result += " -> "
+		result += types.FormatType(resultType)
+	}
+
+	return result
```
eobrain, line 111: If you care about performance you might want to check out http://herman.asia/efficient-string-concatenation-in-go which suggests that something like the following would be much faster:

```go
func formatFunction(resultType *checked.Type, argTypes []*checked.Type, isInstance bool) string {
	var result bytes.Buffer
	if isInstance {
		target := argTypes[0]
		argTypes = argTypes[1:]

		result.WriteString(types.FormatType(target))
		result.WriteString(".")
	}

	result.WriteString("(")
	for i, arg := range argTypes {
		if i > 0 {
			result.WriteString(", ")
		}
		result.WriteString(types.FormatType(arg))
	}
	result.WriteString(")")
	if resultType != nil {
		result.WriteString(" -> ")
		result.WriteString(types.FormatType(resultType))
	}

	return result.String()
}
```

checker/types/mapping.go
```diff
@@ -12,36 +12,36 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-package checker
+package types
 
 import (
 	"fmt"
-	"github.com/google/cel-go/semantics/types"
+	"github.com/google/cel-spec/proto/checked/v1/checked"
```
eobrain, line 19: Suggest using `pb` as short name for proto package.

```go
	pb "github.com/google/cel-spec/proto/checked/v1/checked"
```

eobrain (commented):
LGTM with some optional comments.

checker/errors.go
TristonianJones: Thanks for the tip. I don't particularly care about performance here, but this is good to know.

checker/env.go
TristonianJones: I'll address import naming for protos in another PR. Thanks.

checker/env.go
TristonianJones: Done

checker/errors.go
TristonianJones: Done.

checker/types/mapping.go
TristonianJones: Deferring to another PR.

Issue #13, opened 2018-04-05
Verify that declarations are non-overlapping after type-erasure


TristonianJones:
At type-check time within the env.go, verify that declarations are non-overlapping
with special care to make sure that overload signatures for functions cannot collide
after type erasure.

PR #16, opened 2018-04-09
Check overloads against existing overloads and macros for ambiguity.

ayas:
#13 This implementation pretty much follows the Java one. Note that I did not
add a source location to the error. This seems to be inline with the
Java implementation. If we wanted to add this, we would probably have to
pass the environment the sourceinfo which would lead to more changes

### DIFF
--- a/checker/env.go
+++ b/checker/env.go
@@ -91,6 +91,7 @@ func (e *Env) addOverload(f *checked.Decl, overload *checked.Decl_FunctionDecl_O
 func (e *Env) addFunction(decl *checked.Decl) {
 	current := e.declarations.FindFunction(decl.Name)
 	if current == nil {
+		//Add the function declaration without overloads and check the overloads below.
 		current = decls.NewFunction(decl.Name)
 		e.declarations.AddFunction(current)
 	}
@@ -100,7 +101,6 @@ func (e *Env) addFunction(decl *checked.Decl) {
 	for _, overload := range decl.GetFunction().GetOverloads() {
 		e.addOverload(current, overload)
 	}
-	decl = current
 }
 
 func (e *Env) addIdent(decl *checked.Decl) {
--- a/checker/env.go
+++ b/checker/env.go
@@ -72,8 +72,8 @@ func (e *Env) addOverload(f *checked.Decl, overload *checked.Decl_FunctionDecl_O
 			types.IsAssignable(emptyMappings, existingErased, overloadErased) != nil)
 		if overlap &&
 			overload.GetIsInstanceFunction() == existing.GetIsInstanceFunction() {
-			e.errors.overlappingOverload(common.NoLocation, f.Name, overloadFunction,
-				existingFunction)
+			e.errors.overlappingOverload(common.NoLocation, f.Name, overload.GetOverloadId(), overloadFunction,
+				existing.GetOverloadId(), existingFunction)
 			return
 		}
 	}
@@ -95,9 +95,7 @@ func (e *Env) addFunction(decl *checked.Decl) {
 		current = decls.NewFunction(decl.Name)
 		e.declarations.AddFunction(current)
 	}
-	if current.Name != decl.Name {
-		return
-	}
+
 	for _, overload := range decl.GetFunction().GetOverloads() {
 		e.addOverload(current, overload)
 	}
--- a/checker/env.go
+++ b/checker/env.go
@@ -18,6 +18,7 @@ import (
 	"github.com/google/cel-go/checker/decls"
 	"github.com/google/cel-go/checker/types"
 	"github.com/google/cel-go/common"
+	"github.com/google/cel-go/parser"
 	"github.com/google/cel-spec/proto/checked/v1/checked"
 	expr "github.com/google/cel-spec/proto/v1/syntax"
 )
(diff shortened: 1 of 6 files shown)


checker/env.go
```diff
@@ -57,20 +58,49 @@ func (e *Env) Add(decls ...*checked.Decl) {
...
-		e.declarations.AddFunction(decl)
 	}
+	function.Overloads = append(function.GetOverloads(), overload)
+}
+
+func (e *Env) addFunction(decl *checked.Decl) {
+	current := e.declarations.FindFunction(decl.Name)
+	if current == nil {
+		current = decls.NewFunction(decl.Name)
+		e.declarations.AddFunction(current)
```
TristonianJones: Prefer to return here.

checker/env.go
```diff
@@ -57,20 +58,49 @@ func (e *Env) Add(decls ...*checked.Decl) {
...
+	}
+	if current.Name != decl.Name {
+		return
+	}
+	for _, overload := range decl.GetFunction().GetOverloads() {
+		e.addOverload(current, overload)
+	}
+	decl = current
```
TristonianJones: This line seems to be unused.

checker/env.go
```diff
@@ -57,20 +58,49 @@ func (e *Env) Add(decls ...*checked.Decl) {
...
+			types.IsAssignable(emptyMappings, existingErased, overloadErased) != nil)
+		if overlap &&
+			overload.GetIsInstanceFunction() == existing.GetIsInstanceFunction() {
+			e.errors.overlappingOverload(common.NoLocation, f.Name, overloadFunction,
+				existingFunction)
+			return
+		}
+	}
+
+	for _, macro := range parser.AllMacros {
```
TristonianJones, line 81: We should consider moving the Macros into common/macros/macros.go. Operators and overloads are currently in their own packages there.

checker/env.go
```diff
@@ -57,20 +58,49 @@ func (e *Env) Add(decls ...*checked.Decl) {
...
+	overloadErased := types.Substitute(emptyMappings, overloadFunction, true)
+	for _, existing := range function.GetOverloads() {
+		existingFunction := types.NewFunction(existing.GetResultType(),
+			existing.GetParams()...)
+		existingErased := types.Substitute(emptyMappings, existingFunction, true)
+		overlap := (types.IsAssignable(emptyMappings, overloadErased, existingErased) != nil ||
+			types.IsAssignable(emptyMappings, existingErased, overloadErased) != nil)
+		if overlap &&
+			overload.GetIsInstanceFunction() == existing.GetIsInstanceFunction() {
+			e.errors.overlappingOverload(common.NoLocation, f.Name, overloadFunction,
```
TristonianJones: Consider introducing decls.FormatDecl() to go along with types.FormatType(). Formatting a type results in a human-friendly representation of a proto-based type, so we should have equivalently human-friendly representations of declarations.

TristonianJones (changes requested):
This particular error set won't have source locations since the declarations are defined by the developer checking the expression rather than as part of the expression itself. This is a good start. Let me know if you have any questions about the suggested refactors.

TristonianJones:
For future reference, please fork rather than creating a branch off of master. In general, the main repo should only contain tagged releases and master.

checker/env.go
ayas: maybe decls.FormatDecl() should be done in separate change because we are not using this in this change ? We are passing function types to the error which use types.FormatType to print the function type.

checker/env.go
ayas: I did look into this and did not do it. I thought this would be useful if we could do macros construction only in common/macros/macros.go and expose AllMacros only(similar to what you did with operators and overloads). But macros initialization seem to be tightly integrated with parser because of expander. I could make macros.go use an interface for parserHelper but that did not seem very appealing.

checker/env.go
ayas: Done.

checker/env.go
ayas: I added a comment to clarify this. I think we should still check the added overloads for ambiguity. For this we first add the declaration without overloads and the loop below checks each overload for ambiguity .

checker/env.go
TristonianJones: There are portions of the `parserHelper` that should be made an interface, specifically expression id / source location tracking, but this can wait. Seems like making the macro metadata common and the implementations specific to the parser would be the ideal way to go. I would prefer to not have the checker depend directly on the parser if possible, but let's just add a TODO for now.

checker/env.go
TristonianJones: Printing a function from FormatType appears to be missing from types.go. Please add it and ignore my earlier suggestion of decls.FormatDecl().

I would expect the function type listed in the error message to be formatted as follows:
```
<input>:-1:0: overlapping overload for name  'myfunc' (function(dyn) : dyn) ' 
cannot be distinguished from '(function(dyn) : dyn)'
```

Also, this error message should include the overload id of the candidate and the id of the one it collides with.

checker/env.go
```diff
@@ -57,19 +58,48 @@ func (e *Env) Add(decls ...*checked.Decl) {
...
+}
+
+func (e *Env) addFunction(decl *checked.Decl) {
+	current := e.declarations.FindFunction(decl.Name)
+	if current == nil {
+		//Add the function declaration without overloads and check the overloads below.
+		current = decls.NewFunction(decl.Name)
+		e.declarations.AddFunction(current)
+	}
+	if current.Name != decl.Name {
```
TristonianJones: Under what circumstance would the current.Name and decl.Name not be equal? Consider removing this if-block

TristonianJones (changes requested):

checker/env.go
ayas: Added a TODO.

checker/env.go
ayas: I added formatFunction. Let me know if the formatting of the error message is OK now.

checker/env.go
ayas: I mostly had left this here because it was in the original changes. I agree it looks useless. Removed it.

TristonianJones (approved):

PR #21, opened 2018-04-27
Interpreter benchmarks and related improvements

TristonianJones:
* Benchmarks break down different components of interpretation.
* AST walking now pre-populates the evaluation state with contants
  values encountered during the walk step (no folding yet).
* Eval state switches from maps to lists, yielding a 5x improvement
  in evaluation speed.

Note: `go fmt` was run prior to the creation of this pull request, so
there are a handful of unrelated changes to the ones enumerated
above.


interpreter/astwalker.go
```diff
@@ -522,3 +485,90 @@ func jumpIfEqual(exprId int64, value ref.Value) func(EvalState) bool {
...
+		}
+		switch node.ExprKind.(type) {
+		case *expr.Expr_SelectExpr:
+			count += comprehensionCount(node.GetSelectExpr().GetOperand())
+		case *expr.Expr_CallExpr:
+			call := node.GetCallExpr()
+			count += comprehensionCount(call.GetTarget()) + comprehensionCount(call.GetArgs()...)
+		case *expr.Expr_ListExpr:
+			list := node.GetListExpr()
+			count += comprehensionCount(list.GetElements()...)
```
ayas: node.GetListExpr().GetElements() ?  Similarly below.

interpreter/evalstate.go
```diff
@@ -31,17 +33,21 @@ type MutableEvalState interface {
...
+	exprValues []ref.Value
 }
 
 func (s *defaultEvalState) Value(exprId int64) (ref.Value, bool) {
-	object, found := s.exprValues[exprId]
-	return object, found
+	if exprId >= 0 && exprId < s.exprCount {
+		return s.exprValues[exprId], true
+	}
+	return nil, false
```
ayas: why is this not treated like an error?

interpreter/program.go
```diff
@@ -102,14 +92,36 @@ func (p *exprProgram) GetInstruction(exprId int64) Instruction {
...
+	}
 	return &exprStepper{p, 0}
 }
 
+func (p *exprProgram) MaxInstructionId() int64 {
+	// The max instruction id is computed as the highest expression id + 1
+	// combined with the number of comprehensions times two. Each comprehension
+	// introduces two generated ids (one for an iterator and one for current
+	// iterator value) once the program is initialized.
+	return maxId(p.expression) + comprehensionCount(p.expression)*2
```
ayas: ')*2' -> ') * 2'

interpreter/program.go
```diff
@@ -102,14 +92,36 @@ func (p *exprProgram) GetInstruction(exprId int64) Instruction {
 	return p.instructions[p.revInstructions[exprId]]
 }
 
-func (p *exprProgram) Instructions() InstructionStepper {
+func (p *exprProgram) Init(dispatcher Dispatcher, state MutableEvalState) InstructionStepper {
+	if p.instructions == nil {
```
ayas: Maybe call this Begin() ?

ayas (approved):
Minor comments.

interpreter/evalstate.go
ayas: Does this check imply/impose that the expressions id have to be densely assigned? It looks like a better way could be to check the existence of exprid in the map.

TristonianJones: Using a struct here made a 5x performance difference over maps which was the motivation for the shift.

The code assumes a dense id space. While this is how the parser works now, future changes could result in a more fragmented id space. I've added a TODO to renumber the instructions with runtime identifiers rather than use the ids present in the expression.

interpreter/program.go
TristonianJones: Done, and I split the Init() step from the Begin() step.

Issue #18, opened 2018-04-13
Partial evaluation


TristonianJones:
## Change
Support the evaluation of an expression with partial information. 

### Example
If the expression `a + expensive < 10 || c >= 11` receives values for `a=11`
and `c=10`, but not `expensive` the output of the partial evaluation should
yield:

```
11 + expensive < 10 || false
```

The computed expression can be simplified to: `expensive < -1`. In order to 
support such transformations, the interpreter should generate a stable (error
or non-error) or unknown value with tracking for the state observed for an
expression. The tracked state should be detailed enough to generate a new
expression or minimally detect which inputs were crucial to the computation
of the result.

### Alternatives Considered
CEL could only execute expressions in serial with complete information. This
would simplify the CEL implementation and align with the expectations of
many developers.

PR #24, opened 2018-05-08
Residue computation/constant folding

ayas:
#18 Support for generating a 'residue' expression by constant folding state related
to expression identifiers from a previous evaluation into a new AST. This pull request
also introduces:

* Bug fix for computing the max expression id.
* Bug fix for invocation of `&&`, `||`, and `c ? a : b` with unknown values.
* Helpers to clarify non-value states.

As future work, the `astwalker.go` and the `prune.go` may be refactored into the
following components:

1. Map common sub-expressions
2. Fold constants
3. Prune AST / Program


ayas:
I wanted to checkpoint and have some input on this. I added the basic pruning(prune.go) and added two good todos we should do. So far the tests are "thin". The reason is it looks like we do not handle unknowns in functions.  I added two quick fixes for _&&_ and _||_ so that I can add some tests. These and the ternary are probably the special cases and for the rest, I propose to just return unkown if any of the args are unknown. If we agree, I will do this change and then I will add more tests. I can do these as part of these changes or merge this as is and do a followup. Let me know what you think.

TristonianJones:
There are a couple of small changes for proper unknown support. What you have is a good start. Locally, I had made some added modifications which are helpful here:

1. `instructions.go | type CallExpr` - Add an `IsStrict()` member function which returns true if the `Function` name is neither `operators.LogicalAnd` nor `operators.LogicalOr`. This could just as easily be computed up-front and made a public property.
2. `interpreter.go | func evalCall()` - If the function `IsStrict()` set the result of the function to the first error or unknown argument value (if present). Otherwise, invoke the function.
3. `standard.go` - Add the checks that you've basically added in this pull request. 

At some point we should aggregate unknown and error values, but this will unblock progress.

ayas:
Sounds good. When the changes you mentioned are in master, I will pull them and add some more tests. I believe what we have in prune.go is pretty functional now.

TristonianJones:
Thanks, I'll give prune.go a review.

interpreter/program.go
```diff
@@ -36,7 +36,11 @@ type Program interface {
 	// container is used to resolve type names and identifiers.
 	Container() string
 
-	// GetInstruction returns the instruction at the given expression id.
+	// Return the runtime expression id corresponding to the expression id from
+	// the AST.
+	GetRuntimeExpressionId(exprId int64) int64
+
+	// GetInstruction returns the instruction at the given runtime expression id.
 	GetInstruction(exprId int64) Instruction
```
TristonianJones: Perhaps rename the exprId arg to runtimeId here?

interpreter/program.go
```diff
@@ -73,6 +77,7 @@ type exprProgram struct {
 	instructions    []Instruction
 	metadata        Metadata
 	revInstructions map[int64]int
+	exprIdMappigns  map[int64]int64
```
TristonianJones: nit: s/Mappigns/Map

interpreter/prune.go
```diff
@@ -0,0 +1,126 @@
...
+func ConstantFoldExpr(expr *expr.ParsedExpr, interpreter Interpreter, container string) interface{} {
+	program := NewProgram(expr.Expr, expr.SourceInfo, container)
+	interpretable := interpreter.NewInterpretable(program)
+	result, state := interpretable.Eval(
+		NewActivation(map[string]interface{}{}))
+	// We could prune even in case of error but that is probably not very useful.
+	if !types.IsError(result) {
+		PruneExpr(expr.Expr, program, state)
+	}
+	return result
```
TristonianJones: I think you'll want to preserve the original expression and build a new one since there could be other inputs provided for eval against the expression at a later date.

interpreter/prune.go
```diff
@@ -0,0 +1,126 @@
...
+		}
+	case *expr.Expr_ComprehensionExpr:
+		compre := node.GetComprehensionExpr()
+		p.prune(compre.IterRange)
+	}
+}
+
+func (p *exprPruner) value(id int64) (ref.Value, bool) {
+	if object, found := p.state.Value(id); found {
+		fmt.Printf("index %d value %v \n", id, object)
```
TristonianJones: nit: remove printf here and below.

interpreter/prune.go
```diff
@@ -0,0 +1,126 @@
...
+
+func PruneExpr(expr *expr.Expr, program Program, state EvalState) {
+	pruner := &exprPruner{
+		expr:    expr,
+		program: program,
+		state:   state}
+	pruner.prune(expr)
+}
+
+func ConstantFoldExpr(expr *expr.ParsedExpr, interpreter Interpreter, container string) interface{} {
```
TristonianJones: Maybe provide the EvalState here from a previous run, rather than executing the interpreter in a given container.

TristonianJones (changes requested):
This looks like a promising start. Please refactor with the assumption that EvalState will be provided as input rather than an Interpreter + container.

interpreter/program.go
ayas: I renamed the exprId in GetInstruction to runtimeId

interpreter/program.go
ayas: Done.

interpreter/prune.go
ayas: I think both for PruneExpr and ConstantFoldExpr interfaces, it is better to leave potential cloning to the caller. Cloning might not be necessary or might be necessary in a subset of calls.

interpreter/prune.go
ayas: Done.

interpreter/prune.go
ayas: I listed the typical use case I thought of for PruneExpr and ConstantFoldExpr. I also noticed what we do here is related to caching function call results(especially in case of ConstantFoldExpr). I think it is important to give the users the flexibility to prepare their own runtime environment in constant fold case. They might choose how they provide overloads to enable/disable constant folding of function calls(and effectively enabling/disabling caching function call results). PruneExpr interface is the one that does what you want I think. I thought of that being used in cases where caller prunes tree right after evaluating the expression with some unknown values.

interpreter/instructions.go
```diff
@@ -81,12 +83,19 @@ func (e *CallExpr) String() string {
 		e.GetId())
 }
 
+func checkIsStrict(function string) bool {
```
TristonianJones: nit: please move private helpers after the publicly exported names.

interpreter/program.go
```diff
@@ -36,8 +36,12 @@ type Program interface {
 	// container is used to resolve type names and identifiers.
 	Container() string
 
-	// GetInstruction returns the instruction at the given expression id.
-	GetInstruction(exprId int64) Instruction
+	// Return the runtime expression id corresponding to the expression id from
+	// the AST.
+	GetRuntimeExpressionId(exprId int64) int64
```
TristonianJones: Prefer organizing exported names alphabetically. The code doesn't do this consistently, but it's something I've been working on getting better about doing in recent changes.

interpreter/astwalker.go
```diff
@@ -36,26 +36,28 @@ const (
 func WalkExpr(expression *expr.Expr,
 	metadata Metadata,
 	dispatcher Dispatcher,
-	state MutableEvalState) []Instruction {
+	state MutableEvalState) ([]Instruction, map[int64]int64) {
```
TristonianJones: Just a thought, but would it make more sense to build this mapping in the MutableEvalState? Since this state should be communicable across process boundaries, having more information here beyond a simple map from runtime id to value is probably useful. I believe it would also simplify the prune contract to:
`PruneExpr(expr, state)`

interpreter/prune.go
```diff
@@ -0,0 +1,212 @@
...
+}
+
+// Prunes the given expression based on the given EvalState. Modifies the expression in place.
+// A typical use of this interface would be:
+// 1) Evaluate expr with some unknowns,
+// 2) If result is unknown:
+//   a) Maybe clone the expr(typically would be necessary only for the first iteration) and PruneExpr
+//   b) Goto 1
+// Functional call results which are known would be effectively cached across iterations.
+func PruneExpr(expr *expr.Expr, program Program, state EvalState) {
```
TristonianJones: Taking a step back, the `WalkExpr()` is a special case prune which omits constant values from the generated instruction set. It should be modified to honor previously set state in order to skip unnecessary instruction generation. Right now, `WalkExpr()` does too much. We should probably have a flow that looks as follows:

```go
evalState := NewMutableEvalState()
MapCommonSubexpressions(e *expr.Expr, state MutableEvalState)
FoldConstants(e *expr.Epxr, state MutableEvalState)
program := GenerateInstructions(e *expr.Expr, state EvalState)
```
Optionally a `GenerateAst(e *expr.Expr, state EvalState)` function could be added to output a new AST.

Thoughts?

interpreter/prune.go
```diff
@@ -0,0 +1,212 @@
...
+// A typical use case of this interface would be:
+// 1) Compile the expression (maybe via a service and maybe after checking a compiled expression does not exists in local cache)
+// 2) Prepare the environment and the interpreter
+// 3) ConstantFoldexpr
+// 4) Maybe cache the expression
+//
+// How the environment is prepared in step 2 is flexible. For example, If the caller caches the compiled and constant folded expressions,
+// but is not willing to constant fold( and thus cache results of) some external calls, then he can prepare the overloads accordingly.
+//
+func ConstantFoldExpr(expr *expr.ParsedExpr, interpreter Interpreter, container string) interface{} {
```
TristonianJones: The concern I have with this function is that it seems as too opinionated about how to fold and makes it appear as though it is doing something fundamentally different from the `PruneExpr` which it isn't. That said, I think there is room for improvement in the making the `WalkExpr()` and `PruneExpr()` do less so that they can be better composed together.

TristonianJones (changes requested):

interpreter/instructions.go
ayas: Done.

interpreter/program.go
ayas: Done.

interpreter/astwalker.go
ayas: Done.

interpreter/prune.go
ayas: I agree with the overall thought that generating instructions/a new ast should be driven off from evalstate. I renamed PruneExpr to PruneAst. We can have finding common subexpression decoupled from generating instructions. I left this as a todo. GenerateInstructions could do something versy similar to PruneAst to avoid generating instructions, but I am guessing common use case would be to prune the ast so it can be cached. I removed the ConstantFold interface. As we discussed, users can bring together several tools and do the constant folding their way. I listed a typical uses in PruneAst interface.

interpreter/prune.go
ayas: Removed this interface.

interpreter/prune.go
```diff
@@ -22,60 +22,54 @@ import (
...
-type exprPruner struct {
+type astPruner struct {
 	expr    *expr.Expr
-	program Program
 	state   EvalState
 }
 
-// Prunes the given expression based on the given EvalState. Modifies the expression in place.
-// A typical use of this interface would be:
+// TODO Consider having a separate walk of the AST that finds common subexpressions. This can be called before or after
```
TristonianJones: Note, please keep line length to <= 80 chars

interpreter/prune.go
```diff
@@ -22,60 +22,54 @@ import (
...
+// constant folding to find common subexpressions.
+
+// Prunes the given AST based on the given EvalState and generates a new AST. Modifies the AST in place.
```
TristonianJones: I would really prefer that the implementation not modify the AST in-place. Consider passing an Expr by value and taking a reference to the value before pruning, and then return the pruned result. This should effectively copy on write and allow for the evaluation of many different inputs against the original expression.

interpreter/prune.go
```diff
@@ -188,25 +182,25 @@ func (p *exprPruner) prune(node *expr.Expr) {
...
 	return valueExists && isUnknown(val)
 }
 
-func (p *exprPruner) existsWithKnownValue(id int64) bool {
+func (p *astPruner) existsWithKnownValue(id int64) bool {
 	val, valueExists := p.value(id)
 	return valueExists && !isUnknown(val)
 }
+
+func isUnknown(val ref.Value) bool {
```
TristonianJones: This helper feels a bit redundant.

interpreter/prune.go
```diff
@@ -188,25 +182,25 @@ func (p *exprPruner) prune(node *expr.Expr) {
...
+	return types.IsUnknown(val)
+}
+
```
TristonianJones: Consider adding this function to the 'types' module somewhere. It seems common enough to be generally useful.

TristonianJones (changes requested):

interpreter/prune.go
ayas: Done for all comments.

interpreter/prune.go
ayas: OK. I modified the interface to do copy on write.

interpreter/prune.go
ayas: Done.

interpreter/prune.go
ayas: Moved it types.util.go

common/types/util.go
```diff
@@ -0,0 +1,21 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package types
+
+// Returns whether the input element ref.Type or ref.Value is an ErrType or
```
TristonianJones: nit: go doc comment should indicate the method name first:
`// IsUnknownOrError returns whether ...`

interpreter/prune.go
```diff
@@ -0,0 +1,249 @@
...
+func (p *astPruner) maybePruneAndOr(node *expr.Expr) (bool, *expr.Expr) {
+	if !p.existsWithUnknownValue(node.GetId()) {
+		return false, nil
+	}
+
+	call := node.GetCallExpr()
+
+	// We know result is unknown, so we have at least one unknown arg
+	// and if one side is a known value, we know we can ignore it.
+
```
TristonianJones: nit: extra blank line.

interpreter/prune.go
```diff
@@ -0,0 +1,249 @@
...
+	return newExpr
+}
+
+func (p *astPruner) createLiteral(node *expr.Expr, val *expr.Literal) *expr.Expr {
+	newExpr := *node
+	newExpr.ExprKind = &expr.Expr_LiteralExpr{LiteralExpr: val}
+	return &newExpr
+}
+
+func (p *astPruner) maybePruneAndOr(node *expr.Expr) (bool, *expr.Expr) {
```
TristonianJones: Typically the value is communicated first, and the state second, e.g.
`func (p *astPruner) maybePruneAndOr(...) (*expr.Expr, bool)`

TristonianJones (changes requested):
I have a few minor style requests, but otherwise looks good.

common/types/util.go
ayas: Done

interpreter/prune.go
ayas: Done.

interpreter/prune.go
ayas: Done.

Issue #11, opened 2018-04-05
Support `google.protobuf.Value` types as JSON


TristonianJones:
CEL supports types based on protocol buffers. These types are treated
as object that support field accesses by name. Protocol buffers also
support JSON-like data via `google.protobuf.Value` messages. These
messages are intended to be supported as primitives, maps, and lists
so the message field names should be abstracted away by the
type-checker and interpreter.

PR #26, opened 2018-05-10
Support for protobuf-backed JSON values.

TristonianJones:
#11 Introduce support for protobuf-backed JSON values.

This change adds trivial mappings between JSON primitive and
aggregate values to the relevant CEL primitives (bool, double, string)
and aggregates (list, map).


common/types/double_test.go
```diff
@@ -46,6 +80,9 @@ func TestDouble_ConvertToType(t *testing.T) {
 	if !Double(-4.5).ConvertToType(TypeType).Equal(DoubleType).(Bool) {
 		t.Error("Unsuccessful type conversion to type")
 	}
+	if !IsError(Double(-4.5).ConvertToType(TimestampType)) {
+		t.Error("Unexpected conversion of type double to timestamp.")
```
madi8229: The error message may be "Expected error on conversion of type double to timestamp." since it checks if the conversion returns error.

common/types/json_list_test.go
```diff
@@ -0,0 +1,155 @@
...
+		{&structpb.Value_NumberValue{3}}}})
+	it := list.Iterator()
+	for i := Int(0); it.HasNext() != False; i++ {
+		v := it.Next()
+		if v.Equal(list.Get(i)) != True {
+			t.Errorf("elem[%d] Got '%v', expected '%v'", v, list.Get(i))
+		}
+	}
+
+	if it.HasNext() != False {
```
madi8229, line 149: This condition is the end condition for the previous for loop. If it.HasNext() != False, will the for loop stop and go to this part of the code?

common/types/list_test.go
```diff
@@ -92,58 +266,100 @@ func TestConcatListValue_Iterator(t *testing.T) {
...
 }
 
-func TestListValue_Contains(t *testing.T) {
-	listA := NewDynamicList([]float32{1.0, 2.0}).(traits.Lister)
-	listB := NewDynamicList([]float64{3.0})
-	list := listA.Add(listB).(traits.Lister).Add(listA).(traits.Lister)
-	if list.Contains(Int(3)) == True {
-		t.Error("List contains succeeded with wrong type")
+func TestStringList_Add_Error(t *testing.T) {
+
```
madi8229: Do we need to add  test for this?

common/types/list_test.go
```diff
@@ -92,58 +266,100 @@ func TestConcatListValue_Iterator(t *testing.T) {
...
+		}
+	}
+}
+
+func TestStringList_ConvertToNative(t *testing.T) {
+	list := NewStringList([]string{"h", "e", "l", "p"})
+	_, err := list.ConvertToNative(reflect.TypeOf([]string{}))
+	if err != nil {
+		t.Error("Unable to convert string list to itself.")
+	}
```
madi8229: Do we need to check if the converted content is correct?

common/types/string_test.go
```diff
@@ -33,6 +36,30 @@ func TestString_Compare(t *testing.T) {
...
+	if err != nil ||
+		IsError(val) ||
+		!proto.Equal(val.(proto.Message), pbVal) {
+		t.Error("Error during conversion to json Value type", err, val)
+	}
+}
+
+func TestString_ConvertToNative_String(t *testing.T) {
+	val, err := String("hello").ConvertToNative(reflect.TypeOf(""))
+	if err != nil && val.(string) != "hello" {
```
madi8229: Maybe either of these conditions should be error?

TristonianJones: Done.

common/types/list_test.go
TristonianJones: Done.

common/types/list_test.go
TristonianJones: Done.

common/types/json_list_test.go
TristonianJones: The test case is there to ensure that the HasNext() doesn't loop around from false to true again on subsequent calls.

common/types/double_test.go
TristonianJones: Rephrased as 'Got value, expected error'

madi8229 (approved):

PR #30, opened 2018-05-24
Bug fix for segv following syntax error in parser.

TristonianJones:
Bug fix for SEGV on parse error.

When there is a syntax error in the parser related to an invalid identifier
the `parser.go` will seg fault. Added conditionals to handle these error
scenarios and a test case.

### DIFF
--- a/parser/parser.go
+++ b/parser/parser.go
@@ -35,6 +35,12 @@ import (
 //
 // By default all macros are enabled. For customization of the input data
 // or for customization of the macro set see Parse.
+//
+// Note: syntax errors may produce parse trees of unusual shape which could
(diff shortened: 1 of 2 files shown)


JimLarson (changes requested):
Looks good, but if this isn't a complete solution to the problem of performing all necessary checks on ctx arguments (e.g. sometimes we check ctx == nil, but usually not), the please leave a TODO comment.

TristonianJones:
Syntax errors are relatively well-tested within the parser, but there is still the possibility of edge cases which are not appropriately handled. I've added a note to the parser.go file to indicate what to do if more are encountered.

JimLarson (approved):
LGTM

PR #31, opened 2018-05-24
Fix issue with inverted operands for 'in'

TristonianJones:
The operands to the 'in' operator were flipped causing calls to fail.

Added a fix and test cases.

### DIFF
--- a/interpreter/interpreter_test.go
+++ b/interpreter/interpreter_test.go
@@ -18,6 +18,7 @@ import (
 	"github.com/google/cel-go/common/types"
 	"github.com/google/cel-go/common/types/ref"
 	"github.com/google/cel-go/interpreter/functions"
+	"github.com/google/cel-go/parser"
 	"github.com/google/cel-go/test"
 	expr "github.com/google/cel-spec/proto/v1/syntax"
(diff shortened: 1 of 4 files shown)


interpreter/interpreter_test.go
```diff
@@ -88,6 +89,19 @@ func TestInterpreter_ComprehensionExpr(t *testing.T) {
...
 }
 
+func TestInterpreter_InList(t *testing.T) {
+	parsed, err := parser.ParseText("1 in [1, 2, 3]")
+	if len(err.GetErrors()) != 0 {
+		t.Error(err)
+	}
+	prg := NewProgram(parsed.GetExpr(), parsed.GetSourceInfo(), "")
+	i := interpreter.NewInterpretable(prg)
+	res, _ := i.Eval(NewActivation(map[string]interface{}{}))
```
madi8229, line 99: Do we need to check if there is any error here?

TristonianJones: The eval returns a 'state' object that gives observed expression values during execution, that's what I'm ignoring here.

madi8229 (approved):

PR #32, opened 2018-05-24
Support ref.Value keys in Dynamic Maps

TristonianJones:
The expectation for most dynamic maps is that they are based on some
backing protocol buffer implementation; however, for maps constructed
as part of expression evaluation, the keys are actually already `ref.Value`
instances.

This pull ensures that all types which are suitable as map indices or values
are successfully returned from `ConvertToNative` functions.

### DIFF
--- a/common/types/object.go
+++ b/common/types/object.go
@@ -54,6 +54,10 @@ func (o *protoObj) ConvertToNative(typeDesc reflect.Type) (interface{}, error) {
 	if typeDesc == anyValueType {
 		return ptypes.MarshalAny(o.Value().(proto.Message))
 	}
+	// If the object is already assignable to the desired type return it.
+	if reflect.TypeOf(o).AssignableTo(typeDesc) {
+		return o, nil
+	}
 	return nil, fmt.Errorf("type conversion error from '%v' to '%v'",
 		o.refValue.Type(), typeDesc)
 }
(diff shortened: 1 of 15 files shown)


common/types/object.go
```diff
@@ -54,6 +54,10 @@ func (o *protoObj) ConvertToNative(typeDesc reflect.Type) (interface{}, error) {
 	if typeDesc == anyValueType {
 		return ptypes.MarshalAny(o.Value().(proto.Message))
 	}
+	// If the duration is already assignable to the desired type return it.
```
madi8229: Do you mean "If the object is already assignable to the desired type return it" ?

common/types/string.go
```diff
@@ -60,6 +60,9 @@ func (s String) ConvertToNative(typeDesc reflect.Type) (interface{}, error) {
 			Kind: &structpb.Value_StringValue{
 				StringValue: s.Value().(string)}}, nil
 	}
+	if reflect.TypeOf(s).AssignableTo(typeDesc) {
+		return s, nil
+	}
 	if typeDesc.Kind() != reflect.String {
```
madi8229, line 66: Do wee need to check this after we check the assignableTo for the type? Or we can just throw error for unsupported type?

TristonianJones: It happens that `typeDesc.Kind() == reflect.String` will be `false` when the `typeDesc` refers to an interface (which is what is captured by `AssignableTo()` more or less). So, the check is still valid here, and note that the return value is different when `s` is assignable.

madi8229 (approved):

PR #33, opened 2018-05-24
Emit constant return values with no program steps

TristonianJones:
When a program is generated that contains only constant values,
the state has a single value in it, but the program is empty. This
pull fixes the case where the state would not be inspected to see
if there is a single constant return value for the program at the end
of interpretation.


JimLarson (approved):
LGTM

PR #34, opened 2018-05-24
Export Type declaration helper methods.

TristonianJones:
Export the checker/decls/decls.go type helper methods.

The methods to create types and decls were not exported and so
users were not able to declare non-primitive types within their
checked environment without using the `checked.proto` from
cel-spec directly. This change exposes the relevant helper methods
to improve ease of use.


checker/standard.go
```diff
@@ -23,397 +23,401 @@ import (
...
 			decls.NewParameterizedOverload(overloads.IndexList,
-				[]*checked.Type{listOfA, Int}, paramA,
+				[]*checked.Type{listOfA, decls.Int}, paramA,
 				typeParamAList),
 			decls.NewParameterizedOverload(overloads.IndexMap,
 				[]*checked.Type{mapOfAB, paramA}, paramB,
 				typeParamABList)),
 		//decls.NewOverload(overloads.IndexMessage,
-		//	[]*checked.Type{Dyn, String}, Dyn)),
+		//	[]*checked.Type{Dyn, decls.String}, Dyn)),
```
madi8229: Do we also need to update Dyn to decls.Dyn?
Why do we keep these two commands? Will we use them later?

TristonianJones: I'm not sure. I can delete the commented out lines here. I'm not sure of their purpose at this point.

TristonianJones: On second thought, I recall what this is for and just updated the commented out lines to have valid decls.Dyn references.

madi8229 (approved):
LGTM

PR #38, opened 2018-05-25
Fix unquoting when using single quote versus double.

TristonianJones:
The default `strings.unquote` strategy is to treat single quotes as characters
and double quotes as strings. CEL treats the two as synonymous. This commit
introduces a fallback scenario to handle errors related to single quoting.

### DIFF
--- a/parser/parser.go
+++ b/parser/parser.go
@@ -483,9 +483,10 @@ func (p *parser) extractQualifiedName(e *expr.Expr) (string, bool) {
 }
 
 func (p *parser) unquote(ctx interface{}, value string) string {
-	if text, err := strconv.Unquote(value); err == nil {
-		return text
+	text, err := unescape(value)
+	if err != nil {
+		p.helper.reportError(ctx, err.Error())
+		return value
 	}
-	p.helper.reportError(ctx, "unable to unquote string")
-	return value
+	return text
 }
--- a/parser/unescape.go
+++ b/parser/unescape.go
@@ -1,0 +1,225 @@
+// Copyright 2018 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package parser
+
+import (
+	"fmt"
+	"strings"
+	"unicode/utf8"
+)
+
+// Unescape takes a quoted string, unquotes, and unescapes it.
+//
+// This function performs escaping compatible with GoogleSQL.
+func unescape(value string) (string, error) {
+	// All strings normalize newlines to the \n representation.
+	value = newlineNormalizer.Replace(value)
+	n := len(value)
+
+	// Nothing to unescape / decode.
+	if n < 2 {
+		return value, fmt.Errorf("unable to unescape string")
+	}
+
+	// Raw string preceded by the 'r|R' prefix.
+	isRawLiteral := false
+	if value[0] == 'r' || value[0] == 'R' {
+		value = value[1:]
+		n = len(value)
+		isRawLiteral = true
+	}
+
+	// Quoted string of some form, must have same first and last char.
+	if value[0] != value[n-1] || (value[0] != '"' && value[0] != '\'') {
+		return value, fmt.Errorf("unable to unescape string")
+	}
+
+	// Normalize the multi-line CEL string representation to a standard
+	// Go quoted string.
+	if n >= 6 {
+		if strings.HasPrefix(value, "'''") {
+			if !strings.HasSuffix(value, "'''") {
+				return value, fmt.Errorf("unable to unescape string")
+			}
+			value = "\"" + value[3:n-3] + "\""
+		} else if strings.HasPrefix(value,`"""`) {
+			if !strings.HasSuffix(value, `"""`) {
+				return value, fmt.Errorf("unable to unescape string")
+			}
+			value = "\"" + value[3:n-3] + "\""
+		}
+		n = len(value)
+	}
+	value = value[1 : n-1]
+	// If there is nothing to escape, then return.
+	if isRawLiteral || !strings.ContainsRune(value, '\\') {
+		return value, nil
+	}
+
+	// Otherwise the string contains escape characters.
+	// The following logic is adapted from `strconv/quote.go`
+	var runeTmp [utf8.UTFMax]byte
+	buf := make([]byte, 0, 3*n/2)
+	for len(value) > 0 {
+		c, multibyte, rest, err := unescapeChar(value)
+		if err != nil {
+			return "", err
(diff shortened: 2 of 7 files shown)


parser/parser.go
```diff
@@ -486,6 +486,10 @@ func (p *parser) unquote(ctx interface{}, value string) string {
 	if text, err := strconv.Unquote(value); err == nil {
 		return text
 	}
+	// Handle the unquoting of single quotes.
+	if len(value) >= 2 && value[0] == '\'' && value[len(value)-1] == '\'' {
+		return value[1:len(value)-1]
+	}
```
JimLarson: Add comments for whatever remains to be done for string parsing, e.g. escapes, raw strings, triple-quoted, etc.

Add a unit test in parser_test.

TristonianJones (commented):
I've updated the escaping logic to match that supported by Google SQL. The grammar itself does not yet support all that is in the implementation, so I've added a TODO. 

Note, I think there's a bug in cel-spec with one of the examples. \314\277 (as opposed to (\303\277) decodes to the example Ì¿

parser/unescape_test.go
```diff
@@ -0,0 +1,109 @@
...
+	if err != nil {
+		t.Error(err)
+	}
+	if text != `` {
+		t.Errorf("Got '%v', wanted '%v'", text, ``)
+	}
+}
+
+func TestUnescapeEscapedQuote(t *testing.T) {
+	text, err := unescape(`"\\\""`)
```
JimLarson: Add an explicit comment that the argument the unescape is the four-character string dquote-backslash-dquote-dquote.  (Though this is redundant, strictly speaking, we'll be pushing the corners of the host languages string literals, so the explicit comment will be valuable to the reader and for debugging.)

parser/unescape_test.go
```diff
@@ -0,0 +1,109 @@
...
+	if err != nil {
+		t.Error(err)
+	}
+	if text != `x""x` {
+		t.Errorf("Got '%v', wanted '%v'", text, `x""x`)
+	}
+}
+
+func TestUnescapeMultiOctalSequence(t *testing.T) {
+	text, err := unescape(`"\314\277"`)
```
JimLarson: Add an explicit comment that:
octal 314 -> decimal 204 -> Latin Capital Letter I With Grave
octal 277 -> decimal 191 -> Inverted Question Mark
Also, see fix to langdef.md, example should be octal 303 -> decimal 191 -> &Atilde; to be the same values as the UTF-8 bytes for unicode 255.

parser/unescape.go
```diff
@@ -0,0 +1,215 @@
...
+		value = value[1:]
+		n = len(value)
+		for i := 0; value[0] == value[n-1] && i < 3; i++ {
+			if value[0] != '\'' && value[0] != '"' {
+				break
+			}
+			value = value[1 : n-1]
+			n = len(value)
+		}
+		return value, nil
```
JimLarson: This appears to recognize
r""ABC""
as
ABC
which doesn't match the CEL or Google SQL spec.  It should recognize either one- or three-quote strings.

parser/unescape.go
```diff
@@ -0,0 +1,215 @@
...
+			buf = append(buf, byte(c))
+		} else {
+			n := utf8.EncodeRune(runeTmp[:], c)
+			buf = append(buf, runeTmp[:n]...)
+		}
+	}
+	return string(buf), nil
+}
+
+func unescapeChar(s string) (value rune, multibyte bool, tail string, err error) {
```
JimLarson: Add function comment explaining especially the meaning of the return values.

parser/unescape.go
```diff
@@ -0,0 +1,215 @@
...
+	c := rune(b)
+	switch {
+	case '0' <= c && c <= '9':
+		return c - '0', true
+	case 'a' <= c && c <= 'f':
+		return c - 'a' + 10, true
+	case 'A' <= c && c <= 'F':
+		return c - 'A' + 10, true
+	}
+	return
```
JimLarson: I'd prefer to see the explicit values in the return.

parser/unescape.go
```diff
@@ -0,0 +1,215 @@
...
+		value = '\''
+	case '"':
+		value = '"'
+	case '`':
+		value = '`'
+	case '?':
+		value = '?'
+
+	// 4. Unicode escape sequences, reproduced from `strconv/quote.go`
+	case 'x', 'u', 'U':
```
JimLarson: Google SQL also allows \Xhh, with the same meaning as \xhh.  Make a TODO/bug somewhere to update the CEL langdef.

parser/unescape.go
```diff
@@ -0,0 +1,215 @@
...
+			x, ok := unhex(s[j])
+			if !ok {
+				err = fmt.Errorf("unable to unescape string")
+				return
+			}
+			v = v<<4 | x
+		}
+		s = s[n:]
+		if c == 'x' {
+			// single-byte string, possibly not UTF-8
```
JimLarson: The way I read the Google SQL spec, the hex and octal escapes name Unicode code points for strings, but byte values for bytes.  So the code is okay, but the comment seems wrong.

JimLarson (changes requested):

parser/unescape.go
TristonianJones: There's a TODO listed in the CEL.g4 grammar to make the lexis Google SQL compatible, but it doesn't mention the specific changes. I'll file a GitHub issue with more details. In the mean time I've added 'X' to this block.

parser/unescape.go
TristonianJones: This comment comes from the the `strconv/quote.go` implementation. I'm not entirely sure why it's listed this way, though GoogleSQL does seem to instead always treat this value as a possibly being multibyte, so I've removed the block.

parser/unescape.go
TristonianJones: Sure, happy to.

parser/unescape_test.go
TristonianJones: Done.

parser/unescape_test.go
TristonianJones: Yes, my example changed to match the documented output. I can change it back to 303.

parser/unescape.go
TristonianJones: Done.

parser/unescape.go
TristonianJones: You're quite right. The CEL.g4 grammar only recognizes one or three, I've added a check to ensure odd-quoting.

parser/unescape.go
```diff
@@ -0,0 +1,222 @@
...
+			buf = append(buf, runeTmp[:n]...)
+		}
+	}
+	return string(buf), nil
+}
+
+// unescapeChar takes a string input and returns the following info:
+//
+//   value - the escaped unicode rune at the front of the string.
+//   multibyte - whether the rune value might require multiple bytes to represent.
```
JimLarson, line 94: s/might require/requires/

parser/unescape.go
```diff
@@ -0,0 +1,222 @@
...
+		return c - 'a' + 10, true
+	case 'A' <= c && c <= 'F':
+		return c - 'A' + 10, true
+	}
+	return 0, false
+}
+
+var (
+	newlineNormalizer = strings.NewReplacer("\r\n", "\n", "\r", "\n")
+)
```
JimLarson: No newline at end of file.

JimLarson (approved):

Issue #37, opened 2018-05-25
Type container considered at check-time, but not at interpretation.


TristonianJones:
## Issue Filing Checklist

- [X] There are no issues that match the desired change
- [X] The change is large enough it can't be addressed with a simple [Pull Request](./PULL_REQUEST_TEMPLATE.md) 
- [X] If this is a bug, there are a concrete set of repro steps

## Change
In CEL, type definitions may exist within a specific `container` as indicated on the
last argument of `checker.Check(expr, env, <container>)`. The container is used to resolve
type and enum definitions. At interpretation time, the container is used to resolve enum
names, but not used in conjunction with resolving type names. This leads the checker
to indicate a program is valid, but the interpreter to fail.

### Example
```go
parsed, errors := parser.ParseText("Expr{id: 1}")

provider := types.NewProvider(&expr.Expr{})
env :=  checker.NewStandardEnv(errors, provider)
checked := checker.Check(parsed, env, "google.api.expr.v1")

i := interpreter.NewStandardInterpreter(provider)
eval := i.NewInterpretable(interpreter.NewCheckedProgram(checked, "google.api.expr.v1"))
result, _ := eval.Eval(interpreter.NewActivation(map[string]interface{}{}))
// yields: unknown type 'Expr'
// using fully-qualified names avoids the issue.
```

PR #40, opened 2018-05-25
Fix qualified name resolution.

TristonianJones:
#37 Fix qualified name resolution

This change unifies the name resolution model between the checker and interpreter
in particular:

* `packager.go` bundles up container related features with C++ name resolution
    ordering.
* `interpreter.go` inspects candidate names in resolution order when trying to
   determine whether an unknown qualified name actually exists as a qualified name,
   and ensures that type name references are appropriately resolved before object
   construction.

This change refines the contract between the checker and interpreter, and may be
further refined by ensuring that AST walking for program generation can effeciently
identify qualified enum names which have not gone through type-checking.


interpreter/interpreter.go
```diff
@@ -189,16 +195,19 @@ func (i *exprInterpretable) resolveUnknown(unknown types.Unknown,
...
-			if i.program.Container() != "" {
-				identifier = i.program.Container() + "." + identifier
-			}
-			if object, found := currActivation.ResolveName(identifier); found {
-				i.setValue(selExpr.Id, object)
-			} else if identVal, found := i.interpreter.typeProvider.FindIdent(identifier); found {
-				i.setValue(selExpr.Id, identVal)
-			} else {
-				i.setValue(selExpr.Id, append(types.Unknown{selExpr.Id}, unknown...))
+			pkg := i.interpreter.packager
```
madi8229: I think if use less indents, the code may looks easier to read. For example:
if !validIdent {
   return
}
pkg := i.interpreter.packager
....

madi8229: Also in line 175:
if object, found := currActivation.ResolveReference(selExpr.Id); found {
 		i.setValue(selExpr.Id, object)
                 return
}
validIdent := true
...

madi8229 (approved):
LGTM, some minor comments.

Issue #35, opened 2018-05-25
ConvertToNative does not handle proto2 primitives as pointers.


TristonianJones:
## Issue Filing Checklist

- [X] There are no issues that match the desired change
- [X] The change is large enough it can't be addressed with a simple [Pull Request](./PULL_REQUEST_TEMPLATE.md) 
- [X] If this is a bug, there are a concrete set of repro steps

## Change
For proto3 generated structs, primitive values are set by value. For proto2, primitive values
are set by pointer. `ConvertToNative` needs to handle the change from value to pointer for
primitive values.

### Example
```go
if typeDesc.Kind() == reflect.Ptr && typeDesc.Elem().Kind() == reflect.String {
  return &string(s), nil
}
```

PR #42, opened 2018-06-04
Support conversion of primitive values to pointer types.

TristonianJones:
#35 Support conversion to pointer type for primitive values.

During proto2 message construction, primitive fields are set by pointer rather than
by value (as they are in proto3). This pull request ensures that primitive values can be
converted to pointer references.

### DIFF
--- a/common/types/bool_test.go
+++ b/common/types/bool_test.go
@@ -42,8 +42,10 @@ func TestBool_Compare(t *testing.T) {
 func TestBool_ConvertToNative_Bool(t *testing.T) {
 	refType := reflect.TypeOf(true)
 	val, err := True.ConvertToNative(refType)
-	if err != nil || IsError(val) || !val.(bool) {
-		t.Error("Error during conversion to bool", err, val)
+	if err != nil {
+		t.Error(err)
+	} else if !val.(bool) {
+		t.Error("Error during conversion to bool", val)
 	}
 }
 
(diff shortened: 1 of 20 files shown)


common/types/bool_test.go
```diff
@@ -65,6 +65,15 @@ func TestBool_ConvertToNative_Json(t *testing.T) {
 	}
 }
 
+func TestBool_ConvertToNative_Ptr(t *testing.T) {
+	ptrType := true
+	refType := reflect.TypeOf(&ptrType)
+	val, err := True.ConvertToNative(refType)
+	if err != nil || IsError(val) || !*val.(*bool) {
```
madi8229: I found only bool checks IsError(val). Does other types need to also check IsError(val)?

common/types/double_test.go
```diff
@@ -78,6 +78,22 @@ func TestDouble_ConvertToNative_Json(t *testing.T) {
 	}
 }
 
+func TestDouble_ConvertToNative_Ptr_Float32(t *testing.T) {
+	ptrType := float32(0)
+	val, err := Double(3.1415).ConvertToNative(reflect.TypeOf(&ptrType))
+	if err != nil || *val.(*float32) != 3.1415 {
+		t.Errorf("Got '%v', wanted 3.1415", val)
```
madi8229: err may need to be logged when err != nil

common/types/bool_test.go
TristonianJones: Ah, no, that was leftover from a previous test case that I'd copied. Removed.

madi8229 (approved):
LGTM.

PR #48, opened 2018-06-25
Use descpb package name instead of default import name.

sefk:
We define a Package Name when importing descriptor, but don't use it
below. It works fine here, but causes problems when we re-import this
library back into Google, where we have to use a different import path.

Tested: none.


common/types/pb/pb.go
```diff
@@ -41,7 +41,7 @@ func DescribeFile(message proto.Message) (*FileDescription, error) {
 	if fd, found := revFileDescriptorMap[proto.MessageName(message)]; found {
 		return fd, nil
 	}
-	fileDesc, _ := descriptor.ForMessage(message.(descriptor.Message))
+	fileDesc, _ := descpb.ForMessage(message.(descpb.Message))
```
TristonianJones, line 44: There are issues in the Travis build related to an unused import for 'descriptor', so you'll have to add the alias 'descpb' to that import statement. The default go import strategy is to use the simple name of the last directory in the import which explains the error in the logs.

sefk:
Whoops, didn't intend this to be reviewed @TristonianJones. I just wanted to see if this would result in build issues. Sorry about that!

PR #49, opened 2018-07-03
Implement CelService server.

JimLarson:
Use Go proto libraries exported from cel-spec.

Note that we use a different Go package for the google.rpc.Status message
than is specified in its .proto file. We do this to avoid a different
and incompatible Go proto library that is brought in by the Bazel rules.
We will revert to the right package when this bug is fixed.
See bazelbuild/rules_go#1530

Expand common.Source to subsume interpreter.Metadata, for the latter's
eventual deprecation, and to allow the checker inputs to be constructed
without access to the source text.

Introduce a specialized traits.Lister for arrays of ref.Value values.

Implement CelService server and tests. Includes conversion between
ref.Value types and protos.


JimLarson:
Detailed commit message:

Use Go proto libraries exported from cel-spec.

Note that we use a different Go package for the google.rpc.Status message
than is specified in its .proto file.  We do this to avoid a different
and incompatible Go proto library that is brought in by the Bazel rules.
We will revert to the right package when this bug is fixed.
See https://github.com/bazelbuild/rules_go/issues/1530

Expand common.Source to subsume interpreter.Metadata, for the latter's
eventual deprecation, and to allow the checker inputs to be constructed
without access to the source text.

Introduce a specialized traits.Lister for arrays of ref.Value values.

Implement CelService server and tests.  Includes conversion between
ref.Value types and protos.

JimLarson:
I thought I was up-to-date, but there are some merge conflicts.  Let me rebase.  I'll comment when I've finished the merge.

JimLarson:
Merged.

common/types/list_test.go
```diff
@@ -97,8 +113,7 @@ func TestBaseList_Get(t *testing.T) {
 	}
 }
 
-func TestBaseList_Iterator(t *testing.T) {
-	list := NewDynamicList([]int32{1, 2, 3})
+func validateIterator123(t *testing.T, list traits.Lister) {
 	it := list.Iterator()
```
TristonianJones: Make sure to mark this function as a helper by stating `t.Helper()` as the first line.

common/source.go
```diff
@@ -16,8 +16,13 @@ package common
 
 import (
 	"strings"
+
+	"github.com/google/cel-spec/proto/v1/syntax"
 )
 
+// TODO(jimlarson) "Character offsets" should index the code points
+// within the UTF-8 encoded string.  It currently indexes bytes.
```
TristonianJones: This is more of an issue with the implementation rather than the interface though, correct?

The implementation should probably refer to `contents []rune` rather than a `contents string`, but making that change and all related type conversions should probably do the trick.

common/source.go
```diff
@@ -65,39 +80,50 @@ func NewStringSource(contents string, description string) Source {
...
 }
 
-func (s *StringSource) LocationFromOffset(offset int32) (Location, bool) {
+func (s *sourceImpl) OffsetLocation(offset int32) (Location, bool) {
 	line, lineOffset := s.findLine(offset)
 	return NewLocation(int(line), int(offset-lineOffset)), true
 }
 
-func (s *StringSource) Snippet(line int) (string, bool) {
-	if charStart, found := s.findLineOffset(line); found {
```
TristonianJones, line 100: Prefer instead to early return when the content is empty:

```
if len(s.contents) == 0 {
    return "", false
}
```

server/main.go
```diff
@@ -0,0 +1,36 @@
...
+
+func main() {
+	log.Println("Server opening listening port")
+	lis, err := net.Listen("tcp", "127.0.0.1:")
+	if err != nil {
+		log.Fatalf("failed to listen: %v", err)
+	}
+	log.Println("Server opened port ", lis.Addr())
+
+	fmt.Printf("Listening on %v\n", lis.Addr().String())
```
TristonianJones: s/fmt/log

TristonianJones: The `lis.Addr()` value implements the `fmt.Stringer` interface so there shouldn't be any need to expressly call `.String()` on it.

server/server.go
```diff
@@ -0,0 +1,366 @@
...
+	"github.com/golang/protobuf/ptypes"
+	"github.com/google/cel-go/checker"
+	"github.com/google/cel-go/common"
+	"github.com/google/cel-go/common/packages"
+	"github.com/google/cel-go/common/types"
+	"github.com/google/cel-go/common/types/ref"
+	"github.com/google/cel-go/common/types/traits"
+	"github.com/google/cel-go/interpreter"
+	"github.com/google/cel-go/parser"
+	cs "github.com/google/cel-spec/proto/v1/cel_service"
```
TristonianJones: I'm not a stickler about it, but the common convention for aliasing a proto package is to suffix the alias with pb, as in `cspb`

server/server.go
```diff
@@ -0,0 +1,366 @@
...
+	var macs parser.Macros
+	if in.DisableMacros {
+		macs = parser.NoMacros
+	} else {
+		macs = parser.AllMacros
+	}
+	expr, errs := parser.Parse(src, macs)
+	resp := cs.ParseResponse{ParsedExpr: expr}
+	// Parse() doesn't distinguish the serverity of the errors
+	// it returns.  Heuristic:  if a ParsedExpr was returned,
```
TristonianJones: I like the heuristic, but any error reported by the parse step should be treated as an error even if there was a `ParsedExpr` returned since the parser will inject invalid expression nodes into the AST to ensure that parsing catching as many errors as possible rather than simply halting on the first one.

server/server.go
```diff
@@ -0,0 +1,366 @@
...
+	}
+	if in.ParsedExpr.SourceInfo == nil {
+		st := status.New(codes.InvalidArgument, "No source info.")
+		return nil, st.Err()
+	}
+	typeProvider := types.NewProvider()
+	errs := common.NewErrors(common.NewInfoSource(in.ParsedExpr.SourceInfo))
+	var env *checker.Env
+	if in.NoStdEnv {
+		env = checker.NewEnv(packages.DefaultPackage, typeProvider, errs)
```
TristonianJones: Is there a container argument provided on the CheckRequest? If so that should replace the `packages.DefaultPackage` value.

server/server.go
```diff
@@ -0,0 +1,366 @@
...
+	for _, e := range errs.GetErrors() {
+		resp.Issues = append(resp.Issues, ErrToStatus(e, severity))
+	}
+	return &resp, nil
+}
+
+func (s *CelServer) Eval(ctx context.Context, in *cs.EvalRequest) (*cs.EvalResponse, error) {
+	typeProvider := types.NewProvider()
+	// REVIEWER: should package be computed from in.Container?
+	i := interpreter.NewStandardIntepreter(packages.DefaultPackage, typeProvider)
```
TristonianJones: Ah, yes. See my earlier comment about using `in.Container` for the `Check()` as well.

server/server.go
```diff
@@ -0,0 +1,366 @@
...
+	case *value.Value_Uint64Value:
+		return types.Uint(v.GetUint64Value()), nil
+	case *value.Value_DoubleValue:
+		return types.Double(v.GetDoubleValue()), nil
+	case *value.Value_StringValue:
+		return types.String(v.GetStringValue()), nil
+	case *value.Value_BytesValue:
+		return types.Bytes(v.GetBytesValue()), nil
+	case *value.Value_ObjectValue:
+		any := v.GetObjectValue()
```
TristonianJones, line 317: I'm pretty certain that the `types.NativeToValue` call should work out a little better here since it handles umarshalling and proto-specializations properly. In fact, I'm not opposed to adding this whole set of conversion logic into the `common/types/provider.go` file where the function is declared. It would mean you wouldn't have to specialize handling on maps / lists.

server/server.go
```diff
@@ -0,0 +1,366 @@
...
+			basicType = value.TypeValue_LIST_TYPE
+		case "map":
+			basicType = value.TypeValue_MAP_TYPE
+		case "string":
+			basicType = value.TypeValue_STRING_TYPE
+		case "type":
+			basicType = value.TypeValue_TYPE_TYPE
+		case "uint":
+			basicType = value.TypeValue_UINT_TYPE
+		}
```
TristonianJones: Consider making a constant map for the above.

server/server_test.go
```diff
@@ -0,0 +1,445 @@
...
+		globals.conn = nil
+	}
+	if globals.cmd != nil {
+		globals.cmd.Process.Kill()
+		globals.cmd.Wait()
+		globals.cmd = nil
+	}
+}
+
+var (
```
TristonianJones, line 96: I've tried to create a series of ASTs for testing purposes in the `test/expr.go` file. Consider adding this expected value there? Also please use the helpers in that file to build the AST nodes since it makes it a bit easier to read.

TristonianJones (changes requested):
I haven't gone through the contents of the test cases yet, but the rest makes sense. I've added some feedback to get you started.

common/types/list_test.go
JimLarson: Done.

common/source.go
JimLarson: True, it's about the implementation.  Moved the comment down and included your thoughts.

common/source.go
JimLarson: Inverted the test to give a more linear flow.

server/main.go
JimLarson: Nope - this needs to go to stdout, per the documented protocol.  I'll add a comment to make it clear.

JimLarson: Done.

server/server.go
JimLarson: I'm all about being conventional.  It makes people lower their guard.  Fixed.

server/server.go
JimLarson: Done - in Check and Eval.

server/server.go
JimLarson: Done.

server/server.go
JimLarson: Done, and for the reverse conversion.

server/server_test.go
JimLarson: Those helpers make things much nicer, don't they? :-)  After using the helper, this giant parsed expression is pretty trivial and not worthy of reuse outside this module, IMHO.

server/server.go
JimLarson: Added a TODO.

server/server.go
JimLarson: Per our conversation:
- detecting success/failure by emptiness of errs.GetErrors() instead of nil-ness of proto
- *not* returning proto in case of failure

JimLarson:
PTAL

common/source.go
```diff
@@ -65,49 +81,82 @@ func NewStringSource(contents string, description string) Source {
...
+	}
+	return s.contents[charStart:], true
+}
+
+func (s *sourceImpl) IdOffset(exprId int64) (int32, bool) {
+	offset, found := s.idOffsets[exprId]
+	if !found {
+		return -1, false
+	}
+	return offset, true
```
TristonianJones: It's up to you whether you want to do this or not, but writing the following would make the `offset` and `found` variables scoped to the condition rather than visible within the whole block. It's debatable whether this is better, and I have no preference, but I thought I'd present this as an alternative to consider in the future.

```go
if offset, found := s.idOffsets[exprId]; found {
  return offset, true
}
return -1, false
```

common/source.go
```diff
@@ -59,7 +56,7 @@ type Source interface {
 	IdLocation(exprId int64) (Location, bool)
 }
 
-// Ensure the StringSource implements the Source interface.
+// Ensure the sourceImpl implements the Source interface.
 var _ Source = &sourceImpl{}
```
TristonianJones: You're welcome to remove this line in its entirety as it's accounted for by the `func NewStringSource` signature below.

server/main.go
```diff
@@ -20,7 +20,8 @@ func main() {
 	}
 	log.Println("Server opened port ", lis.Addr())
 
-	fmt.Printf("Listening on %v\n", lis.Addr().String())
+	// Must print to stdout, so the client can find the port.
+	fmt.Printf("Listening on %v\n", lis.Addr())
```
TristonianJones: Note, you probably want to use `log` rather than `fmt` here.

server/server.go
```diff
@@ -38,23 +38,21 @@ func (s *CelServer) Parse(ctx context.Context, in *cs.ParseRequest) (*cs.ParseRe
...
 	} else {
-		severity = cs.StatusDetails_WARNING
-	}
-	for _, e := range errs.GetErrors() {
-		resp.Issues = append(resp.Issues, ErrToStatus(e, severity))
+		// Failure
+		for _, e := range errs.GetErrors() {
+			status := ErrToStatus(e, cspb.StatusDetails_ERROR)
+			resp.Issues = append(resp.Issues, status)
+		}
```
TristonianJones: Please make this a helper function to be shared between service methods.

server/server.go
```diff
@@ -63,40 +61,41 @@ func (s *CelServer) Check(ctx context.Context, in *cs.CheckRequest) (*cs.CheckRe
 		st := status.New(codes.InvalidArgument, "No source info.")
 		return nil, st.Err()
 	}
+	pkg := packages.NewPackage(in.Container)
 	typeProvider := types.NewProvider()
 	errs := common.NewErrors(common.NewInfoSource(in.ParsedExpr.SourceInfo))
```
TristonianJones: This is correct, but makes me wonder if being able to send the string content which generated the parsed expression to the `Check` would make for more consistent error messaging between parse / check. Can be listed as a TODO if you think it's worthwhile.

server/server.go
```diff
@@ -0,0 +1,357 @@
...
+		value.TypeValue_LIST_TYPE:   types.ListType,
+	}
+)
+
+// Convert res, which must not be error or unknown, to a Value proto.
+func RefValueToValue(res ref.Value) (*value.Value, error) {
+	switch res.Type() {
+	case types.BoolType:
+		return &value.Value{
+			Kind: &value.Value_BoolValue{res.Value().(bool)}}, nil
```
TristonianJones, line 199: When the type is a primitive, it is directly convertible to the go-type, e.g. `res.(bool)` is the same as `res.Value().(bool)`. The same is true for `Size()`

server/server.go
```diff
@@ -0,0 +1,357 @@
...
+		return &value.Value{
+			Kind: &value.Value_BytesValue{res.Value().([]byte)}}, nil
+	case types.DoubleType:
+		return &value.Value{
+			Kind: &value.Value_DoubleValue{res.Value().(float64)}}, nil
+	case types.IntType:
+		return &value.Value{
+			Kind: &value.Value_Int64Value{res.Value().(int64)}}, nil
+	case types.ListType:
+		l := res.Value().(traits.Lister)
```
TristonianJones: For `types.ListType` values, the expression `res.(traits.Lister)` is probably what you want. The `res.Value()` may or may not implement the `traits.Lister` interface.

server/server.go
```diff
@@ -0,0 +1,357 @@
...
+		sz := l.Size().Value().(int64)
+		elts := make([]*value.Value, sz)
+		for i := int64(0); i < sz; i++ {
```
TristonianJones: You can also write this as:
```go
elts := make([]*value.Value, l.Size().(int64))
for i := types.Int(0); i < l.Size(); i++ {
   v, err := RefValueToValue(l.Get(i))
   //...
}
```

server/server.go
```diff
@@ -0,0 +1,357 @@
...
+			if err != nil {
+				return nil, err
+			}
+			elts = append(elts, v)
+		}
+		return &value.Value{
+			Kind: &value.Value_ListValue{
+				&value.ListValue{Values: elts}}}, nil
+	case types.MapType:
+		mapper := res.Value().(traits.Mapper)
```
TristonianJones: Here, I think you want the `res.(traits.Mapper)` instead of `res.Value().(traits.Mapper)`

server/server.go
```diff
@@ -0,0 +1,357 @@
...
+			Kind: &value.Value_MapValue{
+				&value.MapValue{Entries: entries}}}, nil
+	case types.NullType:
+		return &value.Value{
+			Kind: &value.Value_NullValue{}}, nil
+	case types.StringType:
+		return &value.Value{
+			Kind: &value.Value_StringValue{res.Value().(string)}}, nil
+	case types.TypeType:
+		typeName := res.Value().(string)
```
TristonianJones: This works, although the the preferred method for accessing the type name is:

```go
res.(types.Type).TypeName()
```

server/server.go
```diff
@@ -0,0 +1,357 @@
...
+		basicType, ok := typeNameToBasicType[typeName]
```
TristonianJones: It would also be acceptable to use `types.Type` as the key in your map, though the string `typeName` is fine. Prefer using `found` when dealing with map lookups and `ok` when dealing with checking a type-assertion.

server/server.go
```diff
@@ -0,0 +1,357 @@
...
+			tv = &value.TypeValue{
+				DesignatorKind: &value.TypeValue_ObjectType{typeName}}
+		}
+		return &value.Value{Kind: &value.Value_TypeValue{tv}}, nil
+	case types.UintType:
+		return &value.Value{
+			Kind: &value.Value_Uint64Value{res.Value().(uint64)}}, nil
+	default:
+		// Object type
+		any, err := ptypes.MarshalAny(res.Value().(proto.Message))
```
TristonianJones: Since this is the default case, make sure to check that the `res.Value()` actually is a proto, and error if not:

```go
pb, ok := res.Value().(proto.Message)
if !ok {
  return nil, status.New(codes.InvalidArg ...)
}
```

server/server.go
```diff
@@ -0,0 +1,357 @@
...
+}
+
+func ExprValueToRefValue(ev *eval.ExprValue) (ref.Value, error) {
+	switch ev.Kind.(type) {
+	case *eval.ExprValue_Value:
+		v := ev.GetValue()
+		return ValueToRefValue(v)
+	case *eval.ExprValue_Error:
+		// Must reduce error set to a single string
+		return types.NewErr("XXX add details later"), nil
```
TristonianJones: Could you add more detail here about what the change set needs to be? Seems like the incoming error details could be provided as the `types.NewErr(<details>)` argument, but perhaps I'm misreading this.

server/server.go
```diff
@@ -0,0 +1,357 @@
...
+	case *eval.ExprValue_Unknown:
+		u := ev.GetUnknown()
+		return types.Unknown(u.Exprs), nil
```
TristonianJones: Prefer to inline the argument value here.

server/server.go
```diff
@@ -0,0 +1,357 @@
...
+			Kind: &value.Value_ObjectValue{any}}, nil
+	}
+	return nil, status.New(codes.InvalidArgument, "unknown ref.Value type").Err()
+}
+
+func ExprValueToRefValue(ev *eval.ExprValue) (ref.Value, error) {
+	switch ev.Kind.(type) {
+	case *eval.ExprValue_Value:
+		v := ev.GetValue()
+		return ValueToRefValue(v)
```
TristonianJones: Prefer to inline the argument value here:

```go
return ValueToRefValue(ev.GetValue())
```

server/server_test.go
TristonianJones: Fair enough. :)

server/server_test.go
```diff
@@ -0,0 +1,422 @@
...
+	pres, err := globals.client.Parse(context.Background(), &preq)
+	if err != nil {
+		t.Fatal(err)
+	}
+	parsedExpr := pres.ParsedExpr
+	if parsedExpr == nil {
+		t.Fatal("Empty parsed expression")
+	}
+
+	intDecl := checked.Decl_Ident{
```
TristonianJones: Consider using `checker/decls/decls.go` helper routines to simplify the proto construction here.

common/source.go
JimLarson: On such short functions, I'm not worried about the scope.  I do like shortening by a line, and being consistent with the rest of the file, so switching here and nearby.

common/source.go
JimLarson: Done.

server/main.go
JimLarson: Nope - this is part of the protocol so the invoker can find the port by parsing stdout.  I've strengthened the comment even further.  If the comment is still insufficient, I will append ", Tristan" to it.

server/server.go
JimLarson: Done.

server/server.go
JimLarson: Hmm, given source info, we can already map expression IDs to line/column coordinates.  Not marking as a TODO right now since it's outside the scope of this change.  Feel free to open this as an issue in cel-spec.

server/server.go
JimLarson: The compiler says:
 impossible type assertion:
	bool does not implement ref.Value (missing ConvertToNative method)

Leaving as-is per our conversation.

server/server.go
JimLarson: Done.

server/server.go
JimLarson: Compiler says
 impossible type assertion:
	int64 does not implement ref.Value (missing ConvertToNative method)

Cleaned up, per our conversation.

server/server.go
JimLarson: Done.

server/server.go
JimLarson: I think you meant ref.Type.  Done.

server/server.go
JimLarson: I think you meant ref.Type.  It's not clear to me what the keys would be then, so I'm leaving as string for now.  Switched "ok" to "found" and used the if-statement scoping idiom.

server/server.go
JimLarson: Done.

server/server.go
JimLarson: There are repeated Status messages, each with a Status message string and repeated details.  We'll eventually want to round-trip these errors through, so evaluating "x" with "x" bound to errors {E1, E2, E3} should yield the same ExprValue on output.  Added the details.

server/server.go
JimLarson: Done, here and above.

server/server.go
JimLarson: Done.

server/server_test.go
JimLarson: Done.  Pretty!

JimLarson:
PTAL

server/server.go
```diff
@@ -119,9 +113,39 @@ func (s *CelServer) Eval(ctx context.Context, in *cspb.EvalRequest) (*cspb.EvalR
...
+	if err == nil {
+		s.Details = append(s.Details, any)
+	}
+	return &s
+}
+
 // TODO(jimlarson): The following conversion code should be moved to
 // common/types/provider.go and consolidated/refactored as appropriate.
-// In partucular, make judicious use of types.NativeToValue.
+// In partucular, make judicious use of types.NativeToValue().
```
TristonianJones: nit: s/partucular/particular

TristonianJones (approved):

server/server.go
JimLarson: Done.

Issue #52, opened 2018-08-07
Fix go lint issues


madi8229:
## Change
Update code to fix lint issues.

PR #61, opened 2018-09-10
replace checked to checkedpb

madi8229:
Fix go lint issues #52
check point: replace checked to checkedpb for lint.


TristonianJones (approved):

madi8229:
Refactor proto alias #67

PR #63, opened 2018-09-12
rename server_bin to cel_server

madi8229:
rename go_binary of server to the same its executable's name.


madi8229:
Fix go lint issues #52

JimLarson (approved):
LGTM

PR #65, opened 2018-09-14
refactor proto alias for duration

madi8229:
Fix go lint issues #52
rename duration proto alias to dpb


TristonianJones (approved):

madi8229:
Refactor proto alias #67

PR #70, opened 2018-09-21
Server test

madi8229:
Update cel-service #69 

Also added tag 'rundir' to the go_test target of server_test. The test is set to execute under the root of the repository, which the test can call the cel_server binary correctly with the location argument from the command line.


madi8229:
Add tag 'rundir' to the go_test target of server_test. The test now run under the root of the repository, which the test can call the cel_server binary correctly with the location argument from the command line.

JimLarson (approved):
LGTM

PR #72, opened 2018-09-26
Import v1alpha1 protos from googleapis.

JimLarson:
Use generated Go libraries for the official google.api.expr.v1alpha protos.
See google/api/expr/v1alpha1 in https://github.com/googleapis/googleapis.
Also use the generated Go library for the google.rpc.status proto.
Besides changing workspace and package imports, adapt to material changes:
- "Literal" reverted to "Constant" in v1alpha1 for compatibility
  (but keep refering to these as literals);
- Type values are represented by string rather than string-or-enum;
- "StatusDetails" now "IssueDetails" using common SourcePosition.


madi8229 (approved):
LGTM

PR #104, opened 2018-10-15
Add mandatory godoc comments to satisfy linter.

### LINKED ISSUE #54: Missing reference documentation (Godoc)
Here's a list of important functions and types that don't have documentation, and probably should:

* [`common.Types.NativeToValue`](https://godoc.org/github.com/google/cel-go/common/types#NativeToValue)
* [`packages`](https://godoc.org/github.com/google/cel-go/common/packages) (the whole package, really. what is it for? an example would be good.)
* [`checker.NewStandardEnv`](https://godoc.org/github.com/google/cel-go/checker#NewStandardEnv)
* [`checker.StandardDeclarations`](https://godoc.org/github.com/google/cel-go/checker#StandardDeclarations)
* [`types.NativeToValue`](https://godoc.org/github.com/google/cel-go/common/types#NativeToValue)

JimLarson:
closes #54

### DIFF
--- a/checker/env.go
+++ b/checker/env.go
@@ -24,6 +24,9 @@ import (
 	expr "google.golang.org/genproto/googleapis/api/expr/v1alpha1"
 )
 
+// Env is the environment for type checking.
+// It consists of a Packager, a Type Provider, declarations,
+// and collection of errors encountered during checking.
 type Env struct {
 	errors       *typeErrors
 	packager     packages.Packager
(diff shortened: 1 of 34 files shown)


checker/env.go
```diff
@@ -65,6 +73,9 @@ func (e *Env) Add(decls ...*expr.Decl) {
 	}
 }
 
+// addOverload adds overload to function declaration f.
```
TristonianJones, line 76: For future reference only the functions that start with a capital letter need a comment.

TristonianJones (approved):

checker/env.go
JimLarson: Right - but comments aren't prohibited for private functions either.  In this case, the error behavior was non-obvious, to me at least, so I captured it in documentation.

PR #105, opened 2018-10-30
use correct constant for defining matches() std fn

hnovikov:
The standard interpreter environment is missing the `matches(str, regexp)` function because the overload name is being used as a function name. This change could break code using `matches_string`


TristonianJones:
LGTM

PR #108, opened 2018-11-19
Go lint fixes for core packages.

TristonianJones:
Address minor variable name and Go style issues identified by `go lint` for the
`checker`, `common`, `parser`, and `interpreter` packages.

### DIFF
--- a/interpreter/evalstate.go
+++ b/interpreter/evalstate.go
@@ -20,9 +20,9 @@ import (
 
 // EvalState tracks the values associated with expression ids during execution.
 type EvalState interface {
-	// GetRuntimeExpressionId returns the runtime id corresponding to the
+	// GetRuntimeExpressionID returns the runtime id corresponding to the
 	// expression id from the AST.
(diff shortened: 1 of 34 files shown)


interpreter/evalstate.go
```diff
@@ -36,35 +36,35 @@ type EvalState interface {
 type MutableEvalState interface {
 	EvalState
 
-	// SetRuntimeExpressionId sets the runtimeId for the given exprId.
-	SetRuntimeExpressionId(exprId int64, runtimeId int64)
+	// SetRuntimeExpressionID sets the runtime id for the given expr id.
```
madi8229, line 39: Do we want to capitalize 'id' here as well?

madi8229 (approved):
LGTM. Just one minor comment.

interpreter/evalstate.go
TristonianJones: Yes, good catch.

Issue #66, opened 2018-09-17
Update CEL.g4 to better handle arithmetic computations


TristonianJones:
## Issue Filing Checklist

- [X] There are no issues that match the desired change
- [X] The change is large enough it can't be addressed with a simple [Pull Request(./PULL_REQUEST_TEMPLATE.md) 

## Change
The lexing of negative numbers makes the parsing of arithmetic expressions
less than ideal. Moving the negative number detection to the parser will address
this usability issue.

### Example
The expression `4--4` is currently ambiguous without additional spaces. This should
read as `call('-', '4', '-4')` once the parser is changed.

PR #109, opened 2018-11-19
Fix arithmetic parsing for negative number sequences

TristonianJones:
#66 Negative number detection moved from lexer to parser to better handle arithmetic operations
such as `4--4` which should parse as `4 - -4`.

### DIFF
--- a/parser/gen/CEL.g4
+++ b/parser/gen/CEL.g4
@@ -45,16 +45,16 @@ calc
     ;
 
 unary
-    : statement                                                     # StatementExpr
-    | (ops+='!')+ statement                                         # LogicalNot
-    | (ops+='-')+ statement                                         # Negate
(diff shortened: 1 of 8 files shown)


parser/gen/CEL.g4
```diff
@@ -78,9 +78,9 @@ mapInitializerList
     ;
 
 literal
-    : tok=NUM_INT   # Int
+    : sign=MINUS? tok=NUM_INT   # Int
```
JimLarson, line 81: Remember to update the lexis in cel-spec.

JimLarson (approved):

parser/gen/CEL.g4
TristonianJones: Thanks, will do!

PR #116, opened 2018-11-29
Use builtin "context" now that it is standard Go.

sefk:
This was taken from a Google-internal large scale change, making it to
the upstream repo instead. The originating change said:

The "context" library added in Go 1.7 was based on Google's and has
the same API. It is now aliased to "context" so the two can be used
interchangeably. This change is a cleanup that standardizes on the
standard library "context" package.

### DIFF
--- a/server/server_test.go
+++ b/server/server_test.go
@@ -1,6 +1,7 @@
 package server
 
 import (
+	"context"
 	"fmt"
 	"log"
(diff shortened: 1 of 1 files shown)


sefk:
Googlers may refer to cl/223210526.

TristonianJones (approved):

PR #117, opened 2018-11-29
NewExhaustiveProgram to evaluate without short-circuiting

jiyunyao:
Standard evaluation will short-circuit conditionals and certain macro comprehensions,
but with this is not always desired. When debugging why a certain expression evaluates
the way it does, the short-cirtcuit is an obstacle to iteratively overcome. This feature
disables short-circuits for the majority of the code paths, but does not yet cover comprehension
macros.

This PR enables #121 for all non-comprehension paths.

### DIFF
--- a/interpreter/astwalker.go
+++ b/interpreter/astwalker.go
@@ -38,7 +38,8 @@ const (
 func WalkExpr(expression *exprpb.Expr,
 	metadata Metadata,
 	dispatcher Dispatcher,
-	state MutableEvalState) []Instruction {
+	state MutableEvalState,
+	forceEval bool) []Instruction {
 	nextID := maxID(expression)
 	walker := &astWalker{
 		dispatcher: dispatcher,
@@ -46,7 +47,8 @@ func WalkExpr(expression *exprpb.Expr,
 		genExprID:  nextID,
 		metadata:   metadata,
 		scope:      newScope(),
-		state:      state}
+		state:      state,
+		forceEval:  forceEval}
 	return walker.walk(expression)
 }
 
@@ -58,6 +60,7 @@ type astWalker struct {
 	metadata   Metadata
 	scope      *blockScope
 	state      MutableEvalState
+	forceEval  bool
 }
 
 func (w *astWalker) walk(node *exprpb.Expr) []Instruction {
@@ -152,11 +155,13 @@ func (w *astWalker) walkCall(node *exprpb.Expr) []Instruction {
 			evalCount += argGroupLens[i]
 			instructions = append(instructions, argGroup...)
 			if i != argCount-1 {
-				instructions = append(instructions,
-					NewJump(
-						argIDs[i],
-						instructionCount-evalCount,
-						jumpIfEqual(argIDs[i], types.Bool(function == operators.LogicalOr))))
+				if w.forceEval{
+					instructions = append(instructions, NewNoOp())
+				} else {
+					instructions = append(instructions,
+						NewJump(argIDs[i], instructionCount-evalCount,
+							jumpIfEqual(argIDs[i], types.Bool(function == operators.LogicalOr))))
+				}
 			}
 			evalCount++
 		}
@@ -185,14 +190,22 @@ func (w *astWalker) walkCall(node *exprpb.Expr) []Instruction {
 			NewJump(conditionID, len(trueVal)+len(falseVal)+3,
 				jumpIfUnknownOrError(conditionID)))
 		// 2: jump to <ELSE> on false.
-		instructions = append(instructions,
-			NewJump(conditionID, len(trueVal)+2,
-				jumpIfEqual(conditionID, types.False)))
+		if w.forceEval {
+			instructions = append(instructions, NewNoOp())
+		} else {
+			instructions = append(instructions,
+				NewJump(conditionID, len(trueVal)+2,
+					jumpIfEqual(conditionID, types.False)))
+		}
 		// 3: <IF> expr
 		instructions = append(instructions, trueVal...)
 		// 4: jump to <END>
-		instructions = append(instructions,
-			NewJump(trueID, len(falseVal)+1, jumpAlways))
+		if w.forceEval {
+			instructions = append(instructions, NewNoOp())
+		} else {
+			instructions = append(instructions,
+				NewJump(trueID, len(falseVal)+1, jumpAlways))
+		}
 		// 5: <ELSE> expr
 		instructions = append(instructions, falseVal...)
 		// 6: <END> ternary
--- a/interpreter/astwalker.go
+++ b/interpreter/astwalker.go
@@ -39,28 +39,28 @@ func WalkExpr(expression *exprpb.Expr,
 	metadata Metadata,
 	dispatcher Dispatcher,
 	state MutableEvalState,
-	forceEval bool) []Instruction {
+	shortCircuit bool) []Instruction {
 	nextID := maxID(expression)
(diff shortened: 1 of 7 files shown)


interpreter/astwalker.go
```diff
@@ -152,11 +155,13 @@ func (w *astWalker) walkCall(node *exprpb.Expr) []Instruction {
 			evalCount += argGroupLens[i]
 			instructions = append(instructions, argGroup...)
 			if i != argCount-1 {
-				instructions = append(instructions,
-					NewJump(
-						argIDs[i],
-						instructionCount-evalCount,
-						jumpIfEqual(argIDs[i], types.Bool(function == operators.LogicalOr))))
+				if w.forceEval{
```
TristonianJones: Here and below, please consider phrasing this as though evaluating all branches is disabled by default.
```
if i != argCount - 1 && w.shortcircuit {
  instructions = append(instructions, NewJump(...))
}
```

interpreter/astwalker.go
```diff
@@ -38,15 +38,17 @@ const (
 func WalkExpr(expression *exprpb.Expr,
 	metadata Metadata,
 	dispatcher Dispatcher,
-	state MutableEvalState) []Instruction {
+	state MutableEvalState,
+	forceEval bool) []Instruction {
```
TristonianJones: Consider renaming from `forceEval` to `shortcircuit`.  Removing the `jump` instructions is all that's needed to evaluate all branches.

interpreter/astwalker.go
```diff
@@ -185,14 +190,22 @@ func (w *astWalker) walkCall(node *exprpb.Expr) []Instruction {
 			NewJump(conditionID, len(trueVal)+len(falseVal)+3,
 				jumpIfUnknownOrError(conditionID)))
 		// 2: jump to <ELSE> on false.
-		instructions = append(instructions,
-			NewJump(conditionID, len(trueVal)+2,
-				jumpIfEqual(conditionID, types.False)))
+		if w.forceEval {
```
TristonianJones:
```
if w.shortcircuit {
   instructions = append(instructions, NewJump(....))
}
```

interpreter/instructions.go
```diff
@@ -197,6 +197,21 @@ func NewJump(exprID int64, instructionCount int, cond func(EvalState) bool) *Jum
 		OnCondition:     cond}
 }
 
+// NoOp represents a no operational placeholder.
```
TristonianJones: Per the earlier comments, I'm not sure you need the `NoOp`. Give it a try without them and if it works, then remove this code, and if it doesn't please just rephrase the code as though short-circuiting is enabled by default, e.g.
```
inst := NewJump(...)
if !w.shortcircuit {
  inst := NewNoOp()
}
instructions = append(instructions, inst)
```

interpreter/interpreter_test.go
```diff
@@ -30,6 +30,51 @@ import (
...
+	program := NewCompleteProgram(
+		test.Conditional.Expr,
+		test.Conditional.Info(t.Name()))
+
+	interpretable := interpreter.NewInterpretable(program)
+	result, state := interpretable.Eval(
+		NewActivation(map[string]interface{}{
+			"a": true,
+			"b": 0.999,
+			"c": types.NewStringList([]string{"hello"})}))
```
TristonianJones, line 46: Would you mind adding a comment about why the test for the value of `c` is being done?

interpreter/interpreter_test.go
```diff
@@ -30,6 +30,51 @@ import (
...
+	ev, _ := state.Value(6)
```
TristonianJones, line 47: Please comment that the identifier for `c` has the expression id `6`.  Same for the later tests where the magic expression identifier is used. You'll want to mention that the next developer should see the test/expr code to figure out what this value should be (or provide a convenient function that gets the expression id for `c` by name so the next person doesn't have to guess).

interpreter/program.go
```diff
@@ -83,7 +84,22 @@ func NewProgram(expression *exprpb.Expr,
...
 		revInstructions: revInstructions,
-		metadata:        newExprMetadata(info)}
+		metadata:        newExprMetadata(info),
+		forceEval:       false,
+	}
+}
+
+// NewCompleteProgram creates a Program from a CEL expression and source
+// information which force evaluating all branches of the expression.
+func NewCompleteProgram(expression *exprpb.Expr,
```
TristonianJones: Maybe call this `NewExhaustiveProgram`. The word `Complete` connotes that somehow the regular program is incomplete.

interpreter/astwalker.go
jiyunyao: Done. Works better~ Thanks

interpreter/astwalker.go
jiyunyao: Done~

interpreter/astwalker.go
jiyunyao: Done. Works better~ Thanks

interpreter/instructions.go
jiyunyao: Done. NoOp is removed

interpreter/interpreter_test.go
jiyunyao: Comments added~ Please let me know if it is sufficient~

interpreter/interpreter_test.go
jiyunyao: Comments added~ Please let me know if it is sufficient~

interpreter/program.go
jiyunyao: Done.

interpreter/interpreter_test.go
```diff
@@ -155,15 +238,18 @@ func TestInterpreter_LogicalOrEquals(t *testing.T) {
 	provider := types.NewProvider(&exprpb.Expr{})
 	i := NewStandardInterpreter(packages.NewPackage("test"), provider)
 	interpretable := i.NewInterpretable(program)
-	result, _ := interpretable.Eval(
+	result, state := interpretable.Eval(
 		NewActivation(map[string]interface{}{
 			"a": true,
 			"b": "b",
 		}))
+	rhv, _ := state.Value(4)
```
TristonianJones, line 249: Please add a comment about where the numeric indices provided to `Value()` calls come from (here and above)

TristonianJones (changes requested):
Looks generally pretty good, but please add a couple of comments to the tests.

interpreter/interpreter_test.go
jiyunyao: done

jiyunyao:
Done.

TristonianJones (commented):
The code present is in good shape, and it allows for exhaustive evaluation over many of the branches
of execution within a CEL expression, but not all. 

For comprehensions, you will need to change comprehension loop to determine the state at each iteration of the loop. The `EvalState` object doesn't capture multiple evaluations of the same
conditions, so the change is more extensive than what is presented here. 

Please create an Issue as follow:

> Introduce an exhaustive evaluation mode
>
> Exhaustive evaluation does not short-circuit conditions, and iterates through all elements
> in the `iterRange` argument to a comprehension. The state of each iteration should also be
> tracked and returned with the call result.

Once the issue is created and a TODO added to the `NewExhaustiveProgram` call, I'll approve
this change and reference the parent issue on merge.

This is great first step. :+1:

jiyunyao:
#121

jiyunyao:
Done.

Issue #55, opened 2018-08-14
Confusing example - how is common.Errors used and why is it passed around?


broady:
The example in the readme...

```
// Parse the expression and returns the accumulated errors.
p, errors := parser.ParseText("a || b && c.exists(x, x > 2)")
if len(errors.GetErrors()) != 0 {
    return nil, fmt.Error(errors.ToDisplayString()))
}

// ...
env := checker.NewStandardEnv(packages.DefaultPackage, typeProvider, errors)
// ...
if len(errors.GetErrors()) != 0 {
    return nil, fmt.Error(errors.ToDisplayString()))
}
```

This is quite confusing - why is the `errors` struct being passed around? What's the reason for returning it from the parser and passing it into the environment?

From my reading, the checker env isn't re-usable or goroutine-safe.

As a newcomer to the package, I'd expect a flow like:

* create environment, provide type information, etc.
* parse and check the expression against the environment
* evaluate the expression

The current interface seems to entangle these operations.

TristonianJones:
The example in the `README.md` expects that users are always parsing and type-checking before evaluating, and `errors` are accumulated between the `parser` and `checker`. I mostly wanted to indicate that one depends on the other, but there are probably better ways to structure this. 

Would you prefer the `checker` return its own instance of `errors` rather than accumulate? I think this addresses your re-use / goroutine safety concerns.

PR #118, opened 2018-11-29
Isolation of `common.Errors` between parser and checker.

TristonianJones:
#55 Sharing `common.Errors` between the `parser` and `checker` is confusing and hampers
the use of CEL in applications which might want to parse, check, and evaluate multiple expressions
concurrently. This CL removes `common.Errors` from the `Env` struct and isolates the use of the
`common.Errors` objects to `checker.go` and `parser.go` files only.

Note: This PR does not attempt to unify the use of a standard environment for use in all CEL
execution phases. Per #55, it should be possible to create and share an `Env` value for multiple
purposes thus simplifying setup and reducing the boilerplate present in the `README.md`

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,7 +60,8 @@ import(
 )
 
 // Parse the expression and returns the accumulated errors.
-p, errors := parser.ParseText("a || b && c.exists(x, x > 2)")
+src := common.NewTextSource("a || b && c.exists(x, x > 2)")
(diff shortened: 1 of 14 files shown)


glerchundi:
This change would solve one of our re-use use case too, awesome!

madi8229 (approved):
LGTM.

jiyunyao (approved):
LGTM

Issue #43, opened 2018-06-04
Introduce non-strict comprehensions


TristonianJones:
## Issue Filing Checklist

- [X] There are no issues that match the desired change
- [X] The change is large enough it can't be addressed with a simple [Pull Request](./PULL_REQUEST_TEMPLATE.md) 
- [X] If this is a bug, there are a concrete set of repro steps

## Change
The termination condition for all comprehensions is `<loop-cond> != false` in order
to ensure that when the loop-condition does not terminate on comprehensions where
the loop-step is computed from a non-strict operator (`&&`, `||`). This can result in
inefficient computation when the loop-step is strict (literally any other operator).

To account for this difference, the parser macros should expand non-strict
loop-conditions to be a function call to a `not_strictly_false(<loop-cond>)` function
which returns true when the loop condition value is error, unknown, or true. The new
function will absorb errors / unknowns just like the non-strict accumulators for `exists`
and `all` macros.

PR #119, opened 2018-11-30
Support not-strict comprehensions

TristonianJones:
#43 Introduce the `not_strictly_false` macro to ensure that the `all` and `exists` macros
accumulate in the same manner as the `||` and `&&` operators would if the macro loops
were unrolled.

The following changes were made: 
*  A new non-strict logical primitive `not_strictly_false` is added that returns
    its argument or true if the argument has an error or is unknown.
*  The macro expansion for ALL and EXISTS uses the  new `not_strictly_false` primitive
    to effectively make the termination condition non-strict as well in these cases.
    In other cases where the comprehension step is naturally strict, the termination
    condition remains strict as well.

During the development of this code an off-by-one error was revealed in the `astwalker.go`
which could cause the program to terminate without moving the final evaluation to the
result register.


common/operators/operators.go
```diff
@@ -19,30 +19,31 @@ package operators
...
-	All           = "all"
-	Exists        = "exists"
-	ExistsOne     = "exists_one"
-	Map           = "map"
-	Filter        = "filter"
+	Conditional      = "_?_:_"
+	LogicalAnd       = "_&&_"
+	LogicalOr        = "_||_"
+	LogicalNot       = "!_"
+	NotStrictlyFalse = "__not_strictly_false__"
```
JimLarson: Add gratuitous punctuation to make it an invalid identifier.

interpreter/astwalker.go
```diff
@@ -255,14 +255,18 @@ func (w *astWalker) walkComprehension(node *exprpb.Expr) []Instruction {
...
-	// 6: jump <END> if !accu
-	// 7: accu = accu && x < 10           # loopStep
-	// 8: jump <LOOP>
-	// 9: result = accu                   # result
-	// 10: comp = result
-	// 11: pop-scope
+	// <LOOP>
+	// 4: hasNext = it.hasNext()          # hasNext()?
+	// 5: jump <END> if !hasNext
+	// 6: iterVar = it.next()             # it.next()
```
JimLarson, line 268: x = it.next()?  or change step 9 to be accu = accu && iterVar < 10?

interpreter/astwalker.go
```diff
@@ -255,14 +255,18 @@ func (w *astWalker) walkComprehension(node *exprpb.Expr) []Instruction {
...
+	// 7: loop = eval(accu, iterVar)      # loopCondition
```
JimLarson: Wot?  Isn't the loop condition just accu?

interpreter/functions/standard.go
```diff
@@ -44,6 +44,16 @@ func StandardOverloads() []*Overload {
 		{
 			Operator: operators.Conditional,
 			Function: conditional},
+		// Not strictly false: IsBool(a) ? true : true
```
JimLarson: Shouldn't this be IsBool(a) ? a : true

interpreter/interpreter_test.go
```diff
@@ -80,17 +80,81 @@ func TestInterpreter_ConditionalExpr(t *testing.T) {
...
+	if len(errors.GetErrors()) != 0 {
+		t.Errorf(errors.ToDisplayString())
+	}
+	pgrm := NewProgram(parsed.Expr, parsed.SourceInfo)
+	eval := interpreter.NewInterpretable(pgrm)
+	result, _ := eval.Eval(emptyActivation)
+	if result != types.True {
+		t.Errorf("Got %v, wanted true", result)
+	}
+}
```
JimLarson, line 181: You're using this pattern enough that it seems to be worth abstracting:

func TestInterpreter_Blah(t *testing.T) {
        EvalExpectingTrue("[0, 2, 4].exists(x, whatever(x))")
}

interpreter/prune_test.go
```diff
@@ -61,16 +61,23 @@ var testCases = []testInfo{
 					test.ExprLiteral(4, int64(1)),
 					test.ExprLiteral(5, uint64(1)),
 					test.ExprLiteral(6, float64(1.0))),
-				"_accu_",
-				test.ExprLiteral(8, false),
-				test.ExprCall(10,
-					operators.LogicalNot,
-					test.ExprIdent(9, "_accu_")),
+				"__result__",
+				test.ExprLiteral(9, false),
```
JimLarson: Make a note saying where expr 8 went.

interpreter/interpreter_test.go
TristonianJones: Go sort of frowns on assertion helpers, but since the evaluation of the source and errors related to it are just a general part of test setup, I've moved them into a helper function called `evalExpr`

interpreter/prune_test.go
TristonianJones: Just missed it is all. :)

common/operators/operators.go
TristonianJones: Unfortunately, I can't change this one since it's referenced in other contexts by this identifier.

JimLarson: Add an alias so that the other contexts can switch.

JimLarson (changes requested):

common/operators/operators.go
TristonianJones: I've made a shift to `@<internal_operator_name>` and preserved the aliases to the old functions in the `interpreter`. Please take another look.

JimLarson (approved):

PR #122, opened 2018-12-04
Fix package structure so `go get` works for CEL-Go

### LINKED ISSUE #51: Issue running `go get` on repo
## Issue Filing Checklist
Issue running `go get` on the parser
screen shot 2018-07-30 at 19 00 45
 package.

TristonianJones:
The current package structure is pretty close to working with `go get`, but the `server` folder needed
to have the `main` package split into its own directory.

closes #51: issue running `go get` on repo


JimLarson (changes requested):
What commands are you running to verify that it's broken?  Running "go get github.com/google/cel-go" retrieves everything, including server/main.go (but gives a spurious warning that there are no .go files at the top level).

Assuming that main.go has to go into a subdir, just call the subdir 'main' and keep the file as 'main.go'.  We have to use 'main', so there's no need for new identifiers.  If there is some strange reason we can't use 'main', use 'entry' instead of 'agent'.

TristonianJones:
I ran `go get github.com/google/cel-go/...` which yields an error on the `server` folder due to the presence of two package declarations. If I try the same command without the elipses, I get an error and the `go list` response is incomplete.

My impression is that all `main` package files go into a single directory, but I'm willing to keep server apart from the other future tools related to parsing, checking, etc. So you'd prefer `main` as the name for the file / binary?

JimLarson (approved):

PR #125, opened 2018-12-05
Remove aggregate type restriction requiring homogeneous element types.

### LINKED ISSUE #56: Remove homogeneous aggregate type restrictions
## Issue Filing Checklist

- [x] The change is large enough it can't be addressed with a simple [Pull Request](./PULL_REQUEST_TEMPLATE.md) 
- [x] If this is a bug, there are a concrete set of repro steps

## Change
CEL is a dynamically typed language. As such there is no way to specify a type-designator
for an aggregate type literal, e.g. `list<int>[1, 2, 3]`. This means that all aggregate types defined
inline within an expression should be treated as `dyn` values. Currently, the type-checker asserts
type-agreement between all elements in a list or map unless expressly cast to `dyn`. This check
should be removed, preferring instead to identify the most general type element in the list.

### Example
Currently fails to type-check: `[1u, 1]` but should succeed as the interpreter handles this
expression just fine.

TristonianJones:
Previously inputs like `[1, 1u]` would result in a type-check error, but since this code
was created, the specification has been updated to indicate that CEL is a dynamically typed
language which implies that aggregate types must be able to support heterogeneous element
types.

closes #56

### DIFF
--- a/checker/checker.go
+++ b/checker/checker.go
@@ -446,8 +446,7 @@ func (c *checker) joinTypes(loc common.Location, previous *exprpb.Type, current 
 		return current
 	}
 	if !c.isAssignable(previous, current) {
-		c.errors.aggregateTypeMismatch(loc, previous, current)
-		return previous
+		return decls.Dyn
 	}
 	return mostGeneral(previous, current)
 }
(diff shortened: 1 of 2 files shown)


madi8229 (approved):
LGTM.

PR #128, opened 2018-12-07
Check time support for JSON and well-known types.

### LINKED ISSUE #47: JSON and well-known types not handled properly during type-resolution
## Issue Filing Checklist

- [X] The change is large enough it can't be addressed with a simple [Pull Request](./PULL_REQUEST_TEMPLATE.md) 
- [X] If this is a bug, there are a concrete set of repro steps

## Change
When a value has a `google.protobuf.Value` type (or derivative), or a well-known type
such as `google.protobuf.Any` or `google.protobuf.Duration`, the type is treated as a
`MESSAGE` by the `checker.go` when it should be mapped to the appropriate built-in
or well-known type value.

TristonianJones:
Built-in CEL types based on protobufs were not properly being identified at `check` time when
referenced by their proto message name. This change sanitizes the environment by mapping
proto messages to their appropriate CEL built-in type when appropriate. 

closes #47

### DIFF
--- a/checker/env.go
+++ b/checker/env.go
@@ -21,6 +21,7 @@ import (
 	"github.com/google/cel-go/checker/decls"
 	"github.com/google/cel-go/common/packages"
 	"github.com/google/cel-go/common/types"
+	"github.com/google/cel-go/common/types/pb"
 	"github.com/google/cel-go/common/types/ref"
 	"github.com/google/cel-go/parser"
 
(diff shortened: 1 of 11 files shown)


checker/env.go
```diff
@@ -142,57 +186,104 @@ func (e *Env) addIdent(decl *exprpb.Decl) errorMsg {
...
+		// If sanitized, replace the overload definition.
+		if sanitized {
+			if o.IsInstanceFunction {
+				overloads[i] =
+					decls.NewInstanceOverload(o.GetOverloadId(), params, rt)
+			} else {
+				overloads[i] =
+					decls.NewOverload(o.GetOverloadId(), params, rt)
+			}
+			// Otherwise, preserve the original overload.
```
JimLarson: Move this comment down after the "else".

JimLarson (approved):

PR #131, opened 2018-12-14
integration test for the 'regex'

kpacha:
This PR adds an integration test to the `tests` package in order to ensure the `matches` overload is working properly


kpacha:
I signed it!

test/integration_test.go
```diff
@@ -0,0 +1,60 @@
+package test
```
TristonianJones: If you wouldn't mind adding the test cases to the `interpreter/interpreter_test.go` file, that would be preferred, and some of the test setup can be done with helper methods. See comments below.

test/integration_test.go
```diff
@@ -0,0 +1,60 @@
...
+	"github.com/google/cel-go/checker"
+	"github.com/google/cel-go/checker/decls"
+	"github.com/google/cel-go/common"
+	"github.com/google/cel-go/common/packages"
+	"github.com/google/cel-go/common/types"
+	"github.com/google/cel-go/interpreter"
+	"github.com/google/cel-go/parser"
+)
+
+func Test_stringOverfloats(t *testing.T) {
```
TristonianJones: Prefer splitting this into `TestInterpreter_Matches` and `TestInterpreter_MatchesAndInList` where the name indicates the file and the operator under test.

test/integration_test.go
```diff
@@ -0,0 +1,60 @@
...
+	for _, expression := range []string{
+		"req_params.Nick.matches('k.*')",
+		"nick.matches('k.*')",
+		"nick in ['klmno', 'kpacha']",
```
TristonianJones: I would combine the expressions in the second and third case: 
`nick.matches('k.*') && nick in ['klmno', 'kpacha']`

test/integration_test.go
```diff
@@ -0,0 +1,60 @@
...
+	} {
+		source := common.NewTextSource(expression)
+		p, errors := parser.Parse(source)
```
TristonianJones: There's a helper method in `interpreter_test.go` called `parseExpr()` which could be used to replace the parsing portion of this test case. However, since you're also checking, I would create a `compileExpr()` helper in `interpreter_test.go` which combines the parse and check steps:

```go
func compileExpr(t *testing.T, src string, decls ...*exprpb.Decl) *exprpb.CheckedExpr {
    t.Helper()
    s := common.NewTextSource(src)
    parsed, errors := parser.Parse(s)
    if len(errors.GetErrors()) != 0 {
        t.Error(errors.ToDisplayString())
        return nil
    }
    env := checker.NewStandardEnv(packages.DefaultPackage, types.NewProvider())
    env.Add(decls...)
    checked, errors := checker.Check(parsed, s, env)
    if len(errors.GetErrors()) != 0 {
       t.Error(errors.ToDisplayString())
       return nil
    }
    return checked
}
```

Then your test setup would probably looks as follows: 

```go
expr := compileExpr(t, "nick in ...", decls.NewIdent("nick", decls.String, nil))
prog := NewCheckedProgram(expr)
eval := interpreter.NewInterpretable(prog)
for _, tc := range []struct { } { ... }
```

test/integration_test.go
```diff
@@ -0,0 +1,60 @@
...
+		checkedExp, errors := checker.Check(p, source, env)
+		if len(errors.GetErrors()) != 0 {
+			t.Error(errors.ToDisplayString())
+			continue
+		}
+
+		interpreterInstance := interpreter.NewStandardInterpreter(packages.DefaultPackage, typeProvider)
+		eval := interpreterInstance.NewInterpretable(interpreter.NewCheckedProgram(checkedExp))
+
+		for _, tc := range []struct {
```
TristonianJones: If you want to define the test input struct as `type testCase struct` that would be fine with me. It might make the setup a bit more readable.

test/integration_test.go
```diff
@@ -0,0 +1,60 @@
...
+			{nick: "foo", res: false},
+			{nick: "bar", res: false},
+			{nick: "klmno", res: true},
+		} {
+			result, _ := eval.Eval(interpreter.NewActivation(map[string]interface{}{
+				"req_params": map[string]string{"Nick": tc.nick},
+				"nick":       tc.nick,
+			}))
+			if v, ok := result.Value().(bool); !ok || v != tc.res {
+				t.Errorf("%s [%s] -> unexpected result: %+v", expression, tc.nick, result.Value())
```
TristonianJones: For error statements like this, the usual style is to write `Got x, wanted y`, so maybe phrase the error as
`Got %v, wanted %v for expr %s with input %s`

TristonianJones (changes requested):
Thanks for the test cases. I've recommended some refactors of the tests. Hopefully you don't find the suggestions helpful.

test/integration_test.go
kpacha: great! thx

test/integration_test.go
kpacha: My original intention was to add some global integration tests that could also be used as an easy complementary introduction to the lib for users (and CEL itself).

test/integration_test.go
kpacha: Since the `InList` test already exists in the `interpreter_test.go`, I'll keep just the `macthes` part

test/integration_test.go
kpacha: refactored to a `map[string]bool` for readability

TristonianJones (approved):
Awesome!

test/integration_test.go
TristonianJones: Ah, yes. We are planning to add a boatload of conformance tests to cel-spec which will be run using the `server/server.go`, but have not yet gotten it done. The idea being that the tests should work across CEL implementations in other languages, e.g. cel-cpp which is also under development.

PR #137, opened 2018-12-18
Implement has() macro in interpreter.

### LINKED ISSUE #133: Macro has() not implemented
**Describe the bug**
It appears to me that has() is only implemented in the checker phase, and the runtime evaluates field selector tests as if they are field selectors. 

**To Reproduce**

- [ ]  parser
- [ ]  checker
- [x]  interpreter

```cel
has(a.b)
```

Test setup:
```go
       // Parse the expression and returns the accumulated errors.
        src := common.NewTextSource("has(a.b)")
        expr, errors := parser.Parse(src)
        if len(errors.GetErrors()) != 0 {
                t.Error(errors.ToDisplayString())
        }

        // Check the expression matches expectations given the declarations for
        // the identifiers a, b, c where the identifiers are scoped to the default
        // package (empty string):
        typeProvider := types.NewProvider()
        env := checker.NewStandardEnv(packages.DefaultPackage, typeProvider)
        env.Add(decls.NewIdent("a", decls.Dyn, nil))
        c, errors := checker.Check(expr, src, env)
        if len(errors.GetErrors()) != 0 {
                t.Error(errors.ToDisplayString())
        }

        // Interpret the checked expression using the standard overloads.
        i := interpreter.NewStandardInterpreter(packages.DefaultPackage, typeProvider)
        eval := i.NewInterpretable(interpreter.NewCheckedProgram(c))
        result, state := eval.Eval(
                interpreter.NewActivation(
                        map[string]interface{}{
                                "a": map[string]string{"b": "c"},
                        }))

        fmt.Printf("%#v\n", state)
        fmt.Println(result)
```

**Expected behavior**
I expect to get "true", but get "c" instead.

TristonianJones:
The `has()` macro was not implemented in the interpreter as it was not a function but a macro
on the select expression that wasn't being checked. Fixed and tested.

closes #133

### DIFF
--- a/common/types/object.go
+++ b/common/types/object.go
@@ -83,6 +83,27 @@ func (o *protoObj) Equal(other ref.Value) ref.Value {
 	return Bool(proto.Equal(o.value, other.Value().(proto.Message)))
 }
 
+func (o *protoObj) IsSet(field ref.Value) ref.Value {
+	if field.Type() != StringType {
+		return NewErr("illegal object field type '%s'", field.Type())
+	}
+	protoFieldName := string(field.(String))
+	if f, found := o.typeDesc.FieldByName(protoFieldName); found {
+		if !f.IsOneof() {
+			return isFieldSet(o.refValue.Elem().Field(f.Index()))
+		}
+
+		getter := o.refValue.MethodByName(f.GetterName())
+		if getter.IsValid() {
+			refField := getter.Call([]reflect.Value{})[0]
+			if refField.IsValid() {
+				return isFieldSet(refField)
+			}
+		}
+	}
+	return NewErr("no such field '%s'", field)
+}
+
 func (o *protoObj) Get(index ref.Value) ref.Value {
 	if index.Type() != StringType {
 		return NewErr("illegal object field type '%s'", index.Type())
@@ -145,9 +166,16 @@ var (
 	protoDefaultInstanceMap = make(map[reflect.Type]ref.Value)
 )
 
-func getOrDefaultInstance(refVal reflect.Value) ref.Value {
-	value := refVal.Interface()
+func isFieldSet(refVal reflect.Value) ref.Value {
 	if refVal.Kind() != reflect.Ptr || !refVal.IsNil() {
+		return True
+	}
+	return False
+}
+
+func getOrDefaultInstance(refVal reflect.Value) ref.Value {
+	if isFieldSet(refVal) == True {
+		value := refVal.Interface()
 		return NativeToValue(value)
 	}
 	return getDefaultInstance(refVal.Type())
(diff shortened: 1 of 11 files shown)


common/types/object.go
```diff
@@ -83,6 +83,27 @@ func (o *protoObj) Equal(other ref.Value) ref.Value {
 	return Bool(proto.Equal(o.value, other.Value().(proto.Message)))
 }
 
+func (o *protoObj) IsSet(field ref.Value) ref.Value {
```
JimLarson, line 87: Add a function comment - keep our lint clean! :-)

common/types/object.go
```diff
@@ -145,9 +166,16 @@ var (
 	protoDefaultInstanceMap = make(map[reflect.Type]ref.Value)
 )
 
-func getOrDefaultInstance(refVal reflect.Value) ref.Value {
-	value := refVal.Interface()
+func isFieldSet(refVal reflect.Value) ref.Value {
 	if refVal.Kind() != reflect.Ptr || !refVal.IsNil() {
```
JimLarson: Consider de Morgan'ing this around to
if refVal.Kind() == reflectPtr && refVal.IsNil() {
        return False
}
return True

common/types/traits/field_definer.go
```diff
@@ -0,0 +1,30 @@
...
+
+import (
+	"github.com/google/cel-go/common/types/ref"
+)
+
+// FieldDefiner indicates if a defined field on an objec type is set to a
+// non-default value.
+//
+// For use with the `has()` macro.
+type FieldDefiner interface {
```
JimLarson: Consider "FieldTester" instead.

common/types/traits/field_definer.go
```diff
@@ -0,0 +1,30 @@
...
+	// IsSet returns true if the field is defined and set to a non-default
+	// value. The method will return false if defiend and not set, and an error
```
JimLarson: s/defiend/defined/

common/types/traits/field_definer.go
```diff
@@ -0,0 +1,30 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package traits
+
+import (
+	"github.com/google/cel-go/common/types/ref"
+)
+
+// FieldDefiner indicates if a defined field on an objec type is set to a
```
JimLarson: s/objec/object/

JimLarson (approved):
Documentation and naming-level feedback only.

kyessenov:
Thanks for fixing the bug. Can this be made merged so we can try it out in istio?

TristonianJones:
Yes, I should be able to later today. Just got back from vacation yesterday.

-Tristan

PR #142, opened 2019-01-08
Fix qualified id resolution.

TristonianJones:
CEL supports qualified / namespaced variable names such as `a.b.c.d` as part of its spec where
the name outright may be associated with a value rather than with a specific pattern of object
traversal. The `checker` handles these fields correctly, but the interpreter was not, instead
halting resolution of the namespaced identifier early. The PR fixes the glitch.

### DIFF
--- a/interpreter/interpreter.go
+++ b/interpreter/interpreter.go
@@ -161,13 +161,13 @@ func (i *exprInterpretable) evalSelect(selExpr *SelectExpr, currActivation Activ
 			i.setValue(selExpr.GetID(), resVal)
 			return
 		}
+		// If the operand is an error, early return.
 		if types.IsError(operand) {
 			i.setValue(selExpr.GetID(), operand)
-		} else {
-			i.setValue(
-				selExpr.GetID(),
-				types.NewErr("Invalid operand for select expression."))
+			return
 		}
+		// Otherwise, create an error.
+		i.setValue(selExpr.GetID(), types.NewErr("invalid operand in select"))
 		return
 	}
 	field := types.String(selExpr.Field)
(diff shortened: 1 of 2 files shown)


TristonianJones:
This PR closes #143

interpreter/interpreter.go
```diff
@@ -155,14 +155,19 @@ func (i *exprInterpretable) evalIdent(idExpr *IdentExpr, currActivation Activati
 func (i *exprInterpretable) evalSelect(selExpr *SelectExpr, currActivation Activation) {
 	operand := i.value(selExpr.Operand)
 	if !operand.Type().HasTrait(traits.IndexerType) {
+		// If the operand is unknown, this could be an identifer.
+		if types.IsUnknown(operand) {
+			resVal := i.resolveUnknown(operand.(types.Unknown), selExpr, currActivation)
+			i.setValue(selExpr.GetID(), resVal)
+			return
+		}
+		// If the oeprand is an error, early return
```
JimLarson: s/oeprand/operand/

interpreter/interpreter.go
```diff
@@ -188,10 +193,9 @@ func (i *exprInterpretable) evalSelect(selExpr *SelectExpr, currActivation Activ
 // qualified identifier name instead of a series of field selections.
```
JimLarson, line 193: Indicate what value is returned.

interpreter/interpreter.go
```diff
@@ -213,22 +217,19 @@ func (i *exprInterpretable) resolveUnknown(unknown types.Unknown,
 		}
 	}
 	if !validIdent {
-		return false
+		return types.NewErr("Invalid identifier encountered.")
 	}
 	pkg := i.interpreter.packager
 	tp := i.interpreter.typeProvider
 	for _, id := range pkg.ResolveCandidateNames(identifier) {
 		if object, found := currActivation.ResolveName(id); found {
```
JimLarson, line 230: Shadows previous "object" - is there a useful adjective to add?

JimLarson (approved):

interpreter/interpreter.go
TristonianJones: When used in the if block like this, the `object` is only visible within the `if`.

interpreter/interpreter.go
TristonianJones: Done.

Issue #146, opened 2019-01-10
Unsupported type conversion accessing proto field with enum value


jf647:
**Describe the bug**

Attempting to evaluate a field of a proto object whose value is a protobuf enum results in an unsupported type conversion error.

**To Reproduce**
Check which components this affects:

- [ ]  parser
- [ ]  checker
- [X]  interpreter

Sample expression and input that reproduces the issue:

```cel
a.repeated_nested_enum[0]
```

(using the TestAllTypes proto)

Test setup:
```go
func TestInterpreter_ObjectEnumField(t *testing.T) {
	src := common.NewTextSource("a.repeated_nested_enum[0]")
	parsed, errors := parser.Parse(src)
	if len(errors.GetErrors()) != 0 {
		t.Errorf(errors.ToDisplayString())
	}

	pkgr := packages.NewPackage("google.api.tools.expr.test")
	provider := types.NewProvider(&test.TestAllTypes{})
	env := checker.NewStandardEnv(pkgr, provider)
	env.Add(decls.NewIdent("a", decls.NewObjectType("google.api.tools.expr.test.TestAllTypes"), nil))
	checked, errors := checker.Check(parsed, src, env)
	if len(errors.GetErrors()) != 0 {
		t.Errorf(errors.ToDisplayString())
	}

	i := NewStandardInterpreter(pkgr, provider)
	eval := i.NewInterpretable(NewCheckedProgram(checked))
	a := &test.TestAllTypes{
		RepeatedNestedEnum: []test.TestAllTypes_NestedEnum{
			test.TestAllTypes_BAR,
		},
	}
	result, state := eval.Eval(NewActivation(map[string]interface{}{
		"a": types.NewObject(a),
	}))
	expected := int64(1)
	got, ok := result.(ref.Value).Value().(int64)
	if !ok {
		t.Fatalf("cannot cast result to int64: result=%v state=%v", result, state)
	}
	if !reflect.DeepEqual(got, expected) {
		t.Errorf("Could not build object properly. Got '%v', wanted '%v'",
			result.(ref.Value).Value(),
			expected)
	}
}
```

**Expected behavior**

I would expect the evaluation to succeed and resolve to a types.Int value equal to TestAllTypes_BAR (1).

**Additional context**

When run, this test gives the following error:

```shell
go test -v ./interpreter -run TestInterpreter_ObjectEnumField
=== RUN   TestInterpreter_ObjectEnumField
[...]

PR #147, opened 2019-01-11
Fix for the treatment of enums as ints within CEL.

TristonianJones:
Issue #146 revealed a bug in the treatment of enums in CEL as ints. Since enums are a kind of `Int32`, but they are not actually `int32` values, the conversion logic was incorrect and reflection needed to be introduced to cover the enum types properly.

### DIFF
--- a/common/types/int.go
+++ b/common/types/int.go
@@ -18,9 +18,10 @@ import (
 	"fmt"
 	"reflect"
 
-	structpb "github.com/golang/protobuf/ptypes/struct"
 	"github.com/google/cel-go/common/types/ref"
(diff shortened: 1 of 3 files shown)


common/types/int.go
```diff
@@ -18,7 +18,7 @@ import (
 	"fmt"
 	"reflect"
 
-	"github.com/golang/protobuf/ptypes/struct"
+	structpb "github.com/golang/protobuf/ptypes/struct"
```
madi8229: Could you move this to the bottom of the imports, in a new section of imports?

madi8229 (approved):
LGTM.

Issue #130, opened 2018-12-13
Better API wrt failures, warnings, and partial results


JimLarson:
**Feature request checklist**

- [X] There are no issues that match the desired change
- [X] The change is large enough it can't be addressed with a simple Pull Request
- [ ] If this is a bug, please file a [Bug Report](./ISSUE_TEMPLATE).

**Change**
Summary of the proposed change and some details about what problem
it helps solve.

The current API for parsing and checking returns:
- an optional parsed or checked AST, and
- an optional list of errors.

In the future, we want to also return warnings.  Other adjacent APIs generalize "errors" to "issues", tagged with a severity.

Also in the future, we'd like to possibly return an AST, even in the case of errors, so that the partial or erroneous AST could be useful for understanding the error.

This means that neither the list of issues, nor the presence/absence of the AST would be a clear indication of whether the operation succeeded.  We don't want to have to grovel through the list of issues to determine success.  We need a different interface to distinguish the following cases:

Successful operation, possibly with warnings:  returns mandatory AST, list of warnings (maybe empty)
Unsuccessful operation, returns optional AST, non-empty list of errors, list of warnings (maybe empty)

**Example**
One possible signature would be (tree *AST, errs []issue, warnings []issue)
where success is recognized by an empty "errs" list, and by convention the tree must be non-nil in this case.

**Alternatives considered**
Go doesn't have convenient algebraic types, otherwise we'd just express the return type directly:
Success of tree * list<issue>
| Failure of opt<tree> * nonemptylist<issue> * list<issue>

PR #148, opened 2019-01-11
High level interface definition for CEL

### LINKED ISSUE #126: Introduce a Shared Env between the Checker and Interpreter
**Feature request checklist**

**Change**
The `checker.Env` is currently only used at check time, but accepts the same parameters
as the `interpreter` construction methods. Ideally, there is a common environment, and it
is constructed once and shared many times with different check and interpret passes.

This simplifies setup, improves code-reuse, and makes it easier to verify that variables
are bound with types compatible with their declarations.

**Example**
```
env := common.NewStandardEnv(provider.NewTypeProvider())
   .Declare('a', types.StringType)
   .Declare('b', types.MapType)
checked, errors := checker.Check(parsed, env)
...
eval := intepreter.NewStandardInterpreter(checked.GetExpr(), env)
ctx := env.NewActivation()
   .Bind('a', "hello")
   .Bind('b', map[string]interface{} {"hello": ["you", "and", "yours"]})
return eval.Eval(ctx)
```

**Alternatives considered**
There are likely other ways to structure the interface, but the idea would be to share
the environment. A similar technique should also be considered with respect to binding
runtime functions via the `dispatcher`.

TristonianJones:
This is a first pass at creating a new high-level interface for CEL.

Core changes:

* `Env` vs. `Program`
  * `Env` is static and used for compiling many expressions.
  * `Program` is the executable view of a compiled AST.
* Functional options to configure the `Env`, `Program`, and per-execution `Eval`.
* `EvalOption` used to optimize the memory use and evaluation behavior.
* `Issues` are now returned from the `Parse` and `Check` calls to support the future
   introduction of non-fatal messages from expression compilation phases.

While the changes do not completely insulate the use of CEL from it's backing
protocol buffer types for expressions, they vastly reduce the surface area of the
library. Future improvements will try to reduce the friction and import set required
to use CEL.

Closes #126, #130


README.md
```diff
@@ -49,48 +49,32 @@ are any errors that need to be reported.
...
-    return nil, fmt.Error(errors.ToDisplayString())
+func main() {
+	// Variables used within this expression environment.
+	decls := cel.Declarations(
+		decls.NewIdent("i", decls.String, nil),
+		decls.NewIdent("you", decls.String, nil))
+	env, _ := cel.NewEnv(decls)
+
+	// Parse and type-check the expression.
+	p, _ := env.Parse(`"Hello " + you + "! I'm " + i + "."`)
```
JimLarson: Here and below, include proper checks of the result of each phase.  Yes, a little more wordy, but makes the example a lot more useful for real code.

cel/BUILD.bazel
```diff
@@ -0,0 +1,50 @@
...
+    ],
+    deps = [
+        "//checker/decls:go_default_library",
+        "//common/types:go_default_library",
+        "//common/types/ref:go_default_library",
+        "//common/types/traits:go_default_library",
+        "//interpreter/functions:go_default_library",
+        "@org_golang_google_genproto//googleapis/api/expr/v1alpha1:go_default_library",
+    ],
+)
```
JimLarson: Missing newline at EOF.

cel/cel_test.go
```diff
@@ -0,0 +1,261 @@
...
+		// Function to generate a greeting from one person to another.
+		//    i.greet(you)
+		decls.NewFunction("greet",
+			decls.NewInstanceOverload("greet_string_string",
+				[]*exprpb.Type{decls.String, decls.String},
+				decls.String)))
+	e, _ := NewEnv(decls)
+
+	// Parse and check the expression.
+	p, _ := e.Parse("i.greet(you)")
```
JimLarson: Especially in the example, illustrate error handling, here and below.

cel/io.go
```diff
@@ -0,0 +1,41 @@
...
+
+package cel
+
+import (
+	"github.com/google/cel-go/common"
+
+	exprpb "google.golang.org/genproto/googleapis/api/expr/v1alpha1"
+)
+
+// CheckedExprToAst converts a checked expression proto message to an Ast.
```
JimLarson, line 25: Please implement the inverse too, and for ParsedExpr as well.

cel/options.go
```diff
@@ -0,0 +1,127 @@
...
+	exprpb "google.golang.org/genproto/googleapis/api/expr/v1alpha1"
+)
+
+// EnvOption is a functional interface for configuring the environment.
+type EnvOption func(e *env) (*env, error)
+
+// ProgramOption is a functional interface for configuring a program.
+type ProgramOption func(p *prog) (*prog, error)
+
+// DisableBuiltIns option removes all standard operators from the environment.
```
JimLarson: Also repeat the note from Declarations:  if also using Declarations(), those must come after.

cel/options.go
```diff
@@ -0,0 +1,127 @@
...
+	"github.com/google/cel-go/common/types/pb"
+	"github.com/google/cel-go/interpreter"
+	"github.com/google/cel-go/interpreter/functions"
+	"github.com/google/cel-go/parser"
+
+	exprpb "google.golang.org/genproto/googleapis/api/expr/v1alpha1"
+)
+
+// EnvOption is a functional interface for configuring the environment.
+type EnvOption func(e *env) (*env, error)
```
JimLarson, line 33: Do you foresee returning a different *env than the one you were given?

cel/env.go
```diff
@@ -0,0 +1,196 @@
...
+	"github.com/google/cel-go/interpreter/functions"
+	"github.com/google/cel-go/parser"
+
+	exprpb "google.golang.org/genproto/googleapis/api/expr/v1alpha1"
+)
+
+// Ast interface representing the checked or unchecked expression, its source, and related metadata
+// such as source position information.
+type Ast interface {
+	IsChecked() bool
```
JimLarson, line 46: Consider putting the docstring on the interface methods and have the implementations reference the interface.

JimLarson (commented):
Checkpointing my comments.

cel/options.go
TristonianJones: I've opted to make the `EnvOption` implementations public, but make it impossible (for now) for external users to create their own. Perhaps in the future I'll open up this functionality.

TristonianJones:
PTAL

README.md
```diff
@@ -49,48 +49,53 @@ are any errors that need to be reported.
...
+	//
+	// Detailed information about which issues/errors were encountered may be
+	// found within the `iss`, but a non-nil `iss.Err()` value will contain a
+	// detailed error string in human-readable form.
+	p, iss := env.Parse(`"Hello " + you + "! I'm " + i + "."`)
+	if iss != nil && iss.Err() {
+		log.Fatalln(iss.Err())
+	}
+	c, iss := env.Check(p)
+	if iss != nil && iss.Err() {
```
JimLarson: Should be `iss.Err() != nil` here too.

README.md
```diff
@@ -49,48 +49,53 @@ are any errors that need to be reported.
...
+		log.Fatalf("environment creation error: %s\n", err)
+	}
+
+	// Parse and type-check the expression.
+	//
+	// Detailed information about which issues/errors were encountered may be
+	// found within the `iss`, but a non-nil `iss.Err()` value will contain a
+	// detailed error string in human-readable form.
+	p, iss := env.Parse(`"Hello " + you + "! I'm " + i + "."`)
+	if iss != nil && iss.Err() {
```
JimLarson: Should be `iss.Err() != nil`.
Meta: audit against the test code.

cel/cel.go
```diff
@@ -0,0 +1,19 @@
+// Copyright 2018 Google LLC
```
JimLarson: s/2018/2019/ for newly-created files?

cel/cel_test.go
```diff
@@ -0,0 +1,297 @@
...
+						"invalid key of type '%v' to obj.get(key, def)",
+						args[1].Type())
+				}
+				defVal := args[2]
+				if attrs.Contains(key) == types.True {
+					return attrs.Get(key)
+				}
+				return defVal
+			}})
+	prg, _ := e.Program(c, funcs, Globals(Vars(map[string]interface{}{
```
JimLarson, line 238: Add a comment to emphasize:
// Global variables can be set in the Program and optionally overriden in Eval.

cel/cel_test.go
```diff
@@ -0,0 +1,297 @@
...
+	}
+	out, err := prg.Eval(Vars(map[string]interface{}{"k": "key", "v": true}))
+	if err != nil {
+		t.Fatalf("runtime error: %s\n", err)
+	}
+	if out.Value().Equal(types.True) != types.True {
+		t.Errorf("Got '%v', expected 'true'", out.Value())
+	}
+
+	// Test to see whether 'v != false' was resolved to a value.
```
JimLarson, line 288: Test that both sides were evaluated - these tests shouldn't depend on whether we evaluate lhs or rhs first.

cel/env.go
```diff
@@ -0,0 +1,223 @@
...
+	// Errors returns the collection of errors encountered in more granular detail.
+	Errors() []common.Error
+}
+
+// NewEnv creates an Env instance suitable for parsing and checking expressions against a set of
+// user-defined constants, variables, and functions.
+//
+// See the EnvOptions for the options that can be used to configure the environment.
+func NewEnv(opts ...EnvOption) (Env, error) {
+	// Macros and the standard built-ins are enabled by default.
```
JimLarson: Move this to the external description.

cel/env.go
```diff
@@ -0,0 +1,223 @@
...
+// such as source position information.
+type Ast interface {
+	// Expr returns the proto serializable instance of the parsed/checked expression.
+	Expr() *exprpb.Expr
+
+	// IsChecked returns whether the Ast value has been successfully type-checked.
+	IsChecked() bool
+
+	// Metadata returns debug information about the expression, such as source position.
+	Metadata() *exprpb.SourceInfo
```
JimLarson: Not happy with the name, since it doesn't abstract over its return type. I'd say change it to SourceInfo(), but you could instead have it return a new Metadata type that abstracts SourceInfo.

cel/env.go
```diff
@@ -0,0 +1,223 @@
...
+
+	// ResultType returns the output type of the expression, or error if the IsChecked() == false.
```
JimLarson: Consider returning Dyn if IsChecked() == false - I'll bet the behavior for checked && Dyn would be the same for unchecked in all cases, so don't make every caller implement this logic.

cel/env.go
```diff
@@ -0,0 +1,223 @@
...
+// evaluable programs from parsed and checked Asts.
+type Env interface {
+	// Check performs type-checking on the input Ast and yields a checked Ast and/or set of Issues.
+	//
+	// Checking has failed if the returned Issues value and its Issues.Err() value is non-nil.
+	// In general Issues should be inspected if they are non-nil, but may not represent a fatal
+	// error.
+	Check(ast Ast) (Ast, Issues)
+
+	// Parse parses the input txt to a Ast and/or a set of Issues.
```
JimLarson: Be clear that "input" means a CEL expression.

cel/options.go
JimLarson: My point is that if you might want to drop the `*env` from the return type if it's never going to be different than what is passed in.

Sometimes you might want to return one of your unmodified arguments for a "builder" pattern:
```
foo := NewFooBuilder().SetName("bar").SetLevel(7).Build()
```
but here returning a pair with `error` prevents that, and it looks like you're going to varargs the options rather than chain them.

cel/env.go
```diff
@@ -0,0 +1,223 @@
...
+	//
+	// Parsing has failed if the returned Issues value and its Issues.Err() value is non-nil.
+	// In general Issues should be inspected if they are non-nil, but may not represent a fatal
+	// error.
+	Parse(txt string) (Ast, Issues)
```
JimLarson, line 81: Add options for parse.  For instance, source location and syntax version.

cel/options.go
```diff
@@ -0,0 +1,190 @@
...
+//
+// Note: This option must be specified after ClearBuiltIns and/or ClearMacros if used together.
+func Macros(macros ...parser.Macro) EnvOption {
+	return func(e *env) (*env, error) {
+		e.macros = append(e.macros, macros...)
+		return e, nil
+	}
+}
+
+// Package sets the container for resolving variable names. Defaults to an empty package.
```
JimLarson: Rename to "Container".  Or declare that we'll standardize on "Package" and rename "Container" to "Package" in our protos, etc.

cel/env.go
```diff
@@ -0,0 +1,223 @@
...
+	// Check performs type-checking on the input Ast and yields a checked Ast and/or set of Issues.
+	//
+	// Checking has failed if the returned Issues value and its Issues.Err() value is non-nil.
+	// In general Issues should be inspected if they are non-nil, but may not represent a fatal
+	// error.
+	Check(ast Ast) (Ast, Issues)
+
+	// Parse parses the input txt to a Ast and/or a set of Issues.
+	//
+	// Parsing has failed if the returned Issues value and its Issues.Err() value is non-nil.
```
JimLarson, line 76: Note that a non-nil Ast might be returned even in the case of a parsing error, in the hope that partial results might be useful, but the returned Ast is not necessarily valid for use in later phases in this case.

cel/env.go
```diff
@@ -0,0 +1,223 @@
...
+	Source() common.Source
+}
+
+// Env defines functions for parsing and type-checking expressions against a set of user-defined
+// constants, variables, and functions. The Env interface also defines a method for generating
+// evaluable programs from parsed and checked Asts.
+type Env interface {
+	// Check performs type-checking on the input Ast and yields a checked Ast and/or set of Issues.
+	//
+	// Checking has failed if the returned Issues value and its Issues.Err() value is non-nil.
```
JimLarson, line 67: Note that a non-nil Ast might be returned even in the case of a check error, in the hope that partial results might be useful, but the returned Ast is not necessarily valid for use in later phases in this case.

cel/env.go
```diff
@@ -0,0 +1,223 @@
...
+}
+
+// Parse implements the Env interface method.
+func (e *env) Parse(txt string) (Ast, Issues) {
+	src := common.NewTextSource(txt)
+	res, errs := parser.ParseWithMacros(src, e.macros)
+	if len(errs.GetErrors()) > 0 {
+		return nil, &issues{errs: errs}
+	}
+	return &astValue{
```
JimLarson, line 204: How about using ParsedExprToAst here?

cel/env.go
```diff
@@ -0,0 +1,223 @@
...
+	res, errs := checker.Check(
+		&exprpb.ParsedExpr{
+			Expr:       ast.Expr(),
+			SourceInfo: ast.Metadata()},
+		ast.Source(),
+		ce)
+	if len(errs.GetErrors()) > 0 {
+		return nil, &issues{errs: errs}
+	}
+	return &astValue{
```
JimLarson, line 187: How about using CheckedExprToAst here?

cel/cel_test.go
```diff
@@ -0,0 +1,297 @@
...
+	}))
+	if err != nil {
+		log.Fatalf("runtime error: %s\n", err)
+	}
+
+	fmt.Println(out.Value())
+	// Output:Hello world! Nice to meet you, I'm CEL.
+}
+
+func Test_ExampleWithBuiltins(t *testing.T) {
```
JimLarson, line 88: Review the use of "builtin" in the API:  we seem to be using it both for the set of all built-in identifiers and macros, and for extension functions added to the run-time.  These should be distinct concepts with distinct names.

parser/macro.go
```diff
@@ -25,102 +25,230 @@ import (
...
+	// Function name to match.
+	Function() string
+
+	// ArgCount for the function call.
+	ArgCount() int
+
+	// IsReceiverStyle returns true if the macro matches a receiver style call.
+	IsReceiverStyle() bool
+
+	// MacroKey returns the macro signatures accepted by this macro.
```
JimLarson, line 80: Give a reference to the notation for "macro signature".

parser/macro.go
```diff
@@ -25,102 +25,230 @@ import (
...
 
-const accumulatorName = "__result__"
+var (
+	// AllMacros includes the list of all spec-supported macros.
+	AllMacros = []Macro{
+		// The macro "has(m.f)" which tests the presence of a field, avoiding the need to specify
+		// the field as a string.
+		NewGlobalMacro(operators.Has, 1, makeHas),
+
+		// The macro "range.all(var, predicate)", which is true if for all elements in range the
```
JimLarson, line 234: We had discussed limiting "comprehensions" to operating on lists only, making the map-key-to-list conversion explicit and deterministic.

cel/env.go
```diff
@@ -0,0 +1,223 @@
...
+
+	// Metadata returns debug information about the expression, such as source position.
+	Metadata() *exprpb.SourceInfo
+
+	// ResultType returns the output type of the expression, or error if the IsChecked() == false.
+	ResultType() (*exprpb.Type, error)
+
+	// Source returns a view of the input used to create the Ast. This source may be complete or
+	// constructed from the Metadata.
+	Source() common.Source
```
JimLarson: Here and below:  so this makes common.Source part of the public API.  Other API methods also return types outside of the cel package.  Do we want to eventually make this API self-contained?

parser/macro.go
```diff
@@ -25,102 +25,230 @@ import (
...
+
+// Macro interface for describing the function signature to match and the MacroExpander to apply.
+//
+// Note: when a Macro should apply to multiple overloads (based on arg count) of a given function,
+// a Macro should be created per arg-count.
+type Macro interface {
+	// Function name to match.
+	Function() string
+
+	// ArgCount for the function call.
```
JimLarson, line 71: Explain what this returns for VarArg macros.

parser/macro.go
```diff
@@ -25,102 +25,230 @@ import (
...
-func (m *Macro) GetArgCount() int {
-	return m.args
+// ArgCount returns the number of arguments the macro expects.
+func (m *macro) ArgCount() int {
+	return m.argCount
 }
 
-// GetIsInstanceStyle returns whether the macro is "instance" (reciever) style.
-func (m *Macro) GetIsInstanceStyle() bool {
+// IsInstanceStyle returns whether the macro is "instance" (reciever) style.
```
JimLarson: s/Instance/Receiver/

parser/macro.go
```diff
@@ -25,102 +25,230 @@ import (
...
+	// optional set of field initializers.
+	NewObject(typeName string, fieldInits ...*exprpb.Expr_CreateStruct_Entry) *exprpb.Expr
+
+	// NewObjectFieldInit creates a new Object field initializer from the field name and value.
+	NewObjectFieldInit(field string, init *exprpb.Expr) *exprpb.Expr_CreateStruct_Entry
+
+	// Fold creates a fold comprehension instruction.
+	//
+	// - iterVar is the iteration variable name.
+	// - iterRange represents the expression that resolves to a list or map where the elements or
```
JimLarson, line 188: Again, need to decide whether to continue direct support for maps.

parser/macro.go
```diff
@@ -25,102 +25,230 @@ import (
...
+	//   keys (repsectively) will be iterated over.
+	// - accuVar is the accumulation variable name, typically parser.AccumulatorName.
```
JimLarson, line 190: Explain constraints on this.  For instance, it shouldn't shadow a variable that you want to access from step or condition.

parser/macro.go
```diff
@@ -25,102 +25,230 @@ import (
...
+		accuVar string,
+		accuInit *exprpb.Expr,
+		condition *exprpb.Expr,
+		step *exprpb.Expr,
+		result *exprpb.Expr) *exprpb.Expr
+
+	// Ident creates an identifier Expr value.
+	Ident(name string) *exprpb.Expr
+
+	// GlobalCall creates a function call Expr value for a global (free) function.
```
JimLarson, line 211: Meh, the member calls are global too.  How about just plain Call() and MemberCall() (or even better, ReceiverCall())?

JimLarson (changes requested):

cel/env.go
TristonianJones: Changed to `SourceInfo`.

cel/env.go
TristonianJones: Good suggestion.

cel/env.go
TristonianJones: The functional option pattern within Go seems to be pretty common. When multiple versions are supported, we should be able to add:

```go
type SourceOption func(cel.Source) (cel.Source, err)
// ... meanwhile ...
Parse(txt string, opts ...SourceOption) (Ast, Issues)
```

I've added a top-level `cel.Source` wrapper around `common.Source` in order to permit this sort of feature in the future.

cel/env.go
TristonianJones: Added a comment about why not, but I did use `AstToParsedExpr` within the `Check`

cel/env.go
TristonianJones: Added a comment about why not, but very reasonable suggestion.

cel/cel_test.go
TristonianJones: This is actually using the builtin operators to perform the same operation as the `Test_Example` does, rather than using extension functions.

cel/env.go
TristonianJones: I created a top-level wrapper for `cel.Source` and opted for it instead.

parser/macro.go
TristonianJones: Folks are already using `map` values as input to these comprehensions, so I think we can improve how they function, but can't remove the feature in the short-term.

TristonianJones:
Thanks for the review PTAL.

JimLarson (approved):

PR #152, opened 2019-01-19
Thread-safe evaluation state

TristonianJones:
Thread-safe evaluation state.

The existing implementation shares a single `MutableEvalState` object across all `Eval` calls.
This is incredibly fast, but not thread-safe as concurrent go-routines might clobber each others
evaluation outcomes. Additionally, the output `EvalState` is dirty, thus making it unsuitable for
multiple evaluations within the context of `Program` instances created with
`NewExhaustiveProgram`.

This PR includes the following changes:

*  Removes the `MutableEvalState` interface.
*  Pools runtime state allocations behind a thread-safe `sync.Pool`
*  Adds a `programFlag` concept which specifies behaviors such as state tracking
    and disabling short-circuiting.
*  Optionally returns a non-nil `EvalState` with the evaluation result if state-tracking
    is flag enabled.

### DIFF
--- a/interpreter/astwalker.go
+++ b/interpreter/astwalker.go
@@ -436,7 +436,7 @@ func (w *astWalker) getID(expr *exprpb.Expr) int64 {
 	id := expr.GetId()
 	if ident := expr.GetIdentExpr(); ident != nil {
 		if altID, found := w.scope.ref(ident.Name); found {
-			w.state.SetRuntimeExpressionID(id, altID)
+			w.state.setRuntimeExpressionID(id, altID)
 			return altID
 		}
 	}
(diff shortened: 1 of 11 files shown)


TristonianJones:
Thanks @kyessenov for reporting this issue. Subsequent changes further improve performance, but this specific issue affects both performance _and_ correctness in concurrent environments.

TristonianJones:
Note this PR will be followed by:
https://github.com/TristonianJones/cel-go/tree/val-or-err
https://github.com/TristonianJones/cel-go/tree/bind-dispatch

The sum total of these changes conflict slightly with some of the code present within pull #148.
Once this trifecta of changes is submitted, I'll update PR #148 to align with this code.

TristonianJones:
This pull request has been replaced by the Recursive Instruction Planner (#157) which is thread-safe and significantly faster.

Issue #149, opened 2019-01-15
Proto2 primitive values not converting to CEL types


TristonianJones:
**Describe the bug**
Proto2 fields are code generated as pointers to primitive values. This appears to be causing issues with field selection. 

Seems related to Issue #35, where `ConvertToNative` did not handle pointer primitives correctly for proto2 values, but in this case with respect to converting the other way using `NativeToValue`.

**To Reproduce**
Check which components this affects:

- [ ]  parser
- [ ]  checker
- [X]  interpreter

Sample expression and input that reproduces the issue:
```cel
// a is a proto2 message
// a == Foo{ a: 1 }
a.level
```

Test setup:
```go
	src := common.NewTextSource("a.level")
	pe, errs := parser.Parse(src)
	if len(errs.GetErrors()) != 0 {
		return errors.New(errs.ToDisplayString())
	}

	typeProvider := types.NewProvider(
		&pb.Foo{},
	)

	env := checker.NewStandardEnv(packages.DefaultPackage, typeProvider)
	err := env.Add(
		decls.NewIdent("a", decls.NewObjectType("Foo"), nil),
	)
	if err != nil {
		return errors.Wrap(err, "add declarations to environment")
	}
	ce, errs := checker.Check(pe, src, env)
	if len(errs.GetErrors()) != 0 {
		return errors.New(errs.ToDisplayString())
	}
	i := interpreter.NewStandardInterpreter(packages.DefaultPackage, typeProvider)
	eval := i.NewInterpretable(interpreter.NewCheckedProgram(ce))
	result, state := eval.Eval(
		interpreter.NewActivation(
			map[string]interface{}{
				"a": types.NewObject(&Foo{bar: proto.Int(1)),
			},
		),
	)
	fmt.Println(result)
	match, ok := result.Value().(bool)
	if !ok {
		return false, fmt.Errorf("expression result is type %q, not bool", result.Type().TypeName())
	}
	return match, nil
```

**Expected behavior**
Expect that `a.level` returns a `types.Int(1)`

PR #153, opened 2019-01-20
Proto2 primitive type conversion fix.

TristonianJones:
Closes issue #149 by adding appropriate `NativeToValue` conversions for `proto2` primitive fields.
In `proto2` primitive fields are generated as pointers to primitive go types, whereas in `proto3` the
simple type is used, e.g. `string` (proto3) vs. `*string` (proto2).


interpreter/interpreter_test.go
```diff
@@ -389,27 +391,128 @@ func TestInterpreter_BuildObject(t *testing.T) {
 	}
 }
 
+func TestInterpreter_GetProto2PrimitiveFields(t *testing.T) {
+	src := common.NewTextSource(`
+	a.single_int32 == a.single_int64 &&
```
madi8229, line 398: Do we have any comments to explain why this works while CEL does not support type conversion? Thank you.

interpreter/interpreter_test.go
```diff
@@ -389,27 +391,128 @@ func TestInterpreter_BuildObject(t *testing.T) {
...
+		SingleDouble: &six,
+		SingleString: &str,
+		SingleBool:   &truth,
+	}
+	result, state := eval.Eval(NewActivation(map[string]interface{}{
+		"input": input,
+	}))
+	got, ok := result.(ref.Value).Value().(bool)
+	if !ok {
+		t.Fatalf("cannot cast result to int64: result=%v state=%v", result, state)
```
madi8229: We have several fields here, why we know if it is not ok, this is the reason? Thank you.

interpreter/interpreter_test.go
TristonianJones: In CEL, 32-bit proto types are widened to 64-bits, so even though the implicit type conversion would be an error normally, the widening happens at the object layer and CEL doesn't consider them different types.

interpreter/interpreter_test.go
TristonianJones: Comment updated to match the actual conditions under test.

madi8229 (approved):
LGTM

TristonianJones:
Closes #149

Issue #155, opened 2019-01-22
conformance: mixed types' equality yields false where it should evaluate to an error


gabrielsz:
**Describe the bug**
The CEL Spec determines that evaluating equality on different types yields a "found no matching overload" (except for specific listed cases). While the type checker behaves correctly, the interpreter yields an incorrect `false` result.

**To Reproduce**
Check which components this affects:

- [ ]  parser
- [ ]  checker
- [X]  interpreter

Sample expression and input that reproduces the issue:
```cel
1.0 == 1
```

Test setup:
```cel-conformance
  test {
    name: "self_test_equals_mixed_types"
    expr: "1.0 == 1"
    disable_check: true # need to make it fail in the evaluation phase
    eval_error : {
      errors: { message: "found no matching overload for '_==_' applied to '(double, int)'" }
    }
  }
  test {
    name: "self_test_not_equals_data_type"
    expr: "[1] == [1.0]"
    disable_check: true # need to make it fail in the evaluation phase
    eval_error: {
      errors: { message: "found no matching overload for '_==_' applied to '(list(int), list(double))'" }
    }
  }
```

**Expected behavior**
The expected behavior for the interpreter is to yield a "found no matching overload" error.

Issue #25, opened 2018-05-09
Unify error codes / messages


TristonianJones:
## Issue Filing Checklist

- [X] There are no issues that match the desired change
- [X] The change is large enough it can't be addressed with a simple [Pull Request](./PULL_REQUEST_TEMPLATE.md) 

## Change
Error messages are formatted strings which are not uniform. While this was
okay for development, the errors should be consolidated into error codes
with a set of associated details. Further, when multiple errors are encountered
they should be aggregated together rather than simply picking the first one.

Proposed error codes:

| Code | Details | Description |
| ----- | ------- | ----------- |
|`ATTRIBUTE_NOT_FOUND` | type, attribute | Map or object access using an unknown attribute. |
| `DIVIDE_BY_ZERO` | n/a | Division by zero, also reported for modulus operations. |
| `DUPLICATE_ATTRIBUTE` | attribute name | Map or object construction supplies same key value more than once. |
| `IDENTIFIER_NOT_FOUND` | identifier name | Variable or function name could not be found. |
| `INVALID_ARGUMENT` | argument value | Invalid argument supplied to a function. |
| `OVERLOAD_NOT_FOUND` | function signature | Function defined, but no matching overload found. |
| `TYPE_NOT_FOUND` | type name | Type definition could not be found. |

### Example

```go
return types.NewError(errors.ATTRIBUTE_NOT_FOUND, value.Type(), field)
```

josephschorr:
Hi @TristonianJones, any updates on this? We would find it incredibly useful to be able to provide better errors messages when CEL evaluation fails.

Happy to help if need be.

PR #157, opened 2019-01-25
Recursive instruction planner

TristonianJones:
The recursive instruction planner walks the `Expr` AST to generate an `Interpretable` call
tree whose branches represent linear evaluation paths. 

This approach offers the following advantages over prior approaches:

* Thread-safe evaluation that leverages `Go` call frames.
* Reduced evaluation overhead by removed a custom alloc to track eval state.
* Specialization of the `&&`, `||`, `? :`, and (in)equality operators.

The improvements in execution are obvious from the benchmarks:

```
benchmark                                       old ns/op     new ns/op     delta
BenchmarkInterpreter/ExprBench/ok_1st-12        615           128           -79.19%
BenchmarkInterpreter/ExprBench/ok_2nd-12        1787          531           -70.29%
BenchmarkInterpreter/ExprBench/not_found-12     1778          535           -69.91%
BenchmarkInterpreter/ExprBench/false_2nd-12     309           11.6          -96.25%

benchmark                                       old allocs     new allocs     delta
BenchmarkInterpreter/ExprBench/ok_1st-12        1              1              +0.00%
BenchmarkInterpreter/ExprBench/ok_2nd-12        5              4              -20.00%
BenchmarkInterpreter/ExprBench/not_found-12     5              4              -20.00%
BenchmarkInterpreter/ExprBench/false_2nd-12     0              0              +0.00%

benchmark                                       old bytes     new bytes     delta
BenchmarkInterpreter/ExprBench/ok_1st-12        8             8             +0.00%
BenchmarkInterpreter/ExprBench/ok_2nd-12        104           88            -15.38%
BenchmarkInterpreter/ExprBench/not_found-12     104           88            -15.38%
BenchmarkInterpreter/ExprBench/false_2nd-12     0             0             +0.00%

```

This PR also includes the following functionality changes:

* Functional `InterpretableDecorator` values may be provided at `Interpretable` creation time
  to layer or replace default `Interpretable` behavior. See `interpreter.ExhaustiveEval`, and
  `interpreter.FoldConstants`.
* `Interpretable` creation now considers the `exprpb.CheckedExpr` metadata to resolve namespaced
   identifiers within `Select` expressions to enum and identifier references.
* Equality checks now consider types inequality to be an error rather than `false`. #155 
* Specialized `map[string]string` handling for field accesses.

This PR replaces #152.


TristonianJones:
@kyessenov the latest performance numbers should be in line with existing Istio CEXL microbenchmarks.

TristonianJones:
Closes #159 and #155

common/types/bool.go
```diff
@@ -43,7 +43,7 @@ var (
 // Compare implements traits.Comparer.Compare.
 func (b Bool) Compare(other ref.Value) ref.Value {
 	if BoolType != other.Type() {
-		return NewErr("unsupported overload")
+		return ValOrErr(other, "no such overload")
```
JimLarson, line 46: Consider defining constants for "no such overload" and the other standard errors, rather than formatting and constructing them each time.  Prevents misspellings.

common/types/double.go
```diff
@@ -119,13 +119,16 @@ func (d Double) Divide(other ref.Value) ref.Value {
 
 // Equal implements ref.Value.Equal.
 func (d Double) Equal(other ref.Value) ref.Value {
-	return Bool(DoubleType == other.Type() && d == other.(Double))
+	if DoubleType != other.Type() {
+		return ValOrErr(other, "no such overload")
+	}
+	return Bool(d == other.(Double))
```
JimLarson, line 126: Add a TODO for handling NaNs.  I think we decided that all NaNs are equal, but I believe that in Go NaNs are unequal even to themselves.

common/types/json_list_test.go
```diff
@@ -134,7 +134,7 @@ func TestJsonListValue_ConvertToType(t *testing.T) {
 
 func TestJsonListValue_Equal(t *testing.T) {
 	listA := NewJSONList(&structpb.ListValue{Values: []*structpb.Value{
-		{Kind: &structpb.Value_StringValue{StringValue: "hello"}},
+		{Kind: &structpb.Value_NumberValue{NumberValue: -3}},
```
JimLarson, line 137: Why are you avoiding the heterogeneous list?

common/types/json_struct.go
```diff
@@ -40,7 +40,7 @@ func NewJSONStruct(st *structpb.Struct) traits.Mapper {
 }
 
 func (m *jsonStruct) Contains(index ref.Value) ref.Value {
-	return !Bool(IsError(m.Get(index).Type()))
+	return Bool(m.Get(index).Type() != ErrType)
```
JimLarson: Why not IsError here?

common/types/json_struct.go
```diff
@@ -112,20 +112,23 @@ func (m *jsonStruct) Equal(other ref.Value) ref.Value {
 	it := m.Iterator()
 	for it.HasNext() == True {
 		key := it.Next()
-		if otherVal := otherMap.Get(key); IsError(otherVal.Type()) {
-			return False
-		} else if thisVal := m.Get(key); IsError(thisVal.Type()) {
+		if otherVal := otherMap.Get(key); IsError(otherVal) {
 			return False
-		} else if thisVal.Equal(otherVal) != True {
+		} else if thisVal := m.Get(key); IsError(thisVal) {
```
JimLarson, line 117: Here and for lists:  I believe that this should result in a propagation of the error, rather than False.

common/types/map.go
```diff
@@ -28,11 +28,24 @@ type baseMap struct {
 	refValue reflect.Value
 }
 
+type stringMap struct {
```
JimLarson, line 34: Document the reason we're making a specialization for string &rarr; string.

common/types/map.go
```diff
@@ -125,17 +145,27 @@ func (m *baseMap) Equal(other ref.Value) ref.Value {
 	it := m.Iterator()
 	for it.HasNext() == True {
 		key := it.Next()
-		if otherVal := otherMap.Get(key); IsError(otherVal.Type()) {
+		if otherVal := otherMap.Get(key); IsError(otherVal) {
 			return False
```
JimLarson, line 152: Might propagate the error here.  See above for maps.

interpreter/activation.go
```diff
@@ -99,15 +96,40 @@ func (a *hierarchicalActivation) ResolveName(name string) (ref.Value, bool) {
...
+	}
+}
+
+type varActivation struct {
+	parent Activation
+	name   string
+	val    ref.Value
+}
+
+func (v *varActivation) ResolveName(name string) (ref.Value, bool) {
```
JimLarson, line 127: Docstring

interpreter/decorators.go
```diff
@@ -0,0 +1,291 @@
...
+	accuCtx.val = fold.accu.Eval(ctx)
+	iterCtx := varActivationPool.Get().(*varActivation)
+	iterCtx.parent = accuCtx
+	iterCtx.name = fold.iterVar
+	it := foldRange.(traits.Iterable).Iterator()
+	for it.HasNext() == types.True {
+		// Modify the iter var in the fold activation.
+		iterCtx.val = it.Next()
+
+		// Evaluate the condition, terminate the loop if false.
```
JimLarson: This doesn't seem to terminate the loop.  Update the comment?

interpreter/evalstate.go
```diff
@@ -20,79 +20,45 @@ import (
...
-	s.exprValues[exprID] = value
-}
-
-func (s *defaultEvalState) Value(exprID int64) (ref.Value, bool) {
+// Value is an implementation of the EvalState interface method.
+func (s *evalState) Value(exprID int64) (ref.Value, bool) {
 	// TODO: The eval state assumes a dense progrma expression id space. While
 	// this is true of how the cel-go parser generates identifiers, it may not
 	// be true for all implementations or for the long term. Replace the use of
 	// parse-time generated expression ids with a dense runtiem identifier.
```
JimLarson: s/runtiem/runtime/

interpreter/planner.go
```diff
@@ -0,0 +1,1089 @@
...
+//  b) creates a field presence test for a select within a has() macro.
+//  c) resolves the select expression to a namespaced identifier.
+func (p *planner) planSelect(expr *exprpb.Expr) (Interpretable, error) {
+	sel := expr.GetSelectExpr()
+	// If the Select was marked TestOnly, this is a presence test.
+	if sel.TestOnly {
+		return p.planTestOnly(expr)
+	}
+	// If the Select id appears in the reference map from the CheckedExpr proto
+	// then it is either a namespaced identifier or enum value.
```
JimLarson: Put in a TODO to ensure checked and parsed ASTs have the same semantics with respect to dotted identifiers.

interpreter/planner.go
```diff
@@ -0,0 +1,1089 @@
...
+
+// planSelect creates an Interpretable with either:
+//  a) selects a field from a map or proto.
+//  b) creates a field presence test for a select within a has() macro.
+//  c) resolves the select expression to a namespaced identifier.
+func (p *planner) planSelect(expr *exprpb.Expr) (Interpretable, error) {
+	sel := expr.GetSelectExpr()
+	// If the Select was marked TestOnly, this is a presence test.
+	if sel.TestOnly {
+		return p.planTestOnly(expr)
```
JimLarson: Strange that this is the only case sent to a helper.  Consider inlining.

interpreter/planner.go
```diff
@@ -0,0 +1,1089 @@
...
+	rVal := or.rhs.Eval(ctx)
+	rBool, rok := rVal.(types.Bool)
+	if rok && rBool == types.True {
+		return types.True
+	}
+	// return if both sides are bool false.
+	if lok && rok {
+		return types.False
+	}
+	// prefer left unknown to right unknown.
```
JimLarson, line 720: Add a TODO to return both if both are unknown or error.

interpreter/planner.go
```diff
@@ -0,0 +1,1089 @@
...
+	if types.IsUnknown(lVal) {
+		return lVal
+	}
+	if types.IsUnknown(rVal) {
+		return rVal
+	}
+	// if the left-hand side is non-boolean return it as the error.
+	return types.ValOrErr(lVal, "Got '%v', expected argument of type 'bool'", lVal)
```
JimLarson: Maybe "no such overload"?

JimLarson (commented):
Thanks for the copyright notices on the server files!

common/types/map.go
TristonianJones: Since there are two reasons why the `Get` could fail, I'm beginning to understand why you might want to consider this an error. Value conversion errors are somewhat uncommon, but still possible through this path. The more common error is the `no such key` error which indicates absence of the key from the map.

I'll add a TODO to refine this behavior. I had played around with alternative representations which might help disambiguate these two error conditions, but it's not ready for incorporation here.

common/types/json_struct.go
TristonianJones: The `Get()` call will error if the key is not found, but since this is an equality test, absence of the key should yield `false`. The error is propagated if the element comparison yields an error though. Same for lists.

common/types/json_list_test.go
TristonianJones: Previously, comparing two values of different types would yield `false`, but per the spec, this should be `no such overload`. The shift to error also affected list equality. I've made the list heterogeneous, but comparable now.

common/types/bool.go
TristonianJones: I filed issue #25 to scrub and standardize the error messaging. It's an older issue that I would be happy to include in a CEL 0.2.0 Milestone. For now I've added a TODO on the `ValOrErr` and `NewErr` functions to make sure this gets looked at more carefully.

common/types/json_struct.go
JimLarson: Perhaps Get() should return (ref.Value, bool) and be used accordingly?

JimLarson (approved):

common/types/json_struct.go
TristonianJones: I need to think about it a bit more, but that seems like a reasonable suggestion. I've added a TODO to refine the logic so I'll get back to this in another PR.

Issue #160, opened 2019-01-29
Tree-balance logical operators at parse-time.


TristonianJones:
**Feature request checklist**

- [X] There are no issues that match the desired change
- [X] The change is large enough it can't be addressed with a simple Pull Request
- [ ] If this is a bug, please file a [Bug Report](./ISSUE_TEMPLATE).

**Change**
Tree-balance `&&`, `||` expressions with repeated terms to reduce proto recursion on serialization
and evaluation.

**Alternatives considered**
Support variadic `&&`, `||` functions in the runtime. This would be preferable, but requires changes in
the parser, type-checker, and runtime rather than in just the parser. Eventually, this will become the
canonical form per https://github.com/google/cel-spec/issues/32 as it is easier to support long-term,
but requires more coordination for rollout.

PR #161, opened 2019-01-30
Tree-balanced logical ops.

TristonianJones:
To ensure that ASTs are represented as compactly as possible, the logical `&&`, `||` operator terms
have been tree-balanced. These operators can create highly skewed DAGs for larger expressions
which can result in issues with the recursion limits present during proto-deserialization. While this
is an unconventional approach that may change the order of evaluation of certain terms, it will not
change the evaluation outcome. 

Closes Issue #160


madi8229 (approved):
LGTM.

Issue #162, opened 2019-01-30
Builtin support for string contains, endsWith, startsWith


TristonianJones:
**Feature request checklist**

- [X] There are no issues that match the desired change
- [X] The change is large enough it can't be addressed with a simple Pull Request

**Change**
Add built-ins for string `contains`, `endsWith`, and `startsWith`.

**Example**
```
'goodbye'.contains('db') && 'goodbye'.startsWith('good') && 'goodbye'.endsWith('bye')
```

**Alternatives considered**
The alternative would be to provide these as extension functions rather than as part of the
standard built-in set; however, their usage seems common enough to warrant inclusion in
the standard set.

PR #163, opened 2019-01-31
Builtins for string contains, endsWith, and startsWith

TristonianJones:
Closes issues #162


common/types/string.go
```diff
@@ -38,7 +40,14 @@ var (
 		traits.AdderType,
 		traits.ComparerType,
 		traits.MatcherType,
+		traits.ReceiverType,
 		traits.SizerType)
+
+	stringOneArgOverloads = map[string]func(String, ref.Value) ref.Value{
```
gabrielsz, line 46: Question w/o a context: Shouldn't the stringMatches be here?

gabrielsz (approved):

common/types/string.go
TristonianJones: There's a builtin `traits.MatcherType` which other types can implement to support a `matches` function as a first class call. These particular functions are pretty specific to strings though.

PR #166, opened 2019-02-05
Fix the eval option init values and decorator lists

TristonianJones:
Decorator lists were growing indefinitely and the options for the CEL eval modes were incorrectly started at 0 rather than 1 making the state tracking always enabled.

### DIFF
--- a/cel/cel_test.go
+++ b/cel/cel_test.go
@@ -156,7 +156,7 @@ func Test_CustomTypes(t *testing.T) {
 				decls.NewObjectType("google.api.expr.v1alpha1.Expr"), nil)))
 
 	p, _ := e.Parse(`
-		expr == Expr{id: 2, 
+		expr == Expr{id: 2,
 			call_expr: Expr.Call{
 				function: "_==_",
 				args: [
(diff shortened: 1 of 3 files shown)


cel/cel_test.go
```diff
@@ -305,3 +305,45 @@ func Test_EvalOptions(t *testing.T) {
...
+	cast, _ := e.Check(past)
+	vars := Vars(map[string]interface{}{
+		"ai": 2,
+		"ar": map[string]string{
+			"foo": "bar",
+		},
+	})
+
+	b.Run("track-state", func(bb *testing.B) {
+		prg, _ := e.Program(cast, EvalOptions(OptTrackState))
```
gabrielsz, line 326: Hi! This could be smaller if you loop over the EvalOptions parameter (or create a closure with it)

cel/options.go
```diff
@@ -168,7 +168,7 @@ type EvalOption int
 
 const (
 	// OptTrackState will cause the runtime to return an immutable EvalState value in the Result.
-	OptTrackState EvalOption = iota << 1
```
gabrielsz, line 171: I've seen in examples that people declare a `_ = iota` to force the numbers to start at 1

TristonianJones: Thanks, I realized I wrote this expression backwards. What I wanted was `1 << iota` rather than the other way around.

cel/cel_test.go
TristonianJones: Done.

PR #167, opened 2019-02-05
CEL Interface change to reduce allocations.

TristonianJones:
Remove `cel.Result` from the CEL interface preferring to return an unwrapped `ref.Val`
with a possible non-nil `cel.EvalDetails`. This subtle change reduces allocations in the
common case.

### DIFF
--- a/cel/cel_test.go
+++ b/cel/cel_test.go
@@ -110,8 +110,9 @@ func Test_ExampleWithBuiltins(t *testing.T) {
 	if err != nil {
 		t.Fatalf("program creation error: %s\n", err)
 	}
+
 	// If the Eval() call were provided with cel.EvalOptions(OptTrackState) the details response
-	// (2nd return) would be non-nill
+	// (2nd return) would be non-nil.
 	out, _, err := prg.Eval(Vars(map[string]interface{}{
 		"i":   "CEL",
 		"you": "world"}))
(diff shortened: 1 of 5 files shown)


cel/cel_test.go
```diff
@@ -110,15 +110,17 @@ func Test_ExampleWithBuiltins(t *testing.T) {
 	if err != nil {
 		t.Fatalf("program creation error: %s\n", err)
 	}
-	out, err := prg.Eval(Vars(map[string]interface{}{
+	// If the Eval() call were provided with cel.EvalOptions(OptTrackState) the details response
```
gabrielsz, line 114: Would you mind adding a `\n` before this comment?

cel/cel_test.go
```diff
@@ -187,7 +189,7 @@ func Test_CustomTypes(t *testing.T) {
 			},
 		},
 	}}))
-	if !out.Value().Equal(types.True).(types.Bool) {
+	if !out.Equal(types.True).(types.Bool) {
```
gabrielsz: If the `Equal` method can return something other than `types.Bool` (you mentioned an error), this could panic. In this case I'd recommend to use the double return and check both

gabrielsz (approved):
LGTM. Just a minor change in a test, which doesn't need a re-review

cel/cel_test.go
TristonianJones: Thanks. This could have been written much more simply. Updated.

PR #168, opened 2019-02-08
Isolated Type Declarations and Descriptor Declarations

JimLarson:
The protocol buffer registry in common/types/pb contains global mutable state.  This state is now pulled into a struct, the PbDb, with a global variable for the default PbDb.  When using "isolated" types, the registry is copied to private state.  Also introduces the ability to provide message declarations via FileDescriptorProto directly.

### DIFF
--- a/cel/cel_test.go
+++ b/cel/cel_test.go
@@ -195,6 +195,43 @@ func Test_CustomTypes(t *testing.T) {
 	}
 }
 
+func Test_IsolatedTypes(t *testing.T) {
+	src := `expr == Expr{id: 2,
+		call_expr: Expr.Call{
+			function: "_==_",
+			args: [
+				Expr{id: 1, ident_expr: Expr.Ident{ name: "a" }},
+				Expr{id: 3, ident_expr: Expr.Ident{ name: "b" }}]
+		}}`
+	e, err := NewEnv(
+		Container("google.api.expr.v1alpha1"),
+		IsolateTypes(),
+		Types(&exprpb.Expr{}),
+		Declarations(
+			decls.NewIdent("expr",
+				decls.NewObjectType("google.api.expr.v1alpha1.Expr"), nil)))
+	if err != nil {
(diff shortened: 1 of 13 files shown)


cel/cel_test.go
```diff
@@ -195,6 +200,57 @@ func Test_CustomTypes(t *testing.T) {
...
+func Test_IsolatedTypes(t *testing.T) {
+	src := `expr == Expr{id: 2,
+		call_expr: Expr.Call{
+			function: "_==_",
+			args: [
+				Expr{id: 1, ident_expr: Expr.Ident{ name: "a" }},
+				Expr{id: 3, ident_expr: Expr.Ident{ name: "b" }}]
+		}}`
+	gzipped := proto.FileDescriptor("google/api/expr/v1alpha1/syntax.proto")
+        r, err := gzip.NewReader(bytes.NewReader(gzipped))
```
gabrielsz: Nit: Seems like indentation is being thrown off by a char or two. Is that something that `gofmt` is doing?

common/types/pb/type_test.go
```diff
@@ -21,42 +22,48 @@ func TestTypeDescription_FieldCount(t *testing.T) {
 }
 
 func TestTypeDescription_Any(t *testing.T) {
-	_, err := DescribeType(".google.protobuf.Any")
+	pbdb := NewPbDb()
```
gabrielsz, line 25: Nit: Do you think these tests could become table tests?

TristonianJones: +1, sounds like a good idea.

TristonianJones (approved):
LGTM. Shifting toward table-driven test cases can be done in a later PR, but is a good suggestion.

cel/cel_test.go
JimLarson: I think it's a matter of tab chars vs 8 spaces.  Ran gofmt to fix it up.

PR #169, opened 2019-02-08
syntax error in sample

ultrasaurus:
I ran the sample by copying it into a file `sample.go` and added  `package main`, 
ran with `go run sample.go` and got error:
```
sample.go:7:32: expected ';', found ','
```
after removing the comma it worked fine

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,7 +52,7 @@ import(
 	"fmt"
 	"log"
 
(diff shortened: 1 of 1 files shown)


gabrielsz:
LGTM

gabrielsz (approved):

PR #170, opened 2019-02-12
Update naming / docs to remove lint and improve naming.

TristonianJones:
Changes include:

* Remove tabs from the `README.md` file.
* Replace `pb.PbDb` (go vet complains about the name stuttering) with `pb.Db`. 
* Remove common typos
* Add comments on exported packages, functions, and variables.


interpreter/planner.go
```diff
@@ -573,7 +573,6 @@ func (p *planner) idResolver(sel *exprpb.Expr_Select) func(Activation) (ref.Val,
 		case *exprpb.Expr_IdentExpr:
 			ident = op.GetIdentExpr().Name + "." + ident
 			resolvedIdent = true
-			break
```
JimLarson, line 576: Take a closer look here.  The "break" IMHO is a nicer way to indicate that we're stopping the iteration of the for-loop, rather than making this intention indirect through a boolean.  Since "validIdent" and "resolvedIdent" aren't used past the for-loop, consider eliminating them, have it be a "for {" loop, and for IdentExpr and default just break.

test/proto2pb/test_all_types.pb.go
```diff
@@ -1,6 +1,7 @@
 // Code generated by protoc-gen-go. DO NOT EDIT.
 // source: test/proto2/test_all_types.proto
 
+// Package proto2pb encapsulates the proto2 message, enum, and fields supported by CEL.
```
JimLarson, line 4: This looks like a generated file - could we perhaps not check it in?

JimLarson (changes requested):

interpreter/planner.go
TristonianJones: The `break` was a mistake on my part as it only applies to the `switch`. I had probably attempted to use it for the `for` loop and then created an infinite loop that I fixed with the boolean switches, but forgot to remove the superfluous `break`.

JimLarson (approved):
All issues resolved in discussion.

test/proto2pb/test_all_types.pb.go
TristonianJones: I've checked this in mostly so that more of the CEL packages work with `go get` and so that the IDEs can easily run tests without having to invoke `bazel` build targets. Maybe not the greatest to check this artifact in, but that's also how we do things with the go genproto files for the `syntax.proto`, etc.

PR #172, opened 2019-02-13
Return the evaluation details even on error.

TristonianJones:
Errors may or may not be expected, depending on how CEL is being used. The `cel.EvalDetails` should be returned if they are available even if the evaluation is in error.

### DIFF
--- a/cel/program.go
+++ b/cel/program.go
@@ -119,7 +119,7 @@ func newProgram(e *env, ast Ast, opts ...ProgramOption) (Program, error) {
 				interpreter: interp}
 			return initInterpretable(clone, ast, decs)
 		}
-		return &progGen{factory: factory}, nil
+		return initProgGen(factory)
 	}
 	// Enable state tracking last since it too requires the factory approach but is less
 	// featured than the ExhaustiveEval decorator.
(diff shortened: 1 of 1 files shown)


cel/program.go
```diff
@@ -203,8 +203,9 @@ func (gen *progGen) Eval(vars interpreter.Activation) (ref.Val, EvalDetails, err
 
 	// Evaluate the input, returning the result and the 'state' as EvalDetails.
 	v, _, err := p.Eval(vars)
+	det := &evalDetails{state: state}
```
gabrielsz: Shouldn't this be returned on line 201 as well?

gabrielsz (approved):
Just one comment/question, otherwise LGTM

cel/program.go
TristonianJones: Good question. The first block of code configures the program execution environment, but didn't do evaluation. The second block actually runs the expression eval

TristonianJones: I've updated the code to always return `EvalDetails` to improve the consistency in behavior. I've also added a factory construction test within the `Program()` call to reduce the likelihood that an error in `Interpretable` construction is deferred until the first `Eval()` call. Thanks for the feedback!

gabrielsz (approved):
LGTM

PR #176, opened 2019-03-04
vendor: add modules support

lopezator:
Add go modules support to vendor dependencies.


lopezator:
I think it would be useful to keep track of what dependencies and what versions of them should be used when using a particular cel-go revision, by adding support for `go modules` on `cel-go`.

Users could blame `cel-go` for a bug that it's outside it's scope, because an incompatibility with the most recent version of a dependency, for example.

This way, the risk of falling into a bug introduced in a version of a specific dependency is minimized.

TristonianJones:
Good suggestion. Thanks for the contribution.

Issue #175, opened 2019-03-04
Data Race error when running tests in parallel


inigohu:
**Describe the bug**
We are using **cel-go** in our project and when we try to run our tests in parallel, we get a `DATA RACE` error. 

I have reproduced the same behavior with your tests, running parser tests in parallel, and I get this error: `WARNING: DATA RACE`     `race detected during execution of test`

**To Reproduce**
Check which components this affects:

- [X]  parser
- [ ]  checker
- [ ]  interpreter

Sample expression and input that reproduces the issue:
```
go test -v parser/parser.go parser/macro.go parser/errors.go parser/helper.go parser/unescape.go parser/parser_test.go -race
```

Test setup:
```go
// test case in go
func TestParse(t *testing.T) {
  for i, tst := range testCases {
    name := fmt.Sprintf("%d %s", i, tst.I)
      t.Run(name, func(tt *testing.T) {
        tt.Parallel()
        ...
```

```
WARNING: DATA RACE
Read at 0x00c000110f00 by goroutine 28:
...
--- FAIL: TestParse (0.03s)
    --- FAIL: TestParse/0_"A" (0.04s)
        testing.go:809: race detected during execution of test
    --- FAIL: TestParse/31_a_>=_b (0.02s)
        testing.go:809: race detected during execution of test
...
```
**Expected behavior**
Should pass all tests

**Additional context**
run tests with `-race`flag to check Data Race

TristonianJones:
I've looked briefly at the error and it looks as though the race comes from the Antlr4 Go runtime library. I'll see if updating the build.bazel WORKSPACE reference to the library addresses the issue.

TristonianJones:
Updating the workspace had no effect. I'll check through the library to see if I can find the source of the issue.

TristonianJones:
It appears this is an issue with the Antlr Go codegen template. I can address the issue manually for now and will add a hefty set of comments about what the change is within the generated source. I'll also see if I can get the Antlr codegen template changed via PR, but I'm not sure what the overhead of the change is or if an alternative would be preferred. This latter step may take a bit more time.

In terms of downstream gotchas, the protobuf libraries also mutate global state, so depending on your use case you may notice data races during type-check and interpretation. The expectation is that you're typically initializing the runtime type-system once per-process. 

Could you tell me a bit more about your use case so I can better understand the desired execution model @inigohu?

inigohu:
[...]

PR #177, opened 2019-03-05
Fix ANTLR related data-race.

TristonianJones:
Issue #175 pointed out a data race within the parser which occurs when the parsing is run in parallel. The issue appears to be related to the [ANTLR Go Codegen](https://github.com/antlr/antlr4/blob/master/tool/resources/org/antlr/v4/tool/templates/codegen/Go/Go.stg) which creates mutable global variables during parser generation.

The best available fix, shy of addressing the Go codegen and ANLTR Go runtime libraries, is to make the global state per-instance. This will likely have a performance impact, but it's difficult to characterize at this time.

### DIFF
--- a/checker/checker_test.go
+++ b/checker/checker_test.go
@@ -1127,6 +1127,7 @@ func TestCheck(t *testing.T) {
 	for i, tst := range testCases {
 		name := fmt.Sprintf("%d %s", i, tst.I)
 		t.Run(name, func(tt *testing.T) {
+			tt.Parallel()
 
 			src := common.NewTextSource(tst.I)
 			expression, errors := parser.Parse(src)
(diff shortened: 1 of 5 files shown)


parser/parser_test.go
```diff
@@ -852,6 +852,7 @@ func TestParse(t *testing.T) {
 	for i, tst := range testCases {
 		name := fmt.Sprintf("%d %s", i, tst.I)
 		t.Run(name, func(tt *testing.T) {
+			tt.Parallel()
```
JimLarson, line 857: Add a comment explaining why it's important to parallelize here.

checker/checker_test.go
```diff
@@ -1127,6 +1127,7 @@ func TestCheck(t *testing.T) {
 	for i, tst := range testCases {
 		name := fmt.Sprintf("%d %s", i, tst.I)
 		t.Run(name, func(tt *testing.T) {
+			tt.Parallel()
```
JimLarson, line 1132: Add a comment explaining why we parallelize here.

JimLarson (approved):

PR #178, opened 2019-03-05
Perform in-order expression id numbering within the parser.

TristonianJones:
Ensure that the expressions are numbered in the same order as the parse elements appear rather than in post-order.


gabrielsz (approved):
LGTM.
The ID generation order seems to match the expected behavior at first sight.

The only concern is that we have many places where we accept a `ctx interface{}` as a way to dynamically accept either an ID or an Expression. As we add more options to that, it could be harder to detect errors at Go compile time.

gabrielsz (approved):
LGTM. Thanks for cleaning `newExpr` out =)

PR #181, opened 2019-03-08
Type provider proto isolation

### LINKED ISSUE #180: Runtime type isolation for `types.Provider`
**Feature request checklist**

**Change**
Parse and type-check isolation is possible using the `cel.IsolateTypes()` and `cel.TypeDesc()` options
when configuring the `cel.Env`. However, the runtime types must all presently share the same `pb.DefaultDb` when using protobuf types. 

The `pb.DefaultDb` holds global mutable state that can trigger race errors in certain use cases like the on described in #175. While the `IndexTypes()` operation can trigger this race, it performs an idempotent write meaning that it will always index the same information. This is a wart in the way proto reflected types are exposed within CEL. It's not clear how often multiple `Provider` configurations are used within the same process, but it should be possible to creation multiple isolated runtime providers.

The fix would be to retain a separate proto reflection mapping at the `pb.Db` level which might result in duplicate work in some cases, but is probably desired for safety and consistency.

**Alternatives Considered**
The `pb.DefaultDb` could use `sync.Map` or mutex locks to protect the map from concurrent modifications on the same keys.

The alternative is a much more surgical change, but it seems like it should be possible (and less confusing) to have runtime type providers that don't share any state.

TristonianJones:
Provide strict isolation between the `types.Provider` instances to make it possible to run multiple CEL expressions together using different environments, etc. This should fix #180 and address issues raised in #175.

Minor changes have also been made to the top-level interface to make sure it is easy to adapt native inputs to CEL types.

### DIFF
--- a/cel/cel_test.go
+++ b/cel/cel_test.go
@@ -78,8 +78,8 @@ func Example() {
 
 	// Evaluate the program against some inputs. Note: the details return is not used.
 	out, _, err := prg.Eval(map[string]interface{}{
-		// CEL values may be supplied directly.
-		"i": types.String("CEL"),
+		// Native values are converted to CEL values under the covers.
+		"i": "CEL",
 		// Values may also be lazily supplied.
 		"you": func() ref.Val { return types.String("world") },
 	})
@@ -120,8 +120,8 @@ func Test_ExampleWithBuiltins(t *testing.T) {
 	// If the Eval() call were provided with cel.EvalOptions(OptTrackState) the details response
 	// (2nd return) would be non-nil.
 	out, _, err := prg.Eval(map[string]interface{}{
-		"i":   types.String("CEL"),
-		"you": types.String("world")})
+		"i":   "CEL",
+		"you": "world"})
 	if err != nil {
 		t.Fatalf("runtime error: %s\n", err)
 	}
@@ -149,7 +149,7 @@ func Test_DisableStandardEnv(t *testing.T) {
 		p, _ := e.Parse("a.b.c")
 		c, _ := e.Check(p)
 		prg, _ := e.Program(c)
-		out, _, _ := prg.Eval(map[string]interface{}{"a.b.c": types.True})
+		out, _, _ := prg.Eval(map[string]interface{}{"a.b.c": true})
 		if out != types.True {
 			t.Errorf("Got '%v', wanted 'true'", out.Value())
 		}
@@ -208,7 +208,7 @@ func Test_HomogeneousAggregateLiterals(t *testing.T) {
 			t.Fatalf("Got issue: %v, expected successful check.", iss.Err())
 		}
 		prg, _ := e.Program(c, funcs)
-		out, _, err := prg.Eval(Vars(map[string]interface{}{"name": "world"}))
+		out, _, err := prg.Eval(map[string]interface{}{"name": "world"})
 		if err != nil {
 			t.Fatalf("Got err: %v, wanted result", err)
 		}
@@ -223,7 +223,7 @@ func Test_HomogeneousAggregateLiterals(t *testing.T) {
 			t.Fatalf("Got issue: %v, expected successful check.", iss.Err())
 		}
 		prg, _ := e.Program(c, funcs)
-		out, _, err := prg.Eval(Vars(map[string]interface{}{"name": "world"}))
+		out, _, err := prg.Eval(map[string]interface{}{"name": "world"})
 		if err != nil {
 			t.Fatalf("Got err: %v, wanted result", err)
 		}
@@ -363,7 +363,7 @@ func Test_GlobalVars(t *testing.T) {
 
 	// Global variables can be configured as a ProgramOption and optionally overridden on Eval.
 	prg, _ := e.Program(c, funcs, Globals(map[string]interface{}{
-		"default": types.String("third"),
+		"default": "third",
 	}))
 
 	t.Run("global_default", func(t *testing.T) {
@@ -409,8 +409,8 @@ func Test_EvalOptions(t *testing.T) {
 	}
 	out, details, err := prg.Eval(
 		map[string]interface{}{
-			"k": types.String("key"),
-			"v": types.True})
+			"k": "key",
+			"v": true})
 	if err != nil {
 		t.Fatalf("runtime error: %s\n", err)
 	}
--- a/cel/cel_test.go
+++ b/cel/cel_test.go
@@ -77,12 +77,12 @@ func Example() {
 	}
 
 	// Evaluate the program against some inputs. Note: the details return is not used.
-	out, _, err := prg.Eval(Vars(map[string]interface{}{
-		// Native values are converted to CEL values under the covers.
-		"i": "CEL",
+	out, _, err := prg.Eval(map[string]interface{}{
+		// CEL values may be supplied directly.
+		"i": types.String("CEL"),
 		// Values may also be lazily supplied.
 		"you": func() ref.Val { return types.String("world") },
-	}))
+	})
 	if err != nil {
 		log.Fatalf("runtime error: %s\n", err)
 	}
@@ -119,9 +119,9 @@ func Test_ExampleWithBuiltins(t *testing.T) {
 
 	// If the Eval() call were provided with cel.EvalOptions(OptTrackState) the details response
 	// (2nd return) would be non-nil.
-	out, _, err := prg.Eval(Vars(map[string]interface{}{
-		"i":   "CEL",
-		"you": "world"}))
+	out, _, err := prg.Eval(map[string]interface{}{
+		"i":   types.String("CEL"),
+		"you": types.String("world")})
 	if err != nil {
 		t.Fatalf("runtime error: %s\n", err)
 	}
@@ -149,7 +149,7 @@ func Test_DisableStandardEnv(t *testing.T) {
 		p, _ := e.Parse("a.b.c")
 		c, _ := e.Check(p)
 		prg, _ := e.Program(c)
-		out, _, _ := prg.Eval(Vars(map[string]interface{}{"a.b.c": true}))
+		out, _, _ := prg.Eval(map[string]interface{}{"a.b.c": types.True})
 		if out != types.True {
 			t.Errorf("Got '%v', wanted 'true'", out.Value())
 		}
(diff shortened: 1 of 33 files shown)


checker/checker_test.go
```diff
@@ -1129,7 +1129,7 @@ func TestCheck(t *testing.T) {
 		t.Run(name, func(tt *testing.T) {
 			// Runs the tests in parallel to ensure that there are no data races
 			// due to shared mutable state across tests.
-			tt.Parallel()
+			// tt.Parallel()
```
lopezator: this is on purpose?

TristonianJones: Sorry, likely something I did during debug. The pull is still a work in progress (comments, code hygiene), but works correctly. I'm going to go over the changes with a fine tooth comb before I request review.

cel/cel_test.go
```diff
@@ -192,13 +192,14 @@ func Test_CustomTypes(t *testing.T) {
...
-	}}))
+	}}
+	out, _, _ := prg.Eval(vars)
 	if out != types.True {
 		t.Errorf("Got '%v', wanted 'true'", out.Value())
 	}
 }
 
-func TestIsolatedTypes(t *testing.T) {
+func Test_IsolatedTypes(t *testing.T) {
```
JimLarson: Rename to Test_TypeIsolation since it no longer calls IsolateTypes().

P.S. Why the underscore?  I've seen Test<Major>_<Minor>, but not _ after Test.

cel/env.go
```diff
@@ -166,6 +169,7 @@ type env struct {
 	macros         []parser.Macro
 	pkg            packages.Packager
 	types          ref.TypeProvider
```
JimLarson: Rename the "types" field to "provider", for symmetry with "TypeAdapter" -> "adapter".

cel/options.go
```diff
@@ -63,6 +64,11 @@ func ClearMacros() EnvOption {
 func CustomTypeProvider(provider ref.TypeProvider) EnvOption {
```
JimLarson, line 73: Needs doc update about this subtle behavior?

common/types/object.go
```diff
@@ -37,31 +38,30 @@ type protoObj struct {
...
+		TypeAdapter: adapter,
+		value:       value,
+		refValue:    reflect.ValueOf(value),
+		typeDesc:    typeDesc,
+		typeValue:   NewObjectTypeValue(typeDesc.Name())}
 }
 
-func (o *protoObj) ConvertToNative(typeDesc reflect.Type) (interface{}, error) {
-	if typeDesc.AssignableTo(o.refValue.Type()) {
+func (o *protoObj) ConvertToNative(refType reflect.Type) (interface{}, error) {
```
JimLarson: Nit: "refType" is too close to "ref.Type".  Change to "refl"?

common/types/pb/pb.go
```diff
@@ -76,14 +67,32 @@ func (pbdb *Db) DescribeDescriptor(fileDesc *descpb.FileDescriptorProto) (*FileD
...
+	typeName := sanitizeProtoName(proto.MessageName(message))
+	if fd, found := pbdb.revFileDescriptorMap[typeName]; found {
+		return fd, nil
+	}
+	fileDesc, _ := descriptor.ForMessage(message.(descriptor.Message))
+	return pbdb.RegisterDescriptor(fileDesc)
+}
+
 // DescribeFile takes a protocol buffer message and indexes all of the message
 // types and enum values contained within the message's file descriptor.
```
JimLarson: Update the description.

common/types/provider.go
```diff
@@ -38,10 +38,10 @@ type protoTypeProvider struct {
 // NewProvider accepts a list of proto message instances and returns a type
 // provider which can create new instances of the provided message or any
 // message that proto depends upon in its FileDescriptor.
-func NewProvider(types ...proto.Message) ref.TypeProvider {
+func NewProvider(types ...proto.Message) *protoTypeProvider {
```
JimLarson: Document why it doesn't return the ref.TypeProvider interface.

common/types/provider.go
```diff
@@ -192,17 +191,79 @@ func (p *protoTypeProvider) registerAllTypes(fd *pb.FileDescription) error {
...
+	// Override the Any type by ensuring that custom proto-types are considered on recursive calls.
+	case *anypb.Any:
+		val := value.(*anypb.Any)
+		unpackedAny := ptypes.DynamicAny{}
+		if ptypes.UnmarshalAny(val, &unpackedAny) != nil {
+			NewErr("Fail to unmarshal any.")
+		}
+		return p.NativeToValue(unpackedAny.Message)
+	// Convert custom proto types to CEL values based on type's presence within the pb.Db.
+	case proto.Message:
```
JimLarson, line 229: Should this be *proto.Message?  If not, document the reason for the difference vs the other message types.

common/types/ref/provider.go
```diff
@@ -41,25 +41,20 @@ type TypeProvider interface {
...
-	// private to this TypeProvider.  Subsequent Describe*() calls
-	// will modify the protobuf registry only in this TypeProvider.
-	// Note that privately-registered protobufs cannot be instantiated
-	// with types.NewObject().
-	IsolateTypes()
+	FindFieldType(messageType string, fieldName string) (*FieldType, bool)
 
 	// NewValue creates a new type value from a qualified name and a map of
 	// field initializers.
 	NewValue(typeName string, fields map[string]Val) Val
```
JimLarson, line 48: Shouldn't this go in TypeAdapter too, since we can't do it for dynamically-provided protos?

interpreter/activation.go
```diff
@@ -34,17 +35,65 @@ type Activation interface {
...
+// Go values to equivalent CEL `ref.Val` objects.
+//
+// The input `bindings` may either be of type `Activation` or `map[string]interface{}`.
+//
+// When the bindings are a `map` the values may be one of the following types:
+//   - `ref.Val`: a CEL value instance.
+//   - `func() ref.Val`: a CEL value supplier.
+//   - other: a native value which must be converted to a CEL `ref.Val` by the `adapter`.
+func NewAdaptingActivation(adapter ref.TypeAdapter,
+	bindings interface{}) (Activation, error) {
```
JimLarson: Formatting?

common/types/ref/provider.go
```diff
@@ -41,25 +41,20 @@ type TypeProvider interface {
...
+}
 
-	// RegisterDescriptor registers the contents of a protocol
-	// buffer FileDescriptor.
+// TypeRegistry allows third-parties to registry custom types. Not all TypeProvider
+// implementations support type-customization, so these features are optional.
+type TypeRegistry interface {
```
JimLarson, line 61: Consider embedding TypeProvider.

common/types/ref/provider.go
```diff
@@ -70,6 +65,12 @@ type TypeProvider interface {
 	RegisterType(types ...Type) error
 }
 
+// TypeAdapter converts native Go values of varying type and complexity to equivalent CEL values.
+type TypeAdapter interface {
```
JimLarson: Consider embedding TypeProvider so you don't have to give both in interpreter.

server/server.go
```diff
@@ -94,7 +94,10 @@ func (s *ConformanceServer) Check(ctx context.Context, in *exprpb.CheckRequest)
 
 // Eval implements ConformanceService.Eval.
 func (s *ConformanceServer) Eval(ctx context.Context, in *exprpb.EvalRequest) (*exprpb.EvalResponse, error) {
-	env, _ := cel.NewEnv(cel.Container(in.Container))
+	tp := types.NewProvider()
```
JimLarson: Comment why we make this explicit.  Instead, perhaps add env.GetProvider() method?

JimLarson (changes requested):

cel/cel_test.go

The rest of the tests in this file have an underscore in them. I think the idea is to indicate that this is a Test of CEL and how it interacts with `feature x`.

cel/env.go
TristonianJones: Done.

cel/options.go
TristonianJones: Good point, I'll make a 'CustomTypeAdapter' option and set them separately with a note that developers may want to set these fields together.

common/types/object.go
TristonianJones: Done.

common/types/pb/pb.go
TristonianJones: Done.

common/types/provider.go
TristonianJones: I think I'll actually create a interface that collects these things together so that the internal type isn't exposed.

common/types/provider.go
TristonianJones: For interface types, there's no real notion of a pointer to it. It can be implemented either by a pointer to a struct or a struct depending on how the interface is defined.

common/types/ref/provider.go
TristonianJones: Done.

common/types/ref/provider.go
TristonianJones: The provider just creates new values, we don't know whether it's handling conversion between native types and CEL types though. Maybe a small distinction, but I can't provide a DefaultTypeAdapter simply if `NewValue()` is part of it since this is only invoked for complex proto message types.

common/types/ref/provider.go
TristonianJones: I think `TypeAdapter` should be standalone, but I switched from making a `NewProvider` to a `NewRegistry` and the registry must implement the `TypeAdapter` interface.

server/server.go
TristonianJones: Fair enough. Added to Env

TristonianJones:
PTAL @JimLarson

cel/options.go
```diff
@@ -58,17 +57,22 @@ func ClearMacros() EnvOption {
 	}
 }
 
+// CustomTypeAdapter swaps the default ref.TypeAdapter implementation with a custom one.
+//
+// Note: This option must be specified before the Types option when used together.
```
JimLarson: ...and TypeDescs().

cel/options.go
```diff
@@ -58,17 +57,22 @@ func ClearMacros() EnvOption {
...
+func CustomTypeAdapter(adapter ref.TypeAdapter) EnvOption {
+	return func(e *env) (*env, error) {
+		e.adapter = adapter
+		return e, nil
+	}
+}
+
 // CustomTypeProvider swaps the default ref.TypeProvider implementation with a custom one.
 //
 // Note: This option must be specified before the Types option when used together.
```
JimLarson: ...and TypeDescs().

JimLarson (approved):

lopezator:
Thank you for doing this!!! 👏

TristonianJones:
You're very welcome.

Issue #144, opened 2019-01-08
Checker option for homogeneous vs. heterogeneous aggregate types


TristonianJones:
**Feature request checklist**

- [X] There are no issues that match the desired change
- [X] The change is large enough it can't be addressed with a simple Pull Request

**Change**
The `dyn` type is used to signify two things:

1. Unknown type to be determined at runtime
1. Mixed type, possibly known statically

When a function is declared with a strongly typed aggregate parameter, e.g. `list(string)` 
it is possible to construct a list literal with mixed element types which passes the check
successfully as the list is marked with a type of `list(dyn)`

```
// Signature of the type assertion:
// func ints(in list<int>) list<int>

// Invocation, check yields ints(list(dyn))~list(int)
// Defers type checking to runtime based on presence of heterogeneous list,
// but should fail.
ints([1, 2u])
```

The introduction of a union type to handle this second case would fix the issue, but introduce
a lot of complexity (i.e. will take a while to implement).

**Alternatives Considered**
There are two alternatives:

1. Checker option to enforce homogeneous types in literals
2. For list literals, test the element types of all known types in the list against the method signature.

Either solution would make the problematic example yield the correct failure at check time.

This issue first appeared in issue #140

PR #183, opened 2019-03-13
Add an option to disable aggregates with dynamic element types.

TristonianJones:
This pull closes feature request #144 and will allow users to disable support for aggregate literals with mixed element types. This feature can be enabled via the following top-level option:
```go
env, err := cel.NewEnv(cel.HomogeneousAggregateLiterals())
```

### DIFF
--- a/cel/options.go
+++ b/cel/options.go
@@ -83,7 +83,7 @@ func Declarations(decls ...*exprpb.Decl) EnvOption {
 // during type-checking.
 //
 // Note, it is still possible to have heterogeneous aggregates when provided as variables to the
-// expression.
+// expression or with unchecked expressions.
 func HomogeneousAggregateLiterals() EnvOption {
 	return func(e *env) (*env, error) {
 		e.enableDynamicAggregateLiterals = false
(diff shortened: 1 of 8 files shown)


glerchundi:
Thanks for making this happen @TristonianJones! 👏

TristonianJones:
You're very welcome, this has come up in a couple of different venues, so I'm glad you'll be putting it to good use.

JimLarson (changes requested):
Clarify in doc comments that it's a check-time only option - or add a TODO if you're going to add a runtime restriction.

TristonianJones:
This is a check-time only restriction. It's still possible to have heterogeneous aggregates at runtime and when using unchecked expressions.

cel/options.go
```diff
@@ -79,6 +79,18 @@ func Declarations(decls ...*exprpb.Decl) EnvOption {
 	}
 }
 
+// HomogeneousAggregateLiterals option ensures that list and map literal entry types must agree
+// during type-checking.
+//
+// Note, it is still possible to have heterogeneous aggregates when provided as variables to the
+// expression or with unchecked expressions.
```
JimLarson: ...or via implicit conversion from well-known types.

JimLarson (approved):

PR #185, opened 2019-03-15
Fix identifier resolution for variable subsumption cases

### LINKED ISSUE #184: Inconsistent resolution of identifiers in unchecked ASTs
**Describe the bug**
When an expression is type-checked it creates metadata which indicates
whether a given select expression (field traversal) refers to a namespaced
identifier. The namespace resolution at check-time is consistent with the
spec, but when evaluating unchecked ASTs (parse-only), the resolution
occurs in reverse order.

**To Reproduce**

- [ ]  parser
- [ ]  checker
- [X]  interpreter

```cel
// a.b and a.b.c are both declarations.
// a.b is map with a key 'c' and a.b.c is a string
// a.b = {'c': 'oops'}
// a.b.c = 'yay'
a.b.c == 'yay' // true for checked AST, false otherwise.
```

Test setup:
```go
// Environment setup.
env, _ := cel.NewEnv(
    cel.Declarations(
       decls.NewIdent("a.b.c", decls.String, nil),
       decls.NewIdent("a.b", decls.NewMapType(decls.String, decls.String), nil))
vars := cel.Vars(map[string]interface{}{
        "a.b": map[string][string]{"c": "oops"},
        "a.b.c": "yay"})

// Parsed eval
ast, _ := env.Parse("a.b.c")
prg, _ := env.Program(ast)
out, _, _ := prg.Eval(vars)
fmt.Printf("parsed eval: %s\n", out)
    
// Checked eval
ast, _ = env.Check(ast)
prg, _ = env.Program(ast)
out, _, _ = prg.Eval(vars)    
fmt.Printf("checked eval: %s\n", out)
```

**Expected behavior**
The output of both checked and parsed eval should agree.

TristonianJones:
This pull closes #184 where qualified identifiers were being resolve differently between parsed and checked expressions. The added tests also highlight the uniform behavior. 

The appropriate cel-spec tests related to this should also be updated in a cel-spec PR.

### DIFF
--- a/interpreter/interpreter_test.go
+++ b/interpreter/interpreter_test.go
@@ -228,6 +228,29 @@ func TestInterpreter_FieldAccess(t *testing.T) {
 	}
 }
 
+func TestInterpreter_SubsumedFieldAccess(t *testing.T) {
+	vars := NewActivation(map[string]interface{}{
+		"a.b":   map[string]types.Int{"c": types.Int(9)},
+		"a.b.c": types.Int(10),
+	})
+
+	parsed := parseExpr(t, `a.b.c == 10`)
+	i, _ := interpreter.NewUncheckedInterpretable(parsed.GetExpr())
+	result := i.Eval(vars)
+	if result != types.True {
+		t.Errorf("Got %v, wanted true", result)
+	}
+
+	checked := compileExpr(t, `a.b.c == 10`,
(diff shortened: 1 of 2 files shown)


madi8229 (approved):
LGTM.

TristonianJones:
Closes #184

PR #186, opened 2019-03-15
Fix off-by-one on getMonth

### LINKED ISSUE #182: getMonth gives 1-based results
**Describe the bug**
The timestamp `getMonth()` function returns months 1-12, but should return zero-based months 0-11 according to the spec.

**To Reproduce**

- [ ]  parser
- [ ]  checker
- [X]  interpreter

```cel
timestamp('2009-02-13T23:31:30Z').getMonth()
```

Test setup:  just evaluate that expression.

**Expected behavior**
Spec says that months are 0-11, but this implementation returns months in 1-12

**Additional context**
See `timestampGetMonth()` in `timestamp.go`

TristonianJones:
Closes #182

### DIFF
--- a/common/types/timestamp.go
+++ b/common/types/timestamp.go
@@ -210,7 +210,7 @@ func timestampGetFullYear(t time.Time) ref.Val {
 	return Int(t.Year())
 }
 func timestampGetMonth(t time.Time) ref.Val {
-	return Int(t.Month())
+	return Int(t.Month() - 1)
 }
 func timestampGetDayOfYear(t time.Time) ref.Val {
 	return Int(t.YearDay())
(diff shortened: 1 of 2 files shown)


common/types/timestamp.go
```diff
@@ -210,7 +210,7 @@ func timestampGetFullYear(t time.Time) ref.Val {
 	return Int(t.Year())
 }
 func timestampGetMonth(t time.Time) ref.Val {
-	return Int(t.Month())
+	return Int(t.Month() - 1)
```
gabrielsz, line 215: Given that our standard library is 1-based for months, but CEL is defined as 0-based, can you comment on the function?

gabrielsz (approved):
LGTM. I only have a nit about adding a documentation line, but it's completely optional.

common/types/timestamp.go
TristonianJones: Done.

Issue #173, opened 2019-02-13
Distinguish unsuccessful evaluation vs successful evaluation to error result


JimLarson:
**Feature request checklist**

- [X] There are no issues that match the desired change
- [ ] The change is large enough it can't be addressed with a simple Pull Request
- [ ] If this is a bug, please file a [Bug Report](./ISSUE_TEMPLATE).

This is probably a small change, but writing this up as a feature request for proper discussion before implementation.

**Change**

Currently, the API for cel.Program.Eval() says:

> If the evaluation is an error, the result will be nil with a non-nil error.

The signature of Eval() is

    Eval(vars interpreter.Activation) (ref.Val, EvalDetails, error)

and "result" refers to the `ref.Val`.  However, this conflates two different cases:
- evaluation proceeded successfully, but to an error result, e.g. "1 / 0";
- evaluation itself failed and we were unable to determine the result of the CEL program.

Some CEL use cases might want to take the ref.Val result of successful evaluation, even if it's an error result, and use it.  Other CEL use cases might want to branch on errors, whether errors in evaluation, or error results from successful evaluation.

Therefore I propose the following tweak to the Eval() API, with no change in signature:
- The ref.Val return is the result of successful evaluation, whether a normal value, an error, or an unknown, or nil if evaluation didn't successfully complete.
- The error return is either the error returned from successful evaluation, or the error which prevented evaluation, or nil if evaluation was successful and the result was a normal value or unknown.

So the return from evaluation will be:
- `val, details, nil` Successful evaluation to a non-error result.
- `val, details, err` Successful evaluation to an error result.
-  `nil, details, err` Unsuccessful evaluation

**Example**

To branch on unsuccessful evaluation, or evaluation to an error
```
result, details, err := program.Eval(activation)
if err != nil {
    log.Fatalf("Evaluation error: %v", err)
}
processResult(result)
```

To use all results of successful evaluation:
```
result, details, err := program.Eval(activation)
if result == nil {
    log.Fatalf("Evaluation error: %v", err)
}
processResult(result)
```

**Alternatives considered**

The original API had no provision for an error return from evaluation, forcing everything into a ref.Val, but this makes it awkward to discern unsuccessful evaluations.

PR #187, opened 2019-03-15
Clarify the evaluation status from Eval

TristonianJones:
Closes issue #173 by ensuring that the `val` output returned from `program.Eval()` is included even if the evaluation results in an `error`. This allow for disambiguation between errors due to the configuration of the program versus errors related to the evaluation of the program itself.

### DIFF
--- a/cel/program.go
+++ b/cel/program.go
@@ -26,8 +26,8 @@ import (
 type Program interface {
 	// Eval returns the result of an evaluation of the Ast and environment against the input vars.
 	//
-	// If the OptTrackState or OptExhaustiveEval is used, the `details` response will be non-nil.
-	// The return state from evaluation will be:
+	// If the `OptTrackState` or `OptExhaustiveEval` flags are used, the `details` response will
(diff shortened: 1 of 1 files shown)


cel/program.go
```diff
@@ -26,9 +26,16 @@ import (
 type Program interface {
 	// Eval returns the result of an evaluation of the Ast and environment against the input vars.
 	//
-	// If the evaluation is an error, the result will be nil with a non-nil error.
+	// If the OptTrackState or OptExhaustiveEval is used, the `details` response will be non-nil.
```
gabrielsz: Nit: "If the _`x`_ or _`y`_ **flags(?) are** used..." (notice the back quote mark on x/y)

cel/program.go
```diff
@@ -26,9 +26,16 @@ import (
...
+	// The return state from evaluation will be:
```
gabrielsz: will -> can/may?

gabrielsz (approved):
LGTM. Just two doc nitpicks. Thanks!

PR #188, opened 2019-03-17
Update the README.md with clearer getting started guidance.

TristonianJones:


README.md
```diff
@@ -1,111 +1,242 @@
...
 
 [![Build Status](https://travis-ci.org/google/cel-go.svg?branch=master)](https://travis-ci.org/google/cel-go) [![Go Report Card](https://goreportcard.com/badge/github.com/google/cel-go)](https://goreportcard.com/report/github.com/google/cel-go)
+[![GoDoc](https://godoc.org/github.com/google/cel-go?status.svg)][6]
 
-This repo contains the Go toolchain for the Common Expression Language (CEL).
-CEL is a non-Turing complete language designed to be portable and fast. It is
-best suited to  applications where sandboxing a full-fledged language like
-JavaScript or Lua would be too resource intensive, but side-effect free dynamic
-computations are strongly desired.
+The Common Expression Language (CEL) is a non-Turing complete language designed
```
JimLarson, line 6: Hyperlink to the cel-spec README.md ASAP.  What's here should summarize the high points of that document - in particular, there should be no new content here that applies to CEL in general, such as the examples below.

README.md
```diff
@@ -1,111 +1,242 @@
...
+for simplicity, speed, safety, and portability. CEL's C-like syntax looks nearly
+identical to equivalent expressions in C++, Go, Java, and TypeScript. e.g.
 
-## Want to learn more?
+```javascript
```
JimLarson: This example suggests that we support sequences of expressions - separate these as different displayed block, i.e. each has its own triple-single-quote start/end.

README.md
```diff
@@ -1,111 +1,242 @@
...
+stack. Parse and check an expression to make sure it's valid. Then evaluate
+the output AST against some input.
 
-### Example
+### Environment Setup
 
-The following example shows the parse, check, and intepretation of a simple
-program. After the parse and type-check steps, the service checks whether there
-are any errors that need to be reported.
+Let's expose `resource.name` and  `auth.claims` variables to CEL using the
```
JimLarson: I don't like encouraging people to use qualified identifiers.  Could you make the example simpler?

README.md
```diff
@@ -1,111 +1,242 @@
...
+    cel.Declarations(
+        decls.NewIdent("resource.name", decls.String, nil),
+        decls.NewIdent("auth.claims",
+            decls.NewMapType(decls.String, decls.Dyn), nil)))
+```
+
+That's it, the environment is ready to be use for parsing and type-checking.
+Note, the `auth.claims` variable is exposed as a map with `string` keys and
+`dyn` values. The `dyn` in this case represents JSON input as the
+`auth.claims` could originate from a JSON Web Token (JWT).
```
JimLarson: Use of JWT seems pretty wild for first basic example.

README.md
```diff
@@ -1,111 +1,242 @@
...
+env, err := cel.NewEnv(
+    cel.Declarations(
+        decls.NewIdent("resource.name", decls.String, nil),
+        decls.NewIdent("auth.claims",
+            decls.NewMapType(decls.String, decls.Dyn), nil)))
+```
+
+That's it, the environment is ready to be use for parsing and type-checking.
+Note, the `auth.claims` variable is exposed as a map with `string` keys and
+`dyn` values. The `dyn` in this case represents JSON input as the
```
JimLarson: Dislike introducing 'dyn' and explaining only what it means in this case.

README.md
```diff
@@ -1,111 +1,242 @@
...
+        "group": "acme.co"}})
+fmt.Println(out) // 'true'
 ```
+
+For more examples of how to use CEL, see [cel_test.go](https://github.com/google/cel-go/tree/master/cel/cel_test.go).
+
+#### Partial State
+
+What if the `auth.claims` hadn't been supplied? Well, CEL is designed for this.
+When some of the input is unknown, the `out` value will contain a list of the
```
JimLarson: ...when some input is bound to a special "Unknown" value.  Current behavior is a bug and shouldn't be documented.

README.md
```diff
@@ -1,111 +1,242 @@
...
+    | .......................................^`,
+```
+
+Both the parsed and checked expressions contain source position information
+about each node that appears in the output AST. This information can be used
+to determine error locations at evaluation time as well.
+
+## Performance
+
+CEL evaluates very quickly. When the expression set does not change frequently,
```
JimLarson: delete "set"

README.md
```diff
@@ -1,111 +1,242 @@
...
+## Common Questions
+
+### Why not JavaScript, Lua, or WASM?
+
+JavaScript and Lua are rich languages that require sandboxing to execute
+safely. Sandboxing is costly and factors into the "what will I let users
+evaluate?" question heavily when the answer is anything more than O(n)
+complexity.
+
+CEL evaluates linearly with respect to the size of the expression and the input
```
JimLarson, line 263: ...if macros are disabled.

JimLarson (changes requested):

README.md
TristonianJones: Linked on "syntax". I'll follow up with the cel-spec to revise the high-level content there as well.

README.md
TristonianJones: I've split the example apart and added a comment that a CEL "program" is a single expression.

README.md
TristonianJones: I switched to primitive types

README.md
TristonianJones: I made this less JWT specific.

README.md
```diff
@@ -1,111 +1,282 @@
...
-| Check     | Type-checks a parsed expression against a given environment.   |
-| Interpret | Evaluates parsed expressions against a set of inputs.          |      |
+## Overview
 
-Type-checking an expression in an optional, but strongly encouraged step that
-can be used to reject some expressions as semantically invalid using static
-analysis. Additionally, the type-check produces some additional metadata
-related to function overload resolution and object field selection which may
-be used by the interpret step to speed up execution.
+Determine the attribute names and functions you want to provide to the CEL
```
JimLarson: "Variable names" instead of "attribute names".  Attributes refer to properties of an underlying entity.

README.md
```diff
@@ -1,111 +1,282 @@
...
+stack. Parse and check an expression to make sure it's valid. Then evaluate
+the output AST against some input.
 
-### Example
+### Environment Setup
 
-The following example shows the parse, check, and intepretation of a simple
-program. After the parse and type-check steps, the service checks whether there
-are any errors that need to be reported.
+Let's expose `name` and  `group` variables to CEL using the standard builtins:
```
JimLarson: ...using the "Declarations" option.

JimLarson (approved):
A few comments, but looks much better.

README.md
```diff
@@ -1,111 +1,282 @@
...
+thread-safe, and cachable.
+
+Type-checking in an optional, but strongly encouraged step that can reject some
+semantically invalid expressions using static analysis. Additionally, the check
+produces metadata which can improve function invocation performance and object
+field selection at evaluation-time.
+
+#### Macros
+
+Macros are enabled by default and may be disabled. The comprehension macros
```
sefk: I've always found macros to be one of the more confusing aspects of CEL. Why are they different than just any other language feature? Why are they called that? I confess that I tried to see how I'd explain and wasn't sure I could.

sefk (approved):
This is a great improvement over what we had already, thanks for adding Tristan!

PR #193, opened 2019-03-28
Move the internal environment construction out of the Check call.

TristonianJones:
The internal checker construction is quite expensive and being performed on every
call to the `cel.Env#Check`. This change moves construction of the internal environment
up to the `cel.NewEnv` call where it ought to be.

Also, there was a previously unchecked error on environment construction that's being
surfaced now

### DIFF
--- a/cel/env.go
+++ b/cel/env.go
@@ -128,6 +128,13 @@ func NewEnv(opts ...EnvOption) (Env, error) {
 			return nil, err
 		}
 	}
+	ce := checker.NewEnv(e.pkg, e.provider)
+	ce.EnableDynamicAggregateLiterals(e.enableDynamicAggregateLiterals)
+	err = ce.Add(e.declarations...)
+	if err != nil {
+		return nil, err
+	}
+	e.chk = ce
 	return e, nil
 }
 
@@ -175,6 +182,7 @@ type env struct {
 	pkg          packages.Packager
 	provider     ref.TypeProvider
 	adapter      ref.TypeAdapter
+	chk          *checker.Env
 	// environment options, true by default.
 	enableBuiltins                 bool
 	enableDynamicAggregateLiterals bool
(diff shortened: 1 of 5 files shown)


TristonianJones:
@kyessenov could you please confirm this has the intended effect on your project?

kyessenov:
Yes, this is a significant improvement for our use case!
Before with 1m benchmark of expression compilation:
```
Benchmark             300         353845577 ns/op        46219153 B/op    1151895 allocs/op
```
After:
```
Benchmark            5000          26936138 ns/op         3113057 B/op      73825 allocs/op
```

cel/env.go
```diff
@@ -175,23 +182,21 @@ type env struct {
 	pkg          packages.Packager
 	provider     ref.TypeProvider
 	adapter      ref.TypeAdapter
+	chk          *checker.Env
```
kyessenov, line 186: This is not race safe I think. There is a list of scopes inside declarations inside checker environment. The list is mutable by Push/Pop. These mutations prevent parallel use of the checker environment.

For my use case, this is fine, but let's document that. We also need to make sure that Push/Pops are balanced.

TristonianJones: Ah, good point, I forgot how this interacts with comprehensions. For the most part the push / pop is used only during environment configuration. Let me work on refactoring this a bit more to be thread-safe since the perf change is substantial, and some projects will really want / need to be able to use the same environment concurrently.

checker/decls/scopes.go
```diff
@@ -16,83 +16,75 @@ package decls
...
+// is nil.
+func (s *Scopes) Pop() *Scopes {
+	if s.parent != nil {
+		return s.parent
+	}
+	// TODO: Consider whether this should be an error / panic.
+	return s
 }
 
-// AddIdent adds the ident Decl in the outermost scope.
```
TristonianJones, line 50: The previous behavior here worked for testing, but could have resulted in cases like the following passing the check when they shouldn't have:

```javascript
// note, x is never declared but it's type is inferred as an int
list.exists(x, x > 10) && x != 5
```

TristonianJones:
PTAL @kyessenov

kyessenov (approved):
/lgtm

PR #197, opened 2019-04-02
Update README.md install instructions

TristonianJones:


TristonianJones:
@lopezator does this look about right for the install instructions?

lopezator:
Depending on the usecase.

What is the most common supposed usage?

Compiling a binary to run the `server/main/main.go`?
Using it as a library on your application?

`go install` would generate a binary on your `$GOBIN` so you can:

```sh
$GOBIN/main
Listening on 127.0.0.1:56623
```

Maybe you don't need this. For a usage as a library this doesn't make sense.

Not sure of what would be the best way to explain this... I would elaborate/discuss all the possible install scenarios/usages to document them all or just strip the install part in the meantime...

As for modules, there are different approachs to install tools a system binary (see `global` approach in the link below), as a project binary (`tools.go` approach), or just as a library (`go get` into your project's `go.mod` and build your own binary as usual).

https://github.com/golang/go/wiki/Modules#why-does-installing-a-tool-via-go-get-fail-with-error-cannot-find-main-module

lopezator:
So... for the `go install` part it would be optional depending if you want a binary from cel-go code or not.

Also `go mod init && go get` stanzas don't make sense if you already are in a modules aware project and just want to start using `cel-go` as a library, then just importing `cel-go` and running a `go build` or `go run` to your code would do all the job for you.

TristonianJones:
I think the common case is to use CEL as a library. Since Go is tackling dependency resolution with modules it seems reasonable to just point folks in that direction. There's a lot of nuance, but `mod init` and adding imports to an application is typically how I see CEL being used. Thoughts?

lopezator:
LGTM now, the previous install instructions might confuse more than help.

TristonianJones:
Thanks David

lopezator:
Thanks to you for your awesome work, CEL is doing our work much easier!! 👏

Issue #204, opened 2019-04-09
Use struct tags to parse function arguments


tobiaszheller:
**Change**

Use struct tags to parse arguments in function definition.

Struct tag should define position of variable in arguments passed to function.

**Example**
```
func funcA(argsVal ...ref.Val) ref.Val {
	var args struct {
		A bool    `position:"0"`
		B float64 `position:"1"`
		C int     `position:"2"`
	}
	err := parseArgs(&args, argsVal...)
}
```
What do you think about that? I am willing to implement that.

TristonianJones:
Hi Tobiaz,

Could you give me a bit more context on the problem this is trying to solve? For example, does this improve guarantees of correctness? Or unify the declaration + definition steps? 

My preference is to avoid two ways of doing the same thing, but I certainly understand there are areas of improvement in the following areas:
1. Working with protos.
2. Separation of declaration from definition, e.g. function signature separate from function impl.
3. Weak guarantees of type-safety in the runtime function definitions.

Thanks,

-Tristan

tobiaszheller:
@TristonianJones 
I think that it will remove a lot of boilerplate code when adding overload functions.
Let's take a look of function with multiple arguments [cel_test.go](https://github.com/google/cel-go/blob/6a046a90da28fd89ea6f11b5834b84d5143ca6db/cel/cel_test.go#L341-L361).
You need to check len of arguments first, then cast every argument to check if it has proper type.
The idea about that feature is to close that boilerplate stuff behind function which will do the magic. It return error if something goes wrong or update variable.

This should be optional but can be helpful if you define a lot of functions overloads.

TristonianJones:
I like the idea of making the overload declaration simpler in the common case. The tradeoff here is between ease of use (tags) and performance (hand-roll). Also, the `planner.go` relies pretty heavily on the specific structure of the existing overloads to optimize the execution plan. 

I think in the ideal case, you would use a feature like this to communicate everything the type-checker might want that it currently gets out of `cel.Declarations` and everything that the runtime might want to enforce the declaration contract in `cel.Functions`. 

I guess I imagine something along the following lines:

```go
func andFunc() cel.Binding {
   // a.and(b) returns true if b is `a` is `true` and `b` is a truthy double. 
   var argsTypes struct {
        a types.Bool      `in: "0", target: "true"`
        b types.Double  `in: "1"`
        c types.Bool      `out: "true"`
    }
[...]

Issue #210, opened 2019-04-29
Eval panic recovery


tobiaszheller:
**Feature request checklist**

- [x] There are no issues that match the desired change
- [ ] The change is large enough it can't be addressed with a simple Pull Request
- [ ] If this is a bug, please file a [Bug Report](./ISSUE_TEMPLATE).

**Change**
Top level panic recovery for `program.Eval` method. 

Based on recent bugs #208 #209, `Eval` method is not so safe and could cause panic. I would expect from expressions interpreter to have top level recovery function and handle panics.

TristonianJones:
That's a very reasonable request. I'll add it to the todo list. :+1:

PR #212, opened 2019-04-29
Improve examples and documentation

tobiaszheller:
Improve examples according to discussion in #204.

### DIFF
--- a/cel/example_goverload_test.go
+++ b/cel/example_goverload_test.go
@@ -50,11 +50,12 @@ func Example_globalOverload() {
 			Binary: func(lhs ref.Val, rhs ref.Val) ref.Val {
 				s1, ok := lhs.(types.String)
 				if !ok {
-					return types.NewErr("unexpected type '%v' passed to shake_hands", lhs.Type())
+					return types.ValOrErr(lhs, "unexpected type '%v' passed to shake_hands", lhs.Type())
 				}
-				// If expression was type checked and dynamic data is not used (JSON, maps),
-				// type checking can be skipped.
-				s2 := rhs.(types.String)
+				s2, ok := rhs.(types.String)
+				if !ok {
+					return types.ValOrErr(rhs, "unexpected type '%v' passed to shake_hands", rhs.Type())
+				}
 				return types.String(
 					fmt.Sprintf("%s and %s are shaking hands.\n", s1, s2))
 			}})
--- a/cel/example_goverload_test.go
+++ b/cel/example_goverload_test.go
@@ -1,0 +1,77 @@
+package cel_test
+
+import (
+	"fmt"
+	"log"
+
+	exprpb "google.golang.org/genproto/googleapis/api/expr/v1alpha1"
+
+	"github.com/google/cel-go/cel"
+	"github.com/google/cel-go/checker/decls"
+	"github.com/google/cel-go/common/types"
+	"github.com/google/cel-go/common/types/ref"
+	"github.com/google/cel-go/interpreter/functions"
+)
+
+// ExampleGlobalOverload demonstrates how to define global overload function.
+func Example_globalOverload() {
+	// Create the CEL environment with declarations for the input attributes and
+	// the desired extension functions. In many cases the desired functionality will
+	// be present in a built-in function.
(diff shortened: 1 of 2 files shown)


cel/cel_test.go
```diff
@@ -45,7 +44,13 @@ func Example() {
 		// Function to generate a greeting from one person to another.
 		//    i.greet(you)
 		decls.NewFunction("greet",
-			decls.NewInstanceOverload("greet_string_string",
+			decls.NewInstanceOverload("greet_i_string_string",
+				[]*exprpb.Type{decls.String, decls.String},
+				decls.String)),
+		// Function to generate shake_hands between two people.
```
TristonianJones: Actually, would you mind creating a separate example for Instance vs. Global overload?

cel/cel_test.go
```diff
@@ -70,6 +75,15 @@ func Example() {
 			Binary: func(lhs ref.Val, rhs ref.Val) ref.Val {
 				return types.String(
 					fmt.Sprintf("Hello %s! Nice to meet you, I'm %s.\n", rhs, lhs))
+			}},
+		&functions.Overload{
+			Operator: "shake_hands",
+			Binary: func(lhs ref.Val, rhs ref.Val) ref.Val {
+				// It is safe to skip ok with assertion if expression was type checked.
```
TristonianJones: In theory, we should still be type-testing the inputs. It's safer to use the values without type testing if you use the checker, but not exactly safe since you could get unexpected input as evidenced by the bugs you filed. Thanks for those reports, btw.

TristonianJones (changes requested):

cel/cel_test.go
tobiaszheller: I think when #210 is solved then it could be safe to use following approach. 
If you are not fine with it, I can use check is assertion was fine in examples however that was my idea how this could be simplified.

TristonianJones: The safety level really depends on how much you trust the input you're sending to the CEL expression. If you're always validating the input types and control the expressions, then type-testing isn't needed. If either the input or the expression leverages dynamic data (e.g. JSON, maps) which could be user provided then a type assertion of the arguments is strongly suggested.

tobiaszheller:
@TristonianJones PTAL

cel/example_goverload_test.go
```diff
@@ -0,0 +1,77 @@
...
+	}
+
+	// Create the program.
+	funcs := cel.Functions(
+		&functions.Overload{
+			Operator: "shake_hands",
+			Binary: func(lhs ref.Val, rhs ref.Val) ref.Val {
+				s1, ok := lhs.(types.String)
+				if !ok {
+					return types.NewErr("unexpected type '%v' passed to shake_hands", lhs.Type())
```
TristonianJones: Prefer instead to use:
```
types.ValOrErr(lhs, "no such overload")
```

cel/example_goverload_test.go
```diff
@@ -0,0 +1,77 @@
...
+				}
+				// If expression was type checked and dynamic data is not used (JSON, maps),
+				// type checking can be skipped.
+				s2 := rhs.(types.String)
```
TristonianJones: Consider doing the same type-assertion as above:

```
s2, ok := rhs.(types.String)
if !ok {
   return types.ValOrErr(rhs, "no such overload")
}
```

If you control the input types, then you're probably correct that the arguments will already be well formed. However, if you expose these functions for use in a customer written expression or evaluate against dynamic data provided by a customer, then there is some possibility that the type-check contract will not line up with the runtime invocation. 

I don't mind documenting this sort of information in a wiki page, but for examples, I would prefer them to be somewhat bullet-proof since people are apt to copy-paste them and not think through the implications.

cel/example_goverload_test.go
tobiaszheller: done

cel/example_goverload_test.go
tobiaszheller: 👍

TristonianJones (approved):

TristonianJones:
Thanks for the contribution!

Issue #208, opened 2019-04-27
in operator causes panic for map


yangminzhu:
**Describe the bug**
I'm not sure if I'm missing anything, the `in` operator causes panic for `map[string]string` in my test code.

**To Reproduce**
Check which components this affects:

- [ ]  parser
- [ ]  checker
- [x]  interpreter

Test setup:
```go
func Test_MyTest(t *testing.T) {
	decls := Declarations(decls.NewIdent("g", decls.NewMapType(decls.String, decls.String), nil))
	env, err := NewEnv(decls)
	if err != nil {
		t.Fatalf("environment creation error: %s\n", err)
	}

	// The following expression will trigger panic
	p, iss := env.Parse(`"g1" in g`)

	// The following expression works as expected, no panic
	// p, iss := env.Parse(`"g1" in {"g1": "g2"}`)

	if iss != nil && iss.Err() != nil {
		t.Fatal(iss.Err())
	}
	c, iss := env.Check(p)
	if iss != nil && iss.Err() != nil {
		t.Fatal(iss.Err())
	}

	prg, err := env.Program(c)
	if err != nil {
		t.Fatalf("program creation error: %s\n", err)
	}

	vars := map[string]interface{}{
		"g": map[string]string{"g1": "g2"},
	}

	// The following vars work as expected, no panic
	// vars := map[string]interface{}{
	// 	 "g": map[string]interface{}{"g1": "g2"},
	// }

	out, _, err := prg.Eval(vars)
	if err != nil {
		t.Fatalf("runtime error: %s\n", err)
	}

	if out != types.True {
		t.Errorf(`Got '%v', wanted true`, out.Value())
	}
}
```

**Actual output**
```
=== RUN   Test_MyTest
--- FAIL: Test_MyTest (0.00s)
panic: reflect: call of reflect.Value.Type on zero Value [recovered]
	panic: reflect: call of reflect.Value.Type on zero Value

goroutine 19 [running]:
testing.tRunner.func1(0xc0001ac100)
	/usr/local/go/src/testing/testing.go:792 +0x387
panic(0x984e00, 0xc000292f40)
	/usr/local/go/src/runtime/panic.go:513 +0x1b9
reflect.Value.Type(0x0, 0x0, 0x0, 0x0, 0xa)
	/usr/local/go/src/reflect/value.go:1743 +0x16c
[...]

PR #213, opened 2019-04-30
Missing contains override fix

TristonianJones:
Missing override fix for the #208 as well as fixes to the `Contains` semantics for list and map
types. The `Contains` behavior, represented as `in` within CEL, is equivalent to iterating over
the entries in the list or map and determining whether any of the entries is `Equal`.  Since lists
and maps may be heterogeneous, it is possible to encounter errors during evaluation of `Contains`
that must not short-circuit the evaluation.

### DIFF
--- a/common/types/list.go
+++ b/common/types/list.go
@@ -83,10 +83,23 @@ func (l *baseList) Add(other ref.Val) ref.Val {
 }
 
 func (l *baseList) Contains(elem ref.Val) ref.Val {
+	if IsUnknownOrError(elem) {
+		return elem
+	}
+	var err ref.Val
 	for i := Int(0); i < l.Size().(Int); i++ {
-		if l.Get(i).Equal(elem) == True {
+		val := l.Get(i)
+		cmp := elem.Equal(val)
+		b, ok := cmp.(Bool)
+		if !ok && err == nil {
+			err = ValOrErr(cmp, "no such overload")
+		}
+		if b == True {
 			return True
 		}
+	}
+	if err != nil {
+		return err
 	}
 	return False
 }
@@ -170,10 +183,10 @@ func (l *baseList) Equal(other ref.Val) ref.Val {
 }
 
 func (l *baseList) Get(index ref.Val) ref.Val {
-	if index.Type() != IntType {
+	i, ok := index.(Int)
+	if !ok {
 		return ValOrErr(index, "unsupported index type '%s' in list", index.Type())
 	}
-	i := index.(Int)
 	if i < 0 || i >= l.Size().(Int) {
 		return NewErr("index '%d' out of range in list size '%d'", i, l.Size())
 	}
--- a/common/types/list.go
+++ b/common/types/list.go
@@ -19,6 +19,7 @@ import (
 	"reflect"
 
 	structpb "github.com/golang/protobuf/ptypes/struct"
+
 	"github.com/google/cel-go/common/types/ref"
 	"github.com/google/cel-go/common/types/traits"
 )
@@ -68,19 +69,20 @@ type baseList struct {
 
 // Add implements the traits.Adder interface method.
 func (l *baseList) Add(other ref.Val) ref.Val {
-	if other.Type() != ListType {
+	otherList, ok := other.(traits.Lister)
+	if !ok {
 		return ValOrErr(other, "no such overload")
 	}
 	if l.Size() == IntZero {
 		return other
 	}
-	if other.(traits.Sizer).Size() == IntZero {
+	if otherList.Size() == IntZero {
 		return l
 	}
 	return &concatList{
 		TypeAdapter: l.TypeAdapter,
 		prevList:    l,
-		nextList:    other.(traits.Lister)}
+		nextList:    otherList}
 }
 
 // Contains implements the traits.Container interface method.
@@ -89,7 +91,8 @@ func (l *baseList) Contains(elem ref.Val) ref.Val {
 		return elem
 	}
 	var err ref.Val
-	for i := Int(0); i < l.Size().(Int); i++ {
+	sz := l.Size().(Int)
+	for i := Int(0); i < sz; i++ {
 		val := l.Get(i)
 		cmp := elem.Equal(val)
 		b, ok := cmp.(Bool)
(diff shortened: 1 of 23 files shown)


TristonianJones:
@yangminzhu Let me know if this change addresses your original report.

common/types/list_test.go
```diff
@@ -18,11 +18,11 @@ import (
 	"reflect"
 	"testing"
 
-	"github.com/golang/protobuf/jsonpb"
-	"github.com/golang/protobuf/proto"
 	"github.com/google/cel-go/common/types/ref"
 	"github.com/google/cel-go/common/types/traits"
 
+	"github.com/golang/protobuf/jsonpb"
```
madi8229, line 24: why move these two import to here?

common/types/list_test.go
```diff
@@ -18,11 +18,11 @@ import (
...
+	"github.com/golang/protobuf/proto"
 	dpb "github.com/golang/protobuf/ptypes/duration"
```
madi8229, line 26: Should there be a seperate blank line between dpb and "github.com/golang/protobuf/proto"?

common/types/list_test.go
```diff
@@ -414,3 +462,40 @@ func getElem(t *testing.T, list traits.Indexer, index Int) interface{} {
...
+	var i = int64(0)
+	for ; it.HasNext() == True; i++ {
+		elem := it.Next()
+		if getElem(t, list, Int(i)) != elem {
+			t.Errorf(
+				"List iterator returned incorrect value: list[%d]: %v", i, elem)
+		}
+	}
+	if it.Next() != nil {
+		t.Errorf("List iterator attempted to continue beyond list size")
```
madi8229, line 496: or HasNext() has something wrong?

common/types/map.go
```diff
@@ -62,10 +62,27 @@ var (
 		traits.SizerType)
 )
 
+// Contains implements the traits.Container interface method.
 func (m *baseMap) Contains(index ref.Val) ref.Val {
-	return Bool(m.Get(index).Type() != ErrType)
+	val, found := m.Find(index)
+	// When the index is not found and val is non-nil, this is an error or unknown value.
+	if !found && val != nil && IsUnknownOrError(val) {
```
madi8229, line 69: what happened if IsUnknownOrError(val) == False but two previous condition is True?

yangminzhu:
@TristonianJones Thank you for fixing this, I patched your fix and tested using the same unit test in #208 and it now passes.

common/types/map_test.go
```diff
@@ -20,6 +20,8 @@ import (
 
 	"github.com/golang/protobuf/jsonpb"
 	"github.com/golang/protobuf/proto"
+	structpb "github.com/golang/protobuf/ptypes/struct"
```
madi8229, line 23: Should we keep the structpb to the last of the imports and have a blank line to separate it with other imports?

common/types/map_test.go
```diff
@@ -115,6 +234,26 @@ func TestBaseMap_Get(t *testing.T) {
...
+	val := mapValue.Get(String("first"))
+	if val.Equal(String("hello")) != True {
+		t.Errorf("Got '%v', wanted 'hello'", val)
+	}
+	if !IsError(mapValue.Get(Int(1))) {
+		t.Error("Got real value, wanted error")
+	}
+	if !IsError(mapValue.Get(String("third"))) {
+		t.Error("Got real value, wanted error")
+	}
```
madi8229, line 256: Do we want different error message for these two check?

TristonianJones:
Excellent, thanks. :)

common/types/string.go
```diff
@@ -22,12 +22,12 @@ import (
 	"strings"
 	"time"
 
+	"github.com/golang/protobuf/ptypes"
+	structpb "github.com/golang/protobuf/ptypes/struct"
```
madi8229, line 26: Why move imports to here?

common/types/list.go
```diff
@@ -210,27 +240,41 @@ type concatList struct {
...
 		prevList:    l,
-		nextList:    other.(traits.Lister)}
+		nextList:    otherList}
 }
 
+// Contains implments the traits.Container interface method.
 func (l *concatList) Contains(elem ref.Val) ref.Val {
-	return Bool(l.prevList.Contains(elem) == True ||
-		l.nextList.Contains(elem) == True)
+	prev := l.prevList.Contains(elem)
```
madi8229, line 265: Can you add a comment here to explain the reuse of the unknown check or check unknown here again?  Either is good. Thank you :D

madi8229 (commented):
I found there are several changes about the imports, is there any reason you change/separate imports in this way? Thank you.

common/types/list_test.go
TristonianJones: I've started trying to make three groups of imports:
- builtins
- cel-go deps
- external deps

I probably haven't done it super consistently yet, but the motivation is to make it easier to understand internal v. external deps.

common/types/list_test.go
TristonianJones: The aliased import is still a part of the external deps list, so I think it probably belongs with the other `github.com/golang/protobuf/...` imports.

common/types/list_test.go
TristonianJones: I suppose. I believe this is an old test that just got moved. The idea would be that the `it.Next()` call should return `nil` once it reaches the end of the input and if it doesn't then, like you said, something is wrong.

common/types/map.go
TristonianJones: The `val` can only be `nil` if `found == false` so it's not possible to have the previous two conditions true.

common/types/map_test.go
TristonianJones: It's a related import to the others in this list. I'm inclined to keep it here.

common/types/map.go
madi8229: I see. Thank you .

common/types/string.go
TristonianJones: I'm trying to separate the imports into three categories. Hopefully my earlier explanation makes sense.

common/types/map_test.go
TristonianJones: The `t.Error()` call will give the correct position of the error in the source, so it's okay to have the same message.

common/types/list.go
TristonianJones: Sure thing. :)

madi8229 (approved):
Thank you for the update and explanation. LGTM.

PR #214, opened 2019-04-30
Add support for converting 'nil' to CEL 'null'

### LINKED ISSUE #209: Panic during runtime when object in vars is nil
**Describe the bug**
`prg.Eval()` panics if referencing proto object (in expression) which is declared as nil pointer (of some type) in vars map.

**To Reproduce**

- [ ]  parser
- [ ]  checker
- [x]  interpreter

```cel
test.name
```

Test setup:
Proto loaded from https://github.com/grpc/grpc-go/tree/master/examples/helloworld/helloworld
```go
func Test_CustomTypes(t *testing.T) {
    env, err := cel.NewEnv(
        cel.Types(&helloworld.HelloRequest{}),
        cel.Declarations(
            decls.NewIdent("test",
                decls.NewObjectType("helloworld.HelloRequest"), nil)))
    if err != nil {
        t.Fatalf("environment creation error: %s\n", err)
    }
    p, iss := env.Parse(`test.name`)
        t.Fatal("Check failed: ", iss)
    }
    c, iss := env.Check(p)
        t.Fatal("Check failed: ", iss)
    }
    prg, _ := env.Program(c)

    var req *helloworld.HelloRequest
    // req = &helloworld.HelloRequest{}
    vars3 := map[string]interface{}{"test": req}

    out, _, _ := prg.Eval(vars3)
}
```

**Expected behavior**
Runtime error.

TristonianJones:
Golang's `nil` is the idiomatic equivalent to CEL `null`. This change adds support for conversion from `nil` to `null`.

Closes #209

### DIFF
--- a/common/types/provider.go
+++ b/common/types/provider.go
@@ -220,9 +220,12 @@ func (p *protoTypeRegistry) NativeToValue(value interface{}) ref.Val {
 	// Override the Any type by ensuring that custom proto-types are considered on recursive calls.
 	case *anypb.Any:
 		val := value.(*anypb.Any)
+		if val == nil {
+			return NewErr("unsupported type conversion: '%T'", value)
+		}
 		unpackedAny := ptypes.DynamicAny{}
 		if ptypes.UnmarshalAny(val, &unpackedAny) != nil {
-			NewErr("Fail to unmarshal any.")
+			NewErr("unknown type: '%s'", val.GetTypeUrl())
 		}
 		return p.NativeToValue(unpackedAny.Message)
 	// Convert custom proto types to CEL values based on type's presence within the pb.Db.
@@ -231,7 +234,7 @@ func (p *protoTypeRegistry) NativeToValue(value interface{}) ref.Val {
 		typeName := proto.MessageName(pbVal)
 		td, err := p.pbdb.DescribeType(typeName)
 		if err != nil {
-			return NewErr("unknown type '%s'", typeName)
+			return NewErr("unknown type: '%s'", typeName)
 		}
 		return NewObject(p, td, pbVal)
 	// Override default handling for list and maps to ensure that blends of Go + proto types
@@ -240,6 +243,9 @@ func (p *protoTypeRegistry) NativeToValue(value interface{}) ref.Val {
 	default:
 		refValue := reflect.ValueOf(value)
 		if refValue.Kind() == reflect.Ptr {
+			if refValue.IsNil() {
+				return NewErr("unsupported type conversion: '%T'", value)
+			}
 			refValue = refValue.Elem()
 		}
 		refKind := refValue.Kind()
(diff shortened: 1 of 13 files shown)


interpreter/activation_test.go
```diff
@@ -53,3 +53,17 @@ func TestHierarchicalActivation(t *testing.T) {
 		t.Error("Activation failed to resolve child value of 'c'")
 	}
 }
+
+func TestActivation_NilValue(t *testing.T) {
+	a, _ := NewActivation(map[string]interface{}{"nil": nil})
```
tobiaszheller: I think this is not solving panic at all. It was trigger only when passing nil pointer to some defined type. For example:
```
var str *string
a, _ := NewActivation(map[string]interface{}{"nil": str})
```

TristonianJones: Ah, well, two bugs, one PR then. :)

I've update the code to do the following things:

* Check for `nil` after the type-assertions for pointers within the type-adapting code.
* For `nil` pointers to simple values (including duration and timestamp), I've chosen
  to return the empty value to be consistent with the CEL spec.
* For `nil` pointers to aggregate values and types, the type adapters will return a
  `types.Err` value as the zero value is not easy to infer and can cause undefined
  behavior in later call sequences.

In general, `nil` pointers shouldn't be provided as input to the program. The PR to
address #210 will handle error recovery in the event that other `panic`s lurk in the
shadows.

common/types/provider.go
```diff
@@ -220,9 +220,12 @@ func (p *protoTypeRegistry) NativeToValue(value interface{}) ref.Val {
 	// Override the Any type by ensuring that custom proto-types are considered on recursive calls.
 	case *anypb.Any:
 		val := value.(*anypb.Any)
+		if val == nil {
+			return NewErr("unsupported type conversion: '%T'", value)
+		}
 		unpackedAny := ptypes.DynamicAny{}
 		if ptypes.UnmarshalAny(val, &unpackedAny) != nil {
-			NewErr("Fail to unmarshal any.")
+			NewErr("unknown type: '%s'", val.GetTypeUrl())
```
gabrielsz: Missing return here

common/types/provider.go
```diff
@@ -338,17 +389,26 @@ func (a *defaultTypeAdapter) NativeToValue(value interface{}) ref.Val {
...
+		return TimestampZero
 	case *anypb.Any:
 		val := value.(*anypb.Any)
+		if val == nil {
+			return NewErr("unsupported type conversion")
+		}
 		unpackedAny := ptypes.DynamicAny{}
 		if ptypes.UnmarshalAny(val, &unpackedAny) != nil {
-			NewErr("Fail to unmarshal any.")
+			NewErr("unknown type: %s", val.GetTypeUrl())
```
gabrielsz: missing return here

gabrielsz (commented):
Hi! 

Please take a look at the 2 comments. If you're considering covering the subtle new cases with unit tests, the tests that would cover this are:
 * `nil` Duration pointer
 * `nil` Timestamp
 * An invalid conversion to `NewJSONList`
 * An invalid conversion to `NewJSONStruct`

Maybe not all of these are of interest, just tried to be exhaustive here =)

interpreter/activation_test.go
TristonianJones: Hmm, actually special handling for the `nil` pointer types is proving problematic. I've opted to instead treat `nil` pointers as errors and return a `types.Err` value when they're encountered.

TristonianJones:
@gabrielsz PTAL

interpreter/activation.go
```diff
@@ -108,15 +97,12 @@ func (a *mapActivation) ResolveName(name string) (ref.Val, bool) {
 		// Resolve a lazily bound value.
 		case func() ref.Val:
 			val := object.(func() ref.Val)()
+			val = a.adapter.NativeToValue(val)
```
gabrielsz, line 107: Will `a.adapter` be non-nil all the time?

gabrielsz (approved):
Thanks for the updates, and the cleanup!

I'm only seeing one point for you to check at, but it might be a false positive. Thanks again!

TristonianJones:
Good point, I should check the input arg to make sure it is

interpreter/activation.go
TristonianJones: Yes, it should be, but I'll add an input test to make sure it's non-nil.

PR #215, opened 2019-05-01
Fix the local variable references within the parser.

TristonianJones:
The parser tests didn't seem to be using appropriate local variable references to verify 
test case behavior. The code has been updated to match the checker_test.go.


JimLarson (approved):
LGTM.  Suggest that you scan the repo for other occurrences of the same pattern.

TristonianJones:
I took a pass and didn't notice any other test cases that suffered from this problem, but I'll add a comment about it for future reference.

PR #216, opened 2019-05-01
Add support for panic recovery in program.Eval()

### LINKED ISSUE #210: Eval panic recovery

TristonianJones:
Recover from panics during evaluation. Note, that there is still room to improve by adding recovery
support into the checker and parser, but this will be tackled as a future PR.

Closes #210 report which discusses panic recovery during Eval.

### DIFF
--- a/cel/cel_test.go
+++ b/cel/cel_test.go
@@ -439,6 +439,33 @@ func Test_EvalOptions(t *testing.T) {
 	}
 }
 
+func Test_EvalRecover(t *testing.T) {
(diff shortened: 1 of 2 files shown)


gabrielsz (approved):
LGTM

PR #217, opened 2019-05-01
Commit generated file descriptor for team.proto

TristonianJones:
The `cel/testdata/team.fds` file is typically generated during a Bazel build, but in order to make
it easier to get started with just `go test` against the `github.com/google/cel-go/cel` directory, the
generated artifact is being added to the repo.

### DIFF
--- a/cel/testdata/team.fds
+++ b/cel/testdata/team.fds
@@ -1,0 +1,13 @@
+
+†
+cel/testdata/mutant.protocel.testdata"S
+Mutant
+name (	Rname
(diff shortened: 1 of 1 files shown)


TristonianJones:
@KristaHyer1 committing this file should unblock you. I'll look into the Bazel build in the mean time.

KristaHyer1:
Thanks for this! The tests are now all passing on my VM.

gabrielsz (approved):
Seems like a reasonable workaround for now =)

PR #218, opened 2019-05-01
Fix simple identifier resolution within a container.

TristonianJones:
Select expressions can represent qualified identifiers within a CEL package / container,
and when they look identifier-like, the code attempts to resolve the name to the most
suitable container. Identifiers have the same property per the CEL spec, but were not 
implemented this way. This PR fixes the glitch.

Note, once committed the cel-spec `broken.textproto` test case which refers to simple
identifier resolution within a container will be moved into the `basic.textproto`.


madi8229 (approved):
LGTM.

PR #220, opened 2019-05-03
Optimize 'in' operations on homogeneous list literals as set membership tests.

TristonianJones:
When CEL evaluates the following expression, the developer's intent is to test whether an element exists within an enumerated value space:

```javascript
elem in ['a', 'b', 'c', 'd']
```

Currently the `in` operator is implemented as a per-element comparison between the `elem` and each value in the list. This is rather inefficient. When the list elements are all of the same type and the list literal is composed of constant values, this expression can be converted to a map key existence test where the value in the list becomes a key in the map.

This optimization only works for primitive types and only when the list literal is defined in CEL. Additionally, this feature must be enabled with the `cel.OptOptimize` flag and a new canonical benchmark was added to demonstrate the overall performance.

**Before**
```
ExprBench/complex-12    1000000    1172 ns/op	152 B/op    10 allocs/op
```

**After**
```
ExprBench/complex-12    2000000     871 ns/op    144 B/op     9 allocs/op
```

### DIFF
--- a/interpreter/decorators.go
+++ b/interpreter/decorators.go
@@ -77,94 +77,104 @@ func decDisableShortcircuits() InterpretableDecorator {
 	}
 }
 
-// decFoldConstants checks whether the arguments the create list and map operations are all
-// constant values and constructs a new aggregate value. Future improvements to this method will
-// also do the same for typed object creations and functions whose arguments are constant.
-func decFoldConstants() InterpretableDecorator {
+// decOptimize optimizes the program plan by looking for common evaluation patterns and
+// conditionally precomputating the result.
+// - build list and map values with constant elements.
+// - convert 'in' operations to set membership tests if possible.
+func decOptimize() InterpretableDecorator {
 	return func(i Interpretable) (Interpretable, error) {
 		switch i.(type) {
 		case *evalList:
-			l := i.(*evalList)
-			for _, elem := range l.elems {
-				_, isConst := elem.(*evalConst)
-				if !isConst {
-					return i, nil
-				}
-			}
-			val := l.Eval(EmptyActivation())
-			return &evalConst{
-				id:  l.id,
-				val: val,
-			}, nil
+			return maybeBuildListLiteral(i)
 		case *evalMap:
-			mp := i.(*evalMap)
-			for idx, key := range mp.keys {
-				_, isConst := key.(*evalConst)
-				if !isConst {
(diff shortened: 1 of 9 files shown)


interpreter/decorators.go
```diff
@@ -76,47 +77,128 @@ func decDisableShortcircuits() InterpretableDecorator {
...
-				id:  mp.id,
-				val: val,
-			}, nil
 		}
 		return i, nil
 	}
 }
 
+func maybeBuildListLiteral(i Interpretable) (Interpretable, error) {
+	l := i.(*evalList)
```
gabrielsz: Would you mind promoting these back to the `switch` statement so that you receive an `evalList` here? The reason is that once this list grows, it will be easier to make a mistake on casting it incorrectly. This same comment would apply to all `func`s called from the switch statement above.

interpreter/decorators.go
```diff
@@ -76,47 +77,128 @@ func decDisableShortcircuits() InterpretableDecorator {
...
+func maybeOptimizeSetMembership(i Interpretable) (Interpretable, error) {
+	call := i.(*evalBinary)
+	if call.overload != overloads.InList {
+		return i, nil
+	}
+	l, isConst := call.rhs.(*evalConst)
+	if !isConst {
+		return i, nil
+	}
+	list := l.val.(traits.Lister)
```
gabrielsz, line 144: Will `l.val` always be cast-eable to `traits.Lister`?

interpreter/decorators.go
```diff
@@ -76,47 +77,128 @@ func decDisableShortcircuits() InterpretableDecorator {
...
+	valueSet := make(map[ref.Val]ref.Val)
+	for it.HasNext() == types.True {
+		elem := it.Next()
+		elemType := elem.Type()
+		if elemType != types.BoolType &&
+			elemType != types.BytesType &&
+			elemType != types.DoubleType &&
+			elemType != types.StringType &&
+			elemType != types.IntType &&
+			elemType != types.UintType {
```
gabrielsz: I see Timestamp and Duration are excluded, as they have the protobuf backing. I see no reason to generalize this type selection behavior for now, but maybe we could have a line explaining something like:

```// non-native types can't be optimized```

gabrielsz (approved):
Great work on the optimizations! Left some optional comments inline.

PR #221, opened 2019-05-15
Refactor of the interpreter benchmarks for consistency.

TristonianJones:
Make it easier and more consistent to specify interpreter benchmarks with varying environments.

### DIFF
--- a/interpreter/interpreter_test.go
+++ b/interpreter/interpreter_test.go
@@ -724,7 +724,7 @@ var (
 	interpreter = NewStandardInterpreter(packages.DefaultPackage, reg, reg)
 	testData    = []testCase{
 		{
-			name: `ok_1st`,
+			name: "or_true_1st",
 			E:    `ai == 20 || ar["foo"] == "bar"`,
 			Env: []*exprpb.Decl{
 				decls.NewIdent("ai", decls.Int, nil),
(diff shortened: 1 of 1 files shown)


interpreter/interpreter_test.go
```diff
@@ -39,7 +39,8 @@ import (
 type testCase struct {
 	name string
 	E    string
-	I    map[string]interface{}
+	Env  []*exprpb.Decl
+	I    interface{}
```
gabrielsz, line 43: All uses below still assume it's a `map[string]interface{}`. Just wanted to double check you'd like to lessen the restriction here.

interpreter/interpreter_test.go
```diff
@@ -791,23 +766,57 @@ var (
...
+			E:    `elems.exists(e, type(e) == uint)`,
+			Env: []*exprpb.Decl{
+				decls.NewIdent("elems", decls.NewListType(decls.Dyn), nil),
+			},
+			I: map[string]interface{}{
+				"elems": []interface{}{0, 1, 2, 3, 4, uint(5), 6},
+			},
+		},
+		{
+			name: `complex`,
```
gabrielsz: Super nitpick, but can you change the backticks to double quotes (or the other way around)?

gabrielsz (approved):
LGTM. Added 2 minor comments

interpreter/interpreter_test.go
TristonianJones: `EmptyActivation()` isn't actually a `map[string]interface{}` in some of the test cases below, so the type-widening was intentional.

interpreter/interpreter_test.go
TristonianJones: Sure

PR #223, opened 2019-05-25
Known unknowns

### LINKED ISSUE #190: Reference to undefined variable should be error instead of unknown
**Describe the bug**
The interpreter currently returns an "unknown" value for undefined variables.  It ought to return an error.

**To Reproduce**

- [ ]  parser
- [ ]  checker
- [X]  interpreter

```cel
// sample expression string
x
```

Test setup:
```go
// test case in go
Evaluate the expression `x` without binding `x`.
```

**Expected behavior**
It should return an error for an undefined variable.

**Additional context**
To evaluate for `x` being unknown (vs undefined), it can be bound to an unknown value.

TristonianJones:
Support for partial evaluation and marking known unknowns.

- `interpreter.PartialActivation` combines unknown attribute paths with a set of known attributes.
- `interpreter.UnknownActivation` treats all attributes as unknown and preserves the v0.2.0 behavior with `EmptyActivation`
- `interpreter.Attribute` interface unifies identifier, select, and index expression behavior.
- `interpreter.Resolver` interface resolves an `Attribute` from an `Activation` or relative object reference.
- (In)equality optimizations for attribute comparisons to constant values.
- Refactored tests to simplify new tests with shared test sets for benchmarks.

Note, while the code in this pull provides a way to listen to attribute resolution, this functionality has not yet been exposed on the top-level interfaces. This will come in a follow up PR.

Closes #190


cel/cel_test.go
```diff
@@ -91,6 +91,70 @@ func Example() {
 	// Output:Hello world! Nice to meet you, I'm CEL.
 }
 
+// ExampleGlobalOverload demonstrates how to define global overload function.
```
JimLarson: Match the function name in the comment.  Also add "a" in:
...how to define *a* global overload function.

cel/options.go
```diff
@@ -236,6 +235,12 @@ const (
 	// creation time. This flag is useful when the expression will be evaluated repeatedly against
 	// a series of different inputs.
 	OptOptimize EvalOption = 1 << iota
+
+	// OptUnknownAttributes indicates to the program that the input may only be partially known at
```
JimLarson, line 239: Define "Attributes" somewhere in API godocs, and in the top-level README.md too.

cel/program.go
```diff
@@ -50,11 +50,43 @@ type EvalDetails interface {
...
+	}
+	if len(unknowns) == 0 {
+		return vars, nil
+	}
+	return interpreter.NewPartialActivation(vars, unknowns)
+}
+
+// Unknown specifies an Attribute whose value is not known at runtime.
+//
+// Not all inputs are created equal, some are cheap to provide and others require I/O or heavy
```
gabrielsz: :+1:

cel/program.go
```diff
@@ -108,6 +136,15 @@ func newProgram(e *env, ast Ast, opts ...ProgramOption) (Program, error) {
 		}
 	}
 
+	// Configure the resolver to support unknown attributes if the eval option is set.
+	resolver := interpreter.NewResolver(e.adapter)
+	if p.evalOpts&OptUnknownAttributes == OptUnknownAttributes {
+		resolver = interpreter.NewUnknownResolver(resolver)
+	}
+	interp := interpreter.NewInterpreter(
```
gabrielsz: Nit: inline `p.interpreter`?

interpreter/interpretable.go
```diff
@@ -748,3 +726,345 @@ func (fold *evalExhaustiveFold) Eval(ctx Activation) ref.Val {
...
+	// the expectation is that the Resolver will return an error.
+	v, found := e.resolver.Resolve(vars, e.attr)
+	if found {
+		return e.adapter.NativeToValue(v)
+	}
+	return types.NewErr("no such attribute")
+}
+
+func (e *evalAttr) addField(id int64, name types.String) attrInst {
+	// treat field and index selection the same way for absoluate attributes.
```
gabrielsz, line 780: "absolute"

interpreter/interpretable.go
```diff
@@ -748,3 +726,345 @@ func (fold *evalExhaustiveFold) Eval(ctx Activation) ref.Val {
...
+
+func (e *evalRelAttr) getResolver() Resolver {
+	return e.resolver
+}
+
+func (e *evalRelAttr) getAttrs() []Attribute {
+	return []Attribute{e.attr}
+}
+
+// evalConstEq describe the case where an attribute is compared against a constant value.
```
gabrielsz, line 929: If `evalConstEq` means "the set of objects that conform to  `evalConstEq`", then replace "the" with "an" in "When the evalConstEq is produced", otherwise add an S to `describe**s**`. Same in evalConstNe

interpreter/interpretable.go
```diff
@@ -748,3 +726,345 @@ func (fold *evalExhaustiveFold) Eval(ctx Activation) ref.Val {
...
+//
+// When the evalConstEq is produced, this implies that the attribute value may be comparable in
+// its Go native form without conversion first to a CEL representation.
+//
+// Note, the following implementation replicates some or all of the equality logic defined on a
```
gabrielsz, line 934: Nit: //TODO-worthy?

gabrielsz (approved):
Thanks for bringing the `unknown` world to CEL! Changes look very readable and I didn't find any issues.

cel/program.go
TristonianJones: I've refactored the code here a bit to reduce the boiler-plate. Hopefully it looks better now.

interpreter/attributes.go
```diff
@@ -0,0 +1,186 @@
...
+type Variable interface {
+	// ID indicates the expression id where the variable appears.
+	//
+	// Non-zero when the Variable is identified during program plan time.
+	//
+	// Zero-ish when the Variable is provided as part of a user-provided Attribute specification
+	// for use with partial state evaluation.
+	ID() int64
+
+	// Name of the variable, may be simple or namespaced, e.g. 'in' or 'cloud.iam.resource'.
```
kyessenov: `in` is a keyword, so cannot be a variable, right?

TristonianJones: Good point, I'll update the comment.

cel/program.go
```diff
@@ -50,11 +50,44 @@ type EvalDetails interface {
...
+	return interpreter.UnknownActivation()
+}
+
 // NoVars returns an empty Activation.
 func NoVars() interpreter.Activation {
 	return interpreter.EmptyActivation()
 }
 
+// Vars returns an Activation containing the program input and an optional list of attributes which
+// are known at evaluation time to be unknown.
```
JimLarson: "Known unknowns" hurts my head. :-)  Maybe "...which are _declared_ at evaluation time to be unknown"?  Also, explain what "program input" is expected to be - the API accepts absolutely anything.

cel/program.go
```diff
@@ -50,11 +50,44 @@ type EvalDetails interface {
...
+//
+// Not all inputs to an expression have the same cost. Some are cheap to provide and others
+// require I/O or heavy compute to resolve. Marking expensive attributes as unknown can be a
+// useful tool for improving the operational efficiency of expression evaluation, as it allows
+// the cheap data to be evaluated quickly and often conclusively. When the expression result
+// cannot be computed, the result will be a types.Unknown value containing the ids of the
+// sub-expressions that resulted in the Unknown value.
+//
+// Unknowns are treated like recoverable errors within CEL and take precedence over types.Err
+// values.
```
JimLarson: Document what pathElems should be - the API allows absolutely anything.  And since this is customer-facing documentation, briefly state that it returns (attr, nil) or (nil, err).

cel/program.go
```diff
@@ -135,36 +148,53 @@ func newProgram(e *env, ast Ast, opts ...ProgramOption) (Program, error) {
...
+type prog struct {
+	*env
+	evalOpts      EvalOption
+	defaultVars   interpreter.Activation
+	dispatcher    interpreter.Dispatcher
+	interpreter   interpreter.Interpreter
+	interpretable interpreter.Interpretable
+}
+
+// Eval implements the Program interface method.
```
JimLarson, line 194: Specify what the argument should be.

interpreter/attributes.go
```diff
@@ -0,0 +1,186 @@
...
+// is compared to the set of known-unknowns which refer to the same top-level variable. When the
+// unknown Attribute matches any part of the resolved attribute, a types.Unknown value is
+// generated. The types.Unknown value refers to the expression id of the last resolved path
+// element that matches the unknown.
+func NewUnknownAttribute(varName string, pathElems ...interface{}) (Attribute, error) {
+	v := &variable{name: varName}
+	p := make([]*PathElem, len(pathElems), len(pathElems))
+	for i, elem := range pathElems {
+		var e ref.Val
+		switch elem.(type) {
```
JimLarson: Consider using the feature where `switch t := elem.(type) {` allows you to use just `t` instead of `elem.(bool)`.

interpreter/resolver.go
```diff
@@ -0,0 +1,564 @@
...
+type defaultResolver struct {
+	adapter ref.TypeAdapter
+}
+
+// Resolve implements the Resolver interface method.
+func (res *defaultResolver) Resolve(vars Activation, attr Attribute) (interface{}, bool) {
+	varName := attr.Variable().Name()
+	attrPath := attr.Path()
+	obj, found := vars.Find(varName)
+	if found {
```
JimLarson: Invert the test to keep the interesting flow outdented.  Consider calling ResolveRelative() to ensure consistency between the two functions.

interpreter/resolver.go
```diff
@@ -0,0 +1,564 @@
...
+	m := obj.(map[string]string)
+	key, ok := k.(types.String)
+	if !ok {
+		return types.ValOrErr(k, "no such overload")
+	}
+	v, found := m[string(key)]
+	if !found {
+		return types.NewErr("no such key")
+	}
+	return v
```
JimLarson, line 221: Ugh, so much repetition!  Maybe you could factor out the key conversion from all these, but I'm not sure it would be worth it.

interpreter/attributes.go
```diff
@@ -0,0 +1,186 @@
...
+	//
+	// If Variable is nil, the Path must contain at least one PathElem.
+	Path() []*PathElem
+
+	// Select a path element from the current Attribute to produce a more fully qualified
+	// Attribute.
+	Select(*PathElem) Attribute
+}
+
+// Variable refers to a type-level expression identifier corresponding to program input.
```
JimLarson, line 44: Rename to "VariableUse", since it's for a particular use of a variable.

interpreter/attributes.go
```diff
@@ -0,0 +1,186 @@
...
+type Variable interface {
+	// ID indicates the expression id where the variable appears.
+	//
+	// Non-zero when the Variable is identified during program plan time.
+	//
+	// Zero-ish when the Variable is provided as part of a user-provided Attribute specification
+	// for use with partial state evaluation.
```
JimLarson: Might users provide attribute specifications for other reasons, e.g. optimization or monitoring?  If so, delete the "...for use with..." qualifier.

interpreter/resolver_test.go
```diff
@@ -0,0 +1,272 @@
...
+			0,
+			uint(1),
+			map[string]string{"str": "what is the answer to the ultimate question?"},
+			map[string]int{"int": 42},
+		},
+	})
+	resolvedAttrs := make(map[int64]string)
+	resolvedStatus := make(map[int64]ResolveStatus)
+	statusListener := func(attr Attribute, status ResolveStatus) {
+		exprID := attr.Variable().ID()
```
JimLarson: Comment that you use the last exprID in the path.

interpreter/resolver_test.go
```diff
@@ -0,0 +1,272 @@
...
+		attrName := attr.Variable().Name()
+		for _, elem := range attr.Path() {
+			exprID = elem.ID
+			elemName := elem.ToValue(vars)
+			attrName += fmt.Sprintf(".%v", elemName)
+		}
+		resolvedAttrs[exprID] = attrName
```
JimLarson: Test that there's no entry in these maps for this exprID already.

Consider just having one map of attrName to status and avoid using exprID at all.

interpreter/resolver_test.go
```diff
@@ -0,0 +1,272 @@
...
+		resolvedStatus[exprID] = status
+	}
+	res := NewListeningResolver(
+		NewUnknownResolver(
+			NewResolver(types.DefaultTypeAdapter)),
+		statusListener)
+
+	// When a top-level variable contains an unknown attribute, then it too becomes
+	// unknown. Otherwise, the use of this value in comparisons might yield invalid
+	// results.
```
JimLarson: Document this clearly in the Godoc for NewUnknownAttribute.

interpreter/resolver_test.go
```diff
@@ -0,0 +1,272 @@
...
+	// containment tests. At present only top-level unknown variables are recommended.
+	res.Resolve(vars, newExprVarAttribute(1, "vars",
+		newExprPathElem(3, types.String("stop"))))
+	if resolvedAttrs[3] != "vars.stop" {
+		t.Errorf("Got %v, wanted 'vars.stop'", resolvedAttrs)
+	}
+	if resolvedStatus[3] != UnknownAttribute {
+		t.Errorf("Got %v, wanted UnknownAttribute", resolvedStatus)
+	}
+}
```
JimLarson, line 269: Add a test that a known-unknown "a" makes resolution of "a.b.c" unknown too.

Also test that even if the activation binds "a.b", listing it as unknown makes it unknown.  (Need this for creating holes in protos.)

interpreter/resolver.go
```diff
@@ -0,0 +1,564 @@
...
+	// When the resolver cannot find the Attribute in the activation it must return nil, false.
+	// The resolution of attributes within checked expressions is relatively simple, but for
+	// unchecked expressions, there may be many alternative Attribute representations to resolve
+	// among.
+	Resolve(Activation, Attribute) (interface{}, bool)
+
+	// ResolveRelative finds a relative Attribute from the input object and Activation.
+	//
+	// If the Attribute cannot be found, the return value must be a types.Err value.
+	ResolveRelative(interface{}, Activation, Attribute) interface{}
```
JimLarson, line 61: Add tests for ResolveRelative (which will help me figure out what this is supposed to do).

interpreter/resolver.go
```diff
@@ -0,0 +1,564 @@
...
+// data.
+//
+// For applications which use proto-based inputs or custom types with relatively static
+// environments are encouraged to hand-roll Resolver implementations suited to their strongly
+// typed objects.
+type Resolver interface {
+	// Resolve finds a top-level Attribute from the given Activation, returning the value
+	// if present.
+	//
+	// When the resolver cannot find the Attribute in the activation it must return nil, false.
```
JimLarson: When the resolver cannot find _the Attribute's variable_ in the activation, it returns nil, false.  If one of the Attribute's path elements cannot be found, returns `e, true` where `e` is a `types.Err` value.

interpreter/resolver.go
```diff
@@ -0,0 +1,564 @@
...
+		obj = res.getElem(obj, elem.ToValue(vars))
+	}
+	return obj
+}
+
+// getElem attempts to get the next 'elem' from the target 'obj' either by using native golang
+// method appropriate for the 'obj' type or by converting the obj to a ref.Val and using the CEL
+// functions.
+func (res *defaultResolver) getElem(obj interface{}, elem ref.Val) interface{} {
+	switch obj.(type) {
```
JimLarson: Add a case for the error type, to make it clear that it's safe to continue to follow the elements even after one has failed.

JimLarson (changes requested):
Dumping an initial set of comments.  I'll continue reading and commenting.

cel/cel_test.go
TristonianJones: This one was a bit odd to me as well, but the naming choice here is mostly for how the comment renders in godoc

interpreter/resolver.go
TristonianJones: I'll see what I can do.

interpreter/resolver.go
```diff
@@ -0,0 +1,564 @@
...
+
+	// NoSuchAttribute indicates that a top-level variable was found, but the referenced Attribute
+	// could not be resolved.
+	NoSuchAttribute ResolveStatus = 1 << iota
+
+	// UnknownAttribute indicates that the Attribute path matched an unknown Attribute path.
+	UnknownAttribute ResolveStatus = 1 << iota
+)
+
+// ResolveListener receives an Attribute and the status of the Resolve call for each Attribute
```
JimLarson, line 96: Either note that this is for testing only, or add a ResolveStatus to the return type for Resolve and ResolveRelative, so that "the status of the Resolve call" has a clear meaning.

cel/program.go
TristonianJones: This is documented on the interface proper.

interpreter/resolver.go
TristonianJones: I've cleaned up a tiny bit using an assignment in the switch, but I'm not sure that the actual key conversion checks are worth the hassle.

interpreter/attributes.go
TristonianJones: I prefer the name as-is. I'm not sure I read this as a 'unique occurrence of a variable' but rather as the variable referenced in the attribute, which may be the whole variable or some sub-property.

interpreter/attributes.go
TristonianJones: Not presently, but I've deleted the qualifying clause.

interpreter/resolver_test.go
TristonianJones: Done.

TristonianJones:
PTAL

interpreter/resolver.go
```diff
@@ -0,0 +1,564 @@
...
+		// TODO: make errors useful by including the expression id where the Error occurs, and then
+		// apply the same logic for getting the ancestor attribute for unknowns to errors.
+		res.listener(attr, NoSuchAttribute)
+		return v, found
+	}
+	res.listener(attr, FoundAttribute)
+	return v, found
+}
+
+// ancestorAttr attempts to determine Attribute or the nearest ancestor that matches the given
```
JimLarson, line 494: Clarify that "nearest" means "closest to the variable" or "with the shortest list of path elements".  If the ID is an expression ID, shouldn't there be only one match?

cel/program.go
```diff
@@ -108,6 +145,14 @@ func newProgram(e *env, ast Ast, opts ...ProgramOption) (Program, error) {
 		}
 	}
 
+	// Configure the resolver to support unknown attributes if the eval option is set.
+	resolver := interpreter.NewResolver(e.adapter)
+	if p.evalOpts&OptUnknownAttributes == OptUnknownAttributes {
```
JimLarson, line 150: Space around binary operators, I think.  Did you run gofmt?

cel/program.go
```diff
@@ -50,11 +50,70 @@ type EvalDetails interface {
...
+//
+// An Attribute is any sequence of a variable, field selection, or key / index lookup. Attributes
+// that exist within an expression are determined during the cel.Program call. An Unknown indicates
+// that some or all of the declared Attributes in a Program will not yet have a value. An Unknown
+// is composed of the following parts:
+//
+//   varName - the name of the top-level identifier in the program
+//   pathElems - optional list of scalar values matching the field and index lookup calls.
+//
+// The path element may only be of string, int, uint, and bool type. The special case '*' indicates
```
JimLarson, line 98: How is marking `a."*"` unknown any different from marking `a` as unknown?  Is it so you can put constraints on sub-paths, such as `a.*.7`?  If so make an explicit comment about that.

How do I distinguish a literal `'*'` from a wildcard `'*'` for a map key?

I assume that a string path element could be used either for a message field or map entry and that an int could be used for either a list element or map entry.  Is such polymorphism intended?  Should be documented regardless.

cel/program.go
```diff
@@ -50,11 +50,70 @@ type EvalDetails interface {
...
+// that the path element should match all inputs. If the path element is not of the correct type,
+// the call will result in an error.
+//
+// Note, any Attribute that overlaps with the Unknown attribute will be treated as Unknown. For
+// example, cel.Unknown("a", "b") will cause the Attributes 'a.b.c' and 'a' to be marked unknown.
```
JimLarson, line 103: I think you mean 'a.b.c' and 'a.b' will be marked unknown, and note that 'a.c' will not be unknown.

interpreter/activation.go
```diff
@@ -19,54 +19,49 @@ import (
...
 	"github.com/google/cel-go/common/types/ref"
 )
 
-// Activation used to resolve identifiers by name and references by id.
-//
-// An Activation is the primary mechanism by which a caller supplies input into a CEL program.
+// Activation describes the variables available to an evaluation scope.
 type Activation interface {
-	// ResolveName returns a value from the activation by qualified name, or false if the name
+	// Find returns a value from the activation by qualified name, or false if the name
```
JimLarson, line 27: ...returns _native_ value?  (To make it clear what the `interface{}` means.)

interpreter/activation.go
```diff
@@ -19,54 +19,49 @@ import (
...
+// A variable or some portion of its attributes may be treated as known-unknowns by CEL when the
+// FindUnknowns function returns a non-empty Attribute set.
+//
+// Note: PartialActivation must be used in conjunction with an interpreter.UnknownResolver to work.
+// Otherwise, the unknowns will be treated as errors.
+type PartialActivation interface {
+	Activation
+
+	// FindUnknowns returns the collection of unknown Attribute values associated with the
+	// variable name provided in the input, if present.
```
JimLarson, line 48: An empty slice adequately represents "not found", so what's the bool mean?

interpreter/activation.go
```diff
@@ -127,18 +150,34 @@ func (a *hierarchicalActivation) Parent() Activation {
...
 }
 
-// NewHierarchicalActivation takes two activations and produces a new one which prioritizes
-// resolution in the child first and parent(s) second.
-func NewHierarchicalActivation(parent Activation, child Activation) Activation {
-	return &hierarchicalActivation{parent, child}
+// partialActivation implements the PartialActivation interface.
+type partialActivation struct {
+	Activation
+	unknowns map[string][]Attribute
```
JimLarson, line 164: Comment that this maps variable names to attributes based on that variable.

interpreter/activation.go
```diff
@@ -19,54 +19,49 @@ import (
...
+// Activation describes the variables available to an evaluation scope.
 type Activation interface {
-	// ResolveName returns a value from the activation by qualified name, or false if the name
+	// Find returns a value from the activation by qualified name, or false if the name
 	// could not be found.
-	ResolveName(name string) (ref.Val, bool)
+	Find(name string) (interface{}, bool)
 
 	// Parent returns the parent of the current activation, may be nil.
 	// If non-nil, the parent will be searched during resolve calls.
```
JimLarson, line 32: If non-nil, the parent will be searched during Find calls?

interpreter/attributes.go
```diff
@@ -0,0 +1,188 @@
...
+
+	"github.com/google/cel-go/common/types"
+	"github.com/google/cel-go/common/types/ref"
+)
+
+// Attribute refers to a Variable and zero or more field or index traversals.
+type Attribute interface {
+	// Variable referenced by the attribute.
+	//
+	// When the Attribute is absolute, the Variable value is non-nil. When the Attribute is
```
JimLarson, line 28: What's an "absolute" attribute?  It sounds like "relative" attributes require you to expand your definition on line 24.

interpreter/attributes.go
```diff
@@ -0,0 +1,188 @@
...
+	Variable() Variable
+
+	// Path of field traversals against reference object, either a variable or a contextual
+	// intermediate value computed during CEL evaluation.
+	//
+	// If Variable is nil, the Path must contain at least one PathElem.
+	Path() []*PathElem
+
+	// Select a path element from the current Attribute to produce a more fully qualified
+	// Attribute.
```
JimLarson, line 40: A _less_ qualified attribute?  It sounds like this lets you go from "a.b.c" to the less-qualified "a.b".  Consider selecting by int index rather than by pointer.

interpreter/attributes.go
```diff
@@ -0,0 +1,188 @@
...
+	// ToValue function reference permits flexible resolution of the cross-reference.
+	ToValue func(Activation) ref.Val
+}
+
+// NewUnknownAttribute creates an Attribute that refers to a top-level variable or one of its
+// descendants fields in a selection path.
+//
+// A selection path element may be of the following scalar types:
+//
+// - bool, types.Bool
```
JimLarson, line 77: ...meaning a map key.

interpreter/attributes.go
```diff
@@ -0,0 +1,188 @@
...
+// - int, int32, int64, types.Int
```
JimLarson, line 78: ...meaning a map key or list index.

interpreter/attributes.go
```diff
@@ -0,0 +1,188 @@
...
+// - string, types.String
```
JimLarson, line 79: ...meaning a map key or message field.

interpreter/attributes.go
```diff
@@ -0,0 +1,188 @@
...
+// - uint, uint32, uint64, types.Uint
```
JimLarson, line 80: ...meaning a map key.

interpreter/attributes.go
```diff
@@ -0,0 +1,188 @@
...
+//
+// The string path element '*' is treated as a wildcard match against any path element value.
+//
+// As absolute Attribute references within an expression are being resolved, the attribute path
+// is compared to the set of declared unknowns which refer to the same top-level variable. When
+// the unknown Attribute matches any part of the resolved attribute, a types.Unknown value is
+// generated. For example, when a top-level variable contains an unknown attribute, then it too
+// becomes unknown. Otherwise, the use of this value in comparisons might yield invalid results.
```
JimLarson, line 88: So marking "a.b.c" unknown makes "a" unknown?  Then is there ever any difference between declaring "a.b.c" to be unknown vs just declaring "a" to be unknown?

cel/cel_test.go
```diff
@@ -91,6 +91,70 @@ func Example() {
...
+		"you": func() ref.Val { return types.String("world") },
+	})
+	if err != nil {
+		log.Fatalf("runtime error: %s\n", err)
+	}
+
+	fmt.Println(out)
+	// Output:CEL and world are shaking hands.
+}
+
```
JimLarson: We need some tests or examples in cel/ to show how to use the new interface for specifying unknowns.

JimLarson (changes requested):

interpreter/resolver.go
TristonianJones: The `ResolveStatus` is for listening to resolve events during the course of execution. I can remove this feature for now as it's not needed yet and probably needs work

interpreter/resolver.go
TristonianJones: I've removed this code for now.

cel/program.go
TristonianJones: Suprisingly, `gofmt` doesn't seem to mind. I've added the spaces though.

TristonianJones:
I'm going to archive this PR since the code is being re-implemented.

Issue #20, opened 2018-04-26
Tool to stringify ASTs


TristonianJones:
## Issue Filing Checklist

- [X] There are no issues that match the desired change
- [X] The change is large enough it can't be addressed with a simple
     [Pull Request](./PULL_REQUEST_TEMPLATE.md) 

## Change
Given an AST generate a human-readable string from it.

### Example
Translate an AST of the form `call('_||_', ident(a), ident(b))`
into the human readable string `a || b`.

The translation step need not honor line-breaks or parentheses
present in the original expression, but must capture the semantic
meaning of the original.

### Alternatives Considered
This functionality does not yet exist. An alternative might be to
record the original string from which the AST was generated, but
this would likely be overkill and serving both the string and AST
form of an expression in production is undesirable.

PR #225, opened 2019-05-31
Expression to string utility.

TristonianJones:
Utility to convert `cel.Ast` back to a string per #20.

Note, comprehension macros are currently not supported and will likely require a change
to `SourceInfo` to record macro expansion replacements.


parser/unparser.go
```diff
@@ -0,0 +1,412 @@
...
+		un.str.WriteString(")")
+	}
+	return nil
+}
+
+func (un *unparser) pos(id int64) int32 {
+	return un.info.GetPositions()[id]
+}
+
+func (un *unparser) pad(id int64) {
```
madi8229, line 401: Can you give a comment for this function?

madi8229 (approved):

PR #226, opened 2019-06-04
Fixed numeric offset timezones

### LINKED ISSUE #191: Support numeric offset timezones
**Describe the bug**
The spec describes an option for timezones to be of the form (+|-)?dd:dd, but this implementation doesn't support them.

**To Reproduce**

- [ ]  parser
- [ ]  checker
- [X]  interpreter

```cel
timestamp('2009-02-13T23:31:30Z').getDayOfMonth('+11:00')
```

Should return 13.

**Expected behavior**
The timezone should be the hours/minutes offset from UTC.

victoriajyang:
Fixed numeric offset timezones and added new tests in buildfile.
Closes #191

### DIFF
--- a/common/types/timestamp.go
+++ b/common/types/timestamp.go
@@ -274,8 +274,6 @@ func timestampGetMillisecondsWithTz(t time.Time, tz ref.Val) ref.Val {
 
 func timeZone(tz ref.Val, visitor timestampVisitor) timestampVisitor {
 	return func(t time.Time) ref.Val {
-		var err error
-		var offset int
 		if StringType != tz.Type() {
 			return ValOrErr(tz, "no such overload")
 		}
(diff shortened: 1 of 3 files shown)


common/types/timestamp.go
```diff
@@ -272,12 +274,33 @@ func timestampGetMillisecondsWithTz(t time.Time, tz ref.Val) ref.Val {
 
 func timeZone(tz ref.Val, visitor timestampVisitor) timestampVisitor {
 	return func(t time.Time) ref.Val {
+		var err error
+		var offset int
```
JimLarson: Don't declare these here.  See below for specifics.

common/types/timestamp.go
```diff
@@ -272,12 +274,33 @@ func timestampGetMillisecondsWithTz(t time.Time, tz ref.Val) ref.Val {
...
 			return ValOrErr(tz, "no such overload")
 		}
-		loc, err := time.LoadLocation(string(tz.(String)))
-		if err == nil {
-			return visitor(t.In(loc))
+		val := string(tz.(String))
+		ind := strings.Index(val, ":")
+		if ind == -1 {
+			loc, err := time.LoadLocation(val)
+			if err == nil {
```
JimLarson: Negate the test and put a "return &Err{err}" inside the "if", and have the happy path "return visitor..." after the "if".  Then, since all paths here end in a return, you can eliminate the "else" and outdent its code.

common/types/timestamp.go
```diff
@@ -272,12 +274,33 @@ func timestampGetMillisecondsWithTz(t time.Time, tz ref.Val) ref.Val {
...
+				return visitor(t.In(loc))
+			}
+		} else {
+			hr, err := strconv.Atoi(string(val[0:ind]))
+			min, err2 := strconv.Atoi(string(val[ind+1]))
+			if err == nil && err2 == nil {
```
JimLarson: Little bug here - if err == nil but err2 is non-nil, we'll end up returning an Err with an empty reason.

Instead, test err right after it's set and split out the error case, then instead of err2, just reuse err for the "min" parsing and also check for errors there.

common/types/timestamp.go
```diff
@@ -272,12 +274,33 @@ func timestampGetMillisecondsWithTz(t time.Time, tz ref.Val) ref.Val {
...
-		if err == nil {
-			return visitor(t.In(loc))
+		val := string(tz.(String))
+		ind := strings.Index(val, ":")
+		if ind == -1 {
+			loc, err := time.LoadLocation(val)
+			if err == nil {
+				return visitor(t.In(loc))
+			}
+		} else {
```
JimLarson, line 304: Start with a comment describing what you're parsing here.

common/types/timestamp.go
```diff
@@ -272,12 +274,33 @@ func timestampGetMillisecondsWithTz(t time.Time, tz ref.Val) ref.Val {
...
+			hr, err := strconv.Atoi(string(val[0:ind]))
+			min, err2 := strconv.Atoi(string(val[ind+1]))
+			if err == nil && err2 == nil {
+				if string(val[0]) == "-" {
+					offset = hr*60 - min
+				} else {
+					offset = hr*60 + min
+				}
+				secondsEastOfUTC := int((time.Duration(offset) * time.Minute).Seconds())
+				timezone := time.FixedZone("Time Zone", secondsEastOfUTC)
```
JimLarson: We don't have a meaningful name here, so just use an empty string.

common/types/timestamp.go
```diff
@@ -272,12 +274,33 @@ func timestampGetMillisecondsWithTz(t time.Time, tz ref.Val) ref.Val {
...
+				curr_time := time.Date(t.Year(), t.Month(), t.Day(), t.Hour(), t.Minute(), t.Second(), t.Nanosecond(), timezone)
```
JimLarson: The timezone is a valid location - any reason not to just use visitor(t.In(timezone))?

JimLarson (changes requested):

JimLarson (approved):

PR #228, opened 2019-06-13
Enforce reserved identifiers.

JimLarson:


parser/helper.go
```diff
@@ -79,10 +80,15 @@ func (p *parserHelper) newLiteralDouble(ctx interface{}, value float64) *exprpb.
 		&exprpb.Constant{ConstantKind: &exprpb.Constant_DoubleValue{DoubleValue: value}})
 }
 
-func (p *parserHelper) newIdent(ctx interface{}, name string) *exprpb.Expr {
+func (p *parserHelper) newIdent(ctx interface{}, name string) (*exprpb.Expr, error) {
+	for _, r := range reservedIds {
+		if r == name {
```
TristonianJones: Since it's possible to provide a leading `.` to designate an identifier as global, consider trimming the leading dot from the input to this check.

parser/helper.go
```diff
@@ -79,10 +80,15 @@ func (p *parserHelper) newLiteralDouble(ctx interface{}, value float64) *exprpb.
...
+			return nil, fmt.Errorf("reserved identifier: %s", name)
```
TristonianJones: Since all calls eventually resolve to `newIdent()` within the parser helper, have you considered reporting the error here rather than returning it? This would mean that you would need to wait until after the `newExpr` call in order to ensure you have an expression id that can be mapped to a source position.

TristonianJones (changes requested):

parser/macro.go
```diff
@@ -205,8 +205,8 @@ type ExprHelper interface {
 		step *exprpb.Expr,
 		result *exprpb.Expr) *exprpb.Expr
 
-	// Ident creates an identifier Expr value.
-	Ident(name string) *exprpb.Expr
+	// Ident creates an identifier Expr value or error for e.g. reserved identifiers.
+	Ident(name string) (*exprpb.Expr, error)
```
TristonianJones: I think it would be preferable to keep the error an implementation detail in this case. Report it internally so that it shows up in the issues list, but don't return it to the caller. Kind of like what you're doing below in `GlobalCall`

TristonianJones: It looks like I was thinking of the `parser.VisitIdentOrGlobalCall` function which can report errors and _should_ be called before ever reaching the `helper.newIdent` function. Moving the checking up before getting into macro territory should be sufficient, I believe.

parser/helper.go
JimLarson: Mooted by move to visitor.

parser/helper.go
JimLarson: Mooted by move to visitor.

parser/macro.go
JimLarson: Mooted by move to visitor.

JimLarson: Yes, things are much cleaner just doing the check in VisitIdentOrGlobalCall().

TristonianJones (approved):

PR #231, opened 2019-07-09
fix: concurrent fields info access

### LINKED ISSUE #230: Panic during type checking phase in `pb.(*TypeDescription).getFieldsInfo`
**Describe the bug**
Type checking expression on proto type concurrently is causing panic (concurrent map writes in `cel-go/common/types/pb.(*TypeDescription).getFieldsInfo`)

**To Reproduce**

- [ ]  parser
- [x]  checker
- [ ]  interpreter

I have simple engine type which initialize env and provides ParseAndCheck function.
```go
type engine struct {
	env cel.Env
}

func New() (*engine, error) {
	e, err := cel.NewEnv(
		cel.Container("google.api.expr.v1alpha1"),
		cel.Types(&exprpb.Expr{}),
		cel.Declarations(
			decls.NewIdent("expr",
				decls.NewObjectType("google.api.expr.v1alpha1.Expr"), nil),
		),
	)
	if err != nil {
		return nil, err
	}
	return &engine{
		env: e,
	}, nil
}

func (e *engine) ParseAndCheck(expr string) (cel.Ast, error) {
	p, iss := e.env.Parse(expr)
		return nil, iss.Err()
	}
	c, iss := e.env.Check(p)
		return nil, iss.Err()
	}
	return c, nil
}
```
Test setup:
```go
func TestParseAndCheck(t *testing.T) {
	engine, err := New()
	if err != nil {
		t.Fatal(err)
	}
	const concurrency = 2
	wgDone := sync.WaitGroup{}
	wgDone.Add(concurrency)

	for i := 0; i < concurrency; i++ {
		go func() {
			defer wgDone.Done()
			expr := fmt.Sprintf("expr.id")
			_, errParse := engine.ParseAndCheck(expr)
			if errParse != nil {
				t.Fatalf("failed to process engine: %v", errParse)
			}
		}()
	}
	wgDone.Wait()
}

```
It can be also observed running above test case with `race` flag.

**Expected behavior**
Run without panic.

krzysztofdrys:
This change makes checker safe for concurrent usage. TypeDescription.getFieldsInfo can be called by multiple goroutines. As it performs lazy initialisation of fields info without any synchronisation mechanizm, concurrent calls will result in concurrent writes to fields map, which will result in panic.

This change makes TypeDescription.getFieldsInfo safe for concurrent usage.

Fixes https://github.com/google/cel-go/issues/230


krzysztofdrys:
I signed CLA and added some tests.

TristonianJones (approved):

TristonianJones:
Thanks for the fix!

krzysztofdrys:
The please is all mine. And thanks for handling this so quickly.

What is your release plan? What are the chances that this fix will get into some release any time soon?

PR #234, opened 2019-07-15
Create Google Cloud Build workflow

victoriajyang:
Add cloudbuild.yaml for GCB trigger and start.sh & zip.sh to convert bazel files into required Google Cloud Storage output for TestGrid


WORKSPACE
```diff
@@ -36,15 +36,15 @@ go_repository(
 
 git_repository(
   name = "com_google_cel_spec",
-  commit = "4df43cd7c60b816eb29301d2f93dbf336808799c", # PR #67
+  commit = "abdfa13f6ccf217f71f8ef677718a965f8af6c47", #PR #97
```
JimLarson: This should be PR 75.

conformance/BUILD.bazel
```diff
@@ -1,9 +1,10 @@
 sh_test(
     name = "ct",
-    srcs = ["@com_google_cel_spec//tests:conftest.sh"],
+    srcs = ["@com_google_cel_spec//tests:conftest-nofail.sh"],
```
JimLarson, line 39: Don't reuse the "ct" target for the dashboard.  We want the "ct" target to fail when a conformance test fails so we can guarantee that new changes don't break the select list of test files in this target.

Create a separate "ct_dashboard" target that a) uses the -nofail.sh script, and b) includes *all* of the simple textproto files, even those that are broken.  This way the dashboard shows the broken-ness of tests, while the "ct" target lets the developer gradually take on tests that they want to keep functional.

It would be nice if we could make the dashboard target "hidden" so that it doesn't get run when you do "bazel test ..." but will get run when you mention it explicitly.  But I can't find a bazel feature to do that.  We could abuse the test size description for this, but that seems inappropriate.

conformance/zip.sh
```diff
@@ -0,0 +1,67 @@
+#!/bin/bash
+mkdir -p artifacts
+touch artifacts/junit_01.xml
+
+input="./bazel-out/k8-fastbuild/testlogs/conformance/ct/test.xml"
```
JimLarson: Add a comment describing what we're expecting to be at the top of the test.xml file.  Note that XML can be substantially reformatted without a change to its meaning, so relying on newlines being in a particular place is an extra assumption - leave plenty of hints for future debuggers.

JimLarson (changes requested):

conformance/BUILD.bazel
```diff
@@ -1,5 +1,41 @@
...
+        "@com_google_cel_spec//tests/simple:testdata/integer_math.textproto",
+        "@com_google_cel_spec//tests/simple:testdata/lists.textproto",
+        "@com_google_cel_spec//tests/simple:testdata/logic.textproto",
+        "@com_google_cel_spec//tests/simple:testdata/macros.textproto",
+        "@com_google_cel_spec//tests/simple:testdata/string.textproto",
+        "@com_google_cel_spec//tests/simple:testdata/timestamps.textproto",
+    ],
+)
+
+# ct_dashboard is a target for the conformance dashboard and includes all simple textproto files, including those that are broken.
```
JimLarson, line 36: So you can add broken.textproto here.

conformance/zip.sh
```diff
@@ -2,9 +2,9 @@
 mkdir -p artifacts
 touch artifacts/junit_01.xml
 
-input="./bazel-out/k8-fastbuild/testlogs/conformance/ct/test.xml"
+input="./bazel-out/k8-fastbuild/testlogs/conformance/ct_dashboard/test.xml"
 
-head -2 $input > artifacts/junit_01.xml
+head -2 $input > artifacts/junit_01.xml # This appends the XML header and <testsuites> to junit_01.xml
```
JimLarson: Consider just writing static strings for this, if you don't expect them to change.

Here and below, put the comment on its own line above what it's describing.

conformance/zip.sh
```diff
@@ -18,15 +18,15 @@ do
...
-    status=$(echo $line | cut -c4-8)
-    name=$(echo $line | tail -c +11 | head -c -9)
+    status=$(echo $line | cut -c4-8) # The first four characters after --- are the pass/fail status of the test)
+    name=$(echo $line | tail -c +11 | head -c -9) # The next string excluding the time is the name of the test
     echo "$testcase$name$end" >> artifacts/junit_01.xml
     if [ $status = "FAIL" ]
     then
       read line1
-      message=$(echo "$line1" | sed 's/</ /g; s/>/ /g; s/"//g')
+      message=$(echo "$line1" | sed 's/</ /g; s/>/ /g; s/"//g') # $message is the failure message excluding <, >, ", which cause the xml file to fail
```
JimLarson: Can you change them to `&lt;` `&gt;` and `&quot;` respectively?

conformance/zip.sh
```diff
@@ -43,9 +43,9 @@ timestamp='"timestamp": '
 time_string="$timestamp$startdate$comma"
 
 result='"result": "'
-test_string=$(tail -n 5 $input)
+test_string=$(tail -n 5 $input)
```
JimLarson: Remove the whitespace at the end of the line.

JimLarson (changes requested):

conformance/zip.sh
```diff
@@ -4,7 +4,8 @@ touch artifacts/junit_01.xml
 
 input="./bazel-out/k8-fastbuild/testlogs/conformance/ct_dashboard/test.xml"
 
-head -2 $input > artifacts/junit_01.xml # This appends the XML header and <testsuites> to junit_01.xml
+echo "<?xml version="1.0" encoding="UTF-8"?>" > artifacts/junit_01.xml
+echo "<testsuites" >> artifacts/junit_01.xml
```
JimLarson: Missing &gt;.

conformance/zip.sh
```diff
@@ -18,15 +19,16 @@ do
...
+  if [ $testline = "---" ] && [ $extendline != '----' ]
   then
     status=$(echo $line | cut -c4-8) # The first four characters after --- are the pass/fail status of the test)
     name=$(echo $line | tail -c +11 | head -c -9) # The next string excluding the time is the name of the test
     echo "$testcase$name$end" >> artifacts/junit_01.xml
     if [ $status = "FAIL" ]
     then
       read line1
-      message=$(echo "$line1" | sed 's/</ /g; s/>/ /g; s/"//g') # $message is the failure message excluding <, >, ", which cause the xml file to fail
+      message=$(echo "$line1" | sed 's/&lt;/ /g; s/&gt;/ /g; s/&quot;/g') # $message is the failure message excluding <, >, ", which cause the xml file to fail
```
JimLarson: What I meant was, change `<` to `&lt;`, etc., so it's displayed correctly.  Change the comment to something like "Quote the characters that XML might interpret.".  Consider quoting `&` as `&amp;` too.  Check an XML reference to see if this is sufficient.

JimLarson (changes requested):

JimLarson (approved):

Issue #227, opened 2019-06-12
Improper conversion of byte literals to byte sequences


phith0n:
**Describe the bug**

When the expression returns a bytes array which contains no UTF-8 character, the result is unexpected.

**To Reproduce**

Check which components this affects:

- [ ]  parser
- [ ]  checker
- [x]  interpreter

Sample expression and input that reproduces the issue:

```cel
b'\x8B'
```

Test setup:

```go
package main

import (
	"github.com/google/cel-go/cel"
	"fmt"
)

func main() {
	env, _ := cel.NewEnv()
	parsed, _ := env.Parse(`b'\x8B'`)
	checked, _ := env.Check(parsed)
	prg, _ := env.Program(checked)
	out, _, _ := prg.Eval(map[string]interface{}{})
	fmt.Println(out)
}
```

**Expected behavior**

Output a byte array, which length is 1, the only character is [139].

**Actual behavior**

Output a byte array, which length is 2, value is [194, 139]

TristonianJones:
Thanks for the report. This is indeed an inconsistency with the spec.

TristonianJones:
@JimLarson would you be able to give this a look while I'm OOO next week? It's an issue with the `unescape.go` converting to `rune` type straight-away and not considering whether the byte sequence coming in is for a byte literal. There's actually a TODO that questions whether the right encoding is being performed.

JimLarson:
Should work now.

PR #235, opened 2019-07-18
Hex and octal escapes for bytes literals.

JimLarson:
Fixes issue #227.

### DIFF
--- a/parser/unescape.go
+++ b/parser/unescape.go
@@ -156,8 +156,16 @@ func unescapeChar(s string, is_bytes bool) (value rune, encode bool, tail string
 			encode = !is_bytes
 		case 'u':
 			n = 4
+			if is_bytes {
+				err = fmt.Errorf("unable to unescape string")
+				return
+			}
 		case 'U':
 			n = 8
+			if is_bytes {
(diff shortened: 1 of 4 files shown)


parser/unescape.go
```diff
@@ -146,9 +149,11 @@ func unescapeChar(s string) (value rune, multibyte bool, tail string, err error)
 	// 4. Unicode escape sequences, reproduced from `strconv/quote.go`
 	case 'x', 'X', 'u', 'U':
 		n := 0
+		encode = true
 		switch c {
 		case 'x', 'X':
 			n = 2
+			encode = !is_bytes
 		case 'u':
```
TristonianJones, line 157: Looks like Unicode encodings aren't supported for byte literals in Google SQL. 
https://cloud.google.com/bigquery/docs/reference/standard-sql/lexical#literals

TristonianJones (changes requested):

parser/unescape.go
JimLarson: Fixed.

TristonianJones (approved):
I ran the test cases against Google SQL and they all match. Nice work!

PR #236, opened 2019-07-23
Move the Example Global Overload to a common location.

TristonianJones:
Minor refactor to move an example into the `cel_test.go` file.


gabrielsz (approved):
LGTM. Ok with the removal based on our offline conversation.

PR #237, opened 2019-08-01
Type param resolution fix

TristonianJones:
When functions are declared with type-params, the checker should ensure the following:

- The types associated with the same type param name are assignable to each other
- The most general type is preferred as the mapping for the type param variable.

For example, the add operation:
```
_+_(A, A) -> A
```
When given the following input:
```
['1', 2] + [true] 
```
Should produce the following result type: `list(dyn)`

This change fixes an issue where the order of the input arguments would result in different result types for the function.

### DIFF
--- a/checker/checker.go
+++ b/checker/checker.go
@@ -289,8 +289,12 @@ func (c *checker) resolveOverload(
 		if len(overload.TypeParams) > 0 {
 			// Instantiate overload's type with fresh type variables.
 			substitutions := newMapping()
+			uniqueParams := map[string]bool{}
 			for _, typePar := range overload.TypeParams {
-				substitutions.add(decls.NewTypeParamType(typePar), c.newTypeVar())
+				if _, found := uniqueParams[typePar]; !found {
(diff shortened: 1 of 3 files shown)


checker/checker.go
```diff
@@ -289,8 +289,15 @@ func (c *checker) resolveOverload(
 		if len(overload.TypeParams) > 0 {
 			// Instantiate overload's type with fresh type variables.
 			substitutions := newMapping()
+			// When a type parameter within an argument bears the same name as another
+			// type parameter in the list it should not be added as an alternative
+			// variable subsitution for the overload result type.
```
JimLarson: I'm confused about how this could happen.  For instance, even for `+: (list<A>, list<A>) --> A`, the type param list is just "A".  Please add to comment explaining how we can wind up in this situation.

TristonianJones: You're right, this was totally unnecessary.

JimLarson (approved):

PR #239, opened 2019-08-01
Update duration/timestamp to reflect CEL documentation

victoriajyang:
timestamp.GetDayOfYear should be 0-based, not 1-based, and string(duration) should return the duration in seconds, not in h/m/s as previously thought.

Depends on google/cel-spec #81

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -36,7 +36,7 @@ go_repository(
 
 git_repository(
   name = "com_google_cel_spec",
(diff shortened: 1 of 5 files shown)


TristonianJones (approved):

TristonianJones:
You'll need to update the WORKSPACE to point to the latest version of https://github.com/google/cel-spec, but then all tests should pass.

victoriajyang:
> You'll need to update the WORKSPACE to point to the latest version of https://github.com/google/cel-spec, but then all tests should pass.

Okay, it's passing now.

PR #240, opened 2019-08-01
Add comments explaining cloudbuild.yaml

victoriajyang:
Add comments to cloudbuild.yaml explaining what is occurring in each step.

### DIFF
--- a/cloudbuild.yaml
+++ b/cloudbuild.yaml
@@ -1,16 +1,16 @@
 steps:
         - name: 'ubuntu'
-          args: ['bash', 'conformance/start.sh', '$COMMIT_SHA']
+          args: ['bash', 'conformance/start.sh', '$COMMIT_SHA'] # writes start time and #PR to started.json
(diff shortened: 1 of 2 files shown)


cloudbuild.yaml
```diff
@@ -1,16 +1,16 @@
 steps:
         - name: 'ubuntu'
-          args: ['bash', 'conformance/start.sh', '$COMMIT_SHA']
+          args: ['bash', 'conformance/start.sh', '$COMMIT_SHA'] # writes start time and #PR to started.json
```
TristonianJones: This is a personal nit, but would you mind moving the end of line comment above the `args` in each case. I find it easier to follow the flow that way.

cloudbuild.yaml
```diff
@@ -1,16 +1,16 @@
 steps:
```
TristonianJones, line 2: Could you give a rough overview of what the build is doing at the top of the file. The line comments indicate how it's doing things, but don't necessarily give the bigger picture.

TristonianJones (changes requested):

TristonianJones (approved):

PR #241, opened 2019-08-02
Wrapper type support

### LINKED ISSUE #205: Support for protobuf wrapper types
**Describe the bug**
The well-known protobuf wrapper types, e.g. `google.protobuf.StringValue` are supported by the type-checker, but not yet supported by the interpreter. This is a feature gap with the current CEL spec. 

**To Reproduce**

- [ ]  parser
- [ ]  checker
- [X]  interpreter

```cel
wrapped_string == "hello" || wrapped_string == null
```

Test setup:
```go
func TestInterpreter_WrapperTypes(t *testing.T) {
  e, _ := cel.NewEnv(
    cel.Declarations(
      decls.NewIdent('wrapped_string', decls.NewWrapperType(decls.String), nil),
    ))
  ast, _ := e.Parse(`wrapped_string == "hello" || wrapped_string == null`)
  prg, _ := e.Program(ast)
  out, _, _ := prg.Eval(map[string]interface{}{'wrapped_string': &pb.StringValue{value: 'hello'}}
  if out != types.True {
      t.Errorf("Got '%v' expected true", out)
  }
  out, _, _ = prg.Eval(map[string]interface{}{'wrapped_string': &pb.StringValue{}}
  if out != types.True {
      t.Errorf("Got '%v' expected true", out)
  }
}
```

**Expected behavior**
The wrapped_string should unwrap to support comparisons to `string` or `null`. 

**Additional context**
Well-known types are unwrapped to their base primitive type at runtime. This causes will result in some surprises though as this could lead to `no such overload` errors since a comparison between a `null` and a `string` is an error. 

It is possible that this is just an issue of implementation. However, if the types were not-unwrapped prior to calling functions the overhead of implementing all of the overload permutations would be quite high. The best fix for this surprise in user expectations would to to implement heterogenous (in)equality: https://github.com/google/cel-spec/issues/54

TristonianJones:
Wrapper type support within the checker and interpreter.

Wrapper types unlike their primitive counter parts may be `null` and support presence testing.

The code also refactors some of the underlying proto type descriptions for ease of use and clarity.

Closes #205


JimLarson (approved):
LGTM.  It's odd that the generated file test_all_types.pb.go is checked in, but it was checked in before, so beyond the scope of a review of this PR.  Please check that you've got tests for the null case of wrapper-type fields.

TristonianJones:
Awesome, thanks. There's a `null` test in the second `interpreter_test.go` case that exercises both the `has` and `null` impacts on wrappers. 

With respect to `test_all_types.pb.go`, the generated code is checked in so that standard go tools can be used to build and run the tests. It feels weird to check in generated code, but it ends up being pretty common practice within go.

Issue #84, opened 2018-10-10
make consistent with the CEL semantics


madi8229:
In parse/macro.go: 167
make consistent with the CEL semantics.

PR #244, opened 2019-08-08
Update conformance BUILD with skip_test flag

victoriajyang:
With the skip_test flag the specific failing tests can be skipped instead of skipping complete files

Depends on cel-spec PR #84


TristonianJones (approved):

TristonianJones (approved):
Looks like a test is failing, but otherwise LGTM

PR #246, opened 2019-08-15
Fix for planner panic.

### LINKED ISSUE #245: Planner doesn't pass TypeProvider to evalIdent
**Describe the bug**
planner.go doesn't pass the type provider to `evalIdent` https://github.com/google/cel-go/blob/v0.2.0/interpreter/planner.go#L195 causing a panic in https://github.com/google/cel-go/blob/v0.2.0/interpreter/planner.go#L621 when using a custom type provider and `NewInterpretable` instead of `NewUncheckedInterpretable`

**To Reproduce**

- [ ]  parser
- [ ]  checker
- [x]  interpreter

Can invest time in extracting a reproducible code sample if this is considered an actual issue. 

**Expected behavior**
Provider is used to resolve the type instead of a panic.

TristonianJones:
Fixed a missing `provider` value within the `evalIdent` on the checked expression paths.

Closes #245

### DIFF
--- a/interpreter/interpreter_test.go
+++ b/interpreter/interpreter_test.go
@@ -828,7 +828,7 @@ func TestInterpreter_SetProto2PrimitiveFields(t *testing.T) {
 	}
 }
 
-func TestInterpreter_CustomTypeProvider(t *testing.T) {
+func TestInterpreter_MissingIdentInSelect(t *testing.T) {
 	src := common.NewTextSource(`a.b.c`)
 	parsed, errors := parser.Parse(src)
 	if len(errors.GetErrors()) != 0 {
@@ -847,7 +847,7 @@ func TestInterpreter_CustomTypeProvider(t *testing.T) {
 	i, _ := interp.NewInterpretable(checked)
 	vars := EmptyActivation()
 	result := i.Eval(vars)
-	if !types.IsUnknownOrError(result) {
+	if !types.IsUnknown(result) {
 		t.Errorf("Got %v, wanted unknown", result)
 	}
 }
--- a/interpreter/interpreter_test.go
+++ b/interpreter/interpreter_test.go
@@ -828,6 +828,30 @@ func TestInterpreter_SetProto2PrimitiveFields(t *testing.T) {
 	}
 }
 
+func TestInterpreter_CustomTypeProvider(t *testing.T) {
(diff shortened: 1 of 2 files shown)


interpreter/interpreter_test.go
```diff
@@ -828,6 +828,30 @@ func TestInterpreter_SetProto2PrimitiveFields(t *testing.T) {
...
+	if len(errors.GetErrors()) != 0 {
+		t.Fatalf(errors.ToDisplayString())
+	}
+
+	interp := NewStandardInterpreter(packages.NewPackage("test"), reg, reg)
+	i, _ := interp.NewInterpretable(checked)
+	vars := EmptyActivation()
+	result := i.Eval(vars)
+	if !types.IsUnknownOrError(result) {
+		t.Errorf("Got %v, wanted unknown", result)
```
JimLarson, line 852: We actually want an error, right?

JimLarson (changes requested):

TristonianJones:
The current behavior will be unknown, but in the future it'll be an error.
I'll change the test name to reflect the issue being exercised

interpreter/interpreter_test.go
```diff
@@ -847,7 +847,7 @@ func TestInterpreter_CustomTypeProvider(t *testing.T) {
 	i, _ := interp.NewInterpretable(checked)
 	vars := EmptyActivation()
 	result := i.Eval(vars)
-	if !types.IsUnknownOrError(result) {
+	if !types.IsUnknown(result) {
```
JimLarson, line 851: Since we want the result to be an error in the future, I was suggesting leaving the test as IsUnknownOrError but changing the message to "wanted error" or "wanted unknown or error".

JimLarson (changes requested):

interpreter/interpreter_test.go
TristonianJones: I've updated the test condition with a TODO since it reflects the current state of the code and will have to be updated in order to address Issue #190.

JimLarson: I would have been happy with the unknown-or-error check, with a note to clean up after the fix, but this is okay for now.

JimLarson (approved):

PR #247, opened 2019-08-15
Change GCS folder to be full date path; add badge

victoriajyang:
UNIX times are hard to read, so folder name is changed to be more human readable, add TestGrid badge

This widget is managed by TestGrid, there is a way to configure our own but I think it would require too much maintenance just to change a name (https://github.com/badges/shields/blob/master/doc/TUTORIAL.md)


JimLarson (approved):

TristonianJones:
Is there any way to adjust the test badge? I like that the conformance dashboard is linked, but the failing status says 'go away'

victoriajyang:
The test badge is pulling directly from TestGrid, and the only statuses it has are healthy, flaky, or failing, so short of creating our own badge I think we can only get those statuses.

TristonianJones:
Would you mind providing a link instead? I don't think we should advertise failing since that sends the wrong message.

victoriajyang:
Okay I'll make another PR

PR #251, opened 2019-08-23
Logical AND / OR fix with error input.

TristonianJones:
The google/cel-spec defines the logical AND / OR operators as commutative and capable
of absorbing errors. The interpreters do not handle this case correctly and will in some
scenarios produce a faulty result. In the majority of cases the bug is a non-issue, but is out
of conformance with the spec.

Once this test is submitted a conformance test will be written to exercise this case and a point
release created.

### DIFF
--- a/interpreter/interpretable.go
+++ b/interpreter/interpretable.go
@@ -177,8 +177,11 @@ func (or *evalOr) Eval(ctx Activation) ref.Val {
 	if types.IsUnknown(rVal) {
 		return rVal
 	}
-	// if the left-hand side is non-boolean return it as the error.
-	return types.ValOrErr(lVal, "no such overload")
+	// If the left-hand side is non-boolean return it as the error.
+	if types.IsError(lVal) {
+		return types.ValOrErr(lVal, "no such overload")
+	}
+	return types.ValOrErr(rVal, "no such overload")
 }
 
 type evalAnd struct {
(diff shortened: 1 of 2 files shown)


interpreter/interpretable.go
```diff
@@ -177,8 +177,11 @@ func (or *evalOr) Eval(ctx Activation) ref.Val {
 	if types.IsUnknown(rVal) {
 		return rVal
 	}
-	// if the left-hand side is non-boolean return it as the error.
-	return types.ValOrErr(lVal, "no such overload")
+	// If the left-hand side is non-boolean return it as the error.
+	if types.IsError(lVal) {
+		return types.ValOrErr(lVal, "no such overload")
```
JimLarson: Just return lVal here, since it's known to be an error.

interpreter/interpretable.go
```diff
@@ -218,8 +221,11 @@ func (and *evalAnd) Eval(ctx Activation) ref.Val {
...
```
JimLarson: Just return lVal here, since it's known to be an error.

interpreter/interpretable.go
```diff
@@ -632,7 +638,11 @@ func (or *evalExhaustiveOr) Eval(ctx Activation) ref.Val {
 	if types.IsUnknown(rVal) {
 		return rVal
 	}
-	return types.ValOrErr(lVal, "no such overload")
+	// If the left-hand side is non-boolean return it as the error.
+	if types.IsError(lVal) {
+		return types.ValOrErr(lVal, "no such overload")
```
JimLarson: Just return lVal.

interpreter/interpretable.go
```diff
@@ -668,7 +678,11 @@ func (and *evalExhaustiveAnd) Eval(ctx Activation) ref.Val {
...
```
JimLarson: Just return lVal.

JimLarson (approved):

TristonianJones:
Correction, the right type of response was being given, but the wrong error might have been returned before the change.

PR #260, opened 2019-09-16
Add missing lexer rule for 'in'

timn:
## What
Add a missing lexer rule for 'in'

## Why
Antlr4 reports an error when separating lexer and parser grammars into separate files. This was a necessity for work on a C++ parser.

## Tests
The parser test suite completes without error (ran `bazel test ...`)

Discussed with @TristonianJones.

### DIFF
--- a/parser/gen/CEL.g4
+++ b/parser/gen/CEL.g4
@@ -93,6 +93,7 @@ literal
 
 EQUALS : '==';
 NOT_EQUALS : '!=';
+IN: 'in';
 LESS : '<';
 LESS_EQUALS : '<=';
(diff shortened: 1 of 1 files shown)


TristonianJones:
I appreciate the update to the lexer rule, but the codegen step is currently buggy in that it unwinds some changes we had to make post-Antlr gen to prevent data races. 

The new lexer rule doesn't actually change the behavior of the generated artifacts, so you can revert everything but the `CEL.g4` change.

timn:
Alright, done. Could these changes that have to be applied some sed-magic to have the generation working again?

TristonianJones (approved):

PR #262, opened 2019-09-17
Fix an unparser error on expressions like -(1 + 2)

timn:
## What
For an expression such as `-(1 + 2)`, the unparser would generate `- 1 + 2 `, which obviously is wrong. This came up during updating the C++ CEL decompiler. Also add some more tests that were useful on the C++ side.

## Why
The issue should be fixed to not produce wrong results in this case.

## Tests
The parser test suite completes without error (ran `bazel test ...`)


parser/unparser.go
```diff
@@ -156,24 +156,18 @@ func (un *unparser) visitCallBinary(expr *exprpb.Expr) error {
 func (un *unparser) visitCallConditional(expr *exprpb.Expr) error {
 	c := expr.GetCallExpr()
 	args := c.GetArgs()
-	// add parens if operand is a conditional itself.
-	nested := isSamePrecedence(operators.Conditional, args[0])
-	err := un.visitMaybeNested(args[0], nested)
+	err := un.visitMaybeParenExpr(args[0])
```
TristonianJones: I don't believe this change is strictly necessary since the conditional is the highest precedence operator and so the following expressions are equivalent:

`(x > 5) ? (x - 5) : 0`  -> `x > 5 ? x - 5 : 0`

I would leave the conditional unmodified, but add a new test case to `TestUnparse_Equivalent`

parser/unparser.go
```diff
@@ -230,7 +224,7 @@ func (un *unparser) visitCallUnary(expr *exprpb.Expr) error {
 		return fmt.Errorf("cannot unmangle operator: %s", fun)
 	}
 	un.str.WriteString(unmangled)
-	return un.visit(args[0])
+	return un.visitMaybeParenExpr(args[0])
```
TristonianJones: Thanks for catching this! When the unary argument is a non-unary operator, the parentheses are incorrect. The current check will add parentheses around standard function calls as well, so perhaps the best choice would be to check whether `args[0]` is an operator:

```
if arg.GetCallExpr() == nil {
  return false
}
fn := arg.GetCallExpr().GetFunction()
_, nested := operators.FindReverse(fn)
un.visitMaybeNested(arg, nested)
```

TristonianJones (changes requested):

parser/unparser.go
timn: Yes, I just ran into that myself testing further. I have now changed it such that it combines both, in order to properly wrap conditionals, but at the same time also provide the slightly more readable inner expressions in parenthesis if applicable.

parser/unparser.go
```diff
@@ -157,23 +157,39 @@ func (un *unparser) visitCallConditional(expr *exprpb.Expr) error {
 	c := expr.GetCallExpr()
 	args := c.GetArgs()
 	// add parens if operand is a conditional itself.
+	// otherwise defer decision to visitMaybeParenExpr.
 	nested := isSamePrecedence(operators.Conditional, args[0])
```
TristonianJones: The precedence and paren expr checking functions are more or less doing the same thing, i.e. is the argument an operator with more than one argument? Instead of checking for precedence, I would instead create a function called `isComplexOperator`:

```go
func isComplexOperator(arg *exprpb.Expr) bool {
  call := arg.GetCallExpr()
  // If the arg is not a call with more than one arg, return false.
  if call == nil || len(call.GetArgs()) < 2 {
    return false
  }
  // Otherwise, return whether the function name is listed as an operator.
  _, isOp := operators.FindReverse(call.GetFunction())
  return isOp
}
```

Then I would recommend writing the code as 

```go
nested := isComplexOperator(args[0])
err = un.visitMaybeNested(args[0], nested)
```

parser/unparser.go
TristonianJones: Prefer using the recommended `isComplexOperator` function in the previous comment with the `maybeVisitNested` call here.

TristonianJones (changes requested):

timn:
Will have another look tomorrow, didn't get to it today.

parser/unparser.go
timn: This function by itself is insufficient, it generates invalid expressions on conditional, since it also has to take non-call functions into account. I have added an additional commit which could go into the direction of what you proposed, but extends it. I'm not sure whether this really improves clarity.

If you run just the version with `isComplexOperator`, you will see that the tests no longer pass, since `a ? (b ? c : d) : e` will be unparsed into the invalid expression `a ? b ? c : d : e`.

TristonianJones: Ah my apologies, I forgot that the `FindReverse` didn't account for ternary operators. Probably some cleanup is due there to make its purpose clearer.

Instead, you'd write the following to get the desired effect:

```go
func isComplexOperator(arg *exprpb.Expr) bool {
	call := arg.GetCallExpr()
	// If the arg is not a call with more than one arg, return false.
	if call == nil || len(call.GetArgs()) < 2 {
		return false
	}
	// Otherwise, return whether the function name is listed as an operator  with any precedence.
	return operators.Precedence(call.GetFunction()) != 0
}
```

I guess my point is that the current function for `isComplexOperator` is more complex than it needs to be.

timn: Made this more clear with the new commits, uses the same structure as the C++ decompiler now and added several more test cases which were problematic before.

timn:
Hi @TristonianJones. I think this is ready for another look. More test coverage, better function naming, distinction of cases, make decisions on nesting early, and alignment with C++ decompiler.

TristonianJones (approved):

PR #263, opened 2019-09-24
Disable dashboard tests hit by GCB builder bug.

JimLarson:
See https://github.com/GoogleCloudPlatform/cloud-builders/issues/548.


madi8229 (approved):
LGTM.

PR #264, opened 2019-09-25
Support function resolution by overload id for custom functions.

TristonianJones:
Update `planner.go` to consider function resolution by overload first, then by function name
as this would be more consistent with the desired behavior and makes custom function
overloading easier:

The `dispatcher.go` errors if more than one custom function has the same name, and reasonably
so, but the function resolution step in the planner was only considering the top-level function name
and not the overload, so it pretty much impossible to implement the overloads separately.

```go
// Desired behavior for overloads of the 'greet' function:
Functions(
  &functions.Overload{
      Operator: "string_greet_string",
      Binary: ...},
  &functions.Overload{
      Operator: "int_greet_int",
      Binary: ...}
)
```

```go
// Previous behavior for overloads of the 'greet' function:
Functions(
  &functions.Overload{
      Operator: "greet",
      Binary: func(...) {
         // type-test and overload internally.
      }},
)
```

### DIFF
--- a/interpreter/planner.go
+++ b/interpreter/planner.go
@@ -254,15 +254,6 @@ func (p *planner) planCall(expr *exprpb.Expr) (Interpretable, error) {
 		len(oRef.GetOverloadId()) == 1 {
 		oName = oRef.GetOverloadId()[0]
 	}
-	// Try to find the specific function by overload id.
-	var fnDef *functions.Overload
-	if oName != "" {
-		fnDef, _ = p.disp.FindOverload(oName)
-	}
-	// If the overload id couldn't resolve the function, try the simple function name.
-	if fnDef == nil {
-		fnDef, _ = p.disp.FindOverload(fnName)
-	}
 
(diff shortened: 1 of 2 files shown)


interpreter/planner.go
```diff
@@ -255,6 +254,15 @@ func (p *planner) planCall(expr *exprpb.Expr) (Interpretable, error) {
...
 	}
+	// Try to find the specific function by overload id.
+	var fnDef *functions.Overload
+	if oName != "" {
+		fnDef, _ = p.disp.FindOverload(oName)
+	}
+	// If the overload id couldn't resolve the function, try the simple function name.
+	if fnDef == nil {
+		fnDef, _ = p.disp.FindOverload(fnName)
+	}
```
gabrielsz: Nit: You can delay the computation until after the specialized functions are accounted for. Won't make a big performance difference unless you make a heavy use of the built-in ones.

gabrielsz (approved):

PR #265, opened 2019-09-25
Fix the 'no such key' error message.

TristonianJones:
Ensure the missing key is properly mentioned in the `no such key` error messages.


madi8229 (approved):
LGTM.

PR #267, opened 2019-09-26
Simple type conversion fix

### LINKED ISSUE #258: Primitive types not assignable to proto wrapper fields
**Describe the bug**

While wrapper types now check and unwrap at runtime with PR #205, they do not rewrap properly during message construction. 

**To Reproduce**

- [ ]  parser
- [ ]  checker
- [X]  interpreter

```cel
// The expression will type check, but will not properly wrap `1` as a `google.protobuf.Int64Value`
TestAllTypes{ single_int64_wrapper: 1 }
```

Test setup:
```go
// interpreter_test.go
{
	name: "literal_pb3_msg_wkt",
	pkg:   "google.expr.proto3.test",
	expr: `TestAllTypes{ single_int64_wrapper: 1, single_uint64_wrapper: null }`,
        out:  &proto3pb.TestAllTypes{
				SingleInt64Wrapper: google.protobuf.Int64Value{ value: 1 },                               
		},
}
```

**Expected behavior**
Primitive and `null` values should rewrap appropriately when being set on proto message
fields of wrapper type. If a value is `null` it should not set the field as the `null` indicates that
the wrapper field was not set on the original message.

TristonianJones:
Add support for the following type conversions:

* [Proto3 to JSON](https://developers.google.com/protocol-buffers/docs/proto3#json) conversions
* Primitives to `google.protobuf.Any`
* Primitives to protobuf wrapper types.

Closes #258

### DIFF
--- a/common/types/bytes.go
+++ b/common/types/bytes.go
@@ -16,11 +16,17 @@ package types
 
 import (
 	"bytes"
+	"encoding/base64"
 	"fmt"
 	"reflect"
 
+	"github.com/golang/protobuf/ptypes"
+
 	"github.com/google/cel-go/common/types/ref"
 	"github.com/google/cel-go/common/types/traits"
+
+	structpb "github.com/golang/protobuf/ptypes/struct"
+	wrapperspb "github.com/golang/protobuf/ptypes/wrappers"
 )
 
 // Bytes type that implements ref.Val and supports add, compare, and size
@@ -33,6 +39,9 @@ var (
 		traits.AdderType,
 		traits.ComparerType,
 		traits.SizerType)
+
+	// byteWrapperType golang reflected type for protobuf bytes wrapper type.
+	byteWrapperType = reflect.TypeOf(&wrapperspb.BytesValue{})
 )
 
 // Add implements traits.Adder interface method by concatenating byte sequences.
@@ -59,6 +68,21 @@ func (b Bytes) ConvertToNative(typeDesc reflect.Type) (interface{}, error) {
 	case reflect.Array, reflect.Slice:
 		if typeDesc.Elem().Kind() == reflect.Uint8 {
 			return b.Value(), nil
+		}
+	case reflect.Ptr:
+		switch typeDesc {
+		case anyValueType:
+			return ptypes.MarshalAny(&wrapperspb.BytesValue{Value: []byte(b)})
+		case byteWrapperType:
+			return &wrapperspb.BytesValue{Value: []byte(b)}, nil
+		case jsonValueType:
+			// proto3 to JSON conversion requires base64 encoding of bytes.
+			str := base64.StdEncoding.EncodeToString([]byte(b))
+			return &structpb.Value{
+				Kind: &structpb.Value_StringValue{
+					StringValue: str,
+				},
+			}, nil
 		}
 	case reflect.Interface:
 		if reflect.TypeOf(b).Implements(typeDesc) {
--- a/common/types/bytes.go
+++ b/common/types/bytes.go
@@ -78,7 +78,9 @@ func (b Bytes) ConvertToNative(typeDesc reflect.Type) (interface{}, error) {
 			// Convert the bytes to a protobuf.BytesValue.
 			return &wrapperspb.BytesValue{Value: []byte(b)}, nil
 		case jsonValueType:
-			// proto3 to JSON conversion requires base64 encoding of bytes.
+			// CEL follows the proto3 to JSON conversion by encoding bytes to a string via base64.
+			// The encoding below matches the golang 'encoding/json' behavior during marshaling,
+			// which uses base64.StdEncoding.
 			str := base64.StdEncoding.EncodeToString([]byte(b))
 			return &structpb.Value{
 				Kind: &structpb.Value_StringValue{StringValue: str},
--- a/common/types/bytes.go
+++ b/common/types/bytes.go
@@ -72,16 +72,16 @@ func (b Bytes) ConvertToNative(typeDesc reflect.Type) (interface{}, error) {
 	case reflect.Ptr:
 		switch typeDesc {
 		case anyValueType:
+			// Primitives must be wrapped before being set on an Any field.
 			return ptypes.MarshalAny(&wrapperspb.BytesValue{Value: []byte(b)})
 		case byteWrapperType:
+			// Convert the bytes to a protobuf.BytesValue.
 			return &wrapperspb.BytesValue{Value: []byte(b)}, nil
 		case jsonValueType:
 			// proto3 to JSON conversion requires base64 encoding of bytes.
 			str := base64.StdEncoding.EncodeToString([]byte(b))
 			return &structpb.Value{
-				Kind: &structpb.Value_StringValue{
-					StringValue: str,
-				},
+				Kind: &structpb.Value_StringValue{StringValue: str},
 			}, nil
 		}
 	case reflect.Interface:
--- a/common/types/double.go
+++ b/common/types/double.go
@@ -18,9 +18,12 @@ import (
 	"fmt"
 	"reflect"
 
-	structpb "github.com/golang/protobuf/ptypes/struct"
 	"github.com/google/cel-go/common/types/ref"
 	"github.com/google/cel-go/common/types/traits"
+
+	"github.com/golang/protobuf/ptypes"
+	structpb "github.com/golang/protobuf/ptypes/struct"
+	wrapperspb "github.com/golang/protobuf/ptypes/wrappers"
 )
 
 // Double type that implements ref.Val, comparison, and mathematical
@@ -36,6 +39,12 @@ var (
 		traits.MultiplierType,
 		traits.NegatorType,
 		traits.SubtractorType)
+
+	// doubleWrapperType reflected type for protobuf double wrapper type.
+	doubleWrapperType = reflect.TypeOf(&wrapperspb.DoubleValue{})
+
+	// floatWrapperType reflected type for protobuf float wrapper type.
+	floatWrapperType = reflect.TypeOf(&wrapperspb.FloatValue{})
 )
 
 // Add implements traits.Adder.Add.
@@ -70,10 +79,19 @@ func (d Double) ConvertToNative(typeDesc reflect.Type) (interface{}, error) {
 	case reflect.Float64:
 		return float64(d), nil
 	case reflect.Ptr:
-		if typeDesc == jsonValueType {
+		switch typeDesc {
+		case anyValueType:
+			return ptypes.MarshalAny(&wrapperspb.DoubleValue{Value: float64(d)})
+		case doubleWrapperType:
+			return &wrapperspb.DoubleValue{Value: float64(d)}, nil
+		case floatWrapperType:
+			return &wrapperspb.FloatValue{Value: float32(d)}, nil
+		case jsonValueType:
 			return &structpb.Value{
 				Kind: &structpb.Value_NumberValue{
-					NumberValue: float64(d)}}, nil
+					NumberValue: float64(d),
+				},
+			}, nil
 		}
 		switch typeDesc.Elem().Kind() {
 		case reflect.Float32:
--- a/common/types/double.go
+++ b/common/types/double.go
@@ -16,6 +16,7 @@ package types
 
 import (
 	"fmt"
+	"math"
 	"reflect"
 
 	"github.com/google/cel-go/common/types/ref"
@@ -87,6 +88,23 @@ func (d Double) ConvertToNative(typeDesc reflect.Type) (interface{}, error) {
 		case floatWrapperType:
 			return &wrapperspb.FloatValue{Value: float32(d)}, nil
 		case jsonValueType:
+			// Handle special cases for proto3 to json conversion.
+			f := float64(d)
+			str := ""
+			if math.IsNaN(f) {
+				str = "NaN"
+			} else if math.IsInf(f, -1) {
+				str = "-Infinity"
+			} else if math.IsInf(f, 0) || math.IsInf(f, 1) {
+				str = "Infinity"
+			}
+			if str != "" {
+				return &structpb.Value{
+					Kind: &structpb.Value_StringValue{
+						StringValue: str,
+					},
+				}, nil
+			}
 			return &structpb.Value{
 				Kind: &structpb.Value_NumberValue{
 					NumberValue: float64(d),
--- a/common/types/double.go
+++ b/common/types/double.go
@@ -23,6 +23,7 @@ import (
 	"github.com/google/cel-go/common/types/traits"
 
 	"github.com/golang/protobuf/ptypes"
+
 	structpb "github.com/golang/protobuf/ptypes/struct"
 	wrapperspb "github.com/golang/protobuf/ptypes/wrappers"
 )
--- a/common/types/double.go
+++ b/common/types/double.go
@@ -16,7 +16,6 @@ package types
 
 import (
 	"fmt"
-	"math"
 	"reflect"
 
 	"github.com/google/cel-go/common/types/ref"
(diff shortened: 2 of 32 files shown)


common/types/double.go
```diff
@@ -70,10 +81,36 @@ func (d Double) ConvertToNative(typeDesc reflect.Type) (interface{}, error) {
...
-		if typeDesc == jsonValueType {
+		switch typeDesc {
+		case anyValueType:
+			return ptypes.MarshalAny(&wrapperspb.DoubleValue{Value: float64(d)})
+		case doubleWrapperType:
+			return &wrapperspb.DoubleValue{Value: float64(d)}, nil
+		case floatWrapperType:
+			return &wrapperspb.FloatValue{Value: float32(d)}, nil
+		case jsonValueType:
+			// Handle special cases for proto3 to json conversion.
```
JimLarson: No need for the special cases - already handled by the marshaling of structpb.Value in jsonpb.

common/types/double_test.go
```diff
@@ -81,6 +98,36 @@ func TestDouble_ConvertToNative_Json(t *testing.T) {
 	} else if !proto.Equal(val.(proto.Message), pbVal) {
 		t.Errorf("Got '%v', expected -1.4", val)
 	}
+
+	val, err = Double(math.NaN()).ConvertToNative(jsonValueType)
+	pbVal = &structpb.Value{Kind: &structpb.Value_StringValue{StringValue: "NaN"}}
```
JimLarson: A google.protobuf.Value should hold a NaN as a double.  You could convert to JSON and check that it's converted to a quoted string, but that would just duplicate the testing of the jsonpb package.  Same for the infinities.

common/types/bytes.go
```diff
@@ -60,6 +69,21 @@ func (b Bytes) ConvertToNative(typeDesc reflect.Type) (interface{}, error) {
...
 			return b.Value(), nil
 		}
+	case reflect.Ptr:
+		switch typeDesc {
+		case anyValueType:
+			return ptypes.MarshalAny(&wrapperspb.BytesValue{Value: []byte(b)})
+		case byteWrapperType:
+			return &wrapperspb.BytesValue{Value: []byte(b)}, nil
+		case jsonValueType:
+			// proto3 to JSON conversion requires base64 encoding of bytes.
```
JimLarson: Change to "CEL follows proto3 by converting to JSON via base64 encoding and note why we're making a fixed choice for non-URL-safe, padded encoding, or add a TODO to add options for those later.

Add a note that we can't just use jsonpb to render the string because it mixes different levels - it would add extra double quotes to the string.

common/types/duration.go
```diff
@@ -103,8 +105,22 @@ func (d Duration) Compare(other ref.Val) ref.Val {
...
 // ConvertToNative implements ref.Val.ConvertToNative.
 func (d Duration) ConvertToNative(typeDesc reflect.Type) (interface{}, error) {
-	if typeDesc == durationValueType {
+	switch typeDesc {
+	case anyValueType:
+		return ptypes.MarshalAny(d.Value().(*dpb.Duration))
+	case durationValueType:
 		return d.Value(), nil
+	case jsonValueType:
+		// proto3 to JSON conversion requires string-formatted durations
```
JimLarson: Change to "CEL follows the proto3 to JSON conversion."  Consider using jsonpb to render the string, though that slightly mixes the levels.

Note that we can't just use jsonpb to render the string because it would add extra double quotes.

common/types/int.go
```diff
@@ -75,14 +86,32 @@ func (i Int) Compare(other ref.Val) ref.Val {
...
+		// Enums are also mapped as int32 derivations.
 		return reflect.ValueOf(i).Convert(typeDesc).Interface(), nil
 	case reflect.Int64:
 		return int64(i), nil
 	case reflect.Ptr:
-		if typeDesc == jsonValueType {
+		switch typeDesc {
+		case anyValueType:
+			return ptypes.MarshalAny(&wrapperspb.Int64Value{Value: int64(i)})
+		case int32WrapperType:
```
JimLarson, line 100: Add a note that we can't export to EnumValue since we don't know which enum we'd be talking about.

common/types/int.go
```diff
@@ -75,14 +86,32 @@ func (i Int) Compare(other ref.Val) ref.Val {
...
+			return &wrapperspb.Int32Value{Value: int32(i)}, nil
+		case int64WrapperType:
+			return &wrapperspb.Int64Value{Value: int64(i)}, nil
+		case jsonValueType:
+			if i.isInt32() {
```
JimLarson: This violates the convention used by proto3 where the number/decimal string choice is made based on the type, not the value.  You can still get JSON numbers if you're working with a message with the 32-bit integer field.

common/types/list.go
```diff
@@ -115,8 +118,16 @@ func (l *baseList) Contains(elem ref.Val) ref.Val {
 
 // ConvertToNative implements the ref.Val interface method.
 func (l *baseList) ConvertToNative(typeDesc reflect.Type) (interface{}, error) {
+	// TODO: Add support for conversion to Any
 	// JSON conversions are a special case since the 'native' type is a proto message.
-	if typeDesc == jsonValueType || typeDesc == jsonListValueType {
+	switch typeDesc {
+	case anyValueType:
+		json, err := l.ConvertToNative(jsonValueType)
```
JimLarson: I think g.p.ListValue should be preferred instead of the less-specific g.p.Value.

common/types/map.go
```diff
@@ -83,17 +87,23 @@ func (m *stringMap) Contains(index ref.Val) ref.Val {
 }
 
 // ConvertToNative implements the ref.Val interface method.
-func (m *baseMap) ConvertToNative(refType reflect.Type) (interface{}, error) {
-	// JSON conversion.
-	if refType == jsonValueType || refType == jsonStructType {
+func (m *baseMap) ConvertToNative(typeDesc reflect.Type) (interface{}, error) {
+	switch typeDesc {
+	case anyValueType:
+		json, err := m.ConvertToNative(jsonValueType)
```
JimLarson: I think g.p.Struct should be preferred here instead of the less-specific g.p.Value.

common/types/map.go
```diff
@@ -102,8 +112,13 @@ func (m *baseMap) ConvertToNative(refType reflect.Type) (interface{}, error) {
 	}
 
 	// Non-map conversion.
-	if refType.Kind() != reflect.Map {
-		return nil, fmt.Errorf("type conversion error from map to '%v'", refType)
+	if typeDesc.Kind() != reflect.Map {
+		return nil, fmt.Errorf("type conversion error from map to '%v'", typeDesc)
+	}
+
+	// If the list is already assignable to the desired type return it.
```
JimLarson: s/list/map/ ?

common/types/null.go
```diff
@@ -38,16 +40,18 @@ func (n Null) ConvertToNative(typeDesc reflect.Type) (interface{}, error) {
 	switch typeDesc.Kind() {
 	case reflect.Ptr:
 		switch typeDesc {
-		case jsonValueType:
-			return &structpb.Value{
-				Kind: &structpb.Value_NullValue{
-					NullValue: structpb.NullValue_NULL_VALUE}}, nil
 		case anyValueType:
 			pb, err := n.ConvertToNative(jsonValueType)
```
JimLarson, line 53: Prefer g.p.NullValue instead of g.p.Value.

common/types/null.go
```diff
@@ -38,16 +40,18 @@ func (n Null) ConvertToNative(typeDesc reflect.Type) (interface{}, error) {
...
 			if err != nil {
 				return nil, err
 			}
 			return ptypes.MarshalAny(pb.(proto.Message))
+		case jsonValueType:
```
JimLarson, line 58: Should also have a case for g.p.NullValue.

common/types/object.go
```diff
@@ -48,19 +52,31 @@ func NewObject(adapter ref.TypeAdapter,
...
+			return pb, nil
+		}
+		return ptypes.MarshalAny(pb)
+	case jsonValueType:
+		jsonTxt, err := (&jsonpb.Marshaler{}).MarshalToString(pb)
+		if err != nil {
+			return nil, err
+		}
+		json := &structpb.Value{}
+		err = jsonpb.UnmarshalString(jsonTxt, json)
```
JimLarson, line 71: This round-trip is worth a comment.

common/types/timestamp.go
```diff
@@ -80,15 +82,29 @@ func (t Timestamp) Compare(other ref.Val) ref.Val {
 
 // ConvertToNative implements ref.Val.ConvertToNative.
 func (t Timestamp) ConvertToNative(typeDesc reflect.Type) (interface{}, error) {
-	if typeDesc == timestampValueType {
+	switch typeDesc {
+	case anyValueType:
+		return ptypes.MarshalAny(t.Timestamp)
+	case jsonValueType:
+		// proto3 to JSON conversion requires string-formatted timestamps
```
JimLarson: We follow proto3 to JSON conversion which formats as an RFC 3339 encoded JSON string.

JimLarson (changes requested):

common/types/int.go
TristonianJones: I'm not sure I follow. We don't have an int32 internally, so this test is to comply with the spirit of documented conversion of int32 as float64 and int64 as string

JimLarson: The documented conversion changes all integers from a 64-bit type to decimal strings, even those that fit in 53 or 32 bits, e.g. "1".  Our int type is 64-bit.  Yes, we do in some cases widen a 32-bit source to 64 bits, but at this point in the code we don't have that context.  None of the proto-to-JSON integer conversions depend on the values, so this is an innovation.

That being said, it's quite possible that most code which deals with the JSON-converted data is resigned to handling strings or numbers for fields which expect an int.  But if that's the case and we're going to make the JSON type depend on the value, then we might as well convert every int up to 2^53 (or 52? and there's a +/- 1 in there too) to a number.

common/types/bytes.go
TristonianJones: We're using the same encoding as the `"encoding/json"` library does with `json.Marshal([]byte(val))` in that it uses `base64.StdEncoding.Encode`. The main difference is the lack of quoting in doing so manually. I can add a comment to this effect.

common/types/int.go
TristonianJones: Ah, fair point. I think it makes sense to adjust to `int53` as the limit.

common/types/null.go
TristonianJones: I don't think I can in this case since the NullValue is an enum and not a `proto.Message` instance.

common/types/null.go
TristonianJones: I'll add a test case for this. It probably needs to sit in the case `reflect.Int32`

TristonianJones:
PTAL

common/types/double_test.go
```diff
@@ -81,6 +98,38 @@ func TestDouble_ConvertToNative_Json(t *testing.T) {
...
+
+	val, err = Double(math.Inf(-1)).ConvertToNative(jsonValueType)
+	pbVal = &structpb.Value{Kind: &structpb.Value_NumberValue{NumberValue: math.Inf(-1)}}
+	if err != nil {
+		t.Error(err)
+	} else if !proto.Equal(val.(proto.Message), pbVal) {
+		t.Errorf("Got '%v', expected -Infinity", val)
+	}
+	val, err = Double(math.Inf(0)).ConvertToNative(jsonValueType)
+	pbVal = &structpb.Value{Kind: &structpb.Value_NumberValue{NumberValue: math.Inf(0)}}
```
JimLarson, line 121: Nit:  the math pkg says Inf(0) should be the same as Inf(1).

common/types/bytes.go
JimLarson: Doesn't look like a comment was added.  After thinking about it (a lot), I don't think we need to explain not using the jsonpb package.  But I think we do need to explain a) that we're following proto-to-json conversion rules here, even though we could choose to do something else (as we do for numbers), b) that we're hardcoding the choice of the non-URL-safe, padded version of base64, even though the protobuf docs say that implementations can support choices for those options, because jsonpb doesn't provide those options either.

Suggested wording:
```
// We follow proto3 to JSON conversion by encoding bytes to a string via base64.
// Following the jsonpb package, we use base64.StdEncoding.

File: common/types/int.go [hunk 15]
```diff
@@ -75,14 +85,38 @@ func (i Int) Compare(other ref.Val) ref.Val {
...
+			// Primitives must be wrapped before being set on an Any field.
+			return ptypes.MarshalAny(&wrapperspb.Int64Value{Value: int64(i)})
+		case int32WrapperType:
+			// Convert the value to a protobuf.Int32Value (with truncation).
+			return &wrapperspb.Int32Value{Value: int32(i)}, nil
+		case int64WrapperType:
+			// Convert the value to a protobuf.Int64Value.
+			return &wrapperspb.Int64Value{Value: int64(i)}, nil
+		case jsonValueType:
+			// JSON can accurately represent 53-bit ints as floating point values.
```
JimLarson: Replace this and the following comment with the following, which  documents that we're intentionally deviating from proto-to-json and the reason why:

```
// The proto-to-JSON conversion rules would convert all 64-bit integer values to JSON decimal strings.
// Because CEL ints might come from the automatic widening of 32-bit values in protos,
// we choose instead to choose the JSON type dynamically based on the value.
// We convert all integers -2^53 < n < 2^53 to JSON numbers,
// the largest interval where such a conversion can round-trip accurately.
// Thus, conversions from a 32-bit source can expect a JSON number as with protobuf.
// Those consuming JSON from a 64-bit source must be able to handle either a JSON number
// or a JSON decimal string, and since that's the case, we might as well
// use JSON numbers whenever possible.
```

common/types/int_test.go
```diff
@@ -75,12 +93,22 @@ func TestInt_ConvertToNative_Int64(t *testing.T) {
 }
 
 func TestInt_ConvertToNative_Json(t *testing.T) {
-	val, err := Int(4147483648).ConvertToNative(jsonValueType)
+	// Value less than int32.
```
JimLarson: "Value can be represented accurately as a JSON number".

common/types/int_test.go
```diff
@@ -75,12 +93,22 @@ func TestInt_ConvertToNative_Int64(t *testing.T) {
...
+	val, err := Int(maxIntJSON).ConvertToNative(jsonValueType)
+	if err != nil {
+		t.Error(err)
+	} else if !proto.Equal(val.(proto.Message),
+		&structpb.Value{Kind: &structpb.Value_NumberValue{NumberValue: 9007199254740991.0}}) {
+		t.Errorf("Got '%v', expected a json number for a 32-bit int", val)
+	}
+
+	// Value greater than max int32.
```
JimLarson: "Value converts to a JSON decimal string".

(Hmm, +/- 2^53 actually *can* be represented accurately as a double, in case you want to expand the range in each direction by one.  But integer maxes are usually one less than a power of 2.)

JimLarson (approved):
LGTM, pending some comment cleanups and optional tweaks.

common/types/int_test.go
TristonianJones: The `maxIntJSON` is copied from the EcmaScript 6 standard, so 2^53-1

common/types/double_test.go
TristonianJones: Ah, I was thrown off by the `math.IsInf()` which can be used to differentiate between positive, negative, or any ole' Infinity. I see the `math.Inf()` call behaves exactly as you mentioned. It's a bit weird that `IsInf(0)` would return infinity for positive or negative, but at any rate, thanks for clearing that up.

common/types/bytes.go
TristonianJones: Updated. I thought I had a comment here, but I must have merged it away on accident.

PR #271, opened 2019-10-09
Add an `Env.Extend` method.

TristonianJones:
Introduce a new method to extend a `cel.Env` with additional `cel.EnvOption` options.

Note, the new method clones the current environment before configuring the extended options.

### DIFF
--- a/cel/env.go
+++ b/cel/env.go
@@ -61,6 +61,9 @@ type Ast interface {
 // constants, variables, and functions. The Env interface also defines a method for generating
 // evaluable programs from parsed and checked Asts.
 type Env interface {
+	// Extend the current environment with additional options to produce a new Env.
+	Extend(opts ...EnvOption) (Env, error)
+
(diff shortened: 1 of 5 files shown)


cel/env.go
```diff
@@ -118,6 +121,25 @@ func NewEnv(opts ...EnvOption) (Env, error) {
...
+func (e *env) Extend(opts ...EnvOption) (Env, error) {
+	ext := &env{
+		declarations:                   e.declarations,
+		macros:                         e.macros,
+		pkg:                            e.pkg,
+		provider:                       e.provider,
+		adapter:                        e.adapter,
+		enableBuiltins:                 e.enableBuiltins,
+		enableDynamicAggregateLiterals: e.enableDynamicAggregateLiterals,
+	}
```
JimLarson: I think you can just say

```
ext := *e
return (&ext).configure(opts...)
```

JimLarson (approved):

PR #272, opened 2019-10-11
Additional tests for the top-level 'cel' package

TristonianJones:
Additional tests for the top-level `cel` package, including:

* Custom Macros
* Custom Type Provider
* IO tests

### DIFF
--- a/cel/env.go
+++ b/cel/env.go
@@ -204,12 +204,8 @@ type env struct {
 
 // Check implements the Env interface method.
 func (e *env) Check(ast Ast) (Ast, Issues) {
-	pe, err := AstToParsedExpr(ast)
-	if err != nil {
-		errs := common.NewErrors(ast.Source())
-		errs.ReportError(common.NoLocation, err.Error())
-		return nil, &issues{errs: errs}
-	}
+	// Note, errors aren't currently possible on the Ast to ParsedExpr conversion.
+	pe, _ := AstToParsedExpr(ast)
 	res, errs := checker.Check(pe, ast.Source(), e.chk)
 	if len(errs.GetErrors()) > 0 {
 		return nil, &issues{errs: errs}
--- a/cel/io_test.go
+++ b/cel/io_test.go
@@ -1,0 +1,70 @@
+// Copyright 2019 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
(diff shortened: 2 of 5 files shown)


TristonianJones:
@tobiaszheller I think this should get the top-level package up to 83% coverage which will satisfy the awesome-go quality standards: https://github.com/avelino/awesome-go/pull/2756

cel/env.go
```diff
@@ -204,12 +204,8 @@ type env struct {
 
 // Check implements the Env interface method.
 func (e *env) Check(ast Ast) (Ast, Issues) {
-	pe, err := AstToParsedExpr(ast)
-	if err != nil {
-		errs := common.NewErrors(ast.Source())
-		errs.ReportError(common.NoLocation, err.Error())
-		return nil, &issues{errs: errs}
-	}
+	// Note, errors aren't currently possible on the Ast to ParsedExpr conversion.
```
gabrielsz, line 207: If this is expected to change ~soon, let's link the github issue for it. If it's a note about a design limitation, it should be OK to leave as-is.

cel/io_test.go
```diff
@@ -0,0 +1,70 @@
...
+	}
+	checked, err := AstToCheckedExpr(ast)
+	if err != nil {
+		t.Fatal(err)
+	}
+	ast3 := CheckedExprToAst(checked)
+	if !proto.Equal(ast3.Expr(), ast.Expr()) {
+		t.Fatalf("Got %v, wanted %v", ast3, ast)
+	}
+
```
gabrielsz: Nit: Extra \n

cel/io_test.go
```diff
@@ -0,0 +1,70 @@
...
+}
+
+func TestIO_AstToString(t *testing.T) {
+	stdEnv, _ := NewEnv()
+	in := "a + b - (c ? (-d + 4) : e)"
+	ast, _ := stdEnv.Parse(in)
```
gabrielsz, line 61: This is assuming all of the strings are parse-able, right?

gabrielsz (approved):
LGTM. I've left low priority nitpicks that don't need a re-review

cel/env.go
TristonianJones: Yep, this is just a design note. For the foreseeable future, it's not possible to produce an error when converting between an `Ast` and a `ParsedExpr`, but it is possible with a `CheckedExpr`. In order to keep the interface symmetric there's an `error` on the return.

cel/io_test.go
TristonianJones: That's right, the test is assuming the string will parse since the purpose of the test is to show the serialization.

PR #273, opened 2019-10-17
Support for parsing embedded CEL expressions within larger files

TristonianJones:
In order to support parsing of CEL expressions within larger source files, the `ParseSource` 
method which accepts a `common.Source` has been added to the `cel.Env` interface.

The `common.Source` value preserves line offsets and provides abstractions for determining
a line and column from an absolute character position. With this change it is possible to embed
CEL expressions within a larger file, e.g. YAML and still refer to the correct line, column for CEL
issues in the embedded expressions.

Note: This is a breaking interface change to `cel.Env`, but future CLs should remove this interface
to ensure better forward compatibility of changes.

### DIFF
--- a/cel/env.go
+++ b/cel/env.go
@@ -75,12 +75,18 @@ type Env interface {
 
 	// Parse parses the input expression value `txt` to a Ast and/or a set of Issues.
 	//
+	// This form of Parse creates a common.Source value for the input `txt` and forwards to the
+	// ParseSource method.
+	Parse(txt string) (Ast, Issues)
+
+	// Parse parses the input source to an Ast and/or set of Issues.
+	//
 	// Parsing has failed if the returned Issues value and its Issues.Err() value is non-nil.
 	// Issues should be inspected if they are non-nil, but may not represent a fatal error.
 	//
 	// It is possible to have both non-nil Ast and Issues values returned from this call; however,
 	// the mere presence of an Ast does not imply that it is valid for use.
-	Parse(txt string) (Ast, Issues)
+	ParseSource(src common.Source) (Ast, Issues)
 
 	// Program generates an evaluable instance of the Ast within the environment (Env).
 	Program(ast Ast, opts ...ProgramOption) (Program, error)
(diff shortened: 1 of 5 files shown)


cel/env.go
```diff
@@ -71,16 +70,22 @@ type Env interface {
...
-	Check(ast Ast) (Ast, Issues)
+	Check(ast Ast) (Ast, *Issues)
 
 	// Parse parses the input expression value `txt` to a Ast and/or a set of Issues.
 	//
+	// This form of Parse creates a common.Source value for the input `txt` and forwards to the
+	// ParseSource method.
+	Parse(txt string) (Ast, *Issues)
+
+	// Parse parses the input source to an Ast and/or set of Issues.
```
JimLarson: s/Parse/ParseSource/

cel/env.go
```diff
@@ -255,25 +252,39 @@ func (e *env) TypeProvider() ref.TypeProvider {
 	return e.provider
 }
 
-// issues is the internal implementation of the Issues interface.
-type issues struct {
+// Issues defines methods for inspecting the error details of parse and check calls.
+//
+// Note: in the future, non-fatal warnings and notices may be inspectable via the Issues struct.
+type Issues struct {
```
JimLarson, line 258: What's the motivation for switching from an interface to a specific struct?  Isn't the type abstraction useful in making the public API more stable?

common/source.go
```diff
@@ -187,3 +179,19 @@ func (s *sourceImpl) findLine(characterOffset int32) (int32, int32) {
 	}
 	return line, s.lineOffsets[line-2]
 }
+
+func (s *sourceImpl) idOffset(exprID int64) (int32, bool) {
```
JimLarson, line 186: Please preserve the documentation for this method.

common/source.go
```diff
@@ -187,3 +179,19 @@ func (s *sourceImpl) findLine(characterOffset int32) (int32, int32) {
...
+	if offset, found := s.idOffsets[exprID]; found {
+		return offset, true
+	}
+	return -1, false
+}
+
+func (s *sourceImpl) idLocation(exprID int64) (Location, bool) {
```
JimLarson, line 196: Please preserve the documentation for this method.

JimLarson (changes requested):

cel/env.go
TristonianJones: One of the Go readability reviewers recommended replace interfaces for structs when the interface exists purely for abstraction. Mostly because customers may try to implement the interface and become suddenly broken with the addition of new methods, but this isn't the case for structs. `Source` is a good interface, but `Issues` isn't.

JimLarson (approved):

PR #278, opened 2019-10-29
Improve type conversion support for lists and maps

TristonianJones:
Improve the `ConvertToNative` support for list and map types.

* Add support for CEL string lists to type `[]interface{}` 
* Support the conversion of CEL maps with string keys to custom structs with exported field names.

Tests have also been added to illustrate some of the different conversion behaviors and limitations.

### DIFF
--- a/common/types/list_test.go
+++ b/common/types/list_test.go
@@ -423,7 +423,7 @@ func TestStringList_ConvertToNative(t *testing.T) {
 		t.Error("Unable to convert string list to itself.")
 	}
 	if !reflect.DeepEqual(val, []string{"h", "e", "l", "p"}) {
-		t.Errorf("Got %v, expected ['h', 'e', 'l', 'p']", val)
+		t.Errorf(`Got %v, expected ["h", "e", "l", "p"]`, val)
 	}
 }
 
@@ -435,7 +435,7 @@ func TestStringList_ConvertToNative_ListInterface(t *testing.T) {
 		t.Error("Unable to convert string list to itself.")
 	}
 	if !reflect.DeepEqual(val, []interface{}{"h", "e", "l", "p"}) {
-		t.Errorf("Got %v, expected ['h', 'e', 'l', 'p']", val)
+		t.Errorf(`Got %v, expected ["h", "e", "l", "p"]`, val)
 	}
 }
 
--- a/common/types/list_test.go
+++ b/common/types/list_test.go
@@ -247,6 +247,20 @@ func TestConcatList_ConvertToNative_Json(t *testing.T) {
 	}
 }
 
+func TestConcatList_ConvertToNative_ListInterface(t *testing.T) {
+	reg := NewRegistry()
+	listA := NewDynamicList(reg, []float32{1.0, 2.0})
+	listB := NewStringList(reg, []string{"3.0"})
+	list := listA.Add(listB)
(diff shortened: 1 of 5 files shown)


common/types/list_test.go
```diff
@@ -413,6 +427,18 @@ func TestStringList_ConvertToNative(t *testing.T) {
...
 
+func TestStringList_ConvertToNative_ListInterface(t *testing.T) {
+	reg := NewRegistry()
+	list := NewStringList(reg, []string{"h", "e", "l", "p"})
+	val, err := list.ConvertToNative(reflect.TypeOf([]interface{}{}))
+	if err != nil {
+		t.Error("Unable to convert string list to itself.")
+	}
+	if !reflect.DeepEqual(val, []interface{}{"h", "e", "l", "p"}) {
+		t.Errorf("Got %v, expected ['h', 'e', 'l', 'p']", val)
```
JimLarson: Nit: the "expected" message seems to say that you want a list of rune literals, rather than a list of strings.  Consider whether the increased clarity of the message is worth the decreased clarity of the code (with all the backslashes).

common/types/map.go
```diff
@@ -111,49 +112,89 @@ func (m *baseMap) ConvertToNative(typeDesc reflect.Type) (interface{}, error) {
 				StructValue: jsonMap}}, nil
 	}
 
+	// Unwrap pointers, but track their use.
+	isPtr := false
+	if typeDesc.Kind() == reflect.Ptr {
```
JimLarson, line 118: Do you want a loop here for pointers-to-pointers, etc?

common/types/map.go
```diff
@@ -111,49 +112,89 @@ func (m *baseMap) ConvertToNative(typeDesc reflect.Type) (interface{}, error) {
...
+		typeDesc = typeDesc.Elem()
+		isPtr = true
+	}
+
 	// Non-map conversion.
-	if typeDesc.Kind() != reflect.Map {
+	if typeDesc.Kind() != reflect.Map && typeDesc.Kind() != reflect.Struct {
```
JimLarson: Is this redundant with the control flow from the switch at line 140?  If so, delete this.  If not, add a little comment explaining why we need to bail out here.

JimLarson (approved):

common/types/map.go
TristonianJones: Yep, thanks.

common/types/map.go
```diff
@@ -111,49 +113,95 @@ func (m *baseMap) ConvertToNative(typeDesc reflect.Type) (interface{}, error) {
...
+			key := it.Next()
+			// Ensure the field name being referenced is exported.
+			fieldName := string(key.ConvertToType(StringType).(String))
+			switch len(fieldName) {
+			case 0:
+				return nil, errors.New("type conversion error, unsupported empty field")
+			case 1:
+				fieldName = strings.ToUpper(fieldName)
+			default:
+				fieldName = strings.ToUpper(fieldName[0:1]) + fieldName[1:]
```
JimLarson, line 187: Sorry to have not caught this the first time around.  I think it's worth a comment explaining why you're capitalizing the field name.

JimLarson (approved):

common/types/map.go
TristonianJones: Ah, yes, I tried to do so with the `Ensure the field name being referenced is exported`, but I can be more clear.

PR #284, opened 2019-11-25
Attribute Resolver for faster data accesses

TristonianJones:
There are three high-level concepts introduced in this PR:

- `Attribute`: a variable or object reference whose use may be qualified zero or more field / index 
   accesses.
- `Qualifier`: a valid protobuf map key type (`bool`, `int`, `string`, `uint`) used to select data from
   a value.
- `Resolver`: a mechanism for determining how to resolve an `Attribute` to a value from an 
  `Activation` or some other context in the expression under evaluation.

The use of the `Attribute` and `Resolver` changes the default evaluation semantics for missing data by producing an error unless expressly instructed to produce an `types.Unknown` value (#190). The performance is also greatly improved for many data accesses.

Before:
```
select_key-8         	  100000	     11079 ns/op	    2433 B/op	     118 allocs/op
select_index-8       	  200000	     11930 ns/op	    2769 B/op	     121 allocs/op
select_field-8       	  100000	     10445 ns/op	    2801 B/op	     103 allocs/op
select_pb2_primitive_fields-8         	  100000	     13927 ns/op	    2513 B/op	     103 allocs/op
select_pb3_wrapper_fields-8           	  200000	     10586 ns/op	    2245 B/op	      77 allocs/op
select_relative-8                     	 1000000	      1665 ns/op	     744 B/op	      17 allocs/op
```

After:
```
select_key-8 	           1000000	      1543 ns/op	     160 B/op	      18 allocs/op
select_index-8             1000000	      1874 ns/op	     416 B/op	      22 allocs/op
select_field-8             2000000	       986 ns/op	     208 B/op	       9 allocs/op
select_pb2_primitive_fields-8         	  500000	      3456 ns/op	     464 B/op	      23 allocs/op
select_pb3_wrapper_fields-8           	 1000000	      1485 ns/op	     404 B/op	       5 allocs/op
select_relative-8                     	 1000000	      1484 ns/op	     728 B/op	      16 allocs/op
```

The After results also come with changes in the test harness which will no longer pre-allocate inputs as `ref.Val`, so the improvements will likely be greater than for most than what's stated here.

Here are two additional points of comparison for single protobuf field comparisons. The first is with the default resolver, the second uses a custom resolver:

```
select_pb3_compare-8          10000000	       176 ns/op	      16 B/op	       2 allocs/op
select_custom_pb3_compare-8   10000000	       135 ns/op	      16 B/op	       2 allocs/op
```


cel/program.go
```diff
@@ -90,11 +91,14 @@ type progGen struct {
 func newProgram(e *env, ast Ast, opts ...ProgramOption) (Program, error) {
 	// Build the dispatcher, interpreter, and default program value.
```
JimLarson, line 92: ...and resolver.

common/types/object.go
```diff
@@ -43,13 +43,14 @@ type protoObj struct {
 // Note:  only uses default Db.
 func NewObject(adapter ref.TypeAdapter,
 	typeDesc *pb.TypeDescription,
+	typeValue *TypeValue,
```
JimLarson, line 49: It's worth a doc comment above describing why the TypeValue might be something unexpected.

common/types/pb/type.go
```diff
@@ -70,11 +72,11 @@ func (td *TypeDescription) ReflectType() reflect.Type {
 
 // DefaultValue returns an empty instance of the proto message associated with the type.
```
JimLarson: ...or nil for wrapper types?

common/types/pb/type.go
```diff
@@ -145,15 +152,28 @@ func (td *TypeDescription) makeMetadata() *typeMetadata {
...
 	index int) *FieldDescription {
+	getterName := fmt.Sprintf("Get%s", prop.Name)
+	getter, _ := tdType.MethodByName(getterName)
+	isProto3 := td.file.desc.GetSyntax() == "proto3"
+	var field *reflect.StructField
+	if isProto3 {
+		if tdType.Kind() == reflect.Ptr {
+			tdType = tdType.Elem()
+		}
+		f, _ := tdType.FieldByName(prop.Name)
```
JimLarson: Test if the field isn't found and panic or something.  It should never happen, but that can't be proven with local information.  The assumed relationship between tdType and prop isn't documented in the method interface.

common/types/pb/type.go
```diff
@@ -204,7 +225,9 @@ func isWrapperType(desc *descpb.FieldDescriptorProto) bool {
 // FieldDescription holds metadata related to fields declared within a type.
 type FieldDescription struct {
 	// getter is the name of the accessor method to obtain the field value.
```
JimLarson: Change to:
getter is the reflected accessor method that obtains the field value.

common/types/pb/type.go
```diff
@@ -244,10 +267,52 @@ func (fd *FieldDescription) CheckedType() *exprpb.Type {
...
+// of proto2 or proto3 accordingly.
+func (fd *FieldDescription) IsSet(target interface{}) bool {
+	t, ok := target.(reflect.Value)
+	if !ok {
+		t = reflect.ValueOf(target)
+	}
+	var fieldVal reflect.Value
+	if fd.field != nil && !fd.IsOneof() {
+		t = reflect.Indirect(t)
+		if t.Kind() == reflect.Struct {
```
JimLarson: Document the that the fall-through with fieldVal left as "no value" is intentional.  Even better, return the desired boolean explicitly rather than falling through.

common/types/pb/type.go
```diff
@@ -244,10 +267,52 @@ func (fd *FieldDescription) CheckedType() *exprpb.Type {
...
+			fieldVal = t.FieldByIndex(fd.field.Index)
+		}
+	} else {
+		fieldVal = fd.getter.Call([]reflect.Value{t})[0]
+	}
+	return isFieldSet(fieldVal)
+}
+
+// GetFrom returns the accessor method associated with the field on the proto generated struct.
```
JimLarson, line 304: Document that target can be a reflect.Value or ... whatever else it can be.

common/types/pb/type.go
```diff
@@ -244,10 +267,52 @@ func (fd *FieldDescription) CheckedType() *exprpb.Type {
...
+// If the field is not set, the proto default value is returned instead.
+func (fd *FieldDescription) GetFrom(target interface{}) (interface{}, error) {
+	t, ok := target.(reflect.Value)
+	if !ok {
+		t = reflect.ValueOf(target)
+	}
+	var fieldVal reflect.Value
+	if fd.field != nil && !fd.IsOneof() {
+		t = reflect.Indirect(t)
+		if t.Kind() == reflect.Struct {
```
JimLarson: As above, either comment that the "no value" fallthrough is intentional, or even better just return the desired results.

common/types/pb/type.go
```diff
@@ -244,10 +267,52 @@ func (fd *FieldDescription) CheckedType() *exprpb.Type {
...
+			fieldVal = t.FieldByIndex(fd.field.Index)
+		}
+	} else {
+		fieldVal = fd.getter.Call([]reflect.Value{t})[0]
+	}
+	if isFieldSet(fieldVal) {
+		return fieldVal.Interface(), nil
```
JimLarson: Check with fieldVal.CanInterface() first, or clearly document why it must be true.

interpreter/activation.go
```diff
@@ -98,21 +82,17 @@ func (a *mapActivation) Parent() Activation {
...
-		default:
-			return a.adapter.NativeToValue(object), true
-		}
+func (a *mapActivation) ResolveName(name string) (interface{}, bool) {
+	obj, found := a.bindings[name]
+	if !found {
+		return nil, false
 	}
-	return nil, false
+	fn, isLazy := obj.(func() ref.Val)
```
JimLarson, line 96: Is there any advantage to allowing lazy objects to be of the more general type `func() interface{}`?

interpreter/activation.go
```diff
@@ -141,6 +121,41 @@ func NewHierarchicalActivation(parent Activation, child Activation) Activation {
...
 
+// UnknownActivation returns an Activation that returns a 'types.Unknown' value for all requests
+// to ResolveName.
+func UnknownActivation() Activation {
+	a, _ := PartialActivation(EmptyActivation())
+	return a
+}
+
+// PartialActivation returns an Activation that will resolve identifier names if present, otherwise
+// will return 'types.Unknown'.
```
JimLarson, line 145: Comment that `bindings` can be `Activation` or `map[string]interface{}`.

interpreter/attributes.go
```diff
@@ -0,0 +1,844 @@
...
+	// AbsoluteAttribute creates an attribute that refers to a top-level variable name.
+	//
+	// Only type-checked expressions generate absolute attributes.
+	AbsoluteAttribute(id int64, name string) Attribute
+
+	// ConditionalAttribute creates an attribute with two Attribute branches, where the Attribute
+	// that is resolved depends on the boolean evaluation of the input 'expr'.
+	ConditionalAttribute(id int64, expr Interpretable, t, f Attribute) Attribute
+
+	// OneofAttribute creates an attribute that refers to either a field selection or a namespaced
```
JimLarson: Nit - I dislike the collision with the proto3 notion of "oneof".  MaybeAttribute?

interpreter/attributes.go
```diff
@@ -0,0 +1,844 @@
...
+	// The qualifier may consider the object type being qualified, if present. If absent, the
+	// qualification should be considered dynamic and the qualification should still work, though
+	// it may be sub-optimal.
+	NewQualifier(objType *exprpb.Type, qualID int64, val interface{}) (Qualifier, error)
+}
+
+// Qualifier marker interface for designating different qualifier values and where they appear
+// within expressions.
+type Qualifier interface {
+	// ID where the qualifier appears within an expression.
```
JimLarson, line 63: Note that it should either be a `Select` or a `Call("_[_]",...)` `Expr` node.

interpreter/attributes.go
```diff
@@ -0,0 +1,844 @@
...
+		res:        r,
+	}
+}
+
+// NewQualifier is an implementation of the AttributeFactory interface.
+func (r *attrFactory) NewQualifier(objType *exprpb.Type,
+	qualID int64,
+	val interface{}) (Qualifier, error) {
+	// Before creating a new qualifier check to see if this is a protobuf message field access.
+	// If so, use the precomputed GetFrom qualification method ratehr than the standard
```
JimLarson: s/ratehr/rather/

interpreter/attributes.go
```diff
@@ -0,0 +1,844 @@
...
+			return nil, fmt.Errorf("no such attribute: %v", typ)
+		}
+		if isUnk {
+			return types.Unknown{a.ID()}, nil
+		}
+	}
+	return nil, fmt.Errorf("no such attribute: %v", a)
+}
+
+// Qualify is an implementation of the AttributeFactory interface method.
```
JimLarson: s/AttributeFactory/Qualifier/

interpreter/attributes.go
```diff
@@ -0,0 +1,844 @@
...
+				adapter:   r.adapter,
+			}, nil
+		}
+	}
+	return newQualifier(r.adapter, qualID, val)
+}
+
+type absoluteAttribute struct {
+	id             int64
+	namespaceNames []string
```
JimLarson, line 189: Describe namespaceNames.

interpreter/attributes.go
```diff
@@ -0,0 +1,844 @@
...
+
+// Resolve iterates through the namespaced variable names until one is found in the Activation,
+// and the the standard qualifier resolution logic is applied.
+//
+// If the variable name is not found an error is returned.
+func (a *absoluteAttribute) Resolve(vars Activation) (interface{}, error) {
+	for _, nm := range a.namespaceNames {
+		op, found := vars.ResolveName(nm)
+		_, isUnk := op.(types.Unknown)
+		if found && !isUnk {
```
JimLarson, line 219: I'm deeply suspicious of the control flow here.  For instance it looks like a variable "x" shadows a type provider identifier "x" - unless variable "x" is bound to unknown, in which case the type constant wins.  Consider the more straightforward flow:
```
op, found := vars.ResolveName(nm)
if found {
    _, isUnk := op.(types.Unknown)
    if isUnk {
        return types.Unknown...
    }
    // handle qualifiers or return err
    return op, nil
}
typ, found := a.provider.FindIdent(nm)
if found {
    // end in returning something
}
// try next namespace name

File: interpreter/attributes.go [hunk 18]
```diff
@@ -0,0 +1,844 @@
...
+	res      AttributeFactory
+}
+
+// ID is an implementation of the Attribute interface method.
+func (a *oneofAttribute) ID() int64 {
+	return a.id
+}
+
+// AddQualifier adds a qualifier to each possible attribute variant in the oneof, and also creates
+// a new namespaced variable from the qualified value.
```
JimLarson, line 341: An example would help greatly here.

interpreter/attributes.go
```diff
@@ -0,0 +1,844 @@
...
+	}, a.attrs...)
+	return a, nil
+}
+
+// Resolve follows the variable resolution rules to determine whether the attribute is a variable
+// or a field selection.
+func (a *oneofAttribute) Resolve(vars Activation) (interface{}, error) {
+	for _, attr := range a.attrs {
+		for _, nm := range attr.namespaceNames {
+			op, found := vars.ResolveName(nm)
```
JimLarson, line 398: Same concerns here with the control flow.

interpreter/attributes.go
```diff
@@ -0,0 +1,844 @@
...
+}
+
+// Qualify implements the Qualifier interface method.
+func (q *intQualifier) Qualify(vars Activation, obj interface{}) (interface{}, error) {
+	i := q.value
+	isMap := false
+	isKey := false
+	isIndex := false
+	switch o := obj.(type) {
+	case map[int]interface{}:
```
JimLarson, line 636: Document clearly why map[int]interface{} is sufficient here, while for a stringQualifier we list map[string]T for all T, in addition to the map[string]interface{} case.

interpreter/attributes.go
```diff
@@ -0,0 +1,844 @@
...
+func (q *fieldQualifier) ID() int64 {
+	return q.id
+}
+
+// Qualify implements the Qualifier interface method.
+func (q *fieldQualifier) Qualify(vars Activation, obj interface{}) (interface{}, error) {
+	if rv, ok := obj.(ref.Val); ok {
+		obj = rv.Value()
+	}
+	v, err := q.FieldType.GetFrom(obj)
```
JimLarson: Consider just `return q.FieldType.GetFrom(obj)`.

interpreter/attributes_test.go
```diff
@@ -0,0 +1,420 @@
...
+	attrs := NewAttributeFactory(packages.DefaultPackage, reg, reg)
+	data := map[string]interface{}{
+		"a": map[int]interface{}{
+			-1: []int32{2, 42},
+		},
+		"b": 1,
+	}
+	vars, _ := NewActivation(data)
+
+	// obj.a[-1][b]
```
JimLarson: It looks like you're repurposing `data` as both the environment and a data structure.  Way too tricky, IMHO.  But if you're going to do so, please be explicit about what's the intended "expression" otherwise it looks like there's a bug here:
```
// {"a": {-1: [2, 42]}, "b": 1}.a[-1][b] in the environment a: {-1: [2, 42]}, b: 1
```

interpreter/attributes_test.go
```diff
@@ -0,0 +1,420 @@
...
+	attrs := NewAttributeFactory(pkg, reg, reg)
+	data := map[string]interface{}{
+		"a": map[int]interface{}{
+			-1: []int32{2, 42},
+		},
+		"acme.b": 1,
+	}
+	vars, _ := NewActivation(data)
+
+	// obj.a[-1][b]
```
JimLarson: Same issue here - too cute.

interpreter/attributes_test.go
```diff
@@ -0,0 +1,420 @@
...
+	data := map[string]interface{}{
+		"a": map[int]interface{}{
+			-1: []int32{2, 42},
+		},
+		"b": []int{0, 1},
+		"c": []interface{}{1, 0},
+	}
+	vars, _ := NewActivation(data)
+
+	// obj[a][-1][(cond ? b : c)[0]]
```
JimLarson: And again.

interpreter/attributes_test.go
```diff
@@ -0,0 +1,420 @@
...
+				"first":  uint(1),
+				"second": uint(2),
+				"third":  uint(3),
+			},
+		},
+		"b": uint32(2),
+	}
+	vars, _ := NewActivation(data)
+
+	// a[-1][{1u: "first", 2u: "second", 3u: "third"}[b]]
```
JimLarson: The comment doesn't match what's evaluated - it should have "obj" or `{"a": , "b": , "c": }` in there somewhere.

JimLarson (changes requested):

interpreter/attributes.go
TristonianJones: You're precisely right that the variable may be getting shadowed by a type. This happens when using the `UnknownActivation` which will return that every name is found and unknown, even when the identifier could actually be a well-established type value. I've kept the logic, but documented why it is this way.

I'll add a TODO to update the logic. I may not keep the unknown activation around beyond the next PR if there's a good way to actually specify unknown attributes with more granularity.

TristonianJones:
PTAL

common/types/pb/type.go
```diff
@@ -160,12 +164,14 @@ func (td *TypeDescription) newFieldDesc(
...
-		}
-		f, _ := tdType.FieldByName(prop.Name)
+	if tdType.Kind() == reflect.Ptr {
+		tdType = tdType.Elem()
+	}
+	f, found := tdType.FieldByName(prop.Name)
+	if found {
 		field = &f
+		// The field will be not found for oneofs, but this is accounted for
+		// by checking the 'desc' value which provides this information.
```
JimLarson: Please move this information about oneof fields in the declaration of FieldDescription.

interpreter/activation.go
```diff
@@ -48,8 +48,14 @@ func EmptyActivation() Activation {
 //
 // The input `bindings` may either be of type `Activation` or `map[string]interface{}`.
 //
-// When the bindings are a `map` form whose values are not of `ref.Val` type, the values will be
-// converted to CEL values (if possible) using the `types.DefaultTypeAdapter`.
+// Lazy bindings may be supplied in either of the following forms:
+// - func() interface{}
```
JimLarson, line 52: In a future, separate change, consider a type alias
```
type native = interface{}
```
with comments documenting what constitutes a native value.  That way the code can have more descriptive types like `func() native` and reserve `interface{}` for situations where we truly mean "anything goes here", even if the type checker can't reliably enforce the restriction.

interpreter/activation.go
```diff
@@ -124,12 +135,25 @@ func NewHierarchicalActivation(parent Activation, child Activation) Activation {
...
+// The input `bindings` may either be of type `Activation` or `map[string]interface{}`.
+//
+// Lazy bindings may be supplied in either of the following forms:
+// - func() interface{}
+// - func() ref.Val
+//
+// The output of the lazy binding will overwrite the variable reference in the internal map.
+//
+// Values which are not represented as ref.Val types on input may be adapted to a ref.Val using
+// the ref.TypeAdapter configured in the environment.
```
JimLarson, line 156: Consider referring to the descrption of `bindings` in `NewActivation`, or define a new type alias
```
type bindings = interface{}
```
and put the comment there.

interpreter/attributes.go
```diff
@@ -324,13 +333,30 @@ type oneofAttribute struct {
...
 }
 
-// AddQualifier adds a qualifier to each possible attribute variant in the oneof, and also creates
+// AddQualifier adds a qualifier to each possible attribute variant, and also creates
 // a new namespaced variable from the qualified value.
-func (a *oneofAttribute) AddQualifier(qual Qualifier) (Attribute, error) {
+//
+// The algorithm for building the maybe attribute is as follows:
+//
+// 1. mb = NewMaybeAttribute(Attribute{[]string{"ns.a", "a"}})
```
JimLarson: The example is confusing - I don't see a NewMaybeAttribute(), I don't know what the string arrays are, etc.

interpreter/attributes.go
JimLarson: Then consider either:

- drop UnknownActivation and PartialActivation right now, since they don't seem to be widely used, or
- check to see whether vars is a partialActivation and handle as a temporary special case within the ideal control-flow, rather than risking spread of the bad precedence for the general case.

interpreter/attributes_test.go
```diff
@@ -192,7 +221,30 @@ func TestAttributes_RelativeAttr_Relative(t *testing.T) {
...
 	vars, _ := NewActivation(data)
 
-	// a[-1][{1u: "first", 2u: "second", 3u: "third"}[b]]
+	// The environment declares the following variables:
+	// {
+	//   a: {
+	//     -1: {
+	//       "first": 1u,
+	//       "second": 2u,
+	//       "third": 3u,
```
JimLarson, line 230: `} }`

JimLarson (changes requested):

cel/program.go
JimLarson: ...and now resolve is attribteFactory, and there's no interpreter.  Maybe just delete the comment.

interpreter/activation.go
TristonianJones: I'd prefer not to use a type alias here since I find `native` misleading. The value can be either CEL's representation, a native value, or a lazy supplier that produces a CEL or native result.

interpreter/attributes.go
TristonianJones: The `UnknownActivation` represents the pre-PR contract for `Activation`. I happen to know of a use case that depends on the ability to do this and would like to ensure that a sync of the latest code can still work correctly with it. 

The `Activation` could be implemented by anyone (and I have leveraged the interface heavily elsewhere). Being able to return unknown for a given variable name is an important capability and I don't particularly want to special case the handling of the response based on the concrete, private, implementation. That said, the shadowing is problematic.

I could try to replace the `UnknownActivation` with a function on the `cel.Env` to generate a `PartialActivation` from the declaration set. I'm happy to do this, but would prefer to do it in a follow-up PR.

interpreter/activation.go
TristonianJones: It seems as though the intent of type aliases was to make it easier to incrementally refactor type names over a series of files. I'd prefer not to add an alias for the sake of documentation.

https://yourbasic.org/golang/type-alias/

common/types/pb/type.go

interpreter/attributes.go
TristonianJones: Updated.

interpreter/attributes_test.go
TristonianJones: Done.

interpreter/attributes.go
TristonianJones: Resolving, conversation continues below.

TristonianJones:
PTAL

interpreter/activation.go
JimLarson: The issue is that `interface{}` means that here, but it might mean something else in another location. E.g. I don't think every consumer of `interface{}` can handle a lazy supplier. But whether in code or commments, this is for the future, so resolving for now.

cel/program.go
```diff
@@ -88,13 +89,16 @@ type progGen struct {
 //
 // If the program cannot be configured the prog will be nil, with a non-nil error response.
 func newProgram(e *env, ast Ast, opts ...ProgramOption) (Program, error) {
-	// Build the dispatcher, interpreter, and default program value.
+	// Build the dispatcher, interpreter, resolver, and default program value.
```
JimLarson: Cleanup - no such thing as a resolver or interpreter any more.  Maybe just delete comment.

JimLarson (approved):

PR #288, opened 2019-12-06
Enable proto-supporting conformance tests

JimLarson:
Build the new protbuf message definitions for conformance tests into the conformance test server.


server/server.go
```diff
@@ -29,6 +29,8 @@ import (
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
 
+	test2pb "github.com/google/cel-spec/proto/test/v1/proto2/test_all_types"
```
TristonianJones, line 32: This can be done in a follow up CL, but if the code generated `.pb.go` files are checked into the cel-spec repo, the copies of these protos in the cel-go repo can be deleted.

TristonianJones (approved):

server/server.go
JimLarson: I'll do that in a followup.

PR #291, opened 2019-12-19
Recursive proto includes

TristonianJones:
Proto includes registered via `proto.Descriptor` or `proto.Message` were not being considered during registration which meant that callers needed to manage the registration themselves. 

The fix was to ensure that dependencies found in the import list of a given proto file are recursively registered.


test/proto3pb/test_all_types.proto
```diff
@@ -94,6 +99,9 @@ message TestAllTypes {
   // Map
   map<string, string> map_string_string = 58;
   map<int64, NestedTestAllTypes> map_int64_nested_type = 59;
+
+  // Imports
+  repeated ImportedGlobalEnum imported_enums = 60;
```
JimLarson, line 104: Can you make the corresponding updates in cel-spec?

test/proto3pb/test_import.proto
```diff
@@ -0,0 +1,11 @@
...
+
+package google.expr.proto3.test;
+
+option go_package = "github.com/google/cel-go/test/proto3pb";
+
+enum ImportedGlobalEnum {
+    IMPORT_FOO = 0;
+    IMPORT_BAR = 1;
+    IMPORT_BAZ = 2;
+}
```
JimLarson: Add newline.

JimLarson (approved):

test/proto3pb/test_all_types.proto
TristonianJones: Sure.

PR #297, opened 2019-12-30
Improve readme docs

tobiaszheller:
I have added `Examples` section into readme. It contains very simple example working only on builtin and two examples with custom instance function and global function. 

I hope that now it will be easier to understand how to use Cel.

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,6 +32,21 @@ scripting language is too resource intensive.
 A dashboard that shows results of cel-go conformance tests can be found
 [here](https://k8s-testgrid.appspot.com/google-cel#cel-go).
 
+---
+
+* [Overview](#overview)
+  * [Environment Setup](#environment-setup)
+  * [Parse and Check](#parse-and-check)
+  * [Macros](#macros)
+  * [Evaluate](#evaluate)
+  * [Errors](#Errors)
+* [Examples](examples/README.md)
+* [Install](#install)
+* [Common Questions](#common-questions)
+* [License](#license)
+
+---
+
 ## Overview
 
 Determine the variables and functions you want to provide to CEL. Parse and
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -206,6 +219,110 @@ Both the parsed and checked expressions contain source position information
 about each node that appears in the output AST. This information can be used
 to determine error locations at evaluation time as well.
 
+## Examples
```
TristonianJones: Thanks for trying to make it easier to get started with CEL!

As a quick note on style for markdown, please makes sure the text is within 80 characters. Code examples aren't subject to this character limit, but hopefully most of the examples fit within that limit anyway. 

With respect to the examples, I'd prefer the code be the source of truth since it's tested on PR publish and submit, but there is a definite place for codelab content. Perhaps this could be the first wiki entry instead of the main README?

TristonianJones (changes requested):

README.md
tobiaszheller: I am afraid that right now file with examples is mixed with both examples and test what makes is not readable, especially for newcomers. 
Moreover I have wanted to extend also macros examples section so this is good time to think where to put it.

Maybe instead of wiki we can make `_examples` directory where we will put both source code examples and REAMDE? We can link it in main REAMDE.

TristonianJones: An examples directory sounds good. If you wouldn't mind structuring the changes that way, I'd appreciate it.

tobiaszheller: @TristonianJones I have moved examples from `cel_tests.go` to `examples` dir and created new readme there. I think it is more readable but right now we won't have examples in Godoc. 🤔 Do you have any suggestions?

TristonianJones: Ah, so just to be clear, I don't mind if the examples are left in the `cel_tests.go` _and_ replicated into an `examples` dir within markdown files, since I expect the examples will be added to over time and contain human-friendly explanations of what's going on.

tobiaszheller: So should I put also source code of examples into examples dir or left it only in cel_test.go?

TristonianJones: I think it's reasonable to copy the examples into the examples dir alongside the documentation you're providing. Thanks @tobiaszheller

tobiaszheller: PTAL, there is now Readme in examples dir along with source of examples and table of contents in main readme which links into examples.

TristonianJones (approved):

PR #298, opened 2020-01-08
Update conformance tests and fix bugs.

JimLarson:
Fixes bugs uncovered by new conformance tests for standard functions.


JimLarson:
Details:

- Checker representation of the type-of-types now a valid proto message.
- Remove "dyn" as a runtime type.
- Add "dyn" type converter as a polymorphic identity function at runtime.
- Strings must be valid unicode.
- Numeric conversions generally throw range errors.
- Logical not does not apply to numbers, negation does not apply to bools.

WORKSPACE
```diff
@@ -36,7 +36,7 @@ go_repository(
 
 git_repository(
   name = "com_google_cel_spec",
-  commit = "3de74f02b6b58fba89a17c4ea0fe32efd560795c", #PR #95
+  commit = "bbd23896f55dc69b7360df5a676e5955bd919c34", # PR #99
   remote = "https://github.com/google/cel-spec.git",
 )
```
TristonianJones, line 42: Don't forget to update the `go.mod` and `go.sum` files as well: https://blog.golang.org/using-go-modules

TristonianJones (approved):

PR #303, opened 2020-01-14
Convert Ast, Env to concrete structs rather than top-level interfaces.

### LINKED ISSUE #301: Remove superfluous interfaces from top-level API
**Describe the bug**
Some of the top-level interfaces on the API would be better replaced by structs since they are not intended for client-side implementation and reflect core assumptions of the Go libraries.

**To Reproduce**

- [X]  parser
- [X]  checker
- [X]  interpreter

**Expected behavior**

**Additional context**

TristonianJones:
Replaces the top-level `Ast` and `Env` interfaces with structs. 

These values were never intended to be implemented by clients, but rather are
core object representations which CEL evaluation depends upon. 

Note: this is a breaking change since the signature of many methods has changed
from interface values for the `Ast` and `Env` to pointers: `*Ast`, `*Env`.

Closes #301

### DIFF
--- a/cel/env.go
+++ b/cel/env.go
@@ -34,67 +34,59 @@ type Source interface {
 	common.Source
 }
 
-// Ast interface representing the checked or unchecked expression, its source, and related metadata
-// such as source position information.
-type Ast interface {
-	// Expr returns the proto serializable instance of the parsed/checked expression.
-	Expr() *exprpb.Expr
+// Ast representing the checked or unchecked expression, its source, and related metadata such as
+// source position information.
+type Ast struct {
+	expr    *exprpb.Expr
+	info    *exprpb.SourceInfo
+	source  Source
+	refMap  map[int64]*exprpb.Reference
+	typeMap map[int64]*exprpb.Type
+}
 
-	// IsChecked returns whether the Ast value has been successfully type-checked.
-	IsChecked() bool
+// Expr returns the proto serializable instance of the parsed/checked expression.
+func (ast *Ast) Expr() *exprpb.Expr {
+	return ast.expr
+}
 
-	// ResultType returns the output type of the expression if the Ast has been type-checked,
-	// else returns decls.Dyn as the parse step cannot infer the type.
-	ResultType() *exprpb.Type
+// IsChecked returns whether the Ast value has been successfully type-checked.
+func (ast *Ast) IsChecked() bool {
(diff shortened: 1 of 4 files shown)


cel/env.go
```diff
@@ -34,156 +34,50 @@ type Source interface {
...
+// constructed from the SourceInfo.
+func (ast *Ast) Source() Source {
 	return ast.source
 }
 
-// env is the internal implementation of the Env interface.
-type env struct {
+// Env defines functions for parsing and type-checking expressions against a set of user-defined
+// constants, variables, and functions. The Env interface also defines a method for generating
+// evaluable programs from parsed and checked Asts.
```
JimLarson: Re-word this now that Env is no longer an interface.  Suggestion:

Env encapsulates the context necessary to perform parsing, type checking, or  generation of evaluable programs for different expressions.

cel/env.go
```diff
@@ -195,8 +89,39 @@ type env struct {
...
+// Extend the current environment with additional options to produce a new Env.
+func (e *Env) Extend(opts ...EnvOption) (*Env, error) {
+	ext := &Env{}
+	*ext = *e
+	return ext.configure(opts...)
+}
+
+// Check performs type-checking on the input Ast and yields a checked Ast and/or set of Issues.
+//
+// Checking has failed if the returned Issues value and its Issues.Err() value is non-nil.
```
JimLarson: s/is non-nil/are non-nil/

cel/program.go
```diff
@@ -67,7 +67,7 @@ func (ed *evalDetails) State() interpreter.EvalState {
 
 // prog is the internal implementation of the Program interface.
 type prog struct {
```
JimLarson, line 63: Do you want to de-interface-ify Program too?

Also how about Source and EvalDetails too?

JimLarson (changes requested):

cel/program.go
TristonianJones: I end up treating the `Source` as an interface elsewhere and it works out pretty well when dealing with CEL embedded in larger files since it's used to provide location information for the input.

I will de-interface `EvalDetails`. I think I had multiple implementations for `Program`, but if not I'll de-interface it too.

TristonianJones:
PTAL

JimLarson (approved):

Issue #292, opened 2019-12-19
Populate the SourceInfo with Macro replacements


TristonianJones:
**Feature request checklist**

- [X] There are no issues that match the desired change
- [X] The change is large enough it can't be addressed with a simple Pull Request
- [ ] If this is a bug, please file a [Bug Report](./ISSUE_TEMPLATE).

**Change**
When macro expansions are performed the original source expression is omitted from the `ParsedExpr` result; however, the original expression is necessary for unparsing the AST back to its original form. Add the macro replacements into the AST.

PR #308, opened 2020-01-22
Top-level support for partial state evaluation

### LINKED ISSUE #299: Ensure UnknownActivation is generated from the declared environment.
**Describe the bug**
Currently the `UnknownActivation` is a catchall object that always returns `types.UnknownSet`
when the variable name requested is not found. This causes issues where a type of the same
name as the 

**To Reproduce**

- [ ]  parser
- [ ]  checker
- [X]  interpreter

```cel
// if an UnknownActivation is used, then `int` (as a variable) would return unknown;
// shadowing of identifier names should be supported, so this toy showcases a problem where
// the shadowing is not honored.
int 
```

**Expected behavior**
The use of an `UnknownActivation` should follow identifier shadowing rules.

**Additional context**

TristonianJones:
This PR introduces top-level support for evaluating partial state and computing residual CEL expressions. 

The following partial attribute evaluation features have been exposed in the `cel` package:

* `OptPartialEval` - Enable partial attribute evaluation for a specific program.
* `PartialVars` - Provide variables, and mark variables or data within a variable's object
     member graph as unknown
* `AttributePattern` - Helper method to create a new attribute pattern which can be used
    to match unknown state for an object or some part of an object.
*  `ResidualAst` - Compute a new parsed / checked `Ast` from an original `Ast` and the
   `EvalDetails` captured from a previous evaluation.

Attribute patterns start with a variable can have an optional set of qualifier patterns which
can be any valid map key type (`string`, `int64`, `uint64`, `bool`) or wildcard. The attribute
pattern behaves similarly to a [`protobuf.FieldMask`](https://godoc.org/google.golang.org/genproto/protobuf/field_mask), but with support for array element
indexing.

Note, computing a residual does not yet work for comprehension macros and depends on the addition of source info for macro replacements (#292).  

Closes #299

### DIFF
--- a/cel/cel_test.go
+++ b/cel/cel_test.go
@@ -29,7 +29,6 @@ import (
 	"github.com/google/cel-go/common/types"
 	"github.com/google/cel-go/common/types/ref"
 	"github.com/google/cel-go/common/types/traits"
-	"github.com/google/cel-go/interpreter"
 	"github.com/google/cel-go/interpreter/functions"
 	"github.com/google/cel-go/parser"
 
@@ -600,18 +599,17 @@ func Test_EvalRecover(t *testing.T) {
 	}
 }
 
-func Test_Pruning(t *testing.T) {
+func Test_ResidualAst(t *testing.T) {
 	e, _ := NewEnv(
 		Declarations(
 			decls.NewIdent("x", decls.Int, nil),
 			decls.NewIdent("y", decls.Int, nil),
 		),
 	)
-	unkVars := e.UnknownActivation()
+	unkVars := e.UnknownVars()
 	ast, _ := e.Parse(`x < 10 && (y == 0 || 'hello' != 'goodbye')`)
 	prg, _ := e.Program(ast,
-		EvalOptions(OptTrackState),
-		PartialAttributes(),
+		EvalOptions(OptTrackState, OptPartialEval),
 	)
 	out, det, err := prg.Eval(unkVars)
 	if !types.IsUnknown(out) {
@@ -620,13 +618,67 @@ func Test_Pruning(t *testing.T) {
 	if err != nil {
 		t.Fatal(err)
 	}
-	pruned := interpreter.PruneAst(ast.Expr(), det.State())
-	ast2 := &exprpb.ParsedExpr{Expr: pruned}
-	expr, err := AstToString(ParsedExprToAst(ast2))
+	residual, err := e.ResidualAst(ast, det)
+	if err != nil {
+		t.Fatal(err)
+	}
+	expr, err := AstToString(residual)
 	if err != nil {
 		t.Fatal(err)
 	}
 	if expr != "x < 10" {
+		t.Errorf("Got expr: %s, wanted x < 10", expr)
+	}
+}
+
+func Test_ResidualAst_Complex(t *testing.T) {
+	e, _ := NewEnv(
+		Declarations(
+			decls.NewIdent("resource.name", decls.String, nil),
+			decls.NewIdent("request.time", decls.Timestamp, nil),
+			decls.NewIdent("request.auth.claims",
+				decls.NewMapType(decls.String, decls.String), nil),
+		),
+	)
+	unkVars, _ := PartialVars(
+		map[string]interface{}{
+			"resource.name": "bucket/my-bucket/objects/private",
+			"request.auth.claims": map[string]string{
+				"email_verified": "true",
+			},
+		},
+		AttributePattern("request.auth.claims").Field("email"),
+	)
+	ast, iss := e.Parse(
+		`resource.name.startsWith("bucket/my-bucket") &&
+		 bool(request.auth.claims.email_verified) == true &&
+		 request.auth.claims.email == "wiley@acme.co"`)
+	if iss != nil && iss.Err() != nil {
+		t.Fatal(iss.Err())
+	}
+	ast, iss = e.Check(ast)
+	if iss != nil && iss.Err() != nil {
+		t.Fatal(iss.Err())
+	}
+	prg, _ := e.Program(ast,
+		EvalOptions(OptTrackState, OptPartialEval),
+	)
+	out, det, err := prg.Eval(unkVars)
+	if !types.IsUnknown(out) {
+		t.Fatalf("Got %v, expected unknown", out)
+	}
+	if err != nil {
+		t.Fatal(err)
+	}
+	residual, err := e.ResidualAst(ast, det)
+	if err != nil {
+		t.Fatal(err)
+	}
+	expr, err := AstToString(residual)
+	if err != nil {
+		t.Fatal(err)
+	}
+	if expr != `request.auth.claims.email == "wiley@acme.co"` {
 		t.Errorf("Got expr: %s, wanted x < 10", expr)
 	}
 }
--- a/cel/cel_test.go
+++ b/cel/cel_test.go
@@ -650,9 +650,9 @@ func Test_ResidualAst_Complex(t *testing.T) {
 		AttributePattern("request.auth.claims").Field("email"),
 	)
 	ast, iss := e.Parse(
-		`resource.name.startsWith("bucket/my-bucket") &&
+		`resource.name.startsWith('bucket/my-bucket') &&
 		 bool(request.auth.claims.email_verified) == true &&
-		 request.auth.claims.email == "wiley@acme.co"`)
+		 request.auth.claims.email == 'wiley@acme.co'`)
 	if iss != nil && iss.Err() != nil {
 		t.Fatal(iss.Err())
 	}
@@ -678,8 +678,8 @@ func Test_ResidualAst_Complex(t *testing.T) {
 	if err != nil {
 		t.Fatal(err)
 	}
-	if expr != `request.auth.claims.email == "wiley@acme.co"` {
-		t.Errorf("Got expr: %s, wanted x < 10", expr)
+	if expr != `request.auth.claims.email == 'wiley@acme.co'` {
+		t.Errorf("Got expr: %s, wanted request.auth.claims.email == 'wiley@acme.co'", expr)
 	}
 }
 
--- a/cel/cel_test.go
+++ b/cel/cel_test.go
@@ -647,12 +647,12 @@ func Test_ResidualAst_Complex(t *testing.T) {
 				"email_verified": "true",
 			},
 		},
-		AttributePattern("request.auth.claims").Field("email"),
+		AttributePattern("request.auth.claims").QualString("email"),
 	)
 	ast, iss := e.Parse(
-		`resource.name.startsWith('bucket/my-bucket') &&
+		`resource.name.startsWith("bucket/my-bucket") &&
 		 bool(request.auth.claims.email_verified) == true &&
-		 request.auth.claims.email == 'wiley@acme.co'`)
+		 request.auth.claims.email == "wiley@acme.co"`)
 	if iss != nil && iss.Err() != nil {
 		t.Fatal(iss.Err())
 	}
@@ -678,8 +678,8 @@ func Test_ResidualAst_Complex(t *testing.T) {
 	if err != nil {
 		t.Fatal(err)
 	}
-	if expr != `request.auth.claims.email == 'wiley@acme.co'` {
-		t.Errorf("Got expr: %s, wanted request.auth.claims.email == 'wiley@acme.co'", expr)
+	if expr != `request.auth.claims.email == "wiley@acme.co"` {
+		t.Errorf("Got expr: %s, wanted request.auth.claims.email == \"wiley@acme.co\"", expr)
 	}
 }
 
--- a/cel/cel_test.go
+++ b/cel/cel_test.go
@@ -21,17 +21,17 @@ import (
 	"sync"
 	"testing"
 
+	"github.com/golang/protobuf/proto"
+	"github.com/google/cel-go/checker/decls"
 	"github.com/google/cel-go/common"
 	"github.com/google/cel-go/common/operators"
 	"github.com/google/cel-go/common/overloads"
-	"github.com/google/cel-go/parser"
-
-	"github.com/golang/protobuf/proto"
-	"github.com/google/cel-go/checker/decls"
 	"github.com/google/cel-go/common/types"
 	"github.com/google/cel-go/common/types/ref"
 	"github.com/google/cel-go/common/types/traits"
+	"github.com/google/cel-go/interpreter"
 	"github.com/google/cel-go/interpreter/functions"
+	"github.com/google/cel-go/parser"
 
 	descpb "github.com/golang/protobuf/protoc-gen-go/descriptor"
 	exprpb "google.golang.org/genproto/googleapis/api/expr/v1alpha1"
@@ -597,6 +597,37 @@ func Test_EvalRecover(t *testing.T) {
 	_, _, err = prgm.Eval(map[string]interface{}{})
 	if err.Error() != "internal error: watch me recover" {
 		t.Errorf("Got '%v', wanted 'internal error: watch me recover'", err)
+	}
+}
+
+func Test_Pruning(t *testing.T) {
+	e, _ := NewEnv(
+		Declarations(
+			decls.NewIdent("x", decls.Int, nil),
+			decls.NewIdent("y", decls.Int, nil),
+		),
+	)
+	unkVars := e.UnknownActivation()
+	ast, _ := e.Parse(`x < 10 && (y == 0 || 'hello' != 'goodbye')`)
+	prg, _ := e.Program(ast,
+		EvalOptions(OptTrackState),
+		PartialAttributes(),
+	)
+	out, det, err := prg.Eval(unkVars)
+	if !types.IsUnknown(out) {
+		t.Fatalf("Got %v, expected unknown", out)
+	}
+	if err != nil {
+		t.Fatal(err)
+	}
+	pruned := interpreter.PruneAst(ast.Expr(), det.State())
+	ast2 := &exprpb.ParsedExpr{Expr: pruned}
+	expr, err := AstToString(ParsedExprToAst(ast2))
+	if err != nil {
+		t.Fatal(err)
+	}
+	if expr != "x < 10" {
+		t.Errorf("Got expr: %s, wanted x < 10", expr)
 	}
 }
 
--- a/cel/env.go
+++ b/cel/env.go
@@ -215,12 +215,15 @@ func (e *Env) UnknownVars() interpreter.PartialActivation {
 // - Optimizing constant expression evaluations away.
 // - Indexing and pruning expressions based on known input arguments.
 // - Surfacing additional requirements that are needed in order to complete an evaluation.
-// - Sharing the evaluation of an expression across multiple processes
+// - Sharing the evaluation of an expression across multiple machines/nodes.
 //
 // For example, if an expression targets a 'resource' and 'request' attribute and the possible
 // values for the resource are known, a PartialActivation could mark the 'request' as an unknown
 // interpreter.AttributePattern and the resulting ResidualAst would be reduced to only the parts
 // of the expression that reference the 'request'.
+//
+// Note, the expression ids within the residual AST generated through this method have no
+// correlation to the expression ids of the original AST.
 //
 // See the PartialVars helper for how to construct a PartialActivation.
 func (e *Env) ResidualAst(a *Ast, details *EvalDetails) (*Ast, error) {
--- a/cel/env.go
+++ b/cel/env.go
@@ -226,6 +226,9 @@ func (e *Env) UnknownVars() interpreter.PartialActivation {
 // correlation to the expression ids of the original AST.
 //
 // See the PartialVars helper for how to construct a PartialActivation.
+//
+// TODO: Consider adding an option to generate a Program.Residual to avoid round-tripping to an
+// Ast format and then Program again.
 func (e *Env) ResidualAst(a *Ast, details *EvalDetails) (*Ast, error) {
 	pruned := interpreter.PruneAst(a.Expr(), details.State())
 	expr, err := AstToString(ParsedExprToAst(&exprpb.ParsedExpr{Expr: pruned}))
--- a/cel/env.go
+++ b/cel/env.go
@@ -189,6 +189,10 @@ func (e *Env) TypeProvider() ref.TypeProvider {
 
 // UnknownActivation returns an interpreter.PartialActivation which marks all variables
 // declared in the Env as unknown AttributePattern values.
+//
+// Note, the UnknownActivation() will behave the same as an interpreter.EmptyActivation()
+// unless the cel.CustomAttributeFactory option is set to a PartialAttributeFactory
+// implementation.
 func (e *Env) UnknownActivation() interpreter.PartialActivation {
 	var unknownPatterns []*interpreter.AttributePattern
 	for _, d := range e.declarations {
--- a/cel/env.go
+++ b/cel/env.go
@@ -23,6 +23,7 @@ import (
 	"github.com/google/cel-go/common/packages"
 	"github.com/google/cel-go/common/types"
 	"github.com/google/cel-go/common/types/ref"
+	"github.com/google/cel-go/interpreter"
 	"github.com/google/cel-go/interpreter/functions"
 	"github.com/google/cel-go/parser"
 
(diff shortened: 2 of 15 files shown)


cel/env.go
```diff
@@ -186,6 +187,62 @@ func (e *Env) TypeProvider() ref.TypeProvider {
...
+
+// ResidualAst takes an Ast and its EvalDetails to produce a new Ast which only contains the
+// attribute references which are unknown.
+//
+// Residual expressions are beneficial in a few scenarios:
+//
+// - Optimizing constant expression evaluations away.
+// - Indexing and pruning expressions based on known input arguments.
+// - Surfacing additional requirements that are needed in order to complete an evaluation.
+// - Sharing the evaluation of an expression across multiple processes
```
JimLarson: Consider changing to "Sharing the evaluation of an expression across concurrent computations."  The expression could, for example, be evaluated across multiple threads in the same process.  Or if you want to highlight that it could be a distributed implementation, then "across multiple machines/nodes".

cel/env.go
```diff
@@ -186,6 +187,62 @@ func (e *Env) TypeProvider() ref.TypeProvider {
...
+//
+// For example, if an expression targets a 'resource' and 'request' attribute and the possible
+// values for the resource are known, a PartialActivation could mark the 'request' as an unknown
+// interpreter.AttributePattern and the resulting ResidualAst would be reduced to only the parts
+// of the expression that reference the 'request'.
+//
+// See the PartialVars helper for how to construct a PartialActivation.
```
JimLarson, line 228: Add a note that the expression IDs of the residual AST have no correlation to the expression IDs of the original AST.

cel/cel_test.go
```diff
@@ -600,6 +599,90 @@ func Test_EvalRecover(t *testing.T) {
...
+	}
+	residual, err := e.ResidualAst(ast, det)
+	if err != nil {
+		t.Fatal(err)
+	}
+	expr, err := AstToString(residual)
+	if err != nil {
+		t.Fatal(err)
+	}
+	if expr != "x < 10" {
```
JimLarson, line 629: I'm a little worried that we're setting precedent that the results of AstToString are stable to the level of whitespace.

cel/program.go
```diff
@@ -48,6 +48,37 @@ func NoVars() interpreter.Activation {
 	return interpreter.EmptyActivation()
 }
 
+// PartialVars returns a PartialActivation which contains variables and a set of AttributePattern
+// values that indicate variables or parts of variables whose value are not yet known.
```
JimLarson, line 52: As always, indicate what sorts of things "vars" could be.

cel/cel_test.go
```diff
@@ -600,6 +599,90 @@ func Test_EvalRecover(t *testing.T) {
...
+	residual, err := e.ResidualAst(ast, det)
+	if err != nil {
+		t.Fatal(err)
+	}
+	expr, err := AstToString(residual)
+	if err != nil {
+		t.Fatal(err)
+	}
+	if expr != `request.auth.claims.email == "wiley@acme.co"` {
+		t.Errorf("Got expr: %s, wanted x < 10", expr)
```
JimLarson: wanted request.auth.claims.email == "wiley@acme.co"

cel/program.go
```diff
@@ -48,6 +48,37 @@ func NoVars() interpreter.Activation {
...
+	unknowns ...*interpreter.AttributePattern) (interpreter.PartialActivation, error) {
+	return interpreter.NewPartialActivation(vars, unknowns...)
+}
+
+// AttributePattern represents a top-level variable with an optional set of qualifier patterns.
+//
+// The variable name must always be a string, and may be a qualified name according to the CEL
+// namespacing conventions, e.g. 'ns.app.a'.
+//
+// The qualifier patterns for attribute matching must be one of the following:
```
JimLarson: This suggests that you can set qualifiers here, which isn't the case.  Either expose the qualifier API, or take the remainder out.

cel/program.go
```diff
@@ -48,6 +48,37 @@ func NoVars() interpreter.Activation {
...
 }
 
+// PartialVars returns a PartialActivation which contains variables and a set of AttributePattern
+// values that indicate variables or parts of variables whose value are not yet known.
+func PartialVars(vars interface{},
+	unknowns ...*interpreter.AttributePattern) (interpreter.PartialActivation, error) {
+	return interpreter.NewPartialActivation(vars, unknowns...)
+}
+
+// AttributePattern represents a top-level variable with an optional set of qualifier patterns.
```
JimLarson: This constructor doesn't allow the optional qualifier patterns.

cel/program.go
```diff
@@ -106,6 +133,13 @@ func newProgram(e *Env, ast *Ast, opts ...ProgramOption) (Program, error) {
 		}
 	}
 
+	// Set the attribute factory after the options have been set.
+	if p.evalOpts&OptPartialEval == OptPartialEval {
```
JimLarson, line 136: Spaces around "&", please.  Did you run gofmt?

interpreter/activation.go
```diff
@@ -132,19 +131,14 @@ func NewHierarchicalActivation(parent Activation, child Activation) Activation {
...
-	return a
-}
-
-// PartialActivation returns an Activation that will resolve identifier names if present, otherwise
-// will return 'types.Unknown'.
+// NewPartialActivation returns an Activation which contains a list of AttributePattern values
+// representing field and index operations that should result in a 'types.Unknown' result.
 //
-// The input `bindings` may either be of type `Activation` or `map[string]interface{}`.
+// The input `bindings` may either be of type `Activation` or `map[string]interface{}`. The
```
JimLarson: What things can the map values be? (It would be great if you had a single place that names and defines this range of types.)

interpreter/attribute_patterns.go
```diff
@@ -0,0 +1,350 @@
...
+func NewAttributePattern(variable string) *AttributePattern {
+	return &AttributePattern{
+		variable:          variable,
+		qualifierPatterns: []*AttributeQualifierPattern{},
+	}
+}
+
+// Field adds a string qualifier pattern to the AttributePattern. The string may be a valid
+// identifier, or string map key including empty string.
+func (apat *AttributePattern) Field(pattern string) *AttributePattern {
```
JimLarson: Consider more consistent naming, e.g. IndexString or QualString.

interpreter/attribute_patterns.go
```diff
@@ -0,0 +1,350 @@
...
+	apat.qualifierPatterns = append(apat.qualifierPatterns,
+		&AttributeQualifierPattern{value: pattern})
+	return apat
+}
+
+// Index adds an int qualifier pattern to the AttributePattern. The index may be either a map or
+// list index.
+func (apat *AttributePattern) Index(pattern int64) *AttributePattern {
```
JimLarson: Consider naming IndexInt instead.

interpreter/attribute_patterns.go
```diff
@@ -0,0 +1,350 @@
...
+
+// IndexBool adds a bool qualifier pattern for a map index operation to the AttributePattern.
+func (apat *AttributePattern) IndexBool(pattern bool) *AttributePattern {
+	apat.qualifierPatterns = append(apat.qualifierPatterns,
+		&AttributeQualifierPattern{value: pattern})
+	return apat
+}
+
+// Wildcard adds a special sentinel qualifier pattern that indicates any value will yeild a
+// qualifier match.
```
JimLarson: Specify whether the wildcard matches a single level or an arbitrary sequence of levels.

interpreter/attribute_patterns.go
```diff
@@ -0,0 +1,350 @@
...
+func (apat *AttributePattern) Wildcard() *AttributePattern {
+	apat.qualifierPatterns = append(apat.qualifierPatterns,
+		&AttributeQualifierPattern{wildcard: true})
+	return apat
+}
+
+// Matches returns true if the variable matches the AttributePattern variable.
+func (apat *AttributePattern) Matches(variable string) bool {
```
JimLarson: Consider "VariableMatches", since it could have qualifiers.  Also specify whether "c" matches "a.b.c" when the namespace is "a.b".

interpreter/attribute_patterns.go
```diff
@@ -0,0 +1,350 @@
...
+
+	"github.com/google/cel-go/common/packages"
+	"github.com/google/cel-go/common/types"
+	"github.com/google/cel-go/common/types/ref"
+)
+
+// AttributePattern represents a top-level variable with an optional set of qualifier patterns.
+//
+// The variable name must always be a string, and may be a qualified name according to the CEL
+// namespacing conventions, e.g. 'ns.app.a'.
```
JimLarson: Clarify whether a variable, qualified or not, will go through the namespace/variable resolution.  E.g. if the namespace is "a.b", does an attribute pattern with variable "c" possibly match a.b.c?  Or is it expected to be the fully-qualified name?

interpreter/attribute_patterns.go
```diff
@@ -0,0 +1,350 @@
...
+	return ok && qve.QualifierValueEquals(qpat.value)
+}
+
+// qualifierValueEquator defines an interface for determining if an input value, of valid map key
+// type, is equal to the value held in the Qualifier. This interface is used by the
+// AttributeQualifierPattern to determine pattern matches for non-wildcard qualifier patterns.
+//
+// Note: Attribute values are also Qualifier values; however, Attriutes are resolved before
+// qualification happens. This is an implementation detail, but one relevant to why the Attribute
+// types do no surface in the list of implementations.
```
JimLarson: "do not"

interpreter/attribute_patterns.go
```diff
@@ -0,0 +1,350 @@
...
+	qve, ok := q.(qualifierValueEquator)
+	return ok && qve.QualifierValueEquals(qpat.value)
+}
+
+// qualifierValueEquator defines an interface for determining if an input value, of valid map key
+// type, is equal to the value held in the Qualifier. This interface is used by the
+// AttributeQualifierPattern to determine pattern matches for non-wildcard qualifier patterns.
+//
+// Note: Attribute values are also Qualifier values; however, Attriutes are resolved before
+// qualification happens. This is an implementation detail, but one relevant to why the Attribute
```
JimLarson, line 129: At the code location that ensures that "attributes are resolved before qualification happens", make a reference back to this comment, so that if that guarantee ever changes we'll know to revisit this section.

interpreter/attribute_patterns.go
```diff
@@ -0,0 +1,350 @@
...
+		},
+		adapter:  fac.adapter,
+		provider: fac.provider,
+		fac:      fac,
+	}
+}
+
+// matchesUnknownPatterns returns true if the variable names and qualifiers for a given
+// Attribute value match any of the ActivationPattern objects in the set of unknown activation
+// patterns on the given PartialActivation.
```
JimLarson, line 216: Please give an example here.  I understand the words individually, but I don't understand when they're put all together.

interpreter/attribute_patterns.go
```diff
@@ -0,0 +1,350 @@
...
+func (fac *partialAttributeFactory) matchesUnknownPatterns(
+	vars PartialActivation,
+	attrID int64,
+	variableNames []string,
+	qualifiers []Qualifier) (types.Unknown, error) {
+	patterns := vars.UnknownAttributePatterns()
+	candIndices := map[int]struct{}{}
```
JimLarson: Please expand "cand" to "candidate", at least in a comment.

interpreter/attribute_patterns.go
```diff
@@ -0,0 +1,350 @@
...
+type attributeMatcher struct {
+	NamespacedAttribute
+	qualifiers []Qualifier
+	fac        *partialAttributeFactory
+}
+
+// AddQualifier implements the Attribute interface method which adds the Qualifier onto the
+// underlying NamespacedAttribute as well as tracks the Qualifier in internal storage. The
+// double accounting for the Qualifier values is to assist with AttributePattern matching in
+// the TryResolve call.
```
JimLarson: Please add an example - I'm not getting the "double accounting".

interpreter/attribute_patterns.go
```diff
@@ -0,0 +1,350 @@
...
+
+// Resolve is an implementation of the Attribute interface method which uses the
+// attributeMatcher TryResolve implementation rather than the embedded NamespacedAttribute
+// Resolve implementation.
+func (m *attributeMatcher) Resolve(vars Activation) (interface{}, error) {
+	obj, found, err := m.TryResolve(vars)
+	if err != nil {
+		return nil, err
+	}
+	if found {
```
JimLarson: Consider flipping the test to get the error return indented instead of the "normal" return.

interpreter/attribute_patterns.go
```diff
@@ -0,0 +1,350 @@
...
+			}
+		}
+		if isUnk {
+			return types.Unknown{matchExprID}, nil
+		}
+	}
+	return nil, nil
+}
+
+type attributeMatcher struct {
```
JimLarson, line 309: Add a comment that this will be an Attribute, and why.

interpreter/attribute_patterns_test.go
```diff
@@ -0,0 +1,284 @@
...
+
+type attr struct {
+	maybe bool
+	name  string
+	quals []interface{}
+}
+
+type patternTest struct {
+	pattern *AttributePattern
+	pkg     string
```
JimLarson: I think we're calling this "container".

interpreter/attribute_patterns_test.go
```diff
@@ -0,0 +1,284 @@
...
+			{name: "ns.var"},
+		},
+	},
+	"var_namespace": {
+		pattern: NewAttributePattern("ns.app.var"),
+		pkg:     "ns.app",
+		matches: []attr{
+			{name: "ns.app.var"},
+			{name: "ns.app.var", quals: []interface{}{int64(0)}},
+			{name: "ns", quals: []interface{}{"app", "var", "foo"}, maybe: true},
```
JimLarson: [Cel-spec issue 53](https://github.com/google/cel-spec/issues/53) calls for variable `ns` to be disallowed if any `ns.app.*` variable is defined.  This isn't _exactly_ the case here, but if we're working in container `ns.app`, we should probably view a variable `ns` with suspicion, if not disallow it outright.

interpreter/attribute_patterns_test.go
```diff
@@ -0,0 +1,284 @@
...
+		},
+		misses: []attr{
+			{name: "ns.var"},
+			{name: "ns", quals: []interface{}{"var"}, maybe: true},
+		},
+	},
+	"var_field": {
+		pattern: NewAttributePattern("var").Field("field"),
+		matches: []attr{
+			{name: "var"},
```
JimLarson, line 86: Make sure that it's clear in the comments in the `cel` package that pattern `a.b` also matches `a` but not `a.c`.

interpreter/attributes.go
```diff
@@ -31,7 +31,7 @@ type AttributeFactory interface {
 	// AbsoluteAttribute creates an attribute that refers to a top-level variable name.
```
JimLarson, line 31: Refer to or copy the description of CandidateVariableNames below.

interpreter/attributes.go
```diff
@@ -81,6 +81,26 @@ type Attribute interface {
 	Resolve(Activation) (interface{}, error)
 }
 
+// NamespacedAttribute values are a variable within a namespace, and an optional set of qualifiers
+// such as field, key, or index accesses.
+type NamespacedAttribute interface {
+	Attribute
+
+	// CandidateVariableNames returns the possible namespaced variable names for this Attribute in
+	// the CEL namespace resolution order.
```
JimLarson, line 97: The type suggests this can be a sequence of arbitrary strings, but that's not what your'e expecting.  Consider changing this to something that represents only what you want, e.g. `[]string` where it represents just the components of the FQVN, or use type abstraction.

JimLarson (changes requested):

interpreter/attribute_patterns_test.go
TristonianJones: Ah, this is a `MaybeAttribute` which represents an unchecked expression whose checked form might be any one of the following:

1. `id: ns.app.var.foo`
2. `select(id: ns.app.var, field: 'foo')`
3. `select(select(id: ns.app, field: 'var'), field: 'foo')`
4. `select(select(select(id: ns, field: 'app'), field: 'var'), field: 'foo')`

The test ensures that even parsed only attributes will match the pattern correctly.

cel/program.go
TristonianJones: I think it was the `gofmt` that actually did this. I had spaces and they disappeared.

cel/program.go
TristonianJones: `AddQualifierPattern` is a method exposed on the `AttributePattern` directly, so I'm not sure it needs additional publication. I opted to try an give a high-level overview of how to use `AttributePattern`, but I could refer to the `AttributePattern` docs instead.

cel/cel_test.go
TristonianJones: Some time back, I normalized the whitespace handling in the `AstToString` call so that it now formats the expression to a single line.

cel/env.go
TristonianJones: I'll emphasize the latter angle. I don't want to give people the idea that the same input could be evaluated in parallel over multiple threads where CEL somehow shards branch evaluations. That would be cool, but we're not really there.

interpreter/attributes.go
TristonianJones: Unfortunately, there is more than one candidate variable name, so this isn't the components of the fully qualified variable name, it's the set of possible FQVN values which could be produced from a parse-only expression.

TristonianJones:
PTAL

cel/program.go
```diff
@@ -50,31 +50,30 @@ func NoVars() interpreter.Activation {
...
+// string field or index qualification `b`. This pattern will match Attributes `a`, and `a.b`,
+// but not `a.c`.
 //
-//   1. ns.myvar["complex-value"]
-//   2. ns.myvar["complex-value"][0]
-//   3. ns.myvar["complex-value"].*.name
+// When using a CEL expression within a container, e.g. a package or namespace, the variable name
+// in the pattern must match the qualified name produced during the variable namespace resolution.
+// For example, when variable `a` is declared within an expression whose container is `ns.app`, the
+// fully qualified variable name may be `ns.app.a`, `ns.a`, or `a` per the CEL namespace resolution
```
JimLarson, line 71: The term "fully qualified" generally means only the longest expansion, e.g. "fully qualified domain name" means "domains.google.com", not one of the prefixes of that string.

interpreter/activation.go
```diff
@@ -134,20 +134,8 @@ func NewHierarchicalActivation(parent Activation, child Activation) Activation {
...
-// Lazy bindings may be supplied in either of the following forms:
-// - func() interface{}
-// - func() ref.Val
-//
-// The output of the lazy binding will overwrite the variable reference in the internal map.
-//
-// Values which are not represented as ref.Val types on input may be adapted to a ref.Val using
-// the ref.TypeAdapter configured in the environment.
+// The `vars` value may be any value type supported by the interpreter.NewActivation call,
+// but is typically either an existing Activation or map[string]interface{}.
```
JimLarson, line 138: There doesn't seem to be a `vars` here.

interpreter/attribute_patterns.go
```diff
@@ -24,8 +24,11 @@ import (
 
 // AttributePattern represents a top-level variable with an optional set of qualifier patterns.
 //
-// The variable name must always be a string, and may be a qualified name according to the CEL
-// namespacing conventions, e.g. 'ns.app.a'.
+// When using a CEL expression within a container, e.g. a package or namespace, the variable name
+// in the pattern must match the qualified name produced during the variable namespace resolution.
+// Variable names provided on input to the AttributePattern are taken at face value. For example,
```
JimLarson: Consider describing this as the "fully qualified variable name" instead of "taken at face value".

interpreter/attribute_patterns_test.go
JimLarson: So it seems that the container setting isn't really relevant to the test - maybe take it out?  Or make it relevant by dropping the "ns.app" prefix from one of the cases?

But again with Issue 53:  it seems like we're declaring that "ns.app.var" is a real variable.  We might decide that this forbids us from having "ns" or "ns.app" as variables.

At least put in the following comment:
```
// Note that https://github.com/google/cel-spec/issues/53 complains about
// variable names that are a dotted prefix of other qualified variable names,
// so this tests some scenarios which might be disallowed in the future.
```

cel/program.go
TristonianJones: Whether `a` is declared globally or within `ns.app` is something that the expression itself does not indicate, but rather is determined by the variable declaration. It's possible the fully qualified declaration name for the identifier `a` is `ns.app.a`, `ns.a`, or `a` depending on what the declaration says. Am I missing something in the feedback here?

interpreter/attribute_patterns_test.go
TristonianJones: However, this test is not exercising subsumption, but rather demonstrating that maybe resolution considers the container when attempting to match a parsed-only expression against the attribute pattern. In the examples I typed out earlier, maybe option number 2 is the one that matches.

interpreter/attribute_patterns.go
TristonianJones: Done.

cel/cel_test.go
JimLarson: My concern is that if we ever make a tweak, we might break not just our tests (bearable), but 3rd-party tests too.  That being said, if you're not trying to be fancy in the formatting, it's probably an acceptable risk. Maybe add a comment in AstToString that it's expected to be stable.

cel/program.go
JimLarson: Ah, I was misreading the comment - it's saying that `a` might refer to 3 different variables, not that one variable has 3 different "fully-qualified" names.  Have the next sentence start "Pick the _fully_ qualified variable name that makes sense...".

interpreter/attribute_patterns.go
```diff
@@ -0,0 +1,382 @@
...
+}
+
+// Wildcard adds a special sentinel qualifier pattern that will match any single qualifier.
+func (apat *AttributePattern) Wildcard() *AttributePattern {
+	apat.qualifierPatterns = append(apat.qualifierPatterns,
+		&AttributeQualifierPattern{wildcard: true})
+	return apat
+}
+
+// VariableMatches returns true if the variable matches the AttributePattern variable.
```
JimLarson: Indicate that the `variable` argument must be fully-qualified.

JimLarson (changes requested):

cel/env.go
```diff
@@ -186,6 +187,65 @@ func (e *Env) TypeProvider() ref.TypeProvider {
...
+// For example, if an expression targets a 'resource' and 'request' attribute and the possible
+// values for the resource are known, a PartialActivation could mark the 'request' as an unknown
+// interpreter.AttributePattern and the resulting ResidualAst would be reduced to only the parts
+// of the expression that reference the 'request'.
+//
+// Note, the expression ids within the residual AST generated through this method have no
+// correlation to the expression ids of the original AST.
+//
+// See the PartialVars helper for how to construct a PartialActivation.
+func (e *Env) ResidualAst(a *Ast, details *EvalDetails) (*Ast, error) {
```
JimLarson, line 232: What do you think about adding an analogous `ResidualProgram` standalone function?  I'd like to be able to compute a residual of the higher-level graph-of-Programs without peeking inside the Program abstraction.

```
func ResidualProgram(p Program, details *EvalDetails) (Program, error)
```

cel/cel_test.go
TristonianJones: I'll add a comment to this effect. For fancy formatting, I think they'll be `FormatOptions` passed to the `AstToString()` call

cel/env.go
TristonianJones: It's worth considering. I'll add a TODO since I think I'd want to generate the residual program a little bit differently than how the `intepreter.PruneAst` is doing it now. Might be a good opportunity to consolidate program generation logic and allow for the creation of an `Ast` from a `Program` rather than just the other way around.

cel/program.go
TristonianJones: Done.

interpreter/attribute_patterns.go
TristonianJones: Done.

interpreter/attribute_patterns_test.go
TristonianJones: I've added a comment about how 'maybe' fits into the test. Perhaps I should describe the test-case a bit differently.

TristonianJones (commented):
PTAL. I've updated the comments and added TODOs

cel/cel_test.go
JimLarson: That would be a great solution, keeping the default behavior stable.

JimLarson (approved):

interpreter/attribute_patterns_test.go
JimLarson: Per our discussion, I see that the matching "ns" plus qualifiers doesn't imply a variable named "ns", since the "unchecked" bit is set, so no warning vs Issue 53 is needed.

PR #310, opened 2020-02-05
Use latest conformance tests and driver.

JimLarson:
Seems to fix flakiness for tests with embedded Any fields.


TristonianJones (commented):
LGTM.

Issue #306, opened 2020-01-16
FR: Built-in substring functions


wlynch:
**Feature request checklist**

- [x] There are no issues that match the desired change
- [x] The change is large enough it can't be addressed with a simple Pull Request
- [x] If this is a bug, please file a [Bug Report](./ISSUE_TEMPLATE).

**Change**
It would be great to see built-in substrings in CEL to complement the existing concatenation and matching.

This is something we're adding ourselves as a custom definition in [our own project](https://github.com/tektoncd/triggers/pull/329), but it'd be great to see something like this upstream!

**Example**
Given a string `s = "0123456789"`:

Expression | Evaluation
--------------- | --------------
s[0] | "0"
s[:5] | "01234"
s[5:] | "56789"
s [1:3] | "12"

**Alternatives considered**
n/a

TristonianJones:
Hi Billy,

I appreciate the feature request. Thanks for using CEL!

With respect to custom syntax for sub-ranges, I'm a fan of it in more powerful languages, but I'm reluctant to add support for it in CEL. The inclusion of such functionality at the syntax level implies that certain, potentially expensive computations, can't be disabled. It's actually for this reason that we're planning string and list concatenation out of the default `+` operator and into their own `concat` (or similar) functions. Ideally, the operator implies the same or similar cpu and memory cost no matter the operand types. 

I would like to make it easier to enable optional modules of functionality for things like string manipulation, time computations, and list/map manipulation. They're not really part of the core language, but the should be simple to enable.

One model I've been considering is to create module like so:

```go
type Module interface {
   CompileOptions() []EnvOption
   ProgramOptions() []ProgramOption
}
```

I would also add a top-level `cel.Compile` function that accepts `Module` values as inputs and perform the proper bindings. For anyone who wants to break apart the parse / check from eval, you could still use the `Module` members directly.

Thoughts?

TristonianJones:
@wlynch Curious to hear your thoughts on the suggested approach when you get a chance.

wlynch:
A module based approach sounds great! I think that would help us bundle our own custom functions for reuse across our own project.

Would this basically be a wrapper around the existing Declarations/ProgramOptions so that they only need to be declared once? e.g. something like:

```go 
env := cel.NewEnv(cel.StringModule(), cel.TimeModule(), ...)
...
p, err := env.Program(checked)
p.Eval(data)
```

TristonianJones:
Something to that effect, yes.

```go
[...]

Issue #290, opened 2019-12-18
Make subsetting the standard environment simple


TristonianJones:
**Feature request checklist**

- [X] There are no issues that match the desired change
- [X] The change is large enough it can't be addressed with a simple Pull Request
- [ ] If this is a bug, please file a [Bug Report](./ISSUE_TEMPLATE).

**Change**
Currently it is easy to extend a standard environment, but painful to subset it. Effectively, the user must copy and paste code from the supported library which is less than ideal.

Make it possible to specify in terms of types and operators which functionality they would like to support and ensure that only this subset is exposed at check-time and evaluation time.

TristonianJones:
Custom environments are difficult to construct since its non-trivial to subset the standard CEL environment. In addressing this feature request an example, as requested in #344, should be added to the `cel_test.go` file.

TristonianJones:
Note, https://github.com/google/cel-policy-templates-go/pull/14 permits the specification of a CEL environment via YAML and would be a good reference for supporting environment subsetting.

PR #311, opened 2020-02-23
Add Library interface and top-level Compile method

TristonianJones:
There are a handful of usability issues which have been pointed out in #290 and #306 which
relate to how CEL programs are configured and specified. The idea of this change is to offer
a path forward where it is easier to specify a subset of the CEL standard environment _and_
easier to build new modules of functionality as `cel.Library` objects.

### DIFF
--- a/cel/cel_test.go
+++ b/cel/cel_test.go
@@ -56,11 +56,7 @@ func Example() {
 	}
 
 	// Parse and check the expression.
-	p, iss := e.Parse("i.greet(you)")
-	if iss != nil && iss.Err() != nil {
-		log.Fatalln(iss.Err())
-	}
-	c, iss := e.Check(p)
+	ast, iss := e.Compile("i.greet(you)")
 	if iss != nil && iss.Err() != nil {
 		log.Fatalln(iss.Err())
 	}
@@ -73,7 +69,7 @@ func Example() {
 				return types.String(
 					fmt.Sprintf("Hello %s! Nice to meet you, I'm %s.\n", rhs, lhs))
 			}})
-	prg, err := e.Program(c, funcs)
+	prg, err := e.Program(ast, funcs)
 	if err != nil {
 		log.Fatalf("program creation error: %s\n", err)
 	}
@@ -114,11 +110,7 @@ func Example_globalOverload() {
 	}
 
 	// Parse and check the expression.
-	p, iss := e.Parse(`shake_hands(i,you)`)
-	if iss != nil && iss.Err() != nil {
-		log.Fatalln(iss.Err())
-	}
-	c, iss := e.Check(p)
+	ast, iss := e.Compile(`shake_hands(i,you)`)
 	if iss != nil && iss.Err() != nil {
 		log.Fatalln(iss.Err())
 	}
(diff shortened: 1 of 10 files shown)


cel/cel_test.go
```diff
@@ -198,24 +186,23 @@ func Test_ExampleWithBuiltins(t *testing.T) {
...
+func Test_CustomEnv(t *testing.T) {
+	e, _ := NewCustomEnv(
 		Declarations(decls.NewIdent("a.b.c", decls.Bool, nil)))
 
 	t.Run("err", func(t *testing.T) {
-		p, _ := e.Parse("a.b.c == true")
-		_, iss := e.Check(p)
+		_, iss := e.Compile("a.b.c == true")
 		if iss == nil || iss.Err() == nil {
 			t.Error("Got successful check, expected error for missing operator '_==_'")
```
JimLarson: Change "Got successful check" to "got successful compile", here and below.  (Go stylists are recommending that error strings are not capitalized.)

cel/cel_test.go
```diff
@@ -269,12 +252,11 @@ func Test_HomogeneousAggregateLiterals(t *testing.T) {
 		},
 	})
 	t.Run("ok_list", func(t *testing.T) {
-		p, _ := e.Parse("name in ['hello', 'world']")
-		c, iss := e.Check(p)
+		ast, iss := e.Compile("name in ['hello', 'world']")
 		if iss != nil && iss.Err() != nil {
 			t.Fatalf("Got issue: %v, expected successful check.", iss.Err())
```
JimLarson: Change to "got issue, expected successful compile".

JimLarson (approved):

cel/env.go
```diff
@@ -138,6 +171,47 @@ func (e *Env) Check(ast *Ast) (*Ast, *Issues) {
...
+// associated issues.
+//
+// If an error is encountered during parsing the CompileSource step will not continue with the
+// Check phase. If non-error issues are encountered during Parse, they may be combined with any
+// issues discovered during Check.
+//
+// Note, for parse-only uses of CEL use Parse.
+func (e *Env) CompileSource(src common.Source) (*Ast, *Issues) {
+	ast, iss := e.ParseSource(src)
+	if iss != nil && iss.Err() != nil {
```
tobiaszheller, line 196: what do you think about changing how method `Err()` on issues works? maybe it could work is similar way as grpc getters.
if issues is nil, return nil.

TristonianJones: I hear you, and while there really isn't any code to show why there is an `Issues` object instead of an `errors` object, it's because we intend to surface deprecation notices through the `Issues` object. There are a couple upcoming standard library changes which will likely warrant a deprecation noticed, but basically even if you're using something deprecated it will still work, it'll just point out that you're using something deprecated and should do something else instead.

tobiaszheller: @TristonianJones I am not sure if you understand me well. 
I understand that you want to return typed Issues. 
However I still think that checking issues could be simplified:
```
func (i *Issues) Err() error {
	if i == nil {
		return nil
	}
	if len(i.errs.GetErrors()) > 0 {
		return errors.New(i.errs.ToDisplayString())
	}
	return nil
}
```
using that approach consumers of API could simply check
```
if iss.Err() != nil {}
```

TristonianJones: Ah, okay, I get it now. I was unfortunately over-indexed on prior feedback of "why not just an error?". I can return an empty issues list when there are no errors. I'll write a PR for it later this week hopefully.

PR #312, opened 2020-02-26
Issues simplification for Parse / Check / Compile 

TristonianJones:
This change ensures the `*Issues` response is always non-nil from the Parse / Check / Compile
calls. 

Examples have been updated to match, along with appropriate copyright notices attached.

### DIFF
--- a/cel/env.go
+++ b/cel/env.go
@@ -150,7 +150,7 @@ func (e *Env) Check(ast *Ast) (*Ast, *Issues) {
 			e.chk = ce
 		}
 	})
-	// The once call will ensure that this value is set or nil for all invocations.
+	// The once call will ensure that this value is set or
 	if e.chkErr != nil {
 		errs := common.NewErrors(ast.Source())
 		errs.ReportError(common.NoLocation, e.chkErr.Error())
@@ -168,7 +168,7 @@ func (e *Env) Check(ast *Ast) (*Ast, *Issues) {
 		expr:    res.GetExpr(),
 		info:    res.GetSourceInfo(),
 		refMap:  res.GetReferenceMap(),
-		typeMap: res.GetTypeMap()}, nil
+		typeMap: res.GetTypeMap()}, noIssues
 }
 
 // Compile combines the Parse and Check phases CEL program compilation to produce an Ast and
@@ -193,12 +193,12 @@ func (e *Env) Compile(txt string) (*Ast, *Issues) {
 // Note, for parse-only uses of CEL use Parse.
 func (e *Env) CompileSource(src common.Source) (*Ast, *Issues) {
 	ast, iss := e.ParseSource(src)
-	if iss != nil && iss.Err() != nil {
+	if iss.Err() != nil {
 		return nil, iss
 	}
 	checked, iss2 := e.Check(ast)
-	if iss != nil && iss2 != nil {
-		iss.Append(iss2)
+	if iss2.Err() != nil {
+		iss = iss.Append(iss2)
 	} else if iss2 != nil {
 		iss = iss2
 	}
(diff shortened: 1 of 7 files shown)


TristonianJones:
Hopefully this is what you had in mind @tobiaszheller

cel/env.go
```diff
@@ -365,23 +365,36 @@ func NewIssues(errs *common.Errors) *Issues {
 
 // Err returns an error value if the issues list contains one or more errors.
 func (i *Issues) Err() error {
-	if len(i.errs.GetErrors()) > 0 {
+	if i.errs != nil && len(i.errs.GetErrors()) > 0 {
```
tobiaszheller: to be honest I would prefer:
```
func (i *Issues) Err() error {
	if i == nil {
		return nil
	}
	if len(i.errs.GetErrors()) > 0 {
		return errors.New(i.errs.ToDisplayString())
	}
	return nil
}
```
right now there is possibility that if someone returns `nil` as no error will trigger panic right?

cel/env.go
```diff
@@ -193,12 +193,12 @@ func (e *Env) Compile(txt string) (*Ast, *Issues) {
...
 		return nil, iss
 	}
 	checked, iss2 := e.Check(ast)
-	if iss != nil && iss2 != nil {
-		iss.Append(iss2)
+	if iss2.Err() != nil {
+		iss = iss.Append(iss2)
 	} else if iss2 != nil {
 		iss = iss2
 	}
```
tobiaszheller: is it valid anymore? right now if `iss` has no error but warning and `iss2` has no error but warning, information from `iss` will be lost. I think that else-if can be completly removed until there are some warning or deprecated notice.

tobiaszheller (commented):
@TristonianJones thanks for doing it, it will simplify using API a lot. Left some comments.

cel/env.go
```diff
@@ -193,12 +193,12 @@ func (e *Env) Compile(txt string) (*Ast, *Issues) {
...
-	if iss != nil && iss.Err() != nil {
+	if iss.Err() != nil {
 		return nil, iss
 	}
 	checked, iss2 := e.Check(ast)
-	if iss != nil && iss2 != nil {
-		iss.Append(iss2)
+	if iss2.Err() != nil {
+		iss = iss.Append(iss2)
 	} else if iss2 != nil {
```
gabrielsz: Probably this case is redundant. The issues list would be empty, so no `nil` checks should be needed

common/errors.go
```diff
@@ -45,9 +45,13 @@ func (e *Errors) GetErrors() []Error {
 	return e.errors[:]
```
gabrielsz, line 45: We might need to do a null check here, depending on how `func (i *Issues) Errors() []common.Error` gets crafted

examples/README.md
```diff
@@ -7,20 +7,17 @@ name.
...
+    "github.com/google/cel-go/cel"
+    "github.com/google/cel-go/checker/decls"
 )
 
-	d := cel.Declarations(decls.NewIdent("name", decls.String, nil))
-	env, err := cel.NewEnv(d)
+    d := cel.Declarations(decls.NewIdent("name", decls.String, nil))
+    env, err := cel.NewEnv(d)
 
     // Check iss for error in both Parse and Check.
```
gabrielsz: Now that you're no longer calling `Parse` and `Check` this comment might need to be updated.

gabrielsz (commented):
Changes look good. Saw a couple of small things to double check before merging.

common/errors.go
TristonianJones: I've made `*Issues` a nil-able object type so whenever the `*Issues` value is non-nil, the `*common.Errors` object will also be non-nil.

examples/README.md
TristonianJones: Done.

cel/env.go
TristonianJones: Thanks, fixed.

cel/env.go
TristonianJones: Removed. Good catch.

cel/env.go
TristonianJones: Done. Thanks.

examples/README.md
```diff
@@ -7,20 +7,20 @@ name.
...
-	c, iss := env.Check(p)
-
-	prg, err := env.Program(c)
-
-	out, _, err := prg.Eval(map[string]interface{}{
+    ast, iss := env.Compile(`"Hello world! I'm " + name + "."`)
+    // Check iss for compilation errors.
+	if iss.Err() != nil {
+		log.Fatalln(iss.Err())
+	}
```
gabrielsz: Seems this is off by a few spaces

gabrielsz (approved):

PR #313, opened 2020-02-26
Feature/decorators add

guptadipanshu:
## Add Decorators as Program Option 
1. Add cel.Decorators function to accept decorators as a Program 
2. Make EvalConst, EvalBinary, EvalWatch public visible so that they can be used in the decorators.

### DIFF
--- a/cel/options.go
+++ b/cel/options.go
@@ -250,3 +250,15 @@ func EvalOptions(opts ...EvalOption) ProgramOption {
 		return p, nil
 	}
 }
+
+// Decorators sets the decorator options which may affect the evaluation or Result.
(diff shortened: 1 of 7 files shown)


TristonianJones:
As discussed offline, let's look at ways to expose more read-only interfaces publicly with helpers to build constants and functions.

TristonianJones:
Any update here @guptadipanshu?

guptadipanshu:
sorry for the delay tristan, we can ignore this CL , currently we got a workaround for supporting IAP. We will be picking this feature request a bit later when supporting IAP troubleshooter. I will put a bug in the backlog so that we don't forget this.

PR #314, opened 2020-02-28
Fix issue with error and unknown propagation on complex type resolution.

TristonianJones:
The majority of the tests for attribute resolution deal with primitive types or declared message types. When the message has not been configured in the environment, an error was being swallowed during attribute resolution for the missing complex type, surfacing `no such overload` instead of the more appropriate error message.

### DIFF
--- a/interpreter/attributes.go
+++ b/interpreter/attributes.go
@@ -860,5 +860,11 @@ func refResolve(adapter ref.TypeAdapter, idx ref.Val, obj interface{}) (ref.Val,
 		}
 		return elem, nil
 	}
+	if types.IsUnknown(celVal) {
+		return celVal, nil
+	}
+	if types.IsError(celVal) {
+		return nil, celVal.Value().(error)
+	}
 	return nil, errors.New("no such overload")
 }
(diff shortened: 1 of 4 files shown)


interpreter/attributes.go
```diff
@@ -860,5 +860,11 @@ func refResolve(adapter ref.TypeAdapter, idx ref.Val, obj interface{}) (ref.Val,
 		}
 		return elem, nil
 	}
+	if types.IsUnknown(celVal) {
+		return celVal, nil
+	}
+	if types.IsError(celVal) {
```
JimLarson, line 869: I thought we were going to have the ability to bind variables to reified errors.  Does this interfere with that?

JimLarson (approved):

interpreter/attributes.go
TristonianJones: The error message would be preserved if the variable is bound to an error, but if we add source positions as extra state carried with the `types.Err` value, then this would be lost. I'll add a `TODO` to make sure this is accounted for in the future. Currently, the error behavior is structured to be consistent with the rest of the code.

PR #316, opened 2020-03-03
Create an extension library for working with strings

### LINKED ISSUE #306: FR: Built-in substring functions

TristonianJones:
This code is based on FR #306 which requested support for built-in substrings. While substrings are not part of the core CEL spec, such functions are a common ask and the following is an implementation of such functions as extensions which can be layered onto the core CEL spec.

Closes #306

### DIFF
--- a/ext/strings.go
+++ b/ext/strings.go
@@ -31,6 +31,7 @@ import (
 )
 
 // Strings returns a cel.EnvOption to configure extended functions for string manipulation.
+// As a general note, all indices are zero-based.
 //
 // CharAt
 //
--- a/ext/strings.go
+++ b/ext/strings.go
@@ -27,6 +27,12 @@ import (
 	exprpb "google.golang.org/genproto/googleapis/api/expr/v1alpha1"
 )
 
+// Strings returns a cel.EnvOption to configure extended string functions.
+//
+//
+//
+//
+//
 func Strings() cel.EnvOption {
 	return cel.Lib(stringLib{})
 }
--- a/ext/strings.go
+++ b/ext/strings.go
@@ -152,11 +152,11 @@ func (stringLib) ProgramOptions() []cel.ProgramOption {
 			},
 			&functions.Overload{
 				Operator: "lower",
-				Unary:    callInStrOutStr(lower),
+				Unary:    callInStrOutStr(strings.ToLower),
 			},
 			&functions.Overload{
 				Operator: "string_lower",
-				Unary:    callInStrOutStr(lower),
+				Unary:    callInStrOutStr(strings.ToLower),
 			},
 			&functions.Overload{
 				Operator: "replace",
@@ -206,19 +206,19 @@ func (stringLib) ProgramOptions() []cel.ProgramOption {
 			},
 			&functions.Overload{
 				Operator: "trim",
-				Unary:    callInStrOutStr(trim),
+				Unary:    callInStrOutStr(strings.TrimSpace),
 			},
 			&functions.Overload{
 				Operator: "string_trim",
-				Unary:    callInStrOutStr(trim),
+				Unary:    callInStrOutStr(strings.TrimSpace),
 			},
 			&functions.Overload{
 				Operator: "upper",
-				Unary:    callInStrOutStr(upper),
+				Unary:    callInStrOutStr(strings.ToUpper),
 			},
 			&functions.Overload{
 				Operator: "string_upper",
-				Unary:    callInStrOutStr(upper),
+				Unary:    callInStrOutStr(strings.ToUpper),
 			},
 		),
 	}
@@ -234,7 +234,7 @@ func after(str, sub string) (string, error) {
 
 func afterOffset(str, sub string, ind int64) (string, error) {
 	i := int(ind)
-	if i >= len(str) {
+	if i < 0 || i >= len(str) {
 		return "", fmt.Errorf("index out of range: %d", ind)
 	}
 	return after(str[i:], sub)
@@ -250,7 +250,7 @@ func before(str, sub string) (string, error) {
 
 func beforeOffset(str, sub string, ind int64) (string, error) {
 	i := int(ind)
-	if i >= len(str) {
+	if i < 0 || i >= len(str) {
 		return "", fmt.Errorf("index out of range: %d", ind)
 	}
 	out, err := before(str[i:], sub)
@@ -262,7 +262,7 @@ func beforeOffset(str, sub string, ind int64) (string, error) {
 
 func charAt(str string, ind int64) (string, error) {
 	i := int(ind)
-	if i >= len(str) {
+	if i < 0 || i >= len(str) {
 		return "", fmt.Errorf("index out of range: %d", ind)
 	}
 	return str[i : i+1], nil
@@ -274,14 +274,10 @@ func indexOf(str, substr string) (int64, error) {
 
 func indexOfOffset(str, substr string, offset int64) (int64, error) {
 	off := int(offset)
-	if off >= len(str) {
+	if off < 0 || off >= len(str) {
 		return -1, fmt.Errorf("index out of range: %d", off)
 	}
 	return offset + int64(strings.Index(str[offset:], substr)), nil
-}
-
-func lower(str string) (string, error) {
-	return strings.ToLower(str), nil
 }
 
 func replace(str, old, new string) (string, error) {
@@ -301,7 +297,7 @@ func splitN(str, sep string, n int64) ([]string, error) {
 }
 
 func substr(str string, start int64) (string, error) {
-	if int(start) >= len(str) {
+	if int(start) < 0 || int(start) >= len(str) {
 		return "", fmt.Errorf("index out of range: %d", start)
 	}
 	return str[int(start):], nil
@@ -312,34 +308,22 @@ func substrRange(str string, start, end int64) (string, error) {
 	if start > end {
 		return "", fmt.Errorf("invalid substring range. start: %d, end: %d", start, end)
 	}
-	if int(start) >= l {
+	if int(start) < 0 || int(start) >= l {
 		return "", fmt.Errorf("index out of range: %d", start)
 	}
-	if int(end) >= l {
+	if int(end) < 0 || int(end) >= l {
 		return "", fmt.Errorf("index out of range: %d", end)
 	}
 	return str[int(start):int(end)], nil
 }
 
-func trim(str string) (string, error) {
-	return strings.TrimSpace(str), nil
-}
-
-func upper(str string) (string, error) {
-	return strings.ToUpper(str), nil
-}
-
-func callInStrOutStr(fn func(string) (string, error)) functions.UnaryOp {
+func callInStrOutStr(fn func(string) string) functions.UnaryOp {
 	return func(val ref.Val) ref.Val {
 		vVal, ok := val.(types.String)
 		if !ok {
 			return types.ValOrErr(val, "no such overload")
 		}
-		out, err := fn(string(vVal))
-		if err != nil {
-			return types.NewErr(err.Error())
-		}
-		return types.String(out)
+		return types.String(fn(string(vVal)))
 	}
 }
 
--- a/ext/strings.go
+++ b/ext/strings.go
@@ -249,7 +249,7 @@ func (stringLib) ProgramOptions() []cel.ProgramOption {
 					if len(values) == 4 {
 						return wrappedReplaceN(values...)
 					}
-					return types.NewErr("no such overload")
+					return types.NoSuchOverloadErr()
 				},
 			},
 			&functions.Overload{
@@ -371,7 +371,7 @@ func callInStrOutStr(fn func(string) string) functions.UnaryOp {
 	return func(val ref.Val) ref.Val {
 		vVal, ok := val.(types.String)
 		if !ok {
-			return types.ValOrErr(val, "no such overload")
+			return types.MaybeNoSuchOverloadErr(val)
 		}
 		return types.String(fn(string(vVal)))
 	}
@@ -381,11 +381,11 @@ func callInStrIntOutStr(fn func(string, int64) (string, error)) functions.Binary
 	return func(val, arg ref.Val) ref.Val {
 		vVal, ok := val.(types.String)
 		if !ok {
-			return types.ValOrErr(val, "no such overload")
+			return types.MaybeNoSuchOverloadErr(val)
 		}
 		argVal, ok := arg.(types.Int)
 		if !ok {
-			return types.ValOrErr(arg, "no such overload")
+			return types.MaybeNoSuchOverloadErr(arg)
 		}
 		out, err := fn(string(vVal), int64(argVal))
 		if err != nil {
@@ -399,11 +399,11 @@ func callInStrStrOutInt(fn func(string, string) (int64, error)) functions.Binary
 	return func(val, arg ref.Val) ref.Val {
 		vVal, ok := val.(types.String)
 		if !ok {
-			return types.ValOrErr(val, "no such overload")
+			return types.MaybeNoSuchOverloadErr(val)
 		}
 		argVal, ok := arg.(types.String)
 		if !ok {
-			return types.ValOrErr(arg, "no such overload")
+			return types.MaybeNoSuchOverloadErr(arg)
 		}
 		out, err := fn(string(vVal), string(argVal))
 		if err != nil {
@@ -417,11 +417,11 @@ func callInStrStrOutListStr(fn func(string, string) ([]string, error)) functions
 	return func(val, arg ref.Val) ref.Val {
 		vVal, ok := val.(types.String)
 		if !ok {
-			return types.ValOrErr(val, "no such overload")
+			return types.MaybeNoSuchOverloadErr(val)
 		}
 		argVal, ok := arg.(types.String)
 		if !ok {
-			return types.ValOrErr(arg, "no such overload")
+			return types.MaybeNoSuchOverloadErr(arg)
 		}
 		out, err := fn(string(vVal), string(argVal))
 		if err != nil {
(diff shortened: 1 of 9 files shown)


TristonianJones:
@wlynch FYI

ext/strings.go
```diff
@@ -0,0 +1,628 @@
...
+	"github.com/google/cel-go/interpreter/functions"
+
+	exprpb "google.golang.org/genproto/googleapis/api/expr/v1alpha1"
+)
+
+// Strings returns a cel.EnvOption to configure extended functions for string manipulation.
+//
+// After
+//
+// Given the target string and a search string, return the substring that occurs after the search
```
JimLarson: Specify whether it takes the remainder after the *first* occurrence or the *last* occurrence, and add an example which illustrates the difference, e.g. `'banana'.after('an') // returns 'ana' or 'a'`.  Consider splitting into two functions with obvious names: `afterFirst` and `afterLast`.

ext/strings.go
```diff
@@ -0,0 +1,628 @@
...
+//     <string>.before(<string>) -> <string>
+//
+// Examples:
+//
+//     'dragon'.before('taco') // returns 'dragon'
+//     'tacocat'.before('cat') // returns 'taco'
+//
+// CharAt
+//
+// Returns the character at the given position. If the position is less than 0 or greater than the
```
JimLarson: "...or greater than or equal to the length of the string."  Or have asking for the end-position character return `''`.

ext/strings.go
```diff
@@ -0,0 +1,628 @@
...
+//     <string>.after(<string>) -> <string>
+//
+// Examples:
+//
+//     'dragon'.after('taco')  // returns ''
+//     'tacocat'.after('taco') // returns 'cat'
+//
+// Before
+//
+// Given the target string and a search string, return the substring that occurs before the search
```
JimLarson: Same question here about first vs last occurrence: `'banana'.before('an') // returns 'b' or 'ban'`.  Consider `beforeFirst` and `beforeLast`.

ext/strings.go
```diff
@@ -0,0 +1,628 @@
...
+//
+// Examples:
+//
+//     'hello'.charAt(4)  // return 'o'
+//     'hello'.charAt(-1) // error
+//     'hello'.charAt(5)  // error
+//
+// IndexOf
+//
+// Returns the integer index of the search string if found, or -1 if not found. Accepts an optional
```
JimLarson: Same issue here with first vs last occurrence, example, and explicitly named alternatives.

ext/strings.go
```diff
@@ -0,0 +1,628 @@
...
+// Examples:
+//
+//     'hello mellow'.indexOf('ello')     // returns 1
+//     'hello mellow'.indexOf('jello')    // returns -1
+//     'hello mellow'.indexOf('ello', 2)  // returns 7
+//     'hello mellow'.indexOf('ello', 20) // error
+//
+// Lower
+//
+// Return a new lowercase version of the target string:
```
JimLarson: Unicode makes this a lot more complicated and opens potential security issues for users of CEL expressions.  For instance, does this transformation do *case-mapping* or *case-folding*?  See [this FAQ](https://unicode.org/faq/casemap_charprop.html).  It's possible, even desirable, to refer to an existing implementation as the standard, e.g. golang's ToLower(), but it will be worth pointing out differences with the behavior of the equivalent function in other languages:  JavaScript, Python, etc.  See also the [security risks of Unicode case transformation](https://0xsha.io/posts/the-fall-of-mighty-django-exploiting-unicode-case-transformations).

ext/strings.go
```diff
@@ -0,0 +1,628 @@
...
+//     <string>.lower() -> <string>
+//
+// Examples:
+//
+//     'User-Agent'.lower() // returns 'user-agent'
+//
+// Replace
+//
+// Produces a new string based on the target, which replaces the occurrences of a search string
+// with a replacement string if present. Accepts an optional argument specifying a limit on the
```
JimLarson: The behavior of the optional argument in the examples doesn't match this description.

ext/strings.go
```diff
@@ -0,0 +1,628 @@
...
+//
+//     'hello hello'.replace('he', 'we')    // returns 'wello wello'
+//     'hello hello'.replace('he', 'we', 1) // returns 'wello hello'
+//     'hello hello'.replace('he', 'we', 0) // returns 'hello wello'
+//     'hello hello'.replace('he', 'we', -1) // returns 'wello wello'
+//
+// Split
+//
+// Produces a list of strings which were split from the input by the given seperator. Accepts an
+// optional argument specifying a limit on the number of substrings produced by the split:
```
JimLarson: Clarify that when limited it will split on the first `n` matches, 0 means no strings returned, and negative means unlimited.

ext/strings.go
```diff
@@ -0,0 +1,628 @@
...
+//
+//     'hello hello hello'.split(' ')     // returns ['hello', 'hello', 'hello']
+//     'hello hello hello'.split(' ', 0)  // returns []
+//     'hello hello hello'.split(' ', 1)  // returns ['hello hello hello']
+//     'hello hello hello'.split(' ', 2)  // returns ['hello', 'hello hello']
+//     'hello hello hello'.split(' ', -1) // returns ['hello', 'hello', 'hello']
+//
+// Substring
+//
+// Returns the substring given a numeric range corresponding to character positions. Optionally
```
JimLarson, line 130: Clarify that character positions are 0-based.  Specify that it's an error for the end position to be before the start position, or for either to be negative or larger than the string length.

ext/strings.go
```diff
@@ -0,0 +1,628 @@
...
+// Examples:
+//
+//     'tacocat'.substring(4)    // returns 'cat'
+//     'tacocat'.substring(0, 4) // returns 'taco'
+//     'tacocat'.substring(-1)   // error
+//     'tacocat'.substring(2, 1) // error
+//
+// Trim
+//
+// Returns a new string which trims the whitespace before and after the target string:
```
JimLarson: Change "before and after" to "at the beginning ad and the end of".  Note what counts as whitespace, e.g. all [Unicode whitespace characters](https://en.wikipedia.org/wiki/Whitespace_character), in which case decide whether to mention that this _doesn't_ include characters like "zero width space U+200B".  Or refer to an existing standard.

ext/strings.go
```diff
@@ -0,0 +1,628 @@
...
+//
+//      <string>.trim() -> <string>
+//
+// Examples:
+//
+//     '    trim    '.trim() // returns 'trim'
+//
+// Upper
+//
+// Returns a new uppercase version of the target string:
```
JimLarson: Again, refer to Unicode case issues from `lower()`.  No titlecase conversion? :-)  Consider just having a case-insensitive comparison function `s1.eqIgnoreCase(s2)`.

ext/strings.go
```diff
@@ -0,0 +1,628 @@
...
+			&functions.Overload{
+				Operator: "string_before_string",
+				Binary:   callInStrStrOutStr(before),
+			},
+			&functions.Overload{
+				Operator: "charAt",
+				Binary:   callInStrIntOutStr(charAt),
+			},
+			&functions.Overload{
+				Operator: "string_chat_at_int",
```
JimLarson: s/chat/char/

ext/strings.go
```diff
@@ -0,0 +1,628 @@
...
+		return "", fmt.Errorf("index out of range: %d", end)
+	}
+	return str[int(start):int(end)], nil
+}
+
+func callInStrOutStr(fn func(string) string) functions.UnaryOp {
+	return func(val ref.Val) ref.Val {
+		vVal, ok := val.(types.String)
+		if !ok {
+			return types.ValOrErr(val, "no such overload")
```
JimLarson: One of these days a `"no such overlaod"` will sneak through.  Consider a constant for at least the string, and perhaps specialized versions of `NewErr()` or `ValOrErr()`.

ext/strings_test.go
```diff
@@ -0,0 +1,275 @@
...
+package ext
+
+import (
+	"fmt"
+	"testing"
+
+	"github.com/google/cel-go/cel"
+)
+
+var stringTests = []struct {
```
JimLarson, line 25: Add a TODO or issue to move these all to conformance tests.

ext/strings_test.go
```diff
@@ -0,0 +1,275 @@
...
+	expr      string
+	err       string
+	parseOnly bool
+}{
+	// Success test cases.
```
JimLarson: Need tests for all of these which venture beyond ASCII characters.

ext/strings_test.go
```diff
@@ -0,0 +1,275 @@
...
+	{expr: `'hello'.after('l') == 'lo'`},
+	{expr: `'hello'.after('none') == ''`},
+	{expr: `'tacocat'.before('c') == 'ta'`},
+	{expr: `'tacocat'.before('none') == 'tacocat'`},
+	{expr: `'tacocat'.charAt(3) == 'o'`},
+	{expr: `'tacocat'.indexOf('a') == 1`},
+	{expr: `'tacocat'.indexOf('a', 3) == 5`},
+	{expr: `'tacocat'.indexOf('none') == -1`},
+	{expr: `'MixedCase'.lower().after('mixed') == 'case'`},
```
JimLarson: Add examples showing that case conversion can change string length.

ext/strings_test.go
```diff
@@ -0,0 +1,275 @@
...
+	{expr: `"{0} days {0} hours".replace("{0}", "2") == "2 days 2 hours"`},
+	{expr: `"{0} days {0} hours".replace("{0}", "2", 1).replace("{0}", "23") == "2 days 23 hours"`},
+	{expr: `"hello world".split(" ") == ["hello", "world"]`},
+	{expr: `"hello world events!".split(" ", 0) == []`},
+	{expr: `"hello world events!".split(" ", 1) == ["hello world events!"]`},
+	{expr: `"hello world events!".split(" ", 2) == ["hello", "world events!"]`},
+	{expr: `"tacocat".substring(4) == "cat"`},
+	{expr: `"tacocat".substring(0, 4) == "taco"`},
+	{expr: `"tacocat".substring(4, 4) == ""`},
+	{expr: `"   trim   ".trim() == "trim"`},
```
JimLarson: Try with more whitespace characters, at least the ASCII ones.

ext/strings_test.go
```diff
@@ -0,0 +1,275 @@
...
+	},
+	{
+		expr:      `"tacocat".substring(0, false) == ""`,
+		err:       "no such overload",
+		parseOnly: true,
+	},
+}
+
+func TestStrings(t *testing.T) {
+	env, err := cel.NewEnv(Strings())
```
JimLarson, line 254: I don't see this new option in this PR.

JimLarson (changes requested):

ext/strings.go
TristonianJones: SGTM

ext/strings.go
TristonianJones: I think I'm going to remove the `before` and `after` functions. I realized as I was writing tests that I would constantly get the two mixed up.

ext/strings.go
TristonianJones: Removed for now.

ext/strings.go
TristonianJones: I'll add a `lastIndexOf` as a complement to the `indexOf`

ext/strings.go
TristonianJones: Fixed.

ext/strings.go
TristonianJones: Case insensitive comparisons are reasonable, though I had not included them since map keys, such as those found in normalized HTTP headers, might be lower-cased or upper-cased. I figured if I added `<str>.lower() == <str>.lower()` that this is probably a reasonable start that works with the map key case I described.

bitsofinfo:
please add a `replace` function for simple string replacements

TristonianJones:
@bitsofinfo I'm happy to report `replace` is in this PR! :)

bitsofinfo:
solid! awesome, so for Tekton to utilize it I assume they as well will have to cut a new release updating the dependency to whatever version this will be in.

ext/strings.go
TristonianJones: I've removed `lower` and `upper` because I can't figure out what exactly the Go implementation is doing for Unicode case mapping here, but it's definitely not what I saw documented in the Unicode 12.0 and 12.1 specifications. It seems like case-folding works though, so perhaps a case insensitive comparison and an ascii-based `lower` and `upper` functions are in the running for future PRs.

ext/strings.go
TristonianJones: Done

ext/strings.go
TristonianJones: Done.

ext/strings.go
TristonianJones: Done.

ext/strings.go
TristonianJones: Done.

ext/strings_test.go
TristonianJones: Done.

ext/strings_test.go
TristonianJones: Done.

ext/strings_test.go
TristonianJones: It's the name of the new library, just missing the `ext` prefix since the test is happening in the `ext` package.

TristonianJones:
PTAL

ext/strings.go
JimLarson: Clarify that a negative argument means unlimited replacements.

ext/strings.go
JimLarson: Still not changed.

ext/strings_test.go
```diff
@@ -0,0 +1,269 @@
...
+
+// TODO: move these tests to a conformance test.
+var stringTests = []struct {
+	expr      string
+	err       string
+	parseOnly bool
+}{
+	// CharAt test.
+	{expr: `'tacocat'.charAt(3) == 'o'`},
+	// Indext of search string tests.
```
JimLarson: s/Indext/Index/

ext/strings_test.go
```diff
@@ -0,0 +1,269 @@
...
+	{expr: `"tacocat".substring(4, 4) == ""`},
+	// Trim tests using the unicode standard for whitespace.
+	{expr: `" \f\n\r\t\vtext  ".trim() == "text"`},
+	{expr: `"\u0085\u00a0\u1680text".trim() == "text"`},
+	{expr: `"text\u2000\u2001\u2002\u2003\u2004\u2004\u2006\u2007\u2008\u2009".trim() == "text"`},
+	{expr: `"\u200atext\u2028\u2029\u202F\u205F\u3000".trim() == "text"`},
+	// Trim test with whitespace-like characters not included.
+	{expr: `"\u180etext\u200b\u200c\u200d\u2060\ufeff".trim()
+				== "\u180etext\u200b\u200c\u200d\u2060\ufeff"`},
+	// Error test cases based on checked expression usage.
```
JimLarson, line 78: Still looking for unicode-containing tests for charAt, indexOf, lastIndexOf, replace, split, and substring.  These can be just single non-edge-case tests.

TristonianJones: I misunderstood, thought you were looking for any tests involving unicode.

ext/strings.go
TristonianJones: Updated.

ext/strings.go
TristonianJones: Done.

TristonianJones (commented):
PTAL

ext/strings.go
```diff
@@ -34,45 +34,52 @@ import (
 //
 // CharAt
 //
-// Returns the character at the given position. If the position is less than 0 or greater than the
-// length of the string, the function will produce an error. CharAt is an instance method with the
-// following signature:
+// Returns the character at the given position. If the position is negative, or greater than
```
JimLarson, line 38: It's probably worth having a package-wide comment that indexes are zero-based throughout, rather than replicate that statement everywhere.

ext/strings_test.go
```diff
@@ -29,13 +29,27 @@ var stringTests = []struct {
...
+	{expr: `'ta©o©αT'.indexOf('©α', 5) == -1`},
+	{expr: `'tacocat'.lastIndexOf('') == 7`},
+	{expr: `'tacocat'.lastIndexOf('at') == 5`},
 	{expr: `'tacocat'.lastIndexOf('none') == -1`},
+	{expr: `'tacocat'.lastIndexOf('', 3) == 3`},
+	{expr: `'tacocat'.lastIndexOf('a', 3) == 1`},
+	{expr: `'ta©o©αT'.lastIndexOf('©') == 4`},
+	{expr: `'ta©o©αT'.lastIndexOf('©', 3) == 2`},
+	{expr: `'ta©o©αT'.lastIndexOf('©α', 4) == 4`},
 	// Replace tests
```
JimLarson, line 53: Add unicode tests for replace and split.  Yes, you're just calling predefined string library functions, so you don't need to go crazy, but just have one unicode character returned to make sure.

TristonianJones: Done.

ext/strings.go
TristonianJones: Done, not as a package comment, but at the top of the comment block for the `Strings` function.

TristonianJones:
PTAL

JimLarson (approved):

bigkevmcd:
Would it be possible to get a doc page with the list of functions / examples?

TristonianJones:
Do you mean for the string extensions or for all functions?

Here's the godoc for the string extensions:
https://godoc.org/github.com/google/cel-go/ext#Strings

bigkevmcd:
@TristonianJones I was hoping for something I could link to from tektoncd/triggers to help people who want to use the functions, they're not necessarily Go developers.

bigkevmcd:
The functions look great, and I'm happy to open a PR with a markdown with a list of functions examples in cel-go rather than in tektoncd/triggers which might help more folks.

TristonianJones:
@bigkevmcd Sure, are you just looking for a markdown file for the strings, or would you like all the CEL functions documented that way. In the fullness of time, I'm sure both are desired, but if you need one sooner than the other, that helps me prioritize.

bigkevmcd:
@TristonianJones Right now, the new string functions would help a lot, as soon as there's a release of cel-go with them, I have a PR ready to go.

TristonianJones:
@bigkevmcd I'll work with @JimLarson to fix a minor issue I found in the library and copy in a markdown file. To use the new functions you'll have to use `ext.Strings()` in the CEL environment setup if you haven't started already.

I let @wlynch and @bobcatfish know offline that I think the `truncate` function won't handle unicode properly, but these extension functions will, so it might be worth updating to `substring` for these cases. I also inquired about static validation of CEL expressions rather than runtime-based validation. If there's some forum I should join to discuss I'd be happy to do so there.

bitsofinfo:
@TristonianJones and @bigkevmcd thanks for brining this together, will be super handy in tekton

Issue #317, opened 2020-03-03
Parsing Fuzzer


lopezator:
**Describe the bug**

I discovered through fuzzing:

https://github.com/dvyukov/go-fuzz  
https://medium.com/@dgryski/go-fuzz-github-com-arolek-ase-3c74d5a3150c  

That some string inputs passed to the parser function end up in a panic:

```go
func TestFuzzCrashers(t *testing.T) {
	var crashers = []string{
		"--",
		"?",
	}
	env, err := NewEnv()
	if err != nil {
		t.Fatal(err)
	}
	for _, crasher := range crashers {
		_, _ = env.Parse(crasher)
	}
}
```

For example, "--", and "?" (and might be more).

The error message is:

```bash
panic: unknown parsetree type: '<nil>': <nil> [<<nil>>]
```

**To Reproduce**
Check which components this affects:

- [x]  parser
- [ ]  checker
- [ ]  interpreter

Tried only with the parser, but we could fuzz also on other places.

**Expected behavior**

To err and not to panic when passing an incorrect string.

I could start a PR if you consider this to be useful.

TristonianJones:
Hi @lopezator, thanks for the report!

A PR for the fuzzer would be greatly appreciated! If you wouldn't mind using https://github.com/google/gofuzz that would be easiest to integrate. 

As for the panic, it originates here: https://github.com/google/cel-go/blob/master/parser/parser.go#L131. A better behavior would be to report the error and return an error node within the parse tree. I can make this quick code change if you'd like to work on the Fuzzer PR in parallel.

What do you think?

TristonianJones:
Once PR #318 is submitted, I'll switch this from a bug to a feature request. If you'd like to contribute a fuzzer, I'd be super happy to have one @lopezator

lopezator:
Wow that's nice, I'll fuzz again once the changes in #318 are addressed to see if more crashes occur.

TristonianJones:
Okay @lopezator the code is all ready for you

TristonianJones:
With the bug now fixed, I've recast this PR as Parser Fuzzing support. Thanks again for filing @lopezator and I look forward to having fuzz tests in the CEL-Go repo.

TristonianJones:
The cel-go parser an interpreter are currently part of the oss-fuzz project. Marking fixed.

PR #318, opened 2020-03-04
Fix to the parser panic

TristonianJones:
Per the report in https://github.com/google/cel-go/issues/317, there was an extant `panic` in the parser which could be triggered if the parse input never produced a single valid expression node. The `panic` has been addressed, but the core issue of fuzzing remains open.

### DIFF
--- a/parser/parser.go
+++ b/parser/parser.go
@@ -123,7 +123,8 @@ func (p *parser) Visit(tree antlr.ParseTree) interface{} {
 		return p.VisitCreateStruct(tree.(*gen.CreateStructContext))
 	}
 
-	// The code only gets to this point if there is an error produced earlier.
+	// Report at least one error if the parser reaches an unknown parse element.
+	// Typically, this happens if the parser has already encountered a syntax error elsewhere.
 	if len(p.errors.GetErrors()) == 0 {
 		txt := "<<nil>>"
 		if tree != nil {
--- a/parser/parser.go
+++ b/parser/parser.go
@@ -17,6 +17,7 @@
 package parser
 
 import (
+	"fmt"
 	"strconv"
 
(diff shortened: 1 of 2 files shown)


lopezator:
Looks good!

parser/parser.go
```diff
@@ -124,11 +122,9 @@ func (p *parser) Visit(tree antlr.ParseTree) interface{} {
 		return p.VisitCreateStruct(tree.(*gen.CreateStructContext))
 	}
 
-	text := "<<nil>>"
-	if tree != nil {
-		text = tree.GetText()
-	}
-	panic(fmt.Sprintf("unknown parsetree type: '%+v': %+v [%s]", reflect.TypeOf(tree), tree, text))
+	// The code only gets to this point if there is an error produced earlier.
```
JimLarson: This doesn't seem accurate.  How about "The code only gets to this point if the tree didn't match one of the above types."?

JimLarson (changes requested):

parser/parser.go
TristonianJones: When the node can't be identified, the `tree` value is `nil` due to a prior error. I haven't seen any cases that would lead me to believe otherwise, but let me test for the presence of other errors, report one if none had been found, and return an empty expression otherwise.

TristonianJones:
PTAL

parser/parser.go
JimLarson: Good, although the line 126 comment is now clearly wrong - there might not be an error recorded at this point.  Maybe just delete the comment - I think the code is clear enough by itself.

JimLarson (approved):

parser/parser.go
TristonianJones: Oops, updated.

JimLarson:
Love it.

PR #319, opened 2020-03-10
Bug and doc fix for lastIndexOf

### LINKED ISSUE #306: FR: Built-in substring functions

TristonianJones:
I noticed a doc issue on `lastIndexOf` which then caused me to notice a bug in the implementation as well. Fixed both. Should fully resolve #306 this time.


ext/strings.go
```diff
@@ -83,7 +83,7 @@ import (
 //     'hello mellow'.lastIndexOf('')         // returns 12
 //     'hello mellow'.lastIndexOf('ello')     // returns 7
 //     'hello mellow'.lastIndexOf('jello')    // returns -1
-//     'hello mellow'.lastIndexOf('ello', 8)  // returns 2
+//     'hello mellow'.lastIndexOf('ello', 6)  // returns 2
```
JimLarson: Shouldn't this return 1?

```
0 1 2 3 4 5 6 7 8 9
 h e l l o _ m e l l o
```

Also:

- specify that we return the index of the start of the matched string;
- the optional index is the _end_ of the search (assuming we're searching from left-to-right, as the name "`lastIndexOf`" implies.

JimLarson (approved):
Fixes for the comments only.

ext/strings.go
TristonianJones: It should be `1`. The optional index represents where to start the substring match starting from the end, rather than where the substring match must end. 

Given input:
```
0 1 2 3 4 5 6 7 8 9 10
h e l l o _ m e l l o
```
Given an index of `7`, the substring match `ello` exactly starts at `7`.  For the index `6` the nearest substring match occurs at `1`.

TristonianJones:
PTAL

ext/strings.go
```diff
@@ -68,12 +68,12 @@ import (
...
-// Returns the integer index of the last occurrence of the search string. If the search string is
-// not found the function returns -1.
+// Returns the integer index at the start of the last occurrence of the search string. If the
+// search string is not found the function returns -1.
 //
-// The function also accepts an optional index which represents where to begin the search in
-// through the string. If the substring is the empty string, the index where the search starts
-// is returned (string length or custom).
+// The function also accepts an optional index which represents where to begin the substring
+// search starting from the end of the string. If the substring is the empty string, the index
```
JimLarson, line 75: I still don't like the sentence.  And there's still some ambiguity about the behavior for overlapping matches.  I *think* what we want is the mirror image of `indexOf()`.

How about this?

The function also accepts an optional index which discards matches starting at or after that index.  The function will return the substring match with the greatest index which is less than the optional limit.

Also add an example/test:  `'bananananana'.lastIndexOf('nana', 8) // returns 6

Issue #322, opened 2020-03-23
Unsafe mutations of shared state


JimLarson:
**Describe the bug**
When creating a new cel.Env with custom protocol buffer types, the code mutates shared state without any locking.

**To Reproduce**
Discovered in thread sanitizer testing with multiple concurrent calls to

```
cel.NewEnv(cel.Types(&some.Message{}))
```

Other examples of mutation of shared state might exist and a thorough search or static analysis might be appropriate.  E.g. separately-created environments share underlying state via `pb.Db` from the `DefaultDb`.

Check which components this affects:

- [X]  parser
- [X]  checker
- [X]  interpreter

**Expected behavior**
Shared state should be immutable or changed only under a mutex.

**Additional context**
Seems to have been introduced in #291 in line 135 of `pb.go`.

JimLarson:
At least one additional instance: the `DefaultDb` in `common/types/pb/pb.go` leads to shared mutable state in new `Db` instances created with `pb.NewDb()`. I will scan for other potential instances.

JimLarson:
I've scanned all other globals, and I don't see anything else that's clearly shared mutable state. I didn't audit all usages, so it's possible that something might, e.g. redefine `types.True` or something perverse like that, but there's nothing that looks like it ought to be mutable.  So yes, this time the pending CL should close this issue.

PR #323, opened 2020-03-23
Avoid mutating shared slice.

JimLarson:
Fixes an instance of Issue #322

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -34,12 +34,10 @@ go_repository(
   importpath = "github.com/antlr/antlr4",
 )
 
-#git_repository(
-local_repository(
+git_repository(
   name = "com_google_cel_spec",
-  path = "/usr/local/google/home/jimlarson/firebase/cel/github/cel-spec",
-  #commit = "a61184970892e985673578119b2263866c4a18d9", # PR #104
-  #remote = "https://github.com/google/cel-spec.git",
+  commit = "a61184970892e985673578119b2263866c4a18d9", # PR #104
+  remote = "https://github.com/google/cel-spec.git",
 )
 
 # Required to use embedded BUILD.bazel file in googleapis/google/rpc
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -34,10 +34,12 @@ go_repository(
   importpath = "github.com/antlr/antlr4",
 )
 
-git_repository(
+#git_repository(
+local_repository(
   name = "com_google_cel_spec",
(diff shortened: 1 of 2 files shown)


common/types/pb/pb.go
```diff
@@ -121,7 +121,12 @@ func CollectFileDescriptorSet(message proto.Message) (*descpb.FileDescriptorSet,
 	fdMap := map[string]*descpb.FileDescriptorProto{}
 	fd, _ := descriptor.ForMessage(message.(descriptor.Message))
 	fdMap[fd.GetName()] = fd
-	deps := fd.GetDependency()
+	// Initialize list of dependencies
```
TristonianJones: I don't think this is doing anything interesting. Were there other changes in other files?

WORKSPACE
```diff
@@ -34,10 +34,12 @@ go_repository(
   importpath = "github.com/antlr/antlr4",
 )
 
-git_repository(
+#git_repository(
+local_repository(
   name = "com_google_cel_spec",
```
TristonianJones: Please revert.

common/types/pb/pb.go
TristonianJones, line 124: I see, you want a copy? Ah, it's the lines below that append to the `deps` list that cause the problem. The `[:]` syntax should make a copy of the slice. This is the simpler way to specify what's being done here:

`deps := fd.GetDependency()[:]`

TristonianJones (changes requested):

common/types/pb/pb.go
TristonianJones: Hmm, nevermind, even though slices are typically copy on write, there's still a chance this would modify the underlying storage.  Let's try:

```go
fileDeps := fd.GetDependency()
deps := make([]string, len(fileDeps), len(fileDeps))
copy(deps, fileDeps)
```

TristonianJones: I've created PR #324 to address what I think are the remainder of the potentially (inadvertent) mutable `append` calls on slices.

WORKSPACE
JimLarson: Done.

common/types/pb/pb.go
JimLarson: More than a "chance" the original suggestion definitely shares storage with the slice returned by `fd.GetDependency()`, but there's a chance that this slice is already at its capacity.  Switched to the more idiomatic copy.

TristonianJones (approved):

PR #324, opened 2020-03-24
Update test parallelism and remove append() calls with side-effects

TristonianJones:
This PR improves the list mutation guarantees and parallelism coverage for the checker and interpreter. 

This PR partially addresses #322. PR #323 should cover the remainder once submitted.

### DIFF
--- a/checker/types.go
+++ b/checker/types.go
@@ -497,10 +497,10 @@ func flattenFunctionTypes(f *exprpb.Type_FunctionType) []*exprpb.Type {
 	if len(argTypes) == 0 {
 		return []*exprpb.Type{f.GetResultType()}
 	}
-	merged := make([]*exprpb.Type, len(argTypes)+1, len(argTypes)+1)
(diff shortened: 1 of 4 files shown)


checker/types.go
```diff
@@ -489,3 +489,16 @@ func substitute(m *mapping, t *exprpb.Type, typeParamToDyn bool) *exprpb.Type {
 func typeKey(t *exprpb.Type) string {
 	return FormatCheckedType(t)
 }
+
+func mergeFunctionTypes(f *exprpb.Type_FunctionType) []*exprpb.Type {
```
JimLarson: Consider naming as `flattenFunctionTypes()` and add the docstring

```
// flattenFunctionTypes takes a function with arg types T1, T2, ..., TN and result type TR
// and returns a slice containing {T1, T2, ..., TN, TR}.
```

JimLarson (approved):

checker/types.go
TristonianJones: Done.

PR #325, opened 2020-03-25
Fix for index attribute resolution with global function call result

TristonianJones:
For index expressions where the argument to the index is a global function call, the expression plan fails with an error, e.g. 

fails: `map[string('key')]`
succeeds: `map['key']`

The solution is to make sure that a `RelativeAttribute` is created for the argument to the index operation when it is not already an `Attribute` or `Constant` value.

### DIFF
--- a/interpreter/interpreter_test.go
+++ b/interpreter/interpreter_test.go
@@ -217,7 +217,7 @@ var (
 		},
 		{
 			name: "index",
-			expr: `m['key'][1] == 42u && m['null'] == null`,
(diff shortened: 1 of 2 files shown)


TristonianJones:
FYI @josephpeacock, we're already planning on building a point release in the next day or so. I'll keep you posted when it's tagged.

JimLarson (approved):

josephpeacock:
Thanks for the quick turnaround @TristonianJones!

PR #326, opened 2020-03-25
Refactor of the proto description logic for safety

### LINKED ISSUE #322: Unsafe mutations of shared state

TristonianJones:
Improvements in the initialization of `pb.FileDescription` and associated `pb.TypeDescription` values.

Notably, the `atomic.Value` members in the `pb.TypeDescription` have been replaced with a `sync.Once` object as the mutations were thread-safe, but should really only have been happening exactly once. 

Additionally, the reverse qualified name to file descriptor lookup is being managed entirely within the `pb.Db` as opposed to spread between the `pb.FileDescription` and `pb.Db` files. The net effect should be safer, simpler code that is better documented and better behaved.

Closes #322

### DIFF
--- a/common/types/pb/enum.go
+++ b/common/types/pb/enum.go
@@ -18,7 +18,8 @@ import (
 	descpb "github.com/golang/protobuf/protoc-gen-go/descriptor"
 )
 
-// NewEnumValueDescription produces a new named enum value description.
+// NewEnumValueDescription produces an enum value description with the fully qualified enum value
+// name and the enum value descriptor.
 func NewEnumValueDescription(name string,
 	desc *descpb.EnumValueDescriptorProto) *EnumValueDescription {
 	return &EnumValueDescription{
@@ -27,13 +28,13 @@ func NewEnumValueDescription(name string,
 	}
 }
 
-// EnumValueDescription maps a qualified enum name to its numeric value.
+// EnumValueDescription maps a fully-qualified enum value name to its numeric value.
 type EnumValueDescription struct {
 	enumValueName string
 	desc          *descpb.EnumValueDescriptorProto
 }
 
-// Name returns the human-readable identifier name for the enum value.
+// Name returns the fully-qualified identifier name for the enum value.
 func (ed *EnumValueDescription) Name() string {
 	return ed.enumValueName
 }
--- a/common/types/pb/enum.go
+++ b/common/types/pb/enum.go
@@ -18,19 +18,27 @@ import (
 	descpb "github.com/golang/protobuf/protoc-gen-go/descriptor"
 )
 
-// EnumDescription maps a qualified enum name to its numeric value.
-type EnumDescription struct {
-	enumName string
-	file     *FileDescription
-	desc     *descpb.EnumValueDescriptorProto
+// NewEnumValueDescription produces a new named enum value description.
+func NewEnumValueDescription(name string,
+	desc *descpb.EnumValueDescriptorProto) *EnumValueDescription {
+	return &EnumValueDescription{
+		enumValueName: name,
+		desc:          desc,
+	}
 }
 
-// Name of the enum.
-func (ed *EnumDescription) Name() string {
-	return ed.enumName
+// EnumValueDescription maps a qualified enum name to its numeric value.
+type EnumValueDescription struct {
+	enumValueName string
+	desc          *descpb.EnumValueDescriptorProto
 }
 
-// Value (numeric) of the enum.
-func (ed *EnumDescription) Value() int32 {
+// Name returns the human-readable identifier name for the enum value.
+func (ed *EnumValueDescription) Name() string {
+	return ed.enumValueName
+}
+
+// Value returns the (numeric) value of the enum.
+func (ed *EnumValueDescription) Value() int32 {
 	return ed.desc.GetNumber()
 }
--- a/common/types/pb/file.go
+++ b/common/types/pb/file.go
@@ -93,9 +93,11 @@ func sanitizeProtoName(name string) string {
 	return name
 }
 
-// fileMetadata is a flattened view of  message types and enum values within a file descriptor.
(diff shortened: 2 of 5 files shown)


common/types/pb/file.go
```diff
@@ -70,44 +85,71 @@ func (fd *FileDescription) GetTypeNames() []string {
...
-	for _, enumType := range enumTypes {
-		for _, enumValue := range enumType.Value {
-			enumValueName := fmt.Sprintf(
-				"%s.%s.%s", pkg, enumType.GetName(), enumValue.GetName())
-			fd.enums[enumValueName] = &EnumDescription{
-				enumName: enumValueName,
-				file:     fd,
-				desc:     enumValue}
-			fd.pbdb.revFileDescriptorMap[enumValueName] = fd
+// fileMetadata is a flattened view of  message types and enum values within a file descriptor.
```
JimLarson: Clarify whether the map keys are fully-qualified.

common/types/pb/enum.go
```diff
@@ -18,19 +18,27 @@ import (
 	descpb "github.com/golang/protobuf/protoc-gen-go/descriptor"
 )
 
-// EnumDescription maps a qualified enum name to its numeric value.
-type EnumDescription struct {
-	enumName string
-	file     *FileDescription
-	desc     *descpb.EnumValueDescriptorProto
+// NewEnumValueDescription produces a new named enum value description.
```
JimLarson: Clarify whether the name is fully qualified.

common/types/pb/enum.go
```diff
@@ -18,19 +18,27 @@ import (
...
-	return ed.enumName
+// EnumValueDescription maps a qualified enum name to its numeric value.
+type EnumValueDescription struct {
+	enumValueName string
+	desc          *descpb.EnumValueDescriptorProto
 }
 
-// Value (numeric) of the enum.
-func (ed *EnumDescription) Value() int32 {
+// Name returns the human-readable identifier name for the enum value.
```
JimLarson: Fully qualified?

common/types/pb/file.go
```diff
@@ -70,44 +85,71 @@ func (fd *FileDescription) GetTypeNames() []string {
...
-			file:     fd,
-			desc:     msgType,
-			metadata: &atomic.Value{},
+		msgName := fmt.Sprintf("%s.%s", container, msgType.GetName())
+		msgTypeMap[msgName] = msgType
+		nestedTypes := msgType.GetNestedType()
+		if len(nestedTypes) == 0 {
+			continue
+		}
+		nestedTypeMap := collectMsgTypes(msgName, nestedTypes)
```
JimLarson: Consider making the type map an argument to `collectMsgTypes()` so it doesn't have to return a throwaway map.  Lines 132-139 could just be:

```
collectMsgTypes(msgName, msgType.GetNestedType(), msgTypeMap)
```

common/types/pb/file.go
```diff
@@ -70,44 +85,71 @@ func (fd *FileDescription) GetTypeNames() []string {
...
+func collectFileMetadata(fileDesc *descpb.FileDescriptorProto) *fileMetadata {
+	pkg := fileDesc.GetPackage()
+	msgTypes := collectMsgTypes(pkg, fileDesc.GetMessageType())
+	enumValues := collectEnumValues(pkg, fileDesc.GetEnumType())
+	for container, msgType := range msgTypes {
+		nestedEnums := msgType.GetEnumType()
+		if len(nestedEnums) == 0 {
+			continue
+		}
+		nestedEnumValues := collectEnumValues(container, nestedEnums)
```
JimLarson: Consider making the enum value map an argument to `collectEnumValue()` so it doesn't have to return a throwaway map.  Then the body of this loop could just be

```
collectEnumValue(container, msgType.getEnumType(), enumValueMap)
```

common/types/pb/type.go
```diff
@@ -18,23 +18,50 @@ import (
...
-	"sync/atomic"
+	"sync"
 
 	"github.com/golang/protobuf/proto"
 	descpb "github.com/golang/protobuf/protoc-gen-go/descriptor"
 	structpb "github.com/golang/protobuf/ptypes/struct"
 	exprpb "google.golang.org/genproto/googleapis/api/expr/v1alpha1"
 )
 
+// NewTypeDescription produces a TypeDescription value for the proto type name with a given
```
JimLarson: Specify whether the type name is fully-qualified.

JimLarson (approved):
LGTM, but with requests for clarification in comments whether names are fully-qualified, and a suggestion for simplification.

common/types/pb/file.go
TristonianJones: Done.

common/types/pb/enum.go
TristonianJones: Done.

common/types/pb/enum.go
TristonianJones: Done.

common/types/pb/file.go
TristonianJones: Done.

common/types/pb/file.go
TristonianJones: Done.

common/types/pb/type.go

PR #327, opened 2020-04-09
Improve type resolution when there are multiple overloads

TristonianJones:
When there is more than one overload that matches a call signature, the current behavior is to mark the function return as `dyn`; however, when all overloads share the same result type, the type inference can be much stronger.

For example, the `in` operator always returns a boolean even though it can accept either `map`, `list`, or `dyn` arguments:

before: `1 in dyn([1, 2, 3])` -> type(dyn)
after: `1 in dyn([1, 2, 3])` -> type(bool)

This issue was surfaced by @francislavoie of the [Caddy](https://github.com/caddyserver) team.

### DIFF
--- a/checker/checker.go
+++ b/checker/checker.go
@@ -296,7 +296,6 @@ func (c *checker) resolveOverload(
 			for _, typePar := range overload.TypeParams {
 				substitutions.add(decls.NewTypeParamType(typePar), c.newTypeVar())
 			}
-
 			overloadType = substitute(substitutions, overloadType, false)
 		}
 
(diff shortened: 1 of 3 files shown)


checker/checker.go
```diff
@@ -308,16 +307,20 @@ func (c *checker) resolveOverload(
...
 			if resultType == nil {
-				// First matching overload, determines result type.
-				resultType = substitute(c.mappings,
-					overloadType.GetFunction().ResultType,
-					false)
+				resultType = fnResultType
+			} else if !isDyn(resultType) &&
+				!isTypeParam(resultType) &&
+				!isTypeParam(fnResultType) &&
+				c.isAssignable(fnResultType, resultType) {
```
JimLarson: Consider whether each of the checks in lines 316-319 are necessary.  Might the mostGeneral() function return the right result without them?  If they are necessary, please include a test case which would fail without having the check.

JimLarson (approved):

checker/checker.go
TristonianJones: Unfortunately, `mostGeneral` relies on the types already being checked for assignability. I did originally just check for type equality which covers a wide range of cases, but probably not the maximal set. Given the complexity of the condition though, I'm almost inclined to revert back to simple equality as the condition for not marking the return `dyn`. Thoughts?

I had originally omitted the `isTypeParam` check and this expression in the `checker_test.go` file is incorrectly tagged as an `int` when it should be a `dyn` since the assignability check adds a type substitution variable where there shouldn't be one: `x.single_value[23] + x.single_struct['y']`. Both `single_value` and `single_struct` are type `dyn`.

JimLarson: The change looks simpler and more clearly correct, so yeah, go with that.

PR #332, opened 2020-04-13
Updates the has() macro to be more consistent with the cel-spec.

### LINKED ISSUE #329: Make the has() macro for proto3 conform to the cel-spec
**Describe the bug**
The `has()` macro specifies the following behaviors for maps, proto2 messages, and proto3 messages:

* maps - `has(map.key)` indicates whether the key is defined in the map.
* proto2 - `has(msg.field)` indicates whether the field is set (or present) on the object.
* proto3 - `has(msg.field)` indicates whether a field of type `message` is set on the object, or whether a field of simple type is set to its zero value. 

The cel-go code isn't properly checking whether proto3 message fields of simple type are set to their zero value and is rejecting expressions that use `has(msg.simple_field)` at check-time and at runtime with the error: `unsupported field selection target`.

**To Reproduce**

- [ ]  parser
- [X]  checker
- [X]  interpreter

```cel
has(msg.int_field)
```

Test setup:
```go
// Simple type field presence test for proto3
{
	name: "select_pb3_primitive_fields",
	expr: `!has(a.single_int32)`,
	env: []*exprpb.Decl{
		decls.NewIdent("a", decls.NewObjectType("google.expr.proto3.test.TestAllTypes"), nil),
	},
	in: map[string]interface{}{
		"a": &proto3pb.TestAllTypes{},
	},
}
```

**Expected behavior**

The expression should check correctly and evaluate true when `int_field` is not equal to 0.

** Additional Context **

This error can be worked around by simply comparing the proto3 simple field to its default zero value: `msg.int_field != 0`

TristonianJones:
The cel-spec specifies the following behavior for the `has()` macro:

```
This means means that has(e.f) ... produces false if f is declared but not set
(or, in proto3, has its default value). Moreover, has(e.f) where e is a map
returns false if f is not defined in the map.
```

In proto2 all non-repeated fields are represented by pointers which will be `nil` if not set. In proto3 all message fields are represented by pointers which will be `nil` if not set. The `has()` macro must also return false if the proto3 primitive field is equal to its zero value. 

In the process of making cel-go more consistent with cel-spec, it seems worth reexamining whether the proto3 test for default-ness should also be extended to repeated types (array, map) for both proto2 and proto3 where:

```
has(e.repeated_f) returns false if repeated_f contains zero elements. 
```

The code has been implemented to reflect this slight deviation from the cel-spec, and an issue for https://github.com/google/cel-spec/issues/113 to address the issue at the spec level.

Closes #329

### DIFF
--- a/checker/BUILD.bazel
+++ b/checker/BUILD.bazel
@@ -50,6 +50,7 @@ go_test(
         "//common/types:go_default_library",
         "//parser:go_default_library",
         "//test:go_default_library",
+        "//test/proto2pb:go_default_library",
         "//test/proto3pb:go_default_library",
         "@com_github_golang_protobuf//proto:go_default_library",
         "@com_github_antlr//runtime/Go/antlr:go_default_library",
--- a/checker/checker.go
+++ b/checker/checker.go
@@ -193,11 +193,6 @@ func (c *checker) checkSelect(e *exprpb.Expr) {
 			messageType.GetMessageType(),
 			sel.Field); found {
 			resultType = fieldType.Type
-			// In proto3, primitive field types can't support presence testing, so the has()
-			// style operation would be invalid in this instance.
-			if sel.TestOnly && !fieldType.SupportsPresence {
-				c.errors.fieldDoesNotSupportPresenceCheck(c.location(e), sel.Field)
-			}
 		}
 	case kindTypeParam:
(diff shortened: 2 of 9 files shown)


JimLarson (approved):
Bravo for simplifying the field descriptor.

PR #334, opened 2020-04-16
Fix shared state in Env extension

### LINKED ISSUE #333: Env.Extend() result shares state
**Describe the bug**
This is a minor issue that prevents using a shared parent Env to build multiple child Envs.

Env.Extend() returns a new Env that shares the underlying slices of the original Env. If I extend the same Env more than once then previous extended Envs can be modified by the current one. This is because they append into the same slice.

**To Reproduce**

- [ ]  parser
- [ ]  checker
- [ ]  interpreter

Test setup:
```go
func TestEnvExtends(t *testing.T) {
	baseEnv, err := cel.NewEnv(cel.Declarations(
		decls.NewIdent("age", decls.Int, nil),
		decls.NewIdent("gender", decls.String, nil),
		decls.NewIdent("country", decls.String, nil)))
	if err != nil {
		t.Fatal(err)
	}

	env1, err := baseEnv.Extend(cel.Declarations(
		decls.NewIdent("name", decls.String, nil)))
	if err != nil {
		t.Fatal(err)
	}

	env2, err := baseEnv.Extend(cel.Declarations(
		decls.NewIdent("group", decls.String, nil)))
	if err != nil {
		t.Fatal(err)
	}

	if _, issues := env2.Compile("size(group) > 10"); issues.Err() != nil {
		t.Fatal(issues.Err())
	}

	// I expect this to work but env1 now has "group" instead of "name"
	if _, issues := env1.Compile("size(name) > 10"); issues.Err() != nil {
		t.Fatal(issues.Err())
	}
}
```
```
--- FAIL: TestEnvExtends (0.01s)
    cel_test.go:37: ERROR: <input>:1:6: undeclared reference to 'name' (in container '')
         | size(name) > 10
         | .....^
```

**Expected behavior**
In the test case I expect env1 to have the declarations of baseEnv + name and env2 to have the declarations of baseEnv + group.

I'd also expect similar behavior for other configurable slice fields.

**Additional context**
I'm reporting this as an Env.Extend() issue but it could also be solved if the EnvOptions made new copies of slices instead of appending.

TristonianJones:
Ensure that slices with the `Env` object are deep-copied during the `Env.Extend` call to ensure extensions are isolated from each other.

Closes #333

### DIFF
--- a/cel/cel_test.go
+++ b/cel/cel_test.go
@@ -705,9 +705,25 @@ func Test_EnvExtension(t *testing.T) {
 				decls.NewObjectType("google.api.expr.v1alpha1.Expr"), nil),
 		),
 	)
-	e2, _ := e.Extend()
+	e2, _ := e.Extend(
+		CustomTypeAdapter(types.DefaultTypeAdapter),
+		Types(&proto3pb.TestAllTypes{}),
+	)
 	if e == e2 {
 		t.Error("got object equality, wanted separate objects")
+	}
+	if e.TypeAdapter() == e2.TypeAdapter() {
+		t.Error("got the same type adapter, wanted isolated instances.")
+	}
+	if e.TypeProvider() == e2.TypeProvider() {
+		t.Error("got the same type provider, wanted isolated instances.")
+	}
+	e3, _ := e2.Extend()
+	if e2.TypeAdapter() != e3.TypeAdapter() {
+		t.Error("got different type adapters, wanted immutable adapter reference")
+	}
+	if e2.TypeProvider() == e3.TypeProvider() {
+		t.Error("got the same type provider, wanted isolated instances.")
 	}
 }
 
--- a/cel/cel_test.go
+++ b/cel/cel_test.go
@@ -730,9 +730,14 @@ func Test_EnvExtensionIsolation(t *testing.T) {
 	if _, issues := env2.Compile("size(group) > 10"); issues.Err() != nil {
 		t.Fatal(issues.Err())
 	}
-	// I expect this to work but env1 now has "group" instead of "name"
+	if _, issues := env2.Compile("size(name) > 10"); issues.Err() == nil {
+		t.Fatal("env2 contains 'name', but should not")
+	}
 	if _, issues := env1.Compile("size(name) > 10"); issues.Err() != nil {
 		t.Fatal(issues.Err())
+	}
+	if _, issues := env1.Compile("size(group) > 10"); issues.Err() == nil {
+		t.Fatal("env1 contains 'group', but should not")
 	}
 }
 
--- a/cel/cel_test.go
+++ b/cel/cel_test.go
@@ -709,6 +709,33 @@ func Test_EnvExtension(t *testing.T) {
 	}
 }
 
+func Test_EnvExtensionIsolation(t *testing.T) {
+	baseEnv, err := NewEnv(Declarations(
+		decls.NewIdent("age", decls.Int, nil),
+		decls.NewIdent("gender", decls.String, nil),
+		decls.NewIdent("country", decls.String, nil)))
+	if err != nil {
+		t.Fatal(err)
+	}
+	env1, err := baseEnv.Extend(Declarations(
+		decls.NewIdent("name", decls.String, nil)))
+	if err != nil {
+		t.Fatal(err)
+	}
+	env2, err := baseEnv.Extend(Declarations(
+		decls.NewIdent("group", decls.String, nil)))
+	if err != nil {
+		t.Fatal(err)
+	}
+	if _, issues := env2.Compile("size(group) > 10"); issues.Err() != nil {
+		t.Fatal(issues.Err())
+	}
+	// I expect this to work but env1 now has "group" instead of "name"
+	if _, issues := env1.Compile("size(name) > 10"); issues.Err() != nil {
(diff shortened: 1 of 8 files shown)


TristonianJones:
Thanks @mctofu for the detailed test case, that made the fix a snap to put together.

cel/cel_test.go
```diff
@@ -709,6 +709,33 @@ func Test_EnvExtension(t *testing.T) {
...
+	}
+	env2, err := baseEnv.Extend(Declarations(
+		decls.NewIdent("group", decls.String, nil)))
+	if err != nil {
+		t.Fatal(err)
+	}
+	if _, issues := env2.Compile("size(group) > 10"); issues.Err() != nil {
+		t.Fatal(issues.Err())
+	}
+	// I expect this to work but env1 now has "group" instead of "name"
```
JimLarson: Remove this comment now that you've fixed the bug.

cel/cel_test.go
```diff
@@ -709,6 +709,33 @@ func Test_EnvExtension(t *testing.T) {
...
+	if _, issues := env1.Compile("size(name) > 10"); issues.Err() != nil {
+		t.Fatal(issues.Err())
+	}
```
JimLarson, line 772: Also test that `env1` does not declare `group` and `env2` does not declare `name`.

cel/env.go
```diff
@@ -209,13 +209,20 @@ func (e *Env) Extend(opts ...EnvOption) (*Env, error) {
...
+	copy(decsCopy, e.declarations)
+	copy(macsCopy, e.macros)
+	copy(progOptsCopy, e.progOpts)
+
 	ext := &Env{
-		declarations:                   e.declarations,
+		declarations:                   decsCopy,
+		macros:                         macsCopy,
+		progOpts:                       progOptsCopy,
 		adapter:                        e.adapter,
```
JimLarson: Check that the adapter isn't mutable.

JimLarson (changes requested):

cel/cel_test.go
TristonianJones: Done.

cel/cel_test.go

cel/env.go
TristonianJones: It's an interface, so its mutability depends on the implementation. Our `ref.TypeRegistry` is inherently mutable, so I've added `Copy` methods to the `ref.TypeRegistry` as well as to the `pb.Db` to make it easier to ensure isolated memory. Note, the `pb.Db` copy is a shallow copy since the `FileDescription` and related objects are intended to be singletons.

TristonianJones:
PTAL

cel/env.go
```diff
@@ -205,18 +205,52 @@ func (e *Env) CompileSource(src common.Source) (*Ast, *Issues) {
...
+	// TypeRegistry as the base implementation are captured below.
+	if isAdapterReg && isProviderReg && adapterReg == providerReg {
+		reg := providerReg.Copy()
+		adapter = reg
+		provider = reg
+	} else if isProviderReg {
+		provider = providerReg.Copy()
+	} else if isAdapterReg {
+		adapter = adapterReg.Copy()
+	}
```
JimLarson, line 244: Looks like `adapter`, `provider`, or both can still be uninitialized/nil at this point.  Should they default to `e.adapter` and `e.provider`?  Or maybe the adapter and provider interfaces should include the Copy method?  (That would complicate the sharing.)

TristonianJones: Oops, I'd started with using e.adapter and e.provider as defaults and accidentally undid it. That's what I'd intended at least. Will fix

TristonianJones: Fixed and added a test case using an immutable adapter.

JimLarson (approved):

cel/env.go
```diff
@@ -205,18 +205,52 @@ func (e *Env) CompileSource(src common.Source) (*Ast, *Issues) {
...
+	// In most cases the provider and adapter will be a ref.TypeRegistry;
+	// however, in the rare cases where they are not, they are assumed to
+	// be immutable. Since it is possible to set the TypeProvider separately
+	// from the TypeAdapter, the possible configurations which could use a
+	// TypeRegistry as the base implementation are captured below.
+	if isAdapterReg && isProviderReg && adapterReg == providerReg {
+		reg := providerReg.Copy()
+		adapter = reg
+		provider = reg
+	} else if isProviderReg {
```
mctofu, line 240: if isAdapterReg && isProviderReg && adapterReg != providerReg should this be making a copy of both providerReg and adapterReg instead of just providerReg?

TristonianJones: The test here is saying that the adapter and provider are the same `ref.TypeRegistry` so only one copy is needed and desired. 

Thanks for following up!

mctofu: The case I was wondering about is if adapter and provider are both `ref.TypeRegistry` but different ones. I'm not sure if that's an expected case but it looks like we wouldn't copy both of them?

TristonianJones:
Ah, I see. I'll amend the code. Thanks

cel/env.go
TristonianJones: Ah, yes. You're right. I have been bouncing between a few different roles here at home and missed that case. It's possible for this to happen, but should be incredibly uncommon. I'll amend the code.

PR #341, opened 2020-04-22
Fix namespaced function resolution

TristonianJones:
Type checking correctly resolves namespaced function names like `base64.encode`; however, this information was being lost at evaluation time with the fragment `base64` being treated as a variable name and `encode` as the function name.

This PR fixes the glitch for namespaced function resolution ensuring the evaluator honors the CEL program container name and qualified function names.

### DIFF
--- a/checker/checker.go
+++ b/checker/checker.go
@@ -23,6 +23,7 @@ import (
 	"github.com/golang/protobuf/proto"
 	"github.com/google/cel-go/checker/decls"
 	"github.com/google/cel-go/common"
+	"github.com/google/cel-go/common/packages"
 	"github.com/google/cel-go/common/types/ref"
 
 	exprpb "google.golang.org/genproto/googleapis/api/expr/v1alpha1"
@@ -143,43 +144,42 @@ func (c *checker) checkNullLiteral(e *exprpb.Expr) {
 }
 
 func (c *checker) checkIdent(e *exprpb.Expr) {
+	// Rewrite the identifier to its fully qualified name.
 	identExpr := e.GetIdentExpr()
-	if ident := c.env.LookupIdent(identExpr.Name); ident != nil {
+	if ident := c.env.LookupIdent(identExpr.GetName()); ident != nil {
 		c.setType(e, ident.GetIdent().Type)
-		c.setReference(e, newIdentReference(ident.Name, ident.GetIdent().Value))
-		e.ExprKind = &exprpb.Expr_IdentExpr{
-			IdentExpr: &exprpb.Expr_Ident{
-				Name: ident.Name,
-			},
-		}
+		c.setReference(e, newIdentReference(ident.GetName(), ident.GetIdent().Value))
+		identExpr.Name = ident.GetName()
 		return
 	}
 
 	c.setType(e, decls.Error)
 	c.errors.undeclaredReference(
-		c.location(e), c.env.packager.Package(), identExpr.Name)
+		c.location(e), c.env.packager.Package(), identExpr.GetName())
 }
 
 func (c *checker) checkSelect(e *exprpb.Expr) {
 	sel := e.GetSelectExpr()
 	// Before traversing down the tree, try to interpret as qualified name.
-	qname, found := toQualifiedName(e)
+	qname, found := packages.ToQualifiedName(e)
 	if found {
 		ident := c.env.LookupIdent(qname)
 		if ident != nil {
 			if sel.TestOnly {
 				c.errors.expressionDoesNotSelectField(c.location(e))
 				c.setType(e, decls.Bool)
-			} else {
-				c.setType(e, ident.GetIdent().Type)
-				c.setReference(e,
-					newIdentReference(ident.Name, ident.GetIdent().Value))
-				identName := ident.Name
-				e.ExprKind = &exprpb.Expr_IdentExpr{
-					IdentExpr: &exprpb.Expr_Ident{
-						Name: identName,
-					},
-				}
+				return
+			}
+			// Rewrite the node to be a variable reference to the resolved fully-qualified
+			// variable name.
+			c.setType(e, ident.GetIdent().Type)
+			c.setReference(e,
+				newIdentReference(ident.GetName(), ident.GetIdent().Value))
+			identName := ident.GetName()
+			e.ExprKind = &exprpb.Expr_IdentExpr{
+				IdentExpr: &exprpb.Expr_Ident{
+					Name: identName,
+				},
 			}
 			return
 		}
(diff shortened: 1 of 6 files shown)


interpreter/planner.go
```diff
@@ -654,3 +642,69 @@ func (p *planner) constValue(c *exprpb.Constant) (ref.Val, error) {
...
+		_, found := p.disp.FindOverload(qualifiedName)
+		if found {
+			fnName = qualifiedName
+			return
+		}
+	}
+	return
+}
+
+func (p *planner) toQualifiedName(operand *exprpb.Expr) (string, bool) {
```
JimLarson, line 727: Doc comment, please.

TristonianJones:
@JimLarson PTAL, the PR now introduces AST rewrites at Check time to drop / consolidate parse-time AST fragments in favor of ASTs that more accurately represent the eval intent.

Note, while the AST is modified, the following features have been preserved to ensure that generated ASTs are compatible with existing evaluators:
- overload_id names remain the same and are preserved in the `CheckedExpr.reference_map`
   (Go, Java)
- qualified function names are packed into the `CallExpr.function` field without a leading '.' as the
   names are expected to be fully-qualified post-check. C++ does not consider the leading '.'
   during name resolution. (Go, C++)
- qualified identifier names are packed into an `IdentExpr.name` field without a leading '.' as the
   names are expected to be fully-qualified post-check. This may result in slightly more resolution
   cost for evaluators that only consider the parsed representation and attempt field resolution
   for backwards compatibility (C++)
- qualified type names are packed into the `CreateStructExpr..message_name`.

In most cases, the net effect will be simpler, more consistent evaluation of checked expressions within evaluators, even when the evaluator ignores the checker metadata produced in the reference_map and type_map.

Once submitted, I'm going to apply a v0.5.0 release label to indicate the potential for a breaking change in the AST representation.

TristonianJones:
PTAL whenever you have a chance @JimLarson. Thanks.

checker/checker.go
```diff
@@ -238,14 +250,17 @@ func (c *checker) checkCall(e *exprpb.Expr) {
 		if qname, found := toQualifiedName(call.Target); found {
 			fn := c.env.LookupFunction(qname + "." + call.Function)
 			if fn != nil {
+				// The function name is actually namespaced and so the target operand would
+				// be an inaccurate representation of the desired evaluation behavior.
+				call.Target = nil
+				call.Function = fn.GetName()
 				resolution = c.resolveOverload(c.location(e), fn, nil, call.Args)
 			}
```
JimLarson: Note on the "else" path that we do not allow qualified names for receiver-style functions, so for `a.b.c()` either `a.b.c` is a function, or `c` is a function with target `a.b`, but it's never the case that `b.c` is a function with target `a`.

checker/checker.go
```diff
@@ -147,6 +147,11 @@ func (c *checker) checkIdent(e *exprpb.Expr) {
 	if ident := c.env.LookupIdent(identExpr.Name); ident != nil {
 		c.setType(e, ident.GetIdent().Type)
 		c.setReference(e, newIdentReference(ident.Name, ident.GetIdent().Value))
+		e.ExprKind = &exprpb.Expr_IdentExpr{
```
JimLarson: Comment "Rewrite the node to include the resolved fully-qualified variable name."

Since it's already an Expr_Ident node, can't we just overwrite the Name field?

checker/checker.go
```diff
@@ -169,6 +174,12 @@ func (c *checker) checkSelect(e *exprpb.Expr) {
 				c.setType(e, ident.GetIdent().Type)
 				c.setReference(e,
 					newIdentReference(ident.Name, ident.GetIdent().Value))
+				identName := ident.Name
+				e.ExprKind = &exprpb.Expr_IdentExpr{
```
JimLarson: Comment "Rewrite the node to be a variable reference to the resolved fully-qualified variable name."

checker/checker.go
```diff
@@ -228,6 +239,7 @@ func (c *checker) checkCall(e *exprpb.Expr) {
 	if call.Target == nil {
 		// Regular static call with simple name.
 		if fn := c.env.LookupFunction(call.Function); fn != nil {
+			call.Function = fn.Name
```
JimLarson: Comment "Overwrite with fully-qualified resolved function name".

checker/checker.go
```diff
@@ -238,14 +250,17 @@ func (c *checker) checkCall(e *exprpb.Expr) {
 		if qname, found := toQualifiedName(call.Target); found {
 			fn := c.env.LookupFunction(qname + "." + call.Function)
 			if fn != nil {
+				// The function name is actually namespaced and so the target operand would
+				// be an inaccurate representation of the desired evaluation behavior.
+				call.Target = nil
+				call.Function = fn.GetName()
```
JimLarson: Be consistent with `fn.GetName()` vs `fn.Name` above.

interpreter/planner.go
```diff
@@ -555,15 +552,7 @@ func (p *planner) planCreateStruct(expr *exprpb.Expr) (Interpretable, error) {
...
-	typeName := obj.MessageName
-	var defined bool
-	for _, qualifiedTypeName := range p.pkg.ResolveCandidateNames(typeName) {
-		if _, found := p.provider.FindType(qualifiedTypeName); found {
-			typeName = qualifiedTypeName
-			defined = true
-			break
-		}
-	}
+	typeName, defined := p.resolveTypeName(obj.MessageName)
```
JimLarson, line 555: So this all looks okay, but it's largely duplicating the logic in the checker.  I would much rather see the planner and the checker both call a separate pass of walking the AST to do the qualified identifier rewrites, with a pluggable qualified name tester that checks either declarations (checker) or bound variables and dispatchable functions (planner).  Yet another function can rewrite a CheckedExpr based on its reference map, returning early if it sees that the rewrite has already been performed.

Oh, nasty thought - can the iteration variable in a comprehension (fold) be a qualified name?  What if we are in container "a.b" and write `mylist.exists(a.b.c, c % 2 == 0)`?

JimLarson (changes requested):

interpreter/planner.go
TristonianJones: The comprehensions aren't permitted to have a qualified iteration variable (thank goodness). 

I've restructured the name qualification code to be right next to the `packager`, but it's a bit painful to disentangle the rest into a common unit. I could do as you suggest, but it seems easier to understand the similarities and differences by mirroring the conventions used in each file. At any rate, let me know what you think of the latest update.

checker/checker.go
```diff
@@ -143,32 +144,41 @@ func (c *checker) checkNullLiteral(e *exprpb.Expr) {
 }
 
 func (c *checker) checkIdent(e *exprpb.Expr) {
+	// Rewrite the identifier to its fully qualified name.
```
JimLarson: Consider putting this comment at line 150, in the block where we're actually doing this rewrite.

checker/checker.go
```diff
@@ -217,50 +227,81 @@ func (c *checker) checkSelect(e *exprpb.Expr) {
...
-			fn := c.env.LookupFunction(qname + "." + call.Function)
-			if fn != nil {
-				resolution = c.resolveOverload(c.location(e), fn, nil, call.Args)
-			}
+				c.location(e), c.env.packager.Package(), fnName)
+			c.setType(e, decls.Error)
+			return
 		}
+		// Check to see whether the overload resolves.
+		call.Function = fn.GetName()
```
JimLarson, line 255: Comment "Overwrite function name with fully-qualified resolved name."

JimLarson (approved):

PR #345, opened 2020-04-30
Fix the doc comments referring to the extant ClearBuiltins() option

### LINKED ISSUE #344: cel.ClearBuiltins was removed but doc comments still reference it
**Describe the bug**

The doc comments for [`cel.ClearMacros`](https://pkg.go.dev/github.com/google/cel-go/cel?tab=doc#ClearMacros), [`cel.Declarations`](https://pkg.go.dev/github.com/google/cel-go/cel?tab=doc#Declarations), and [`cel.Macros`](https://pkg.go.dev/github.com/google/cel-go/cel?tab=doc#Macros) mention `cel.ClearBuiltins` which was replaced by `cel.NewCustomEnv` in #311 

**Expected behavior**

These doc comments shouldn’t have references to the removed function. They can simply be deleted.

Perhaps it would be helpful to mention `NewCustomEnv`, but I would suggest that this be an `ExampleCustomEnv`.

TristonianJones:
Doc comments have been updated, but no additional godoc level `Example` tests have been added yet because I'd like the example to be meaningful and most of the tests around custom environments are toy examples. 

When #290 is implemented, I can easily see the `NewCustomEnv` warranting its own example as it will be much easier to subset CEL. Comment added to the FR #290 to this effect.

Closes #344

### DIFF
--- a/cel/options.go
+++ b/cel/options.go
@@ -65,9 +65,9 @@ func CustomTypeProvider(provider ref.TypeProvider) EnvOption {
 
 // Declarations option extends the declaration set configured in the environment.
 //
-// Note: Declarations will by default be appended to the pre-existing declaration set cnofigured
-// for the environment. The cel.NewEnv call builds on top of the standard CEL declarations. For
-// a purely custom set of declarations use cel.NewCustomEnv.
+// Note: Declarations will by default be appended to the pre-existing declaration set configured
+// for the environment. The NewEnv call builds on top of the standard CEL declarations. For a
+// purely custom set of declarations use NewCustomEnv.
 func Declarations(decls ...*exprpb.Decl) EnvOption {
 	// TODO: provide an alternative means of specifying declarations that doesn't refer
 	// to the underlying proto implementations.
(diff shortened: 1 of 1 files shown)


cel/options.go
```diff
@@ -68,7 +65,9 @@ func CustomTypeProvider(provider ref.TypeProvider) EnvOption {
 
 // Declarations option extends the declaration set configured in the environment.
 //
-// Note: This option must be specified after ClearBuiltIns if both are used together.
+// Note: Declarations will by default be appended to the pre-existing declaration set cnofigured
```
sfllaw: Spelling:
```suggestion
// Note: Declarations will by default be appended to the pre-existing declaration set configured
```

cel/options.go
```diff
@@ -68,7 +65,9 @@ func CustomTypeProvider(provider ref.TypeProvider) EnvOption {
...
+// for the environment. The cel.NewEnv call builds on top of the standard CEL declarations. For
+// a purely custom set of declarations use cel.NewCustomEnv.
```
sfllaw: This is the only place in the documentation where you use the namespaced name in the doc comment. Since this function also lives in `cel`, perhaps it is unnecessary?

```suggestion
// for the environment. The NewEnv call builds on top of the standard CEL declarations. For
// a purely custom set of declarations use NewCustomEnv.
```

sfllaw (commented):
It seems that you might have missed this mention:
https://github.com/TristonianJones/cel-go/blob/doc-fix/cel/options.go#L95

TristonianJones:
@sfllaw Thanks for the review.

JimLarson (approved):

Issue #346, opened 2020-04-30
Checker performance improvements


TristonianJones:
In assessing Checker performance there were three factors which dominated performance:
- `proto.Equal` (46%)
- ANTLR Lexer / Parser initialization (33%)
- checker declaration conflict detection (17%)
- other (4%)

The dominant use of `proto.Equal` was actually introduced in #250 to fix an infinite recursion bug. The second issue was introduced in #177 to address data races within the ANTLR runtime. 

Upon further review, the use of `proto.Equal` in #250 could be much more narrowly scoped and the substitution logic simplified, thus significantly improving the expression compilation performance within larger applications like https://github.com/google/cel-policy-templates-go.

Before:
```
BenchmarkCompiler_Template/canonical-8         	      12	  96157074 ns/op	26354824 B/op	  683280 allocs/op
```

After:
```
BenchmarkCompiler_Template/canonical-8         	      19	  57289719 ns/op	21251336 B/op	  407321 allocs/op
```

TristonianJones:
Note, I've submitted PR https://github.com/antlr/antlr4/pull/2816 to make ANTLR's Go codegen like what we've hacked together in CEL, and to export a method that will allow us to use ANTLR Lexer / Parser objects within a `sync.Pool` in order amortize the construction cost while still being thread-safe.

TristonianJones:
The bulk of performance improvements came with #520

PR #347, opened 2020-04-30
Copy validated declarations between checker.Env instances on extension

TristonianJones:
This change introduces several small changes to improve environment extension and CEL bootstrapping.

### Type-substitution Utility

Introduce a common helper method for performing type substitution checks. The helper method scopes
the use of `proto.Equal` within `internalIsAssignable` to only the cases where recursive type substitutions
occur. All other branches progressively narrow the types under consideration and use simpler checks
appropriate for the protobuf type.

This change improves check-time performance roughly 40% per issue #346. 

#### `EagerlyValidateDeclarations` Option

When this option is enabled, a guaranteed to compile expression is parsed and type-checked ensuring that
the internal instance of the `checker.Env` is initialized and all declarations provided as options to the
`cel.NewEnv()` call are checked to ensure there are no overlapping / colliding function declarations. 

Using this option in combination with `Extend` also makes it possible to copy the already validated
declarations into the cloned `checker.Env` of the extended environment. The cloned declarations are
shallowly copied on `checker.Env` initialization. If a declaration provided to the derived environment
needs to be merged with a function type declared in an ancestor environment, the ancestor function
is copied prior to being modified to ensure changes are isolated to the derived environment.

#### NewEnv() Extends by Default

The `NewEnv()` call now extends a globally initialized copy of the CEL standard environment whose
declarations have been validated as part of an `init()` call. This means in the common case that when
only a few declarations or parameters are changed as part of the CEL standard environment, the internal
`checker.Env` setup cost is kept to a minimum.

#### Filtered Overloads

The support for `CrossNumericTypes` was filtering overloads on addition to the `checker.Env`. This
behavior has now been shifted to filtering on access of the overload set within the type-checker.
Without this change it was impossible to toggle `CrossNumericTypes` support between `Extend`
calls.

#### Performance

The performance impact of these changes is dramatic, dropping the cost of the simple case by >95%

```
benchmark                          old ns/op     new ns/op     delta
BenchmarkNewEnvLazy-8              14748         13535         -8.22%
BenchmarkNewEnvEager-8             2506297       61987         -97.53%
BenchmarkEnvExtendEager-8          2658853       62218         -97.66%
BenchmarkEnvExtendEagerTypes-8     2742668       65600         -97.61%
BenchmarkEnvExtendEagerDecls-8     2650621       102302        -96.14%

benchmark                          old allocs     new allocs     delta
BenchmarkNewEnvLazy-8              97             39             -59.79%
BenchmarkNewEnvEager-8             20283          209            -98.97%
BenchmarkEnvExtendEager-8          20300          204            -99.00%
BenchmarkEnvExtendEagerTypes-8     20205          221            -98.91%
BenchmarkEnvExtendEagerDecls-8     20409          474            -97.68%

benchmark                          old bytes     new bytes     delta
BenchmarkNewEnvLazy-8              10489         7634          -27.22%
BenchmarkNewEnvEager-8             578535        18031         -96.88%
BenchmarkEnvExtendEager-8          581087        17801         -96.94%
BenchmarkEnvExtendEagerTypes-8     576225        20088         -96.51%
BenchmarkEnvExtendEagerDecls-8     579028        27177         -95.31%
```

### DIFF
--- a/cel/env.go
+++ b/cel/env.go
@@ -109,6 +109,10 @@ type Env struct {
 // See the EnvOption helper functions for the options that can be used to configure the
 // environment.
 func NewEnv(opts ...EnvOption) (*Env, error) {
+	// Extend the statically configured standard environment, disabling eager validation to ensure
+	// the cost of setup for the environment is still just as cheap as it is in v0.11.x and earlier
+	// releases. The user provided options can easily re-enable the eager validation as they are
+	// processed after this default option.
 	stdOpts := append([]EnvOption{EagerlyValidateDeclarations(false)}, opts...)
 	return stdEnv.Extend(stdOpts...)
 }
--- a/cel/env.go
+++ b/cel/env.go
@@ -155,29 +155,8 @@ func (e *Env) Check(ast *Ast) (*Ast, *Issues) {
 	pe, _ := AstToParsedExpr(ast)
 
 	// Construct the internal checker env, erroring if there is an issue adding the declarations.
-	e.chkOnce.Do(func() {
-		chkOpts := []checker.Option{}
-		chkOpts = append(chkOpts, e.chkOpts...)
-		chkOpts = append(chkOpts,
-			checker.HomogeneousAggregateLiterals(
-				e.HasFeature(featureDisableDynamicAggregateLiterals)),
-			checker.CrossTypeNumericComparisons(
-				e.HasFeature(featureCrossTypeNumericComparisons)))
-
-		ce, err := checker.NewEnv(e.Container, e.provider, chkOpts...)
-		if err != nil {
-			e.chkErr = err
-			return
-		}
-		err = ce.Add(e.declarations...)
-		if err != nil {
-			e.chkErr = err
-			return
-		}
-		e.chk = ce
-	})
-	// The once call will ensure that this value is set or nil for all invocations.
-	if e.chkErr != nil {
+	err := e.initChecker()
+	if err != nil {
 		errs := common.NewErrors(ast.Source())
 		errs.ReportError(common.NoLocation, e.chkErr.Error())
 		return nil, NewIssues(errs)
@@ -466,13 +445,38 @@ func (e *Env) configure(opts []EnvOption) (*Env, error) {
 	// The simplest way to eagerly validate declarations on environment creation is to compile
 	// a dummy program and check for the presence of e.chkErr being non-nil.
 	if e.HasFeature(featureEagerlyValidateDeclarations) {
-		e.Compile("'validate'")
-		if e.chkErr != nil {
-			return nil, e.chkErr
+		err := e.initChecker()
+		if err != nil {
+			return nil, err
 		}
 	}
 
 	return e, nil
+}
+
+func (e *Env) initChecker() error {
+	e.chkOnce.Do(func() {
+		chkOpts := []checker.Option{}
+		chkOpts = append(chkOpts, e.chkOpts...)
+		chkOpts = append(chkOpts,
+			checker.HomogeneousAggregateLiterals(
+				e.HasFeature(featureDisableDynamicAggregateLiterals)),
+			checker.CrossTypeNumericComparisons(
+				e.HasFeature(featureCrossTypeNumericComparisons)))
+
+		ce, err := checker.NewEnv(e.Container, e.provider, chkOpts...)
+		if err != nil {
+			e.chkErr = err
+			return
+		}
+		err = ce.Add(e.declarations...)
+		if err != nil {
+			e.chkErr = err
+			return
+		}
+		e.chk = ce
+	})
+	return e.chkErr
 }
 
 // Issues defines methods for inspecting the error details of parse and check calls.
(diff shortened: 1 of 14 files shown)


checker/types.go
```diff
@@ -265,6 +248,32 @@ func internalIsAssignable(m *mapping, t1 *exprpb.Type, t2 *exprpb.Type) bool {
 	}
 }
 
+// isValidTypeSubstitution returns whether t2 (or its type substitution) is a valid type
+// substitution for t1.
```
JimLarson: ...and whether t2 has a type substitution in mapping `m`.

TristonianJones:
@JimLarson Do you think you'd be able to take a look at this? I realize more extensive refactors are probably necessary, but looking this change again, it does make the code easier to decipher (imo).

JimLarson (approved):
I don't think I'll be able to reload the state in my head to make sense of this anytime soon, so approving to get it off my review list and unblock you.

checker/env.go
```diff
@@ -99,6 +101,9 @@ func NewEnv(container *containers.Container, provider ref.TypeProvider, opts ...
 	if envOptions.crossTypeNumericComparisons {
 		filteredOverloadIDs = make(map[string]struct{})
 	}
+	if envOptions.validatedDeclarations != nil {
+		declarations = envOptions.validatedDeclarations.Copy()
```
jpbetz, line 105: It's not possible with protos to somehow have a "copy on write" type here is it?  Being able to share the declarations across multiple environments would be a potentially use memory optimization here...

checker/env.go
```diff
@@ -212,14 +217,14 @@ func (e *Env) addFunction(decl *exprpb.Decl) []errorMsg {
 	if current == nil {
```
jpbetz, line 217: rename this surrounding function to `setFunction` as well?

checker/env.go
```diff
@@ -212,14 +217,14 @@ func (e *Env) addFunction(decl *exprpb.Decl) []errorMsg {
...
+		// Copy on write since we don't know where this original definition came from.
+		current = proto.Clone(current).(*exprpb.Decl)
 	}
+	e.declarations.SetFunction(current)
 
 	errorMsgs := make([]errorMsg, 0)
 	for _, overload := range decl.GetFunction().GetOverloads() {
-		if _, found := e.filteredOverloadIDs[overload.GetOverloadId()]; found {
-			continue
-		}
```
jpbetz, line 222: Just to double check. Removing this code is intentional because it allows for set semantics here, right?

jpbetz (commented):
LGTM

I added a few comments, mostly questions and one optional change.

The test coverage and benchmarks look solid. This covers the use case needed by Kubernetes really well.

I did not review the checker/types.go. I don't feel like I understand the specifics well enough there to be helpful. All the validatedDeclarations code made sense to me.

checker/env.go
TristonianJones: In this case only a shallow copy is made of the map containing the protos. The actual protos themselves for function declarations are `proto.Clone`d if there's an overload being added to a function that's already in the declaration set.

checker/env.go
TristonianJones: I simply moved the filtering to the `checker.resolveOverloadOrErr` calls, ensuring that disabled (filtered) overloads are not in the set considered during type-checking.

checker/env.go
TristonianJones: Done.

cel/env.go
```diff
@@ -432,6 +462,16 @@ func (e *Env) configure(opts []EnvOption) (*Env, error) {
 	if err != nil {
 		return nil, err
 	}
+
```
jcking, line 444: This seems...hacky. But okay.

checker/env.go
```diff
@@ -236,6 +241,13 @@ func (e *Env) addIdent(decl *exprpb.Decl) errorMsg {
 	return ""
 }
 
+func (e *Env) isOverloadDisabled(overloadID string) bool {
+	if _, found := e.filteredOverloadIDs[overloadID]; found {
```
jcking: Just return found and get rid of the if?

checker/types.go
```diff
@@ -180,43 +165,26 @@ func isEqualOrLessSpecific(t1 *exprpb.Type, t2 *exprpb.Type) bool {
 
 /// internalIsAssignable returns true if t1 is assignable to t2.
 func internalIsAssignable(m *mapping, t1 *exprpb.Type, t2 *exprpb.Type) bool {
-	// A type is always assignable to itself.
-	// Early terminate the call to avoid cases of infinite recursion.
-	if proto.Equal(t1, t2) {
-		return true
-	}
 	// Process type parameters.
 	kind1, kind2 := kindOf(t1), kindOf(t2)
```
jcking, line 169: Is there a need just check if kind1 == kind2 early? Not sure if covered already by the caller.

cel/env.go
TristonianJones: Ah, yes, the compile is a bit hacky. Let me just do the checker construction here.

checker/types.go
TristonianJones: Done.

checker/env.go
TristonianJones: Done.

jcking (approved):

TristonianJones:
@jpbetz Once the release is cut would you mind just updating the #519 ticket with the updated performance graph?

Closes #346

jpbetz:
> @jpbetz Once the release is cut would you mind just updating the #519 ticket with the updated performance graph?

I'd be happy to.

PR #354, opened 2020-05-12
Implement a feature flag mechanism.

### LINKED ISSUE #353: Extensible boolean feature flags
**Feature request checklist**

**Change**
As we begin to flag-protect behavior changes we need a more convenient mechanism for controlling optional behavior. Rather than adding ad-hoc boolean fields everywhere, we'd like to have a generic structure to query with enumerated simple feature flag values.

Feature flags should indicate legacy or optional behavior, with standard behavior being the flag-less default.  Flags must be explicitly set as options.

**Example**
Instead of

```
env := cel.NewEnv(cel.CustomOption1(), cel.CustomOption2())
...
if env.customOption1 {
    option_behavior()
} else {
    nromal_behavior()
}
return
```

do the following:

```
env := cel.NewEnv(cel.Features(cel.FEATURE_OPT_1, cel.FEATURE_OPT_2))
...
if env.HasFeature(cel.FEATURE_OPT_1) {
    overrides_normal_behvior()
    return
}
normal_behavior_not_indented()
return
```

**Alternatives considered**
Considered having the default be legacy behavior, rather than the desired standard behavior, but this rewards inertia. Instead, applications which wish to update their software but retain old behavior will be saddled with the responsibility of maintaining the feature flags they use.

JimLarson:
Closes #353

### DIFF
--- a/cel/options.go
+++ b/cel/options.go
@@ -29,6 +29,20 @@ import (
 	exprpb "google.golang.org/genproto/googleapis/api/expr/v1alpha1"
 )
 
+// These constants beginning with FEATURE_ enable optional behavior in
+// the library.  See the documentation for each constant to see its
+// effects, compatibility restrictions, and standard conformance.
+const (
+	_ = iota
+
+	// Disallow heterogeneous aggregate (list, map) literals.
+	// Note, it is still possible to have heterogeneous aggregates when
+	// provided as variables to the expression, as well as via conversion
+	// of well-known dynamic types, or with unchecked expressions.
+	// Affects checking.  Provides a subset of standard behavior.
+	FEATURE_DISABLE_DYNAMIC_AGGREGATE_LITERALS
+)
+
 // EnvOption is a functional interface for configuring the environment.
 type EnvOption func(e *Env) (*Env, error)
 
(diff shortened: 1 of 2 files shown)


cel/options.go
```diff
@@ -29,6 +29,20 @@ import (
...
+// effects, compatibility restrictions, and standard conformance.
+const (
+	_ = iota
+
+	// Disallow heterogeneous aggregate (list, map) literals.
+	// Note, it is still possible to have heterogeneous aggregates when
+	// provided as variables to the expression, as well as via conversion
+	// of well-known dynamic types, or with unchecked expressions.
+	// Affects checking.  Provides a subset of standard behavior.
+	FEATURE_DISABLE_DYNAMIC_AGGREGATE_LITERALS
```
TristonianJones: This was counter-intuitive to me too, but the convention within the Go standard libraries for exported constants is `CamelCase`, and I believe that's the convention used elsewhere within CEL.

cel/options.go
```diff
@@ -29,6 +29,20 @@ import (
 	exprpb "google.golang.org/genproto/googleapis/api/expr/v1alpha1"
 )
 
+// These constants beginning with FEATURE_ enable optional behavior in
+// the library.  See the documentation for each constant to see its
+// effects, compatibility restrictions, and standard conformance.
+const (
+	_ = iota
```
TristonianJones, line 36: You could collapse these two lines, but I leave it to you to decide what's most readable here:

```
FeatureDisableDynamicAggregateLiterals = iota + 1
```

TristonianJones (changes requested):

cel/options.go
JimLarson: Done.

cel/options.go
JimLarson: I'll keep it as-is so the first definition looks like the rest, and so we can freely reorder them.

JimLarson:
PTAL

TristonianJones (approved):

Issue #359, opened 2020-05-22
Check the timestamp format at compile time


lopezator:
Correct me if I'm wrong, but the only way to know if the timestamp format is right, is to evaluate the expression?

Wouldn't be nice to know that before, at parse/check time?

e.g.

```go
func TestTimestamp(t *testing.T) {
	// create env
	env, err := cel.NewEnv(
		cel.Declarations(
			decls.NewVar("date", decls.Timestamp)))
	if err != nil {
		t.Error(err)
	}

	// compile (parse+check)
	_, issues := env.Compile(`date > timestamp('wrong timestamp')`)
	if issues != nil && issues.Err() != nil {
		t.Fatal("I want to err here!")
	}
}
```

We are using parse+check to generate an AST, using that AST as an input to a filter component, who generates SQL from it.

This would allow us to support timestamp without having to evalute the expression just for that, or introducing custom code to post-check the AST, if the format passed to `timestamp("wrong")` is wrong.

I suppose this could apply for duration as well.

TristonianJones:
Hi @lopezator thanks for filing the request. There's sort of a general category of lint-style checks that we've been thinking about adding for CEL. Checking string literal inputs to regexes, timestamps, durations, etc. definitely fall into this category. 

These sort of checks can be a bit brittle when new functionality is introduced which is often why they're considered warnings rather than hard errors; however, I can see the cases you mentioned being errors.

I'm happy to consider this a future enhancement, but I don't have a delivery timeline in mind just yet.

lopezator:
Thank you for your response @TristonianJones !!

I think we could apply that logic in our end the meantime, I just wanted to be sure I wasn't missing something obvious.

Just to be sure, what do you exactly mean by warnings? Do CEL return any warning in use cases like this? How and where?

Thank you for your work in this exceptional library, it's being very useful for us.

Best.

TristonianJones:
@lopezator This is a good feature request, so thank you for filing it. You're not missing anything obvious at all. 

The `cel.Issues` object has an `Err()` method exposed which can be tested to check for problems in the syntactic or semantic correctness. The object also contains a list of errors, but in theory could contain a list of non-error severity issues such as warnings and deprecation notices. It's very likely the literal checks you mentioned would still have error severity, but I can see other cases where a less severe issue should be returned and then the application, such as yours, could decide whether the warning should be a hard-stop fail or an intended use of the library.

[...]

PR #360, opened 2020-05-26
Optimize type conversion functions. 

TristonianJones:
Unary type-conversion functions with literal arguments should be evaluated
eagerly when the `cel.EvalOptions(cel.OptOptimize)` flag is set. The conversion
will also ensure that if an error is encountered when evaluating these functions,
such as when constructing timestamps or durations from strings, the evaluation
will result in an error in the `env.Program` call.

This is not a perfect fix for #359, but it can be used to assist with correctness checks.

### DIFF
--- a/common/overloads/overloads.go
+++ b/common/overloads/overloads.go
@@ -280,6 +280,7 @@ func IsTypeConversionFunction(function string) bool {
 		TypeConvertBytes,
 		TypeConvertDouble,
 		TypeConvertDuration,
+		TypeConvertDyn,
 		TypeConvertInt,
 		TypeConvertString,
 		TypeConvertTimestamp,
--- a/common/overloads/overloads.go
+++ b/common/overloads/overloads.go
@@ -271,3 +271,22 @@ const (
 	HasNext  = "@hasNext"
 	Next     = "@next"
 )
+
+// IsTypeConversionFunction returns whether the input function is a standard library type
+// conversion function.
+func IsTypeConversionFunction(function string) bool {
+	switch function {
+	case TypeConvertBool,
+		TypeConvertBytes,
+		TypeConvertDouble,
+		TypeConvertDuration,
+		TypeConvertInt,
+		TypeConvertString,
+		TypeConvertTimestamp,
+		TypeConvertType,
+		TypeConvertUint:
+		return true
+	default:
+		return false
+	}
+}
--- a/interpreter/interpreter_test.go
+++ b/interpreter/interpreter_test.go
@@ -33,6 +33,7 @@ import (
 	"github.com/google/cel-go/interpreter/functions"
 	"github.com/google/cel-go/parser"
 
+	dpb "github.com/golang/protobuf/ptypes/duration"
 	structpb "github.com/golang/protobuf/ptypes/struct"
 	tpb "github.com/golang/protobuf/ptypes/timestamp"
 	wrapperspb "github.com/golang/protobuf/ptypes/wrappers"
@@ -277,11 +278,6 @@ var (
 				"c": []string{"hello"},
 			},
 			out: types.False,
-		},
-		{
-			name: "in_empty_list",
-			expr: `6 in []`,
-			out:  types.False,
 		},
 		{
 			name: "in_list",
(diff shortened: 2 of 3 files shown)


glerchundi:
This is great! Correct me if i'm wrong, this implementation would contract the AST from a `timestamp('correct-ts-value')` into `*exprpb.Constant_TimestampValue`?

lopezator:
Yahoo! 🎉

TristonianJones:
@glerchundi well, not exactly. The code will convert from `timestamp(<rfc_3339_str>)` to a CEL `ref.Val`. The AST is left unmodified which is why this is an imperfect fix for your needs. You could use it as a work-around in the short term.

glerchundi:
Ok, thanks anyway :D. Good job!

TristonianJones:
@glerchundi I'll eventually get to supporting linting. This will at least tell you that the checked-AST has `timestamp(<str>)` and `duration(<str>)` calls with well-formed string inputs. It's not perfect, but it should meet your needs in the short-term. Just make sure to turn on optimization.

I have a few more tests that I need to add to catch malformed inputs at optimization time, but then I'll send this out for review.

glerchundi:
I guess it does partially.

Currently we're parsing the expression and the walking the AST to build the SQL expression. Having an evaluation process where it executes the corresponding sanity check before proceeding sounds good but it's not enough as we need to continue walking the unmodified AST and taking care of the call expressions in it (like `timestamp`). 

Ideally I would expect `OptOptimize` to construct (i don't care the mechanism `newAst := PruneAst(details, oldAst)`) a cleaned up tree in order to avoid dealing with convert overloads on our side.

We're already doing it but makes the implementation more complicated that I would like it to be.

glerchundi:
Does your decorator implementation in #363 helps converting/contracting `ref.Val` (with `timestamp(const)`) into another `ref.Val` (with timestamp_const_value). I must admit that i'm a bit lost there :D

TristonianJones:
Hi @glerchundi, #363 is a decorator that participates in an AST walk done before evaluation. It's not really what you're looking for since it doesn't modify the AST; however, what I could do is expose a `ParserDecorator` which would allow you to observe, alter, or replace the `Expr` AST nodes at parse time. It would at least simplify your existing logic.

common/overloads/overloads.go
```diff
@@ -271,3 +271,22 @@ const (
...
+	switch function {
+	case TypeConvertBool,
+		TypeConvertBytes,
+		TypeConvertDouble,
+		TypeConvertDuration,
+		TypeConvertInt,
+		TypeConvertString,
+		TypeConvertTimestamp,
+		TypeConvertType,
+		TypeConvertUint:
```
JimLarson, line 288: Consider whether TypeConvertDyn should also give a positive answer. If not, add a comment to explain the discrepancy versus the list of `TypeConvert*` constants above.

Note that the language spec lacks type conversion functions for `bool()` and `type()`, while it contains `list()` and `map()` conversions. Should the former be added to the spec?  Should the latter be added to the list here?  Looks like we implement a string-to-bool conversion that's not in the spec.

interpreter/interpreter_test.go
```diff
@@ -278,6 +278,11 @@ var (
 			},
 			out: types.False,
 		},
+		{
+			name: "in_empty_list",
+			expr: `6 in []`,
+			out:  types.False,
+		},
```
JimLarson: This test seems unrelated to this PR - maybe move to another PR?

interpreter/interpreter_test.go
```diff
@@ -1164,6 +1169,49 @@ func TestInterpreter_MissingIdentInSelect(t *testing.T) {
 	}
 }
 
+func TestInterpreter_TypeConversionOpt_Errors(t *testing.T) {
```
JimLarson: Add some positive tests that ensure that the optimization actually happens.

JimLarson (changes requested):

common/overloads/overloads.go
TristonianJones: It looks like `list()` and `map()` type conversions were never implemented. I'm actually having a hard time making sense of what they were intended to do. :/ They sort of look like identity functions:

```
list(T) -> list(T) // ???
```

With respect to the places where cel-go goes beyond the spec, the type conversions are usually pretty reasonable. I may have picked a strange name for `type(uint)` as `TypeConvertType`, but it's the standard type function: `type(uint) == type`, `type(null_type) == type`. Stuff like that.

I'll add support for `dyn()` as well.

interpreter/interpreter_test.go
TristonianJones: Ah, sure thing. It must have gotten picked up when I was looking over coverage.

common/overloads/overloads.go
TristonianJones: Per https://github.com/google/cel-spec/issues/123, I'll forego the `list()` and `map()` conversion functions, but will add `dyn()`. 

With respect to `bool()` I believe the identity function should be supported and I'm not opposed to supporting `string(bool("true")) == "true"` since this seems like reasonable round-tripping behavior even if its infrequently used.

TristonianJones:
PTAL

glerchundi:
That is a fantastic idea @TristonianJones!

I'll create an issue asking for exactly that and in case we found a free time slot & we see ourselves with the capacity to make it happen without messing the code we'll send a PR.

JimLarson (approved):

PR #361, opened 2020-05-26
The type map references when keyed by attribute may be incorrect.

### LINKED ISSUE #364: Attribute ids may result in sub-optimal proto field resolution
**Describe the bug**
The CEL `planner.go` periodically references the `checked.proto` reference map and type map to optimize the program plan. The keys to these maps are the expression identifiers related to the operation in question. When `Attribute` values were introduced, it was possible for the wrote id to be used with these maps and the optimizations to fail.

**To Reproduce**

- [ ]  parser
- [ ]  checker
- [X]  interpreter

```cel
proto.field1.field2
```

Test setup:
```go
	name:  "nested_proto_field",
	expr:  `pb3.single_nested_message.bb`,
	env: []*exprpb.Decl{
		decls.NewVar("pb3",
			decls.NewObjectType("google.expr.proto3.test.TestAllTypes")),
	},
	in: map[string]interface{}{
		"pb3": &proto3pb.TestAllTypes{
			NestedType: &proto3pb.TestAllTypes_SingleNestedMessage{
				SingleNestedMessage: &proto3pb.TestAllTypes_NestedMessage{
					Bb: 1234,
				},
			},
		},
	},
	out: types.Int(1234),
```

**Expected behavior**
The program evaluates the same way, but the expectation is that the field type of 'bb' would be found and extracted in a more optimal manner from its parent proto.

**Additional context**
n/a

TristonianJones:
Attribute ids refer to the first operand in the qualified field path; however, the type map and reference map ids refer to the last field in the qualified attribute field path. Ensure that the types and references are correctly determined by referring directly to the id as it comes from the `checked.proto` AST.

Closes #364

### DIFF
--- a/interpreter/interpreter_test.go
+++ b/interpreter/interpreter_test.go
@@ -540,6 +540,25 @@ var (
 			},
 		},
 		{
+			name:  "nested_proto_field",
+			expr:  `pb3.single_nested_message.bb`,
+			types: []proto.Message{&proto3pb.TestAllTypes{}},
+			env: []*exprpb.Decl{
+				decls.NewVar("pb3",
+					decls.NewObjectType("google.expr.proto3.test.TestAllTypes")),
+			},
+			in: map[string]interface{}{
+				"pb3": &proto3pb.TestAllTypes{
+					NestedType: &proto3pb.TestAllTypes_SingleNestedMessage{
+						SingleNestedMessage: &proto3pb.TestAllTypes_NestedMessage{
+							Bb: 1234,
+						},
+					},
+				},
+			},
+			out: types.Int(1234),
+		},
+		{
 			name: "or_true_1st",
 			expr: `ai == 20 || ar["foo"] == "bar"`,
 			env: []*exprpb.Decl{
--- a/interpreter/interpreter_test.go
+++ b/interpreter/interpreter_test.go
@@ -559,6 +559,28 @@ var (
 			out: types.Int(1234),
 		},
 		{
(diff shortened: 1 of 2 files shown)


interpreter/interpreter_test.go
```diff
@@ -515,6 +515,25 @@ var (
...
+				"pb3": &proto3pb.TestAllTypes{
+					NestedType: &proto3pb.TestAllTypes_SingleNestedMessage{
+						SingleNestedMessage: &proto3pb.TestAllTypes_NestedMessage{
+							Bb: 1234,
+						},
+					},
+				},
+			},
+			out: types.Int(1234),
+		},
```
JimLarson, line 560: So as I understand the comment, this test would pass without these changes. Could you have a test that fails without the changes? Also for an expression that uses indexing.

JimLarson (changes requested):

interpreter/interpreter_test.go
TristonianJones: The code is remarkably robust in that I only noticed the issue when looking at alloc counts from benchmarks. I'll see if I can add a test, but I think the nice thing about the current arrangement is that almost no matter what I do, the code will work. :)

TristonianJones:
PTAL, I was able to add a test which checks for the specialized qualifier type `fieldQualifier` which can only appear when the optimization paths are built correctly.

JimLarson (approved):
Great, thanks!

PR #367, opened 2020-06-11
Aliased identifier support for CEL

### LINKED ISSUE #340: Resolve multiple packages with cel.Container
**Feature request checklist**

**Change**

Currently, `cel.Container` can be used to specify a namespace for symbol resolution. For example:

``` go
env, _ := cel.NewEnv(
	cel.Container("a"),
	cel.Types(&apb.A{}),
)
ast, _ := env.Compile("A{}")
prg, _ := env.Program(ast)
prg.Eval(map[string]interface{}{})
```

In order to write programs that reference multiple protobuf packages, we are forced to use fully qualified names for types outside the container:

``` go
env, _ := cel.NewEnv(
	cel.Container("a"),
	cel.Types(&apb.A{}, &bpb.B{}),
)
ast, _ := env.Compile("b.B{}")
prg, _ := env.Program(ast)
prg.Eval(map[string]interface{}{})
```

**Example**

We should provide the ability to resolve multiple namespaces. There are two options:

1. `Container` could take multiple package names:
  ``` go
  func Container(pkgs ...string) EnvOption
  ```

2. `Env` could be initialized with a custom `packages.Packager`:
  ``` go
  func Package(pkg packages.Packager) EnvOption
  ```

**Alternatives considered**

It’s possible that both these methods would be useful. Here is a sample implementation:

``` go
package cel

func Container(pkgs ...string) EnvOption {
	return func(e *Env) (*Env, error) {
		return e.Package(packages.NewPackage(pkgs...))
	}
}

func Package(pkg packages.Packager) EnvOption {
	return func(e *Env) (*Env, error) {
		e.pkg = pkg
		return e, nil
	}
}

//////

package packages

type defaultPackage struct {
	pkgs []string
}

func NewPackage(pkgs ...string) packages.Packager {
	seen := make(map[string]bool, len(pkgs))

	var p defaultPackage
	for _, pkg := range pkgs {
		if seen[pkg] {
			continue
		}
		p.pkgs = append(p.pkgs, p.pkgs)
		seen[pkg] = true
	}
	return &p
}

func (p *defaultPackage) Package() string {
	return strings.Join(p.pkgs, " ")
}

func (p *defaultPackage) ResolveCandidateNames(name string) []string {
	if strings.HasPrefix(name, ".") {
		return []string{name[1:]}
	}

	if len(p.pkg) == 0 {
		return []string{name}
	}

	for _, nextPkg := range p.pkgs {
		candidates := []string{nextPkg + "." + name}
		for i := strings.LastIndex(nextPkg, "."); i >= 0; i = strings.LastIndex(nextPkg, ".") {
			nextPkg = nextPkg[:i]
			candidates = append(candidates, nextPkg+"."+name)
		}
	}

	return append(candidates, name)
}
```

TristonianJones:
This PR introduces support for aliased identifiers within CEL programs. The feature and its nuances are described in detail in #340; however, there are a few items worth noting in this PR that aren't described in the spec:

- The `common/packages` directory has been replaced with `common/containers`.
- The `Packager` interface has been replaced by the concrete struct `Container`.
- The `cel.Env` variable exports its `Container` variable for use in creating new custom containers within `cel.EnvOption` functions provided outside of the CEL core library.

Closes #340.


TristonianJones:
@sfllaw FYI

cel/cel_test.go
```diff
@@ -189,6 +190,119 @@ func Test_ExampleWithBuiltins(t *testing.T) {
...
+	}
+	ast, iss := env.Compile(`"hello "+ name`) // alias resolved here.
+	if iss.Err() != nil {
+		t.Fatal(iss.Err())
+	}
+	prg, err := env.Program(ast)
+	if err != nil {
+		t.Fatal(err)
+	}
+    out, _, err := prg.Eval(
```
sfllaw: :pick: It looks like `go fmt` might have missed this?

cel/options.go
```diff
@@ -126,9 +126,73 @@ func Macros(macros ...parser.Macro) EnvOption {
...
+		var containerOpts []containers.ContainerOption
+		for alias, qualifiedName := range aliasSet {
+			containerOpts = append(containerOpts, containers.AliasAs(qualifiedName, alias))
+		}
+		containerOpts = append(containerOpts, containers.Aliases(qualifiedNames...))
+		cont, err := containers.NewContainer(e.Container.Name(), containerOpts...)
+		if err != nil {
+			return nil, err
+		}
+		e.Container = cont
```
sfllaw, line 184: `Container` and `Aliases` are almost the same, which I assume is because `containers.Container` is immutable.

Maybe this is OK for now, but I get a weird code smell from this. If CEL (or a user) wants to add a new `EnvOption` that manipulates `Env.Container`, they will need to duplicate this code again.

In addition, isn’t calling `NewContainer` all the time going to be expensive?

:question: Perhaps `containers.Container` ought to be copy-on-write, which is already implemented by `Env.Extend`? We could introduce `Container.Extend` and `ExtendName`:

```go
func Container(name string) EnvOption {
	return func(e *Env) (*Env, error) {
		e.Container = e.Container.ExtendName(name)
		return e, nil
	}
}

func Aliases(qualifiedNames ...string) EnvOption {
	return func(e *Env) (*Env, error) {
		cont, err := e.Container.Extend(containers.Aliases(qualifiedNames...))
		if err != nil {
			return nil, err
		}
		e.Container = cont
		return e, nil
	}
}
```

common/containers/container.go
```diff
@@ -0,0 +1,224 @@
...
+import (
+	"fmt"
+	"strings"
+
+	exprpb "google.golang.org/genproto/googleapis/api/expr/v1alpha1"
+)
+
+var (
+	// DefaultContainer has an empty container name.
+	DefaultContainer = &Container{aliases: make(map[string]string)}
```
sfllaw: :pick: Should we DRY and use `NewContainer` instead?
```suggestion
	DefaultContainer = NewContainer("")
```

common/containers/container.go
```diff
@@ -0,0 +1,224 @@
...
+			return nil, fmt.Errorf(
+				"alias collides with existing reference: name=%s, alias=%s, existing=%s",
+				qualifiedName, alias, aliasRef)
+		}
+		if strings.HasPrefix(c.Name(), alias+".") || c.Name() == alias {
+			return nil, fmt.Errorf(
+				"alias collides with container name: name=%s, alias=%s, container=%s",
+				qualifiedName, alias, c.Name())
+		}
+		c.aliases[alias] = qualifiedName
```
sfllaw, line 276: :bug: Can we make this work with a nil `c.aliases` so that `&Container{}` doesn’t panic? (Also, this means that `NewContainer` won’t have to make the map if there are no aliases.)
```suggestion
		if c.aliases == nil {
			c.aliases = make(map[string]string)
		}
		c.aliases[alias] = qualifiedName
```

common/containers/container.go
```diff
@@ -0,0 +1,224 @@
...
+// the container wins as the container will continue to evolve over time and the program must be
+// forward compatible with changes in the container.
+func Aliases(qualifiedNames ...string) ContainerOption {
+	return func(c *Container) (*Container, error) {
+		for _, qn := range qualifiedNames {
+			ind := strings.LastIndex(qn, ".")
+			alias := qn
+			if ind > 0 || ind < len(qn)-1 {
+				alias = qn[ind+1:]
+			}
```
sfllaw: :bug: Did you mean to `&&` this test?
```suggestion
			if ind > 0 && ind < len(qn)-1 {
				alias = qn[ind+1:]
			}
```
As written, this conditional will almost always evaluate true, which I don’t think you want. Perhaps you want to add tests for `Aliases(".foo")` and `Aliases("bar.")`?

sfllaw (changes requested):
Great job! :tada: 

I have found some bugs :bug:, nitpicks :pick:, and a design question :question:.

cel/cel_test.go
TristonianJones: Sure did. There's some lag on my laptop on between save / reformat and every once in a while this happens and I don't notice. Fixed.

cel/options.go
TristonianJones: There's definitely a tradeoff w.r.t. to clarity / ease of use here. I don't expect containers or environments to be configured frequently, and typically most applications pick one `Env` and stick with it. I'm not too worried about the mem / cpu overhead as I believe it's about the same for either example and it tends to be amortized over a large expression / eval set.

I like the latter example of `Extend` and I'll export options to swap the container name, add aliases, and reset them in order to concentrate the exported method set on functional options. This way developers will be able to either create a container from scratch with `NewContainer` or extend the configuration of an existing one with `Extend`.

common/containers/container.go
TristonianJones: The `NewContainer` call returns two values, so this isn't possible here. However, the suggestion to lazily gave me an idea about how to make `DefaultContainer *Container = nil` work here instead.

common/containers/container.go
TristonianJones: Good catch. I intended `&&`. Fixed with updated tests.

common/containers/container.go
TristonianJones: Sure, sounds good to me.

cel/options.go
```diff
@@ -126,9 +126,62 @@ func Macros(macros ...parser.Macro) EnvOption {
...
+		cont, err := e.Container.Extend(containers.Name(name))
+		if err != nil {
+			return nil, err
+		}
+		e.Container = cont
+		return e, nil
+	}
+}
+
+// Aliases configures a set of simple names as aliases for fully-qualified names.
```
JimLarson: It's not really an "alias", since you don't have the choice of a short name. It's really an abbreviation, so the option could be `Abbreviation()`, `Abbrev()` (whoa, so meta), or as an imperative `Abbreviate()`.

We've got so many feature names that are a near-miss for the thing described ("macro", "comprehension", "constant") that if we add another it becomes our brand.

common/containers/container.go
```diff
@@ -0,0 +1,281 @@
...
+	}
+	// If an alias exists for the name, ensure it is searched last.
+	alias, found := c.aliasSet()[name]
+	if found {
+		return append(candidates, alias)
+	}
+	return candidates
+}
+
+// ContainerOption specifies a functional configuration option for a Container.
```
JimLarson, line 169: Add a note that ContainerOptions must handle a nil container pointer.

common/containers/container.go
```diff
@@ -0,0 +1,281 @@
...
+	aliases map[string]string
+}
+
+// Extend creates a new Container with the existing settings and applies a series of
+// ContainerOptions to further configure the new container.
+func (c *Container) Extend(opts ...ContainerOption) (*Container, error) {
+	if c == nil {
+		return NewContainer(opts...)
+	}
+	// Copy the name and aliases of the existing container.
```
JimLarson, line 63: Consider copying a nil container pointer to a nil pointer, rather than allocating a new struct.

common/containers/container.go
```diff
@@ -0,0 +1,281 @@
...
+func (c *Container) ResolveCandidateNames(name string) []string {
+	if strings.HasPrefix(name, ".") {
+		qn := name[1:]
+		return c.candidatesWithAlias([]string{qn}, qn)
+	}
+	if c.Name() == "" {
+		return c.candidatesWithAlias([]string{name}, name)
+	}
+
+	nextCont := c.name
```
JimLarson: Could `c` be nil here? Documented assumptions are dangerously absent in this module. Might as well save the result from `c.Name()` which was just called a few lines above.

common/containers/container.go
```diff
@@ -0,0 +1,281 @@
...
+	if c == nil || c.aliases == nil {
+		return noAliases
+	}
+	return c.aliases
+}
+
+// candidatesWithAliases returns the resolved candidates from the container with any applicable
+// aliases appended on the end of the list.
+func (c *Container) candidatesWithAlias(candidates []string, name string) []string {
+	if len(c.aliasSet()) == 0 {
```
JimLarson: This special case seems unnecessary.

common/containers/container.go
```diff
@@ -0,0 +1,281 @@
...
+		if c.aliases == nil {
+			c.aliases = make(map[string]string)
+		}
+		c.aliases[alias] = qualifiedName
+		return c, nil
+	}
+}
+
+// Name sets the fully-qualified name of the Container.
+func Name(name string) ContainerOption {
```
JimLarson, line 282: It should probably be an error if name is empty or starts with ".", but it seems to do something reasonable even if this is the case.

common/containers/container.go
```diff
@@ -0,0 +1,281 @@
...
+	if found {
+		return append(candidates, alias)
+	}
+	return candidates
+}
+
+// ContainerOption specifies a functional configuration option for a Container.
+type ContainerOption func(*Container) (*Container, error)
+
+// Aliases configures a set of simple names as aliases for fully-qualified names.
```
JimLarson: "Aliases" -> "Abbrevs".

common/containers/container.go
```diff
@@ -0,0 +1,281 @@
...
+// poorly chosen protobuf message / package names.
+//
+// Note: all of the rules that apply to Aliases also apply to AliasAs.
+func AliasAs(qualifiedName, alias string) ContainerOption {
+	return func(c *Container) (*Container, error) {
+		if len(alias) <= 0 || strings.Contains(alias, ".") {
+			return nil, fmt.Errorf(
+				"alias names must be non-empty and simple (not qualified): alias=%s", alias)
+		}
+		ind := strings.LastIndex(qualifiedName, ".")
```
JimLarson, line 254: Also error if qualifiedName starts with "."?

common/containers/container.go
```diff
@@ -0,0 +1,281 @@
...
+// ToQualifiedName converts an expression AST into a qualified name if possible, with a boolean
+// 'found' value that indicates if the conversion is successful.
+func ToQualifiedName(e *exprpb.Expr) (string, bool) {
+	switch e.ExprKind.(type) {
+	case *exprpb.Expr_IdentExpr:
+		id := e.GetIdentExpr()
+		return id.Name, true
+	case *exprpb.Expr_SelectExpr:
+		sel := e.GetSelectExpr()
+		if qual, found := ToQualifiedName(sel.Operand); found {
```
JimLarson, line 311: Must also ensure that `sel.testOnly()` is `false`.

common/containers/container_test.go
```diff
@@ -0,0 +1,218 @@
...
+func TestContainer_ResolveCandidateNames_EmptyContainer(t *testing.T) {
+	names := DefaultContainer.ResolveCandidateNames("R.s")
+	want := []string{"R.s"}
+	if !reflect.DeepEqual(names, want) {
+		t.Errorf("got %v, wanted %v", names, want)
+	}
+}
+
+func TestContainer_Aliases(t *testing.T) {
+	c, err := NewContainer(Name("a.b.c"), Aliases("my.alias.R"))
```
JimLarson: Need some tests for abbrevs with the default container too. (I see there's one below with a low-level invocation of AliasAs(), but also need to test the top-level API.)

common/containers/container_test.go
```diff
@@ -0,0 +1,218 @@
...
+	}
+	if c.aliasSet()["alias"] != "test.alias" {
+		t.Errorf("got alias %v wanted 'test.alias'", c.aliasSet())
+	}
+	c, err = c.Extend(Name("with.container"))
+	if err != nil {
+		t.Fatal(err)
+	}
+	if c.Name() != "with.container" {
+		t.Errorf("got container name %s, wanted 'goodbye.container'", c.Name())
```
JimLarson: Wanted `'with.container'`.

common/containers/container_test.go
```diff
@@ -0,0 +1,218 @@
...
+	}
+}
+
+func TestContainers_Extend_Name(t *testing.T) {
+	c, err := DefaultContainer.Extend(Name(""))
+	if err != nil {
+		t.Fatal(err)
+	}
+	if c != nil {
+		t.Error("got non-nil container, wanted nil")
```
JimLarson: Ehhh, I think it's up to the implementation whether to use `nil` or `Container{name: ""}`. Just check `c.Name()` here.

JimLarson (changes requested):
I suggest changing the top-level option name to `Abbrev()`, but keeping the underlying feature as "alias".  Can change `AsAlias()` to just `Alias()` without risk of collision with `Aliases()`.

common/containers/container.go
TristonianJones: I don't think it can be, but the code would be safer if it re-used `c.Name()`

common/containers/container.go
TristonianJones: Yeah, it's only covering one or the other, but not both. I'll refactor this.

common/containers/container.go
TristonianJones: Interesting, I just moved this code from another location. It's likely that `TestOnly` checks are happening before the method is invoked, but I'll do another pass through the code to make sure.

common/containers/container_test.go
TristonianJones: Code-wise `Aliases` proxies to `AliasAs`, but you're right that this is not guaranteed by the top-level API and could change some day.

common/containers/container.go
TristonianJones: The only minor annoyance with this change is the subtle shift in the error message based on whether `Abbrev` or `Alias` is used. I can make it work though.

common/containers/container.go
TristonianJones: An empty name is totally valid, but a leading '.' is not. I'll add an error check for leading '.'.

TristonianJones:
PTAL

cel/cel_test.go
```diff
@@ -189,6 +190,119 @@ func Test_ExampleWithBuiltins(t *testing.T) {
...
+			"external.Expr": "wrong expr",
+		},
+	)
+	if err != nil {
+		t.Fatal(err)
+	}
+	want := &exprpb.Expr{Id: 1}
+	got, _ := out.ConvertToNative(reflect.TypeOf(want))
+	if !proto.Equal(got.(*exprpb.Expr), want) {
+		t.Errorf("got %v, wanted 'string expr'", out)
```
JimLarson: "wanted Expr proto"

common/containers/container.go
```diff
@@ -0,0 +1,296 @@
...
+}
+
+// ContainerOption specifies a functional configuration option for a Container.
+//
+// Note, ContainerOption implementations must be able to handle nil container inputs.
+type ContainerOption func(*Container) (*Container, error)
+
+// Abbrevs configures a set of simple names as abbreviations for fully-qualified names.
+//
+// An alias can be useful when working with variables, functions, and especially types from
```
JimLarson: "An abbrev can be useful..."

cel/options.go
```diff
@@ -126,9 +126,63 @@ func Macros(macros ...parser.Macro) EnvOption {
...
+//    Abbrevs("qual.pkg.version.ObjTypeName", "alt.container.ver.FieldTypeName")
+//    // Simplified Object construction
+//    ObjTypeName{field: FieldTypeName{value: ...}}
+//
+// There are a few rules for the qualified names and the simple name abbreviations generated from
+// them:
+// - Qualified names must be dot-delimited, e.g. `package.subpkg.name`.
+// - The last element in the qualified name is the simple name used as an abbreviation.
+// - Abbreviations must not collide with each other.
+// - The abbreviation must not collide with the root-level container name.
```
JimLarson: What is a "root-level container name"?  Is it an unqualfied name?

common/containers/container.go
```diff
@@ -0,0 +1,296 @@
...
+//    // CEL Go option
+//    Abbrevs("qual.pkg.version.ObjTypeName", "alt.container.ver.FieldTypeName")
+//    // Simplified Object construction
+//    ObjTypeName{field: FieldTypeName{value: ...}}
+//
+// There are a few rules for the qualified names and the simple abbreviations generated from them:
+// - Qualified names must be dot-delimited, e.g. `package.subpkg.name`.
+// - The last element in the qualified name is the simple name used as the abbreviation.
+// - Abbreviations names must not collide with each other.
+// - The abbreviation name must not collide with the root-level container name.
```
JimLarson: "must not collide with unqualified names in use".

common/containers/container.go
```diff
@@ -0,0 +1,296 @@
...
+//
+// Abbrevs are distinct from container-based references in the following important ways:
+// - Containers follow C++ namespace resolution rules with searches from the most qualified name
+//   to the least qualified name.
+// - Container references within the CEL program may be relative, and are resolved to fully
+//   qualified names at either type-check time or program plan time, whichever comes first.
+// - Abbreviations must resolve to a fully-qualified name.
+// - Resolved abbreviations do not participate in namespace resolution.
+// - Resolved abbreviations are searched after container names, including container names in the
+//   global scope.
```
JimLarson: I'm not sure what "container names in the global scope" means. Unless you want to introduce new terminology, "container name" sounds like it should be the name of the container. Maybe "containerized name"? Consider rewriting this bullet item as: "Abbreviation resolution happens after the container path has been searched for matching identifiers, including the global scope."

common/containers/container.go
```diff
@@ -0,0 +1,296 @@
...
+//
+// If there is ever a case where an identifier could be in both the container and in the alias,
+// the container wins as the container will continue to evolve over time and the program must be
+// forward compatible with changes in the container.
```
JimLarson: Don't we want this to go the other way for forward compatibility?  If we've written an expression with container `a.b.c` to use `x` as an abbrev for `z.y.x`, then we certainly don't want `x` to silently switch meaning if `a.b.c.x` or `a.b.x` or `a.x` is added to proto definitions somewhere. If we want to use that new `a.b.c.x`, we can use the fully-qualified name for it.  We want an error if "x" is in the global (unqualified) scope, as that's probably a declared variable.  On the other hand, if we've already written expressions relying on container path search to have `x` mean `a.b.x`, then it's the responsibility of the integrator to search for collisions before introducing a new abbreviation.  The integrator has control over their set of aliases, but might not have control over the proto definitions they include.

common/containers/container.go
```diff
@@ -0,0 +1,296 @@
...
+func Abbrevs(qualifiedNames ...string) ContainerOption {
+	return func(c *Container) (*Container, error) {
+		for _, qn := range qualifiedNames {
+			alias := qn
```
JimLarson: Not used - can initialize `alias` on line 200 below.

common/containers/container.go
TristonianJones: The container is a higher-order concept which may not be (completely) within the developer's control; whereas the alias is within the developer's control. In the ideal case the container wins, but if you want to ensure the alias is picked, using the leading '.' will ensure the proper behavior.

cel/options.go
TristonianJones: I think I mean, the first component of a qualified container name.

common/containers/container.go
TristonianJones: Done

common/containers/container.go
TristonianJones: Done.

cel/cel_test.go

common/containers/container.go
TristonianJones: Done.

common/containers/container.go
JimLarson: I'm confused - how is the container not in the developer's control? Do you mean the contents of the proto namespace? If so, that's the point: with non-alias names first, a reference to an alias can silently change underneath you, but with alias first, all shadowing is deliberate. Yes, a leading "." will work, but then maybe recommend that and show it in examples?

JimLarson: I'll LGTM if you're really sure this is the way you'd like it to go, but it might be worth writing up some guidance, maybe in the integration guide, of what to check for when adding new aliases, and how to update your proto definitions when you have stored some compiled ASTs.

JimLarson (approved):
Approved, conditional on the discussion of alias vs non-alias resolution order.

common/containers/container.go
TristonianJones: The current behavior is what we discussed at the language council and is what's present in the design doc. Also, the examples showcase the use of a leading '.' for disambiguation. With respect to compiled ASTs, the aliases would have been resolved ahead of time, so the program behavior should be preserved even with new containers. Parsed ASTs would be subject to some amount of variability. I think that's the most compelling argument for switching to resolve aliases ahead of containerized names.

Both the `container` and `alias` are extra-lingual which is part of what makes this choice difficult. If the alias can shadow a containerized name it becomes difficult to reason about the source of truth for the identifier. Does it come from the container or the alias? If the alias wins, then it can shadow and render unusable parts of the container. If the container wins, it can change the meaning of the alias, unless the alias has a leading '.'. Given that each direction could shadow the other and they're both developer owned, it's hard to say what the best outcome should be.

TristonianJones: Per our offline conversation, if we are recommending recompilation of an expression to address AST generation errors, then a stored AST is not sufficient to preserve the program meaning in light of changes in the container structure. To preserve the the meaning between recompilations, the aliases must be resolved first, and disambiguation will require using the fully qualified names rather than abbreviated forms.

TristonianJones:
PTAL. I've changed the alias resolution order to reflect our recent design updates.

cel/options.go
```diff
@@ -156,26 +156,25 @@ func Container(name string) EnvOption {
...
 //
-// There are a few rules for the qualified names and the simple name abbreviations generated from
-// them:
+// There are a few rules for the qualified names and the simple abbreviations generated from them:
 // - Qualified names must be dot-delimited, e.g. `package.subpkg.name`.
-// - The last element in the qualified name is the simple name used as an abbreviation.
-// - Abbreviations must not collide with each other.
-// - The abbreviation must not collide with the first component of a qualified container name.
+// - The last element in the qualified name is the simple name used as the abbreviation.
+// - Abbreviations names must not collide with each other.
```
JimLarson: Establish terminology up above  so it's clear that "abbreviation names" refers to the pre-expansion strings.  Maybe "name" and "expansion". Or "abbreviation/abbrev" and "expansion" since "name" might be hopelessly ambiguous.

cel/options.go
```diff
@@ -156,26 +156,25 @@ func Container(name string) EnvOption {
...
+// - The abbreviation must not collide with unqualified names in use.
 //
-// Abbreviations are distinct from container-based references in the following important ways:
+// Abbrevs are distinct from container-based references in the following important ways:
 // - Containers follow C++ namespace resolution rules with searches from the most qualified name
```
JimLarson, line 170: See if the comment flows better moving the first two items (lines 166-169) to the end of the group - discuss abbrevs first because they are resolved first.

common/containers/container.go
```diff
@@ -105,25 +105,31 @@ func (c *Container) Name() string {
 //     a.R.s
 //     R.s
 //
-// If aliases or abbreviations are configured for the container, then alias names appear after
-// the potential set of container-based names.
+// If aliases or abbreviations are configured for the container, then alias names will take
+// precedence over containerized names.
```
JimLarson, line 109: Consider "...then alias expansion is done instead of searching the container path."

common/containers/container.go
```diff
@@ -134,15 +140,31 @@ func (c *Container) aliasSet() map[string]string {
 	return c.aliases
 }
 
-// candidatesWithAlias returns the resolved candidates from the container with any applicable
-// alias appended on the end of the list.
-func (c *Container) candidatesWithAlias(candidates []string, name string) []string {
+// findAlias takes a name as input and returns an aliased name if one exists.
```
JimLarson: Note that the name must not have a leading dot.

common/containers/container.go
```diff
@@ -134,15 +140,31 @@ func (c *Container) aliasSet() map[string]string {
...
+	alias, found := c.aliasSet()[simple]
 	if found {
-		return append(candidates, alias)
+		if qualifier != "" {
+			alias = alias + qualifier
+		}
+		return alias, true
 	}
-	return candidates
+	return "", false
```
JimLarson: Consider replacing 161-167 with:

```
if !found {
        return "", false
}
return alias + qualifier, true
```

JimLarson (approved):
Just a few notes on comments and small refactoring.

Issue #356, opened 2020-05-19
Late Function Bindings in Go Cel Evaluation


mswest46:
Currently, function definitions must be provided in the parsing step, so data relevant to their definition must be available at that step. This request is to support binding a function at activation time instead. As a use case example, I'd like to write a function `has_tag :: string -> bool` whose implementation checks membership in a list `tags` that is only available at evaluation time and remains hidden from the expression writer.

JimLarson:
We're discussing a generalization of this - allow functions to access the full Activation, not just the supplied variables. This way you could bind `tags` as a variable at runtime in the normal way, but it would be hidden from the expression if you omit it from the declarations at type check time. This would require a separate flavor of overload which would accept something like `func(interpreter.Activation, ...ref.Val) ref.Val`.

hyp0th3rmi4:
I have stumbled upon a use case for this feature which would be elegantly solved by allowing to override some of the bindings used during program evaluation. By digging down into the source code the function bindings are resolved when the program is created here:

https://github.com/google/cel-go/blob/master/cel/program.go#L180-L203

and become part of the interpreter that is used to create the interpretable. Out of function calls found in the expression, the planner will create one of the following Interpretable implementations (reference: https://github.com/google/cel-go/blob/master/interpreter/planner.go#L211-L285):

- `evalZeroArity (impl: functions.FunctionOp)`
- `evalUnary (impl: functions.FunctionUnaryOp)`
- `evalBinary (impl: functions.FunctionBinaryOp)`
- `evalVarArgs (impl: functions.FunctionOp)`

because all of them retain a `function` and `overload` identifier, their corresponding evaluation method can lookup the activation and eventually replace such binding with the late bound supplied version of the same function signature.

I think we should still be checking during `program.Eval(....)` that all signatures and bindings that are overridden match so that any calculation that has been done before and validation still remains valid.

Another option would be passing a decorator to the `program.Eval`, to do this job by scanning the interpretable and perform the replacement.

I don't think that the code changes are massive, they seem rather contained to a couple of files at first site. If this is a viable solution I can prepare a PR that implement one of the two approaches below:

### Approach 1:  With Activation and Internal CustomDecorator

- Define a new type of activation that contains overridden bindings. 
- Define a custom decorator that is injected if the Activation contains binding definitions. 
- The decorator can then check the interpretables whose types match the above `evalXXX` types and perform the replacements as needed.
[...]

PR #368, opened 2020-06-19
Asynchronous function invocation support

TristonianJones:
The PR introduces support for asynchronous functions as CEL extensions.

The general algorithm for evaluation relies on the `types.Unknown` value to indicate when
async calls relevant to the outcome of an expression. Async calls arguments are tracked,
deduped, and memoized based on the overload id associated with the implementation. 

It is possible to use partial state evaluation with async evaluation, though async calls will
be resolved before any unknown attribute patterns. Using both features together may be
inefficient; however, there is a general expectation that async function implementations may
implement their own caching logic in order to handle repeated invocations of the same
async function with the same arguments.

### DIFF
--- a/cel/env.go
+++ b/cel/env.go
@@ -307,14 +307,13 @@ func (e *Env) ParseSource(src common.Source) (*Ast, *Issues) {
 
 // Program generates an evaluable instance of the Ast within the environment (Env).
 func (e *Env) Program(ast *Ast, opts ...ProgramOption) (Program, error) {
-	optSet := e.progOpts
-	if len(opts) != 0 {
-		mergedOpts := []ProgramOption{}
-		mergedOpts = append(mergedOpts, e.progOpts...)
-		mergedOpts = append(mergedOpts, opts...)
-		optSet = mergedOpts
-	}
-	return newProgram(e, ast, optSet)
+	return e.newProgram(ast, opts /* async= */, false)
+}
+
+// AsyncProgram generates an evaluable instance of the Ast with support for asynchronous extension
+// functions.
+func (e *Env) AsyncProgram(ast *Ast, opts ...ProgramOption) (AsyncProgram, error) {
+	return e.newProgram(ast, opts /* async= */, true)
 }
 
 // SetFeature sets the given feature flag, as enumerated in options.go.
@@ -427,8 +426,8 @@ func (i *Issues) Err() error {
 	if i == nil {
 		return nil
 	}
-	if len(i.errs.GetErrors()) > 0 {
-		return errors.New(i.errs.ToDisplayString())
+	if len(i.Errors()) > 0 {
+		return errors.New(i.String())
 	}
 	return nil
 }
--- a/cel/BUILD.bazel
+++ b/cel/BUILD.bazel
@@ -52,6 +52,7 @@ go_test(
         "//common/types/ref:go_default_library",
         "//common/types/traits:go_default_library",
         "//interpreter/functions:go_default_library",
+        "//test:go_default_library",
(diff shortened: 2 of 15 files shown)


TristonianJones:
While this PR does not specifically attempt to address https://github.com/google/cel-go/issues/356, it is possible to use the `functions.AsyncOp` implementation to solve this problem. The only caveat is that there is likely a 2x performance penalty for using the `functions.AsyncOp` purely for plumbing `Activation` data into a synchronous function.

TristonianJones:
FYI @mswest46

cel/env.go
```diff
@@ -307,14 +307,13 @@ func (e *Env) ParseSource(src common.Source) (*Ast, *Issues) {
...
 func (e *Env) Program(ast *Ast, opts ...ProgramOption) (Program, error) {
-	optSet := e.progOpts
-	if len(opts) != 0 {
-		mergedOpts := []ProgramOption{}
-		mergedOpts = append(mergedOpts, e.progOpts...)
-		mergedOpts = append(mergedOpts, opts...)
-		optSet = mergedOpts
-	}
-	return newProgram(e, ast, optSet)
+	return e.newProgram(ast, opts /* async= */, false)
```
JimLarson, line 310: It's odd to have the comma between the comment and the field it documents, here and below.

cel/env.go
```diff
@@ -307,14 +307,13 @@ func (e *Env) ParseSource(src common.Source) (*Ast, *Issues) {
...
+}
+
+// AsyncProgram generates an evaluable instance of the Ast with support for asynchronous extension
+// functions.
```
JimLarson, line 314: It's unclear what an "asynchronous extension function" is.

JimLarson (changes requested):
Per our conversation:

It looks like there are a few separate features here which can largely be  used independently:

- a result cache which can be used by expensive functions;
- parallel execution for high-latency calls;
- a potentially arbitrary argument passed in at the top level and propagated to specially-declared extension functions.
    - Due to strange golang magic, restricted to be just `context.Context`, but hey, as long as you're digging one tunnel, why not build a parallel shaft and plumb through a `interface{}`?  Consider it.
    - Even though `context.Context` allows arbitrary key/value pairs to piggyback on it, use of this as a general-purpose optional argument is contra-indicated in its docs.

At a minimum, consider allowing these functions to be used in the API independently.

Consider also having the result cache be implemented as a wrapper layer exploiting the plumbed-through `interface{}` for the cache. Yes, for peak efficiency you'd want this a little more integrated when you're using it with parallelism, so you can avoid creating a new goroutine when there'd be a cache hit, but in cases where the cache is useful this overhead is probably negligible, so it might be worthwhile for the cleaner stratification of features. OTOH, the deeper integration might not be that much more complex.

Consider having these new features just be enabled/disabled at each extension function, maintaining just one kind of `Program`.

Lastly, these all need better names and documentation. It's not an `async function` since you're never returning from the call stack. The docs need to make it clear what's being implemented here.

mneff-roblox:
Are there still any plans to revive this? Based on the issues and PR descriptions I read this seems like the preferred approach for propagating context / activation to a extension function binding.

TristonianJones:
> Are there still any plans to revive this? Based on the issues and PR descriptions I read this seems like the preferred approach for propagating context / activation to a extension function binding.

I would really like to revive this, but we've just had a lot of other competing asks for our time. That said, #1117 is working toward supporting late bindings which we could use to capture a context within the dynamically created function binding. The PR is only the first half, but the second half will come soon.

TristonianJones:
Fixed by https://github.com/cel-expr/cel-go/pull/1369

PR #369, opened 2020-06-22
Add heuristic evaluation cost to expressions

jinmmin:
The primary goal for the change is for CEL policy templates where a 
cost limit can be configured and enforced at compile time and runtime.

I am uncertain about some of the cost estimations, but feel it is easier
 to discuss with a PR.

Below are the points that might be worth discussing:

* The cost for all operators and functions is 1, which may be 
customizable.
* The resulting cost is an interval, mainly for expression that contains
 short-circuitable operations, such logical and/or, condition ternary.
* The cost for all existing qualifiers is 1. The interface added there
 is more for extensibility (e.g. a custom qualifier) but we can also
  remove it.
* The actual cost for some expressions (`fold` specifically) may rely
 on the input size, which is not available at compile time. Currently I
  just return the maximum possible range. Wdyt? (this is also the
 reason I am using `int64` as the cost as it is more obvious when the
  cost overflows).

### DIFF
--- a/cel/cel_test.go
+++ b/cel/cel_test.go
@@ -876,3 +876,38 @@ func Test_CustomInterpreterDecorator(t *testing.T) {
 		t.Errorf("got %v as the last observed constant, wanted 1", lastConst)
 	}
 }
+
+func Test_Cost(t *testing.T) {
+	e, err := NewEnv()
+	if err != nil {
+		log.Fatalf("environment creation error: %s\n", err)
+	}
+	ast, iss := e.Compile(`"Hello, World!"`)
+	if iss.Err() != nil {
+		log.Fatalln(iss.Err())
+	}
+
+	wantedCost := []int64{0, 0}
+
+	// Test standard evaluation cost.
+	prg, err := e.Program(ast)
+	if err != nil {
+		log.Fatalf("program creation error: %s\n", err)
+	}
+	min, max := EstimateCost(prg)
+	if min != wantedCost[0] || max != wantedCost[1] {
+		log.Fatalf("Got cost interval [%v, %v], wanted %v",
+			min, max, wantedCost)
+	}
+
+	// Test the factory-based evaluation cost.
+	prg, err = e.Program(ast, EvalOptions(OptExhaustiveEval))
+	if err != nil {
+		t.Fatalf("program creation error: %s\n", err)
+	}
+	min, max = EstimateCost(prg)
+	if min != wantedCost[0] || max != wantedCost[1] {
+		log.Fatalf("Got cost interval [%v, %v], wanted %v",
+			min, max, wantedCost)
+	}
+}
--- a/cel/program.go
+++ b/cel/program.go
@@ -16,6 +16,7 @@ package cel
 
 import (
 	"fmt"
+	"math"
 
 	"github.com/google/cel-go/common/types"
 	"github.com/google/cel-go/common/types/ref"
@@ -260,6 +261,11 @@ func (p *prog) Eval(input interface{}) (v ref.Val, det *EvalDetails, err error) 
 	return
 }
 
+// Cost implements the Coster interface method.
+func (p *prog) Cost() (min, max int64) {
+	return estimateCost(p.interpretable)
+}
+
 // Eval implements the Program interface method.
 func (gen *progGen) Eval(input interface{}) (ref.Val, *EvalDetails, error) {
 	// The factory based Eval() differs from the standard evaluation model in that it generates a
@@ -282,4 +288,31 @@ func (gen *progGen) Eval(input interface{}) (ref.Val, *EvalDetails, error) {
 		return v, det, err
 	}
 	return v, det, nil
+}
+
+// Cost implements the Coster interface method.
+func (gen *progGen) Cost() (min, max int64) {
+	// Use an empty state value since no evaluation is performed.
+	p, err := gen.factory(emptyEvalState)
+	if err != nil {
+		return 0, math.MaxInt64
+	}
+	return estimateCost(p)
+}
+
+var (
+	emptyEvalState = interpreter.NewEvalState()
+)
+
+// EstimateCost returns the heuristic cost interval for the program.
+func EstimateCost(p Program) (min, max int64) {
+	return estimateCost(p)
+}
+
+func estimateCost(i interface{}) (min, max int64) {
+	c, ok := i.(interpreter.Coster)
+	if !ok {
+		return 0, math.MaxInt64
+	}
+	return c.Cost()
 }
--- a/interpreter/attributes.go
+++ b/interpreter/attributes.go
@@ -17,6 +17,7 @@ package interpreter
 import (
 	"errors"
 	"fmt"
+	"math"
 
 	"github.com/google/cel-go/common/packages"
 	"github.com/google/cel-go/common/types"
@@ -227,6 +228,18 @@ func (a *absoluteAttribute) ID() int64 {
 	return a.id
 }
 
+// Cost implements the Coster interface method.
+func (a *absoluteAttribute) Cost() (min, max int64) {
+	for _, q := range a.qualifiers {
+		minQ, maxQ := estimateCost(q)
+		min += minQ
+		max += maxQ
+	}
+	min++ // For object retrieval.
+	max++
+	return
+}
+
 // AddQualifier implements the Attribute interface method.
 func (a *absoluteAttribute) AddQualifier(qual Qualifier) (Attribute, error) {
 	a.qualifiers = append(a.qualifiers, qual)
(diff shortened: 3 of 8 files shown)


jinmmin:
@TristonianJones

cel/program.go
```diff
@@ -41,6 +42,9 @@ type Program interface {
 	// or `ProgramOption` values used in the creation of the evaluation environment or executable
 	// program.
 	Eval(vars interface{}) (ref.Val, *EvalDetails, error)
+
+	// Cost returns the estimated cost interval for the evaluation.
+	Cost() (min, max int64)
```
TristonianJones: Maybe make this a separate interface called `Coster`? You wouldn't need to embed the interface into other interfaces since the behavior of cost-ing is orthogonal to interpretation / attribute calculation. In theory the `Program`, `Interpretable` steps wouldn't be the only objects with a heuristic `Cost`, e.g. you could imagine adding it to a custom function implementation. 

If you make it a separate interface I'd wrap `Cost()` computations into a helper within the `interpreter` package:

```go
func estimateCost(i interface{}) (min, max int64) {
    c, ok := i.(Coster)
    if !ok {
       return 0, math.MaxInt64
    }
    return c.Cost()
}
```

cel/program.go
```diff
@@ -283,3 +292,13 @@ func (gen *progGen) Eval(input interface{}) (ref.Val, *EvalDetails, error) {
 	}
 	return v, det, nil
 }
+
+// Cost implements the Program interface method.
+func (gen *progGen) Cost() (min, max int64) {
+	state := interpreter.NewEvalState()
```
TristonianJones: Consider using a private empty state value for computing cost since no evaluation is performed.

cel/program.go
TristonianJones: Note, if you created a separate interface, you could also expose a helper on the `cel` package: `cel.EstimateCost(Program) (min, max int64)`

TristonianJones (changes requested):
This is really cool and plays nicely into heuristic branch predictions where we re-organize expression evaluation to perform less costly evaluations first. 

I have a few comments for making this model a bit more extensible and for providing a cleaner separation of concerns, but overall looks good.

As part of the policy for merging PRs on the core CEL repos, we ask that a GitHub issue be filed before merging the change.

cel/program.go
jinmmin: refactored. PTAL.

I created an issue https://github.com/google/cel-go/issues/371 but does not seem to have permission to link it here.

cel/program.go
jinmmin: like this?

cel/program.go
```diff
@@ -283,3 +289,37 @@ func (gen *progGen) Eval(input interface{}) (ref.Val, *EvalDetails, error) {
...
+// Cost implements the Coster interface method.
+func (gen *progGen) Cost() (min, max int64) {
+	p, err := gen.factory(&emptyEvalState{})
+	if err != nil {
+		return 0, math.MaxInt64
+	}
+	return estimateCost(p)
+}
+
+type emptyEvalState struct{}
```
TristonianJones: I was actually thinking you might create an `EvalState` object and assign it to an unexported `var`:

```go
var (
  emptyEvalState := interpreter.NewEvalState()
)
```

interpreter/attributes.go
```diff
@@ -324,6 +337,14 @@ func (a *conditionalAttribute) ID() int64 {
 	return a.id
 }
 
+// Cost implements the Coster interface method.
```
TristonianJones: Where appropriate, describe anything interesting about the heuristic cost computation on the implementations:

```go
// Cost provides the heuristic cost of a ternary operation <expr> ? <t> : <f>
// The cost is computed as cost(expr) plus the min/max costs of evaluating either
// `t` or `f`.
```

interpreter/attributes.go
```diff
@@ -650,6 +707,11 @@ func (q *stringQualifier) Value() ref.Val {
 	return q.celValue
 }
 
+// Cost implements the Coster interface method.
+func (q *stringQualifier) Cost() (min, max int64) {
```
TristonianJones, line 714: Perhaps mention, `Cost returns zero for constant field qualifiers`

interpreter/coster.go
```diff
@@ -0,0 +1,31 @@
...
+package interpreter
+
+import "math"
+
+// Coster calculates the heuristic cost incurred during evaluation.
+type Coster interface {
+	Cost() (min, max int64)
+}
+
+// EstimateCost returns the heuristic cost interval for the program.
```
TristonianJones: nit: make sure this comment is lower-cased unless you want to export it.

interpreter/interpretable.go
```diff
@@ -124,7 +126,14 @@ func (test *evalTestOnly) Eval(ctx Activation) ref.Val {
 		return container.Contains(test.field)
 	}
 	return types.ValOrErr(obj, "invalid type for field selection.")
+}
 
+// Cost implements the Coster interface method.
+func (test *evalTestOnly) Cost() (min, max int64) {
+	min, max = estimateCost(test.op)
```
TristonianJones, line 134: Make sure you comment on why the cost for test only is at least `1` when the cost fo field accesses is `0`

interpreter/interpretable.go
```diff
@@ -873,6 +1028,27 @@ func (fold *evalExhaustiveFold) Eval(ctx Activation) ref.Val {
 	return res
 }
 
+// Cost implements the Coster interface method.
+func (fold *evalExhaustiveFold) Cost() (min, max int64) {
```
TristonianJones, line 1042: Make sure you document the logic for how cost is computed for the fold, can be internal to the function here given the complexity.

TristonianJones (changes requested):

cel/program.go
jinmmin: I see, like singletons

interpreter/attributes.go
jinmmin: done

interpreter/coster.go
jinmmin: done

interpreter/interpretable.go
jinmmin: done

interpreter/interpretable.go
jinmmin: done

interpreter/attributes.go
jinmmin: done

cel/cel_test.go
```diff
@@ -90,6 +90,14 @@ func Example() {
 
 	fmt.Println(out)
 	// Output:Hello world! Nice to meet you, I'm CEL.
+
+	// Calculates the estimate cost for the program.
+	min, max := EstimateCost(prg)
```
TristonianJones: Since this is an example, maybe don't include the cost test here. I'd be happy to have another cost-specific test.

interpreter/attributes_test.go
```diff
@@ -304,6 +332,7 @@ func TestAttributes_ConditionalAttr_TrueBranch(t *testing.T) {
 	vars, _ := NewActivation(data)
 
 	// (true ? a : b.c)[-1][1]
+
```
TristonianJones: nit: extra newline.

interpreter/attributes_test.go
```diff
@@ -499,3 +548,7 @@ func (q *nestedMsgQualifier) Qualify(vars Activation, obj interface{}) (interfac
 	pb := obj.(*proto3pb.TestAllTypes_NestedMessage)
 	return pb.GetBb(), nil
 }
+
+func (q *nestedMsgQualifier) Cost() (min, max int64) {
```
TristonianJones, line 552: Don't forget to add a comment here.

TristonianJones (approved):
Minor nits, but otherwise LGTM

interpreter/attributes_test.go
jinmmin: added

interpreter/attributes_test.go
jinmmin: removed

cel/cel_test.go
jinmmin: moved to a separate test

cel/program.go
```diff
@@ -283,3 +289,29 @@ func (gen *progGen) Eval(input interface{}) (ref.Val, *EvalDetails, error) {
 	}
 	return v, det, nil
 }
+
+// Cost implements the Coster interface method.
+func (gen *progGen) Cost() (min, max int64) {
+	p, err := gen.factory(emptyEvalState)
```
TristonianJones, line 296: Could you add one last comment here about why it's okay to be empty?

jinmmin: done

TristonianJones:
Closes #371

PR #372, opened 2020-07-02
Functions for upper / lower casing ASCII characters in strings.

### LINKED ISSUE #370: strings: add string casing functions
Would be great to have some string casing functions in strings.co extension https://github.com/google/cel-go/tree/master/ext

such as `toUpper`, `toLower` etc @TristonianJones

TristonianJones:
This change introduces extension functions for upper / lower casing ASCII characters in strings.

Closes #370

### DIFF
--- a/ext/README.md
+++ b/ext/README.md
@@ -65,7 +65,7 @@ Examples:
 
 ### LowerAscii
 
-Produces a new string where all ASCII characters are lower-cased.
+Returns a new string where all ASCII characters are lower-cased.
 
 This function does not perform Unicode case-mapping for characters outside the
 ASCII range.
@@ -79,7 +79,7 @@ Examples:
 
 ### Replace
 
-Produces a new string based on the target, which replaces the occurrences of a
+Returns a new string based on the target, which replaces the occurrences of a
 search string with a replacement string if present. The function accepts an
 optional limit on the number of substring replacements to be made.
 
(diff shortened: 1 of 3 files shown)


TristonianJones:
FYI @bitsofinfo regarding https://github.com/tektoncd/triggers/issues/638

ext/README.md
```diff
@@ -64,6 +63,20 @@ Examples:
     'hello mellow'.lastIndexOf('ello', 6)  // returns 1
     'hello mellow'.lastIndexOf('ello', -1) // error
 
+### LowerAscii
+
+Produces a new string where all ASCII characters are lower-cased.
```
JimLarson: "Returns" instead of "produces", for consistency. (Okay, I see "produces" elsewhere, but let's start converging on "returns".) Here and elsewhere.

ext/strings.go
```diff
@@ -386,6 +437,17 @@ func lastIndexOfOffset(str, substr string, offset int64) (int64, error) {
 	return -1, nil
 }
 
+func lowerASCII(str string) string {
+	runes := []rune(str)
+	for i, r := range runes {
+		if r&unicode.MaxASCII == r {
```
JimLarson: `if r <= unicode.MaxASCII` seems to be more idiomatic, but I'm okay with it as-is. Here and below.

JimLarson (approved):
Just some minor changes to consider.

ext/strings.go
TristonianJones: Done.

ext/README.md
TristonianJones: Done.

PR #373, opened 2020-07-06
Encoders extensions with base64 encode/decode support

TristonianJones:
Extensions for base64 encoding and decoding strings and, as appropriate, bytes.

### DIFF
--- a/ext/README.md
+++ b/ext/README.md
@@ -3,6 +3,35 @@
 CEL extensions are a related set of constants, functions, macros, or other
 features which may not be covered by the core CEL spec.
 
+## Encoders
+
+Encoding utilies for marshalling data into standardized representations.
+
+### Base64.Decode
+
+Decodes a base64-encoded string to an unencoded string value.
+
+This function will return an error if the input string is not base64-encoded.
+
+    base64.decode(<string>) -> <string>
+
+Examples:
+
+    base64.decode('aGVsbG8=')  // return 'hello'
+    base64.decode('aGVsbG8')   // error
+
+### Base64.Encode
+
+Encodes a bytes or string input to a base64-encoded string.
+
+    base64.encode(<bytes>) -> <string>
+    base64.encode(<string>) -> <string>
+
+Examples:
+
+    base64.encode('hello')  // return 'aGVsbG8='
+    base64.encode(b'hello') // return 'aGVsbG8='
+
 ## Strings
 
 Extended functions for string manipulation. As a general note, all indices are
--- a/ext/README.md
+++ b/ext/README.md
@@ -9,28 +9,31 @@ Encoding utilies for marshalling data into standardized representations.
 
 ### Base64.Decode
 
-Decodes a base64-encoded string to an unencoded string value.
+Decodes bytes or a base64-encoded string to a sequence of bytes.
 
-This function will return an error if the input string is not base64-encoded.
+This function will return an error if the UTF-8 string input is not
+base64-encoded.
 
-    base64.decode(<string>) -> <string>
+    base64.decode(<bytes>)  -> <bytes>
+    base64.decode(<string>) -> <bytes>
 
 Examples:
 
-    base64.decode('aGVsbG8=')  // return 'hello'
+    base64.decode(b'aGVsbG8=') // return b'hello'
+    base64.decode('aGVsbG8=')  // return b'hello'
     base64.decode('aGVsbG8')   // error
 
 ### Base64.Encode
 
-Encodes a bytes or string input to a base64-encoded string.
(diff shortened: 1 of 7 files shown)


TristonianJones:
Closes #374

ext/README.md
```diff
@@ -3,6 +3,35 @@
 CEL extensions are a related set of constants, functions, macros, or other
 features which may not be covered by the core CEL spec.
 
+## Encoders
+
+Encoding utilies for marshalling data into standardized representations.
+
+### Base64.Decode
+
+Decodes a base64-encoded string to an unencoded string value.
```
JimLarson: I guess I'm showing my age - I first read that as "uuencoded", which seemed like an odd choice.

Base64 decodes to arbitrary binary data, so this should be bytes.  If it's actually ASCII or UTF8 data, the call can be wrapped with `string()`.

ext/README.md
```diff
@@ -3,6 +3,35 @@
...
+    base64.decode(<string>) -> <string>
+
+Examples:
+
+    base64.decode('aGVsbG8=')  // return 'hello'
+    base64.decode('aGVsbG8')   // error
+
+### Base64.Encode
+
+Encodes a bytes or string input to a base64-encoded string.
```
JimLarson: Specify that the string-accepting version converts to utf-8 bytes first. Consider accepting only bytes input and rely composition with explicit `bytes()` conversion, as this might be less error-prone.

JimLarson (changes requested):

ext/README.md
TristonianJones: Fair enough, I'll decode to `bytes`

ext/README.md
TristonianJones: I think I'll still provide functions for both `string` and `bytes` inputs for simplicity on the encoding side since the composition would suffer from the same issues as the `string` based overload.

TristonianJones:
PTAL

ext/README.md
```diff
@@ -3,6 +3,38 @@
...
+Encoding utilies for marshalling data into standardized representations.
+
+### Base64.Decode
+
+Decodes bytes or a base64-encoded string to a sequence of bytes.
+
+This function will return an error if the UTF-8 string input is not
+base64-encoded.
+
+    base64.decode(<bytes>)  -> <bytes>
```
JimLarson, line 17: Specify that the bytes are interpreted as a utf8-encoded string.

ext/README.md
```diff
@@ -3,6 +3,38 @@
...
+    base64.decode(b'aGVsbG8=') // return b'hello'
+    base64.decode('aGVsbG8=')  // return b'hello'
+    base64.decode('aGVsbG8')   // error
+
+### Base64.Encode
+
+Encodes a bytes or string input to base64-encoded bytes.
+
+    base64.encode(<bytes>)  -> <bytes>
+    base64.encode(<string>) -> <bytes>
```
JimLarson, line 32: Specify the the string is converted to bytes via utf8, then encoded.

ext/encoders.go
```diff
@@ -0,0 +1,154 @@
...
+// Encoders returns a cel.EnvOption to configure extended functions for string, byte, and object
+// encodings.
+//
+// Base64.Decode
+//
+// Decodes bytes or a base64-encoded string to a sequence of bytes.
+//
+// This function will return an error if the UTF-8 string input is not base64-encoded.
+//
+//     base64.decode(<bytes>)  -> <bytes>
```
JimLarson, line 38: Same comments from the Readme apply here.

JimLarson (approved):
Looks good modulo some doc/comment improvements. While the CEL language gives utf8 the preferred treatment in the library for the standard string-to-bytes and bytes-to-string conversions, it's not necessarily true that the runtime representation of strings will be utf8 in all CEL implementations.

ext/README.md
TristonianJones: Actually, the `bytes` are just treated as a raw `byte` sequence with no assumptions of utf-8 encoding.

ext/README.md
TristonianJones: Done.

ext/encoders.go
TristonianJones: Done.

ext/README.md
JimLarson: Then how do you interpret the sequence of bytes as characters without assuming the character encoding?  What if I give it some base64-encoded text, using utf16 as the text-to-binary encoding? If the Go library decodes a sequence of bytes, it must do it by assuming a text encoding. That's the assumption that must be surfaced.

PR #375, opened 2020-07-07
Support hexadecimal integer literals

jcking:
Implements support for hexadecimal integer literals, as defined in the [language lexis](https://github.com/google/cel-spec/blob/master/doc/langdef.md#syntax). Tests are being added to [cel-spec](https://github.com/google/cel-spec/pull/136).

### DIFF
--- a/parser/parser.go
+++ b/parser/parser.go
@@ -19,6 +19,7 @@ package parser
 import (
 	"fmt"
 	"strconv"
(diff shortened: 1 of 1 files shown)


TristonianJones:
Closes https://github.com/google/cel-go/issues/376

TristonianJones (approved):

PR #387, opened 2020-08-20
Fixed Residual computation to be non-mutating

### LINKED ISSUE #384: ResidualAst mutates input *Ast
**Describe the bug**
`ResidualAst` mutates the input `*Ast` which causes subsequent evaluations to fail or return incorrect results.

**To Reproduce**

- [ ]  parser
- [ ]  checker
- [x]  interpreter

Test setup:
```go
func Test_ResidualAst_Modified(t *testing.T) {
	e, _ := cel.NewEnv(
		cel.Declarations(
			decls.NewVar("x", decls.NewMapType(decls.String, decls.Int)),
			decls.NewVar("y", decls.Int),
		),
	)
	ast, _ := e.Parse("x == y")
	prg, _ := e.Program(ast,
		cel.EvalOptions(cel.OptTrackState, cel.OptPartialEval),
	)
	for _, x := range []int{123, 456} {
		vars, _ := cel.PartialVars(map[string]interface{}{
			"x": x,
		}, cel.AttributePattern("y"))
		out, det, err := prg.Eval(vars)
		if !types.IsUnknown(out) {
			t.Fatalf("got %v, expected unknown", out)
		}
		if err != nil {
			t.Fatal(err)
		}
		residual, err := e.ResidualAst(ast, det)
		if err != nil {
			t.Fatal(err)
		}
		orig, err := cel.AstToString(ast)
		if err != nil {
			t.Fatal(err)
		}
		if orig != "x == y" {
			t.Errorf("parsed ast: got expr: %s, wanted x == y", orig)
		}
		expr, err := cel.AstToString(residual)
		if err != nil {
			t.Fatal(err)
		}
		want := fmt.Sprintf("%d == y", x)
		if expr != want {
			t.Errorf("residual ast: got expr: %s, wanted %s", expr, want)
		}
	}
}
```

```
=== RUN   Test_ResidualAst_Modified
    filter_test.go:83: parsed ast: got expr: 123 == y, wanted x == y
    filter_test.go:83: parsed ast: got expr: 123 == y, wanted x == y
    filter_test.go:91: residual ast: got expr: 123 == y, wanted 456 == y
--- FAIL: Test_ResidualAst_Modified (0.00s)
```

**Expected behavior**
`ResidualAst` should not be mutating it's input `*Ast` and should instead return a copy.

**Additional context**
~Either `ResidualAst` or `interpreter.PruneAst` should be making a copy of the expr that can then be safely mutated.~

TristonianJones:
Ensure that the Residual expression generation step does not mutate the input AST

Closes #384

### DIFF
--- a/interpreter/prune.go
+++ b/interpreter/prune.go
@@ -105,8 +105,7 @@ func (p *astPruner) maybeCreateLiteral(id int64, val ref.Val) (*exprpb.Expr, boo
 	}
 
 	// Attempt to build a list literal.
-	list, isList := val.(traits.Lister)
-	if isList {
+	if list, isList := val.(traits.Lister); isList {
 		sz := list.Size().(types.Int)
 		elemExprs := make([]*exprpb.Expr, sz)
 		for i := types.Int(0); i < sz; i++ {
@@ -131,8 +130,7 @@ func (p *astPruner) maybeCreateLiteral(id int64, val ref.Val) (*exprpb.Expr, boo
 	}
 
 	// Create a map literal if possible.
-	mp, isMap := val.(traits.Mapper)
-	if isMap {
+	if mp, isMap := val.(traits.Mapper); isMap {
 		it := mp.Iterator()
 		entries := make([]*exprpb.Expr_CreateStruct_Entry, mp.Size().(types.Int))
 		i := 0
@@ -383,10 +381,12 @@ func (p *astPruner) existsWithKnownValue(id int64) bool {
 }
 
 func (p *astPruner) nextID() int64 {
-	_, found := p.state.Value(p.nextExprID)
-	for found {
+	for true {
+		_, found := p.state.Value(p.nextExprID)
+		if found {
+			break
+		}
 		p.nextExprID++
-		_, found = p.state.Value(p.nextExprID)
 	}
 	next := p.nextExprID
 	p.nextExprID++
--- a/interpreter/prune.go
+++ b/interpreter/prune.go
@@ -18,15 +18,18 @@ import (
 	"github.com/google/cel-go/common/operators"
 	"github.com/google/cel-go/common/types"
 	"github.com/google/cel-go/common/types/ref"
+	"github.com/google/cel-go/common/types/traits"
 
 	structpb "github.com/golang/protobuf/ptypes/struct"
 
+	expr "google.golang.org/genproto/googleapis/api/expr/v1alpha1"
 	exprpb "google.golang.org/genproto/googleapis/api/expr/v1alpha1"
 )
 
 type astPruner struct {
-	expr  *exprpb.Expr
-	state EvalState
+	expr       *exprpb.Expr
+	state      EvalState
+	nextExprID int64
 }
 
 // TODO Consider having a separate walk of the AST that finds common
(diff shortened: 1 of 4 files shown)


interpreter/prune.go
```diff
@@ -162,79 +161,124 @@ func (p *astPruner) prune(node *exprpb.Expr) (*exprpb.Expr, bool) {
...
+		call := node.GetCallExpr()
+		args := call.GetArgs()
+		newArgs := make([]*exprpb.Expr, len(args))
+		newCall := &exprpb.Expr_Call{
+			Function: call.GetFunction(),
+			Target:   call.GetTarget(),
+			Args:     newArgs,
+		}
+		for i, arg := range args {
+			newArgs[i] = args[i]
```
JimLarson: Or `newArgs[i] = arg`.

interpreter/prune.go
```diff
@@ -162,79 +161,124 @@ func (p *astPruner) prune(node *exprpb.Expr) (*exprpb.Expr, bool) {
...
 	case *exprpb.Expr_ListExpr:
-		newList := *node.GetListExpr()
+		elems := node.GetListExpr().GetElements()
+		newElems := make([]*exprpb.Expr, len(elems))
 		var prunedList bool
-		var prunedElem bool
-		for i, elem := range node.GetListExpr().Elements {
-			if newList.Elements[i], prunedElem = p.prune(elem); prunedElem {
+		for i, elem := range elems {
+			if newElem, prunedElem := p.prune(elem); prunedElem {
```
JimLarson, line 290: It looks like you also need to copy the non-pruned elements too, just in case a later element in the list is pruned. Otherwise your list will have a bunch of nulls. Please add a test for this case.

interpreter/prune.go
```diff
@@ -162,79 +161,124 @@ func (p *astPruner) prune(node *exprpb.Expr) (*exprpb.Expr, bool) {
...
-				prunedEntry = true
-				newEntry.GetKeyKind().(*exprpb.Expr_CreateStruct_Entry_MapKey).MapKey = newKey
+		entries := node.GetStructExpr().GetEntries()
+		messageType := node.GetStructExpr().GetMessageName()
+		newEntries := make([]*exprpb.Expr_CreateStruct_Entry, len(entries))
+		for i, entry := range entries {
+			newKey, prunedKey := p.prune(entry.GetMapKey())
+			newValue, prunedValue := p.prune(entry.GetValue())
+			if !prunedKey && !prunedValue {
+				newEntries[i] = entry
```
JimLarson: I think you want a `continue` here.

interpreter/prune.go
```diff
@@ -162,79 +161,124 @@ func (p *astPruner) prune(node *exprpb.Expr) (*exprpb.Expr, bool) {
...
 		}
 	case *exprpb.Expr_ComprehensionExpr:
-		if newIterRange, pruned := p.prune(node.GetComprehensionExpr().IterRange); pruned {
-			newExpr := *node
-			newCompre := *newExpr.GetComprehensionExpr()
-			newCompre.IterRange = newIterRange
-			newExpr.GetExprKind().(*exprpb.Expr_ComprehensionExpr).ComprehensionExpr = &newCompre
-			return &newExpr, true
+		compre := node.GetComprehensionExpr()
+		if newRange, pruned := p.prune(compre.GetIterRange()); pruned {
```
JimLarson, line 348: Add a comment explaining why it's sufficient to only test the `iter_range` but not any of the other sub-expressions.

JimLarson (changes requested):

interpreter/prune.go
TristonianJones: Done.

interpreter/prune.go
TristonianJones: This was definitely an oversight. Thanks. Test case added.

interpreter/prune.go
TristonianJones: Done.

interpreter/prune.go
TristonianJones: Will do. We don't go further because the state tracking information only captures a single iteration of the loop so the residual might be inaccurate if attempted.

TristonianJones (commented):
PTAL, I had to adjust the relative attribute construction to ensure that the ids for relative select statements were calculated appropriately.

interpreter/prune.go
```diff
@@ -60,21 +63,118 @@ type astPruner struct {
...
+		return p.createLiteral(id,
+			&exprpb.Constant{ConstantKind: &exprpb.Constant_BytesValue{BytesValue: val.Value().([]byte)}}), true
+	case types.NullType:
+		return p.createLiteral(id,
+			&exprpb.Constant{ConstantKind: &exprpb.Constant_NullValue{NullValue: val.Value().(structpb.NullValue)}}), true
+	}
+
+	// Attempt to build a list literal.
+	list, isList := val.(traits.Lister)
+	if isList {
```
JimLarson: Consider combining lines 108-109 into a single line. Same for `isMap` below.

interpreter/prune.go
```diff
@@ -298,3 +381,14 @@ func (p *astPruner) existsWithKnownValue(id int64) bool {
 	val, valueExists := p.value(id)
 	return valueExists && !types.IsUnknown(val)
 }
+
+func (p *astPruner) nextID() int64 {
+	_, found := p.state.Value(p.nextExprID)
+	for found {
+		p.nextExprID++
+		_, found = p.state.Value(p.nextExprID)
+	}
```
JimLarson, line 390: Consider, even if only briefly, the following idiom (which makes up for the [control structure that's missing](https://en.wikipedia.org/wiki/Control_flow#Loop_with_test_in_the_middle) in most languages):

```
for true {
        _, found := p.state.Value(p.nextExprID)
        if found {
                break
        }
        p.nextExprID++
}
```

interpreter/prune_test.go
```diff
@@ -17,126 +17,129 @@ package interpreter
...
-			test.ExprCall(3,
-				operators.Less,
-				test.ExprIdent(2, "b"),
-				test.ExprLiteral(4, 1.2)),
-			test.ExprCall(6,
-				operators.Equals,
-				test.ExprIdent(5, "c"),
-				test.ExprList(7, test.ExprLiteral(7, "hello")))),
-		P: `_<_(b,1.2)`,
+		expr: `true && false`,
```
JimLarson, line 35: So much nicer!

JimLarson (approved):

PR #394, opened 2020-10-01
Update the CheckedExprToAst utility to handle constant expressions 

TristonianJones:
Constant expressions do not have a reference map and were causing the AST objects not to be recognized as having been checked.


TristonianJones:
Closes #395

JimLarson (approved):

Issue #398, opened 2020-10-20
Golang Protobuf APIv2 Support


jarreds:
**Feature request checklist**

- [x] There are no issues that match the desired change
- [x] The change is large enough it can't be addressed with a simple Pull Request
- [ ] If this is a bug, please file a [Bug Report](./ISSUE_TEMPLATE).

**Change**
Golang recently released [protobuf APIv2](https://blog.golang.org/protobuf-apiv2). This included the new [`protoreflect`](https://pkg.go.dev/google.golang.org/protobuf/reflect/protoreflect) and the associated APIs to dynamically generate protobuf descriptors.

The current CEL reflection based access patterns do not work on `protoreflect` generated messages. For example, the [`getter`](https://github.com/google/cel-go/blob/master/common/types/pb/type.go#L190-L191) defined for current `pb` types is attempting to reflect a method that does not exist for dynamic messages.

It appears a large chunk of protobuf functionality built within CEL is now provided by the new APIv2. Whether it could be wholey subsumed by APIv2, I'm not sure. As it is now though, there is at least some work that needs to be done to allow CEL to use dynamically generated protobufs.

_Edit_

It wasn't clear that we're attempting to use protobufs generated with [`google.golang.org/protobuf/types/dynamicpb`](https://pkg.go.dev/google.golang.org/protobuf/types/dynamicpb).

jarreds:
Shoot, I hadn't seen #385. This is a little more of a feature request than that. It won't hurt my feelings if you close this in favor that.

jarreds:
Amazing. Thank you!

PR #399, opened 2020-11-11
Protobuf API v2 support

### LINKED ISSUE #385: protobuf: update deprecated github.com/golang/protobuf/proto to google.golang.org/protobuf/proto
We are using [golangci-lint](https://github.com/golangci/golangci-lint) with several linters in our project to spot bugs. One of them is complaining with tjhe following message:

> SA1019: package github.com/golang/protobuf/proto is deprecated: Use the "google.golang.org/protobuf/proto" package instead.  (staticcheck)

As we are using this in some part of our project:

```
typeProvider := types.NewRegistry(mymessage)
```

I found myself in the situation where I cannot update my own code to `google.golang.org/protobuf/proto`, because the above line would complain about expecting `github.com/golang/protobuf/proto` not `google.golang.org/protobuf/proto`.

P.D. I would recommend using golangci-lint also here, I made a run over the project and found several apparently unused things, deprecations, etc.

TristonianJones:
Support for the Golang protobuf v2 API

### Value v. ConvertToNative

Calling `Value()` on a proto-based object will likely return a `*dynamicpb.Message` instance rather than the strongly typed proto message. Using `ConvertToNative(reflect.TypeOf(&<DesiredType>{})` will produce the desired result.

### DescriptorProto v. Descriptor

Top-level interfaces have shifted away from using the `DescriptorProto` and `FileDescriptorProto` messages directly in favor of the protobuf v2 API `protoreflect.Descriptor` and `protoreflect.FileDescriptor`. 

### Duration and Timestamp

The `duration.go` and `timestamp.go` object types are now derived from go-native `time.Duration` and `time.Time` respectively as the dependence on the protobuf counterparts was not actually necessary.

## Performance

* Proto operations are now roughly 50% - 100% slower in most cases, but some operations are significantly faster.
* In general this PR improves the evaluation speed by about 10% across the board due to changes in `NativeToValue` conversion.
* The `map.go` and `list.go` implementations were completely refactored to reduce repetitive `ConvertToNative` call paths. Many bugs were fixed, and there is now a slightly higher allocation overhead for these object types, but the overall code is much simpler to maintain.
* Specializations were added for `map[ref.Val]ref.Val` and `[]ref.Val` as these are the values of CEL literals.

## Bug Fixes

* Additional coverage has been added and a number of minor inconsistencies in the `ConvertToNative` call paths of `common/types` objects have been fixed as a result of testing for conformance with the new proto API.
* CEL `map.go` values can be converted to Go-native structs; however, support for configuring a Go native struct from a `proto.Message` field of type `map` is not yet implemented.

## Benchmarks

These were the changes in evaluation speed which had a notable performance difference (>10%):

```
Interpreter/call_varargs-8                       417ns ± 6%     320ns ± 8%   -23.34%
Interpreter/cond-8                               209ns ± 4%     172ns ± 6%   -17.41%
Interpreter/in_map-8                             792ns ± 3%      80ns ± 5%   -89.87%
Interpreter/index-8                              240ns ± 7%     195ns ± 9%   -18.86%
Interpreter/index_relative-8                    2.19µs ± 5%    0.69µs ± 7%   -68.71%
Interpreter/literal_pb3_msg-8                   2.26µs ± 1%    2.89µs ± 3%   +27.96%
Interpreter/literal_pb_enum-8                   2.13µs ± 2%    2.66µs ± 3%   +25.24%
Interpreter/timestamp_eq_timestamp-8             707ns ± 2%      22ns ± 2%   -96.89%
Interpreter/timestamp_lt_timestamp-8            69.7ns ± 1%    58.6ns ± 4%   -15.97%
Interpreter/timestamp_le_timestamp-8            70.5ns ± 0%    60.1ns ± 6%   -14.74%
Interpreter/timestamp_gt_timestamp-8            71.0ns ± 0%    59.8ns ± 5%   -15.77%
Interpreter/timestamp_ge_timestamp-8            75.5ns ±10%    59.9ns ± 3%   -20.62%
Interpreter/macro_all_non_strict_var-8          1.91µs ± 3%    1.61µs ± 1%   -15.98%
Interpreter/macro_has_map_key-8                 1.67µs ± 1%    0.30µs ± 1%   -82.07%
Interpreter/macro_has_pb2_field-8               12.8µs ±71%     7.9µs ± 6%   -37.96%
Interpreter/nested_proto_field-8                 533ns ± 1%     435ns ± 4%   -18.27%
Interpreter/nested_proto_field_with_index-8      900ns ± 1%    1816ns ± 1%  +101.89%
Interpreter/or_true_1st-8                       62.8ns ± 2%    44.7ns ± 4%   -28.91%
Interpreter/or_true_2nd-8                        226ns ± 3%     175ns ± 4%   -22.53%
Interpreter/or_false-8                           221ns ± 1%     196ns ±20%   -11.18%
Interpreter/pkg_qualified_id-8                  58.1ns ± 3%    41.1ns ± 6%   -29.20%
Interpreter/pkg_qualified_id_unchecked-8         332ns ± 1%     101ns ± 4%   -69.63%
Interpreter/pkg_qualified_index_unchecked-8     1.18µs ± 2%    0.28µs ± 6%   -76.60%
Interpreter/select_key-8                        1.49µs ± 1%    1.16µs ± 2%   -22.43%
Interpreter/select_bool_key-8                   4.41µs ± 3%    5.26µs ± 4%   +19.37%
Interpreter/select_index-8                      1.89µs ± 1%    1.37µs ± 2%   -27.33%
Interpreter/select_field-8                      1.04µs ± 3%    1.46µs ± 1%   +41.41%
Interpreter/select_pb2_primitive_fields-8       4.14µs ± 4%    1.84µs ± 2%   -55.66%
Interpreter/select_pb3_wrapper_fields-8         1.62µs ± 2%    3.38µs ± 2%  +108.76%
Interpreter/select_pb3_compare-8                 194ns ± 1%     167ns ± 6%   -13.89%
Interpreter/select_custom_pb3_compare-8          116ns ± 1%      75ns ± 6%   -35.17%
Interpreter/select_subsumed_field-8             50.6ns ± 3%    32.5ns ± 5%   -35.77%
```

Closes #385 #398 #400


glerchundi:
This is awesome @TristonianJones, keep up the good work!

achew22:
To echo what @glerchundi said, great work @TristonianJones

TristonianJones:
Thanks @achew22 and @glerchundi. I'm happy to finally get this done.

TristonianJones:
@JimLarson I think it's ready for a first pass. I couldn't enable proto3 `optional` fields yet because I can't figure out how to get bazel to recognize a very specific experimental flag that needs to be set to use them publicly. Once this is better understood, I'll follow up with more tests.

WORKSPACE
```diff
@@ -1,117 +1,127 @@
...
-  name = "org_golang_google_genproto",
-  build_file_proto_mode = "disable",
-  commit = "bd91e49a0898e27abb88c339b432fa53d7497ac0",
-  importpath = "google.golang.org/genproto",
+http_archive(
+    name = "rules_proto",
+    sha256 = "8e7d59a5b12b233be5652e3d29f42fba01c7cbab09f6b3a8d0a57ed6d1e9a0da",
+    strip_prefix = "rules_proto-7e4afce6fe62dbff0a4a03450143146f9f2d7488",
+    urls = [
+        "https://mirror.bazel.build/github.com/bazelbuild/rules_proto/archive/7e4afce6fe62dbff0a4a03450143146f9f2d7488.tar.gz",
```
JimLarson, line 29: Here and for the other references that use a cryptic hash instead of a meaningful build identifier: add a comment explaining why or how this revision was chosen, e.g. "latest version as of 2020-11-19 to include fix for issue X".

cel/options.go
```diff
@@ -228,30 +229,50 @@ func Types(addTypes ...interface{}) EnvOption {
...
+//
+// Note that messages instantiated from these descriptors will be *dynamicpb.Message values
+// rather than the concrete message type.
 func TypeDescs(descs ...interface{}) EnvOption {
 	return func(e *Env) (*Env, error) {
 		reg, isReg := e.provider.(ref.TypeRegistry)
 		if !isReg {
 			return nil, fmt.Errorf("custom types not supported by provider: %T", e.provider)
 		}
+		var fds *descpb.FileDescriptorSet
```
JimLarson, line 249: Add comment: "Scan descs for FileDescriptorProtos and synthesize a FileDescriptorSet from them because..." and add the reason why. Also indicate whether it's important for this step to happen first.

cel/options.go
```diff
@@ -228,30 +229,50 @@ func Types(addTypes ...interface{}) EnvOption {
...
 
-// TypeDescs adds type declarations for one or more protocol buffer
-// FileDescriptorProtos or FileDescriptorSets.  Note that types added
-// via descriptor will not be able to instantiate messages, and so are
-// only useful for Check() operations.
+// TypeDescs adds type declarations from any protoreflect.FileDescriptor, protoregistry.Files,
+// google.protobuf.FileDescriptorProto or google.protobuf.FileDescriptorSet provided.
+//
+// Note that messages instantiated from these descriptors will be *dynamicpb.Message values
+// rather than the concrete message type.
```
JimLarson, line 236: Overarching concern: Expr Service relies on different CEL Envs being hermetic with respect to registering new FileDescriptorSets. We need to capture that invariant in documentation and in a test.

lopezator:
Awesome!

common/types/bool.go
```diff
@@ -66,26 +65,28 @@ func (b Bool) Compare(other ref.Val) ref.Val {
...
+			return anypb.New(wrapperspb.Bool(bool(b)))
 		case boolWrapperType:
 			// Convert the bool to a protobuf.BoolValue.
-			return &wrapperspb.BoolValue{Value: bool(b)}, nil
+			return wrapperspb.Bool(bool(b)), nil
 		case jsonValueType:
-			return &structpb.Value{
-				Kind: &structpb.Value_BoolValue{BoolValue: bool(b)},
-			}, nil
+			return structpb.NewBoolValue(bool(b)), nil
```
JimLarson, line 79: Note that it returns google.protobuf.Value?

common/types/bool.go
```diff
@@ -66,26 +65,28 @@ func (b Bool) Compare(other ref.Val) ref.Val {
...
-		return bool(b), nil
+		return reflect.ValueOf(b).Convert(typeDesc).Interface(), nil
 	case reflect.Ptr:
 		switch typeDesc {
 		case anyValueType:
 			// Primitives must be wrapped before being set on an Any field.
-			return ptypes.MarshalAny(&wrapperspb.BoolValue{Value: bool(b)})
+			return anypb.New(wrapperspb.Bool(bool(b)))
 		case boolWrapperType:
 			// Convert the bool to a protobuf.BoolValue.
```
JimLarson: google.protobuf.BoolValue

common/types/err.go
```diff
@@ -42,6 +42,12 @@ func NoSuchOverloadErr() ref.Val {
 	return NewErr("no such overload")
 }
 
+// UnsupportedTypeConversionErr returns a types.NewErr instance with a no such type conversion
+// message that indicates that the native value could not be converted to a CEL ref.Val.
+func UnsupportedTypeConversionErr(val interface{}) ref.Val {
+	return NewErr("no such type conversion to ref.Val: (%T)%v", val, val)
```
JimLarson: The phrase "type conversion" suggests a conversion from type T1 to type T2, which further suggests that the error message should include both a "from" and a "to" type for the value. Consider calling this a "ref.Val conversion"?

common/types/map.go
```diff
@@ -68,18 +102,21 @@ var (
...
-// Contains implements the traits.Container interface method.
-func (m *baseMap) Contains(index ref.Val) ref.Val {
-	val, found := m.Find(index)
-	// When the index is not found and val is non-nil, this is an error or unknown value.
-	if !found && val != nil && IsUnknownOrError(val) {
-		return val
-	}
-	return Bool(found)
+type mapAccessor interface {
+	Find(ref.Val) (ref.Val, bool)
```
JimLarson, line 114: Comment to explain meaning of argument and returned values.

common/types/map.go
```diff
@@ -263,59 +257,382 @@ func (m *baseMap) Equal(other ref.Val) ref.Val {
...
+	}
+}
+
+type jsonStructAccessor struct {
+	ref.TypeAdapter
+	st map[string]*structpb.Value
+}
+
+// Find searches the json struct field map for the input key value and returns the value and 'true'
+// if found.
```
JimLarson: Returns <value>,true if found, nil,false if not found, and <err>,false on error. Hoist this up to where Find() is declared and reference it there.

common/types/map.go
```diff
@@ -263,59 +257,382 @@ func (m *baseMap) Equal(other ref.Val) ref.Val {
...
+		return nil, false
+	}
+	return a.NativeToValue(keyVal), true
+}
+
+// Iterator creates a new traits.Iterator from the set of JSON struct field names.
+func (a *jsonStructAccessor) Iterator() traits.Iterator {
+	mapKeys := make([]string, len(a.st))
+	i := 0
+	for k := range a.st {
```
JimLarson, line 338: Consider `for k, _ := range a.st {` to make it explicit that we're ignoring the map values.

common/types/map.go
```diff
@@ -263,59 +257,382 @@ func (m *baseMap) Equal(other ref.Val) ref.Val {
...
+	keyVal, found := a.st[string(strKey)]
+	if !found {
+		return nil, false
+	}
+	return a.NativeToValue(keyVal), true
+}
+
+// Iterator creates a new traits.Iterator from the set of JSON struct field names.
+func (a *jsonStructAccessor) Iterator() traits.Iterator {
+	mapKeys := make([]string, len(a.st))
```
JimLarson, line 336: Comment that we copy the keys to make their order stable, as this is another surprising Go feature.

common/types/map.go
```diff
@@ -263,59 +257,382 @@ func (m *baseMap) Equal(other ref.Val) ref.Val {
...
 }
 
-// Get implements the traits.Indexer interface method.
-func (m *baseMap) Get(key ref.Val) ref.Val {
-	v, found := m.Find(key)
+// Iterator creates a new traits.Iterator from the string key set of the map.
+func (a *stringMapAccessor) Iterator() traits.Iterator {
+	mapKeys := make([]string, len(a.mapVal))
+	i := 0
+	for k := range a.mapVal {
```
JimLarson, line 459: Consider adding an explicit `_` to make it clear we're skipping the map values.

common/types/map.go
```diff
@@ -263,59 +257,382 @@ func (m *baseMap) Equal(other ref.Val) ref.Val {
...
-	return String(val), true
+	return String(keyVal), true
 }
 
-// Get implements the traits.Indexer interface method.
-func (m *baseMap) Get(key ref.Val) ref.Val {
-	v, found := m.Find(key)
+// Iterator creates a new traits.Iterator from the string key set of the map.
+func (a *stringMapAccessor) Iterator() traits.Iterator {
+	mapKeys := make([]string, len(a.mapVal))
```
JimLarson, line 457: Comment that we copy the keys to make iteration stable.

common/types/map.go
```diff
@@ -263,59 +257,382 @@ func (m *baseMap) Equal(other ref.Val) ref.Val {
...
 	if !found {
-		return ValOrErr(v, "no such key: %v", key)
+		return nil, false
+	}
+	return a.NativeToValue(keyVal), true
+}
+
+// Iterator creates a new traits.Iterator from the string key set of the map.
+func (a *stringIfaceMapAccessor) Iterator() traits.Iterator {
+	mapKeys := make([]string, len(a.mapVal))
```
JimLarson, line 501: Same comments as above for `Iterator()`.

common/types/pb/file.go
```diff
@@ -124,28 +111,30 @@ func collectFileMetadata(fileDesc *descpb.FileDescriptorProto) *fileMetadata {
 
 // collectMsgTypes recursively collects messages and nested messages into a map of fully
 // qualified message names to message descriptors.
```
JimLarson: Update docstring to indicate that it populates the enum map too.

common/types/pb/file.go
```diff
@@ -96,26 +91,18 @@ func sanitizeProtoName(name string) string {
 // fileMetadata is a flattened view of message types and enum values within a file descriptor.
 type fileMetadata struct {
 	// msgTypes maps from fully-qualified message name to descriptor.
-	msgTypes map[string]*descpb.DescriptorProto
+	msgTypes map[string]protoreflect.MessageDescriptor
 	// enumValues maps from fully-qualified enum value to enum value descriptor.
-	enumValues map[string]*descpb.EnumValueDescriptorProto
+	enumValues map[string]protoreflect.EnumValueDescriptor
```
JimLarson, line 96: We'll want to collect the enum types too, if we want to create distinct checker-time types for them and named conversion functions.

common/types/pb/pb.go
```diff
@@ -17,38 +17,36 @@
...
+	emptypb "google.golang.org/protobuf/types/known/emptypb"
+	structpb "google.golang.org/protobuf/types/known/structpb"
+	tspb "google.golang.org/protobuf/types/known/timestamppb"
+	wrapperspb "google.golang.org/protobuf/types/known/wrapperspb"
 )
 
 // Db maps from file / message / enum name to file description.
 type Db struct {
 	revFileDescriptorMap map[string]*FileDescription
+	files                []*FileDescription
```
JimLarson: Document what the "files" field is for. Is it the de-duped set of values from the reverse map?

common/types/pb/file.go
```diff
@@ -67,11 +64,9 @@ func (fd *FileDescription) GetEnumNames() []string {
 
 // GetTypeDescription returns a TypeDescription for a qualified type name
```
JimLarson: Feel free to ignore this, since it's sort of outside the scope of this PR, but it's not clear what a "type" is in this context, since it appears to be only protobuf message types.

common/types/pb/type.go
```diff
@@ -17,280 +17,182 @@ package pb
...
+	return unwrap(td, msg)
 }
 
-// ReflectType returns the reflected struct type of the generated proto struct.
-func (td *TypeDescription) ReflectType() reflect.Type {
-	if td.getMetadata().reflectedType == nil {
-		return nil
-	}
-	return *td.getMetadata().reflectedType
+// Name of the type.
```
JimLarson: (fully qualified)

common/types/pb/type.go
```diff
@@ -17,280 +17,182 @@ package pb
...
-func (td *TypeDescription) FieldCount() int {
-	// The number of keys in the field indices map corresponds to the number
-	// of fields on the proto message.
-	return len(td.getMetadata().fieldIndices)
+// FieldMap returns a string field name to FieldDescription map.
+func (td *TypeDescription) FieldMap() map[string]*FieldDescription {
+	return td.fieldMap
 }
 
 // FieldByName returns the FieldDescription associated with a field name.
```
JimLarson: Bool indicates whether a field of that name exists.

common/types/pb/type.go
```diff
@@ -17,280 +17,182 @@ package pb
...
-
-	// td is the type description for message typed fields.
-	td *TypeDescription
-
-	// proto descriptor data.
-	desc      *descpb.FieldDescriptorProto
-	index     int
-	prop      *proto.Properties
-	oneofProp *proto.OneofProperties
+	// KeyType returns the key FieldDescription for map fields.
```
JimLarson: s/returns/holds/ here and below

common/types/pb/type.go
```diff
@@ -299,203 +201,141 @@ func (fd *FieldDescription) CheckedType() *exprpb.Type {
...
-	fieldVal := fd.getter.Call([]reflect.Value{t})[0]
-	return isFieldSet(fieldVal)
 }
 
 // GetFrom returns the accessor method associated with the field on the proto generated struct.
 //
 // If the field is not set, the proto default value is returned instead.
 //
-// The input target may either be a reflect.Value or Go struct type.
+// The input target may either be a proto.Message or protoreflect.Message.
```
JimLarson: The next few lines seem to say that it must be a proto.Message.

common/types/pb/type.go
```diff
@@ -299,203 +201,141 @@ func (fd *FieldDescription) CheckedType() *exprpb.Type {
...
-		fd.TypeName(), fd.OrigName(), fd.IsOneof())
+// ReflectType returns the Golang reflect.Type for this field.
+func (fd *FieldDescription) ReflectType() reflect.Type {
+	return fd.reflectType
 }
 
-// Type returns the TypeDescription for the field.
-func (fd *FieldDescription) Type() *TypeDescription {
-	return fd.td
+// String returns a struct-like field definition string.
```
JimLarson: It's not clear what this means - give an example?

common/types/pb/type.go
```diff
@@ -518,24 +358,170 @@ func checkedWrap(t *exprpb.Type) *exprpb.Type {
...
+		msg = desc.Zero()
+	}
+	// In order to ensure that these wrapped types match the expectations of the CEL type system
+	// the dynamicpb.Message must be merged with an protobuf instance of the well-known type value.
+	typeName := string(refMsg.Descriptor().FullName())
+	switch typeName {
+	case "google.protobuf.Any":
+		unwrapped := &anypb.Any{}
+		proto.Merge(unwrapped, msg)
+		return unwrapped, true
```
JimLarson: You need to iterate/recurse on the unwrapped value, as it might be subject to further unwrapping.

common/types/provider.go
```diff
@@ -59,14 +57,23 @@ func NewRegistry(types ...proto.Message) ref.TypeRegistry {
 		TimestampType,
 		TypeType,
 		UintType)
+	if err != nil {
+		return nil, err
+	}
 
+	for _, fd := range p.pbdb.FileDescriptions() {
```
JimLarson, line 64: Comment to explain why this is necessary.

common/types/provider.go
```diff
@@ -302,237 +246,280 @@ var (
...
+		if v != nil {
+			return Uint(*v), true
 		}
 	case *uint64:
-		if ptr := value.(*uint64); ptr != nil {
-			return Uint(*ptr)
+		if v != nil {
+			return Uint(*v), true
 		}
 	case []byte:
```
JimLarson, line 349: Comment that there are still many reasonable native slice and map types which could be supported, and support will be added as requested.

common/types/provider.go
```diff
@@ -302,237 +246,280 @@ var (
...
-	case *structpb.ListValue:
-		if ptr := value.(*structpb.ListValue); ptr != nil {
-			return NewJSONList(a, ptr)
+		unpackedAny, err := v.UnmarshalNew()
+		if err != nil {
+			return NewErr("anypb.UnmarshalNew() failed for type %q: %v", v.GetTypeUrl(), err), true
 		}
-	case structpb.NullValue, *structpb.NullValue:
-		return NullValue
+		return a.NativeToValue(unpackedAny), true
```
JimLarson, line 372: Might this call result in an error? If so, detect and return _, false.

JimLarson (changes requested):

common/types/provider.go
TristonianJones: If an error occurs here, it will be returned to the caller which is the desired behavior. Fixing this behavior to more clearly spell out an error upon conversion might be desired, but I'm not sure it's needed and it would be a breaking change.

common/types/provider.go
TristonianJones: Done.

common/types/provider.go
TristonianJones: Done.

common/types/pb/type.go
TristonianJones: Comment added. Let me know what you think. I'm on the fence about it only because it seems a little ugly to change, but maybe I just need to mull it over more.

common/types/pb/type.go
TristonianJones: Reworded.

common/types/pb/type.go
TristonianJones: Comments added.

common/types/pb/type.go

common/types/pb/type.go

common/types/pb/type.go

common/types/pb/file.go
TristonianJones: Comment minorly updated.

common/types/pb/pb.go
TristonianJones: Added, could probably be improved in the future.

common/types/pb/file.go
TristonianJones: TODO added.

common/types/pb/file.go
TristonianJones: Done.

common/types/map.go
TristonianJones: Done.

common/types/map.go
TristonianJones: Done.

common/types/map.go
TristonianJones: Done.

common/types/map.go
TristonianJones: The `go lint` tool complains about the unused `_` here.

common/types/map.go
TristonianJones: This language is what's used on the `traits.Mapper`. I've added it to the `mapAccessor` and other map implementations within this file as well.

common/types/map.go
TristonianJones: Done.

common/types/err.go
TristonianJones: Updated.

common/types/bool.go
TristonianJones: Done.

common/types/bool.go
TristonianJones: Done.

cel/options.go
TristonianJones: I'll update the docs, but luckily we do have a test to ensure the hermetic nature of the environments.

cel/options.go
TristonianJones: Done.

WORKSPACE
TristonianJones: Oddly enough, this is the recommended embed style, but I hear you: https://github.com/bazelbuild/rules_proto. I added some comments.

TristonianJones:
PTAL

common/types/bool.go
JimLarson: I'm still not seeing the `google.` prefix. Or maybe refer to it by its golang-level name `wrapperspb.BoolValue`?

common/types/bool.go
JimLarson: Still missing the `google.` prefix. Or maybe refer to it by its golang-level name `structpb.Value`?

common/types/pb/type.go
JimLarson: I don't think the comment is sufficient - it tries to put the responsibility on the callers, yet the callers just one page up in this very same file don't attempt further unwrapping. Ugly or not this package needs to have a handy way of bottoming out our automatic proto conversions.

JimLarson (changes requested):
Almost done. I added replies on naming the right returned type in `bool.go` and handling Any messages wrapping other wrappers in `type.go`.

common/types/bool.go
TristonianJones: Opted for `wrapperspb` and updated other references.

common/types/pb/type.go
TristonianJones: I updated the code to do the unwrapping and removed a specialization for wrapper types in dynamic protobufs which is probably just fine.

TristonianJones:
PTAL

JimLarson (approved):

PR #411, opened 2021-01-29
Ensure protodesc.FileDescriptor references are sourced from the global proto registry

TristonianJones:
Round-tripping from a `protoreflect.FileDescriptor` to a `FileDescriptorProto` and back results in a `protoreflect.FileDescriptor` value which is incompatible with the original. This is by design. Based on this choice, all `protoreflect.FileDescriptor` values must be checked to see if they are also defined in the `protoregistry.GlobalFiles` and if so, sourced from the global proto registry rather than taken as-is.

### DIFF
--- a/common/types/pb/pb.go
+++ b/common/types/pb/pb.go
@@ -19,6 +19,7 @@ package pb
 import (
 	"google.golang.org/protobuf/proto"
 	"google.golang.org/protobuf/reflect/protoreflect"
+	"google.golang.org/protobuf/reflect/protoregistry"
 
 	anypb "google.golang.org/protobuf/types/known/anypb"
 	durpb "google.golang.org/protobuf/types/known/durationpb"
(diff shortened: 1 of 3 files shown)


common/types/pb/pb.go
```diff
@@ -93,6 +94,18 @@ func (pbdb *Db) RegisterDescriptor(fileDesc protoreflect.FileDescriptor) (*FileD
...
 		return fd, nil
 	}
+	// Make sure to search the global registry to see if a protoreflect.FileDescriptor for
+	// the file specified has been linked into the binary. If so, use the copy of the descriptor
+	// from the global cache.
+	//
+	// Note: Proto reflection relies on descriptor values being object equal rather than object
+	// equivalence. This choice means that a FieldDescriptor generated from a FileDescriptorProto
+	// will be incompatible with the FieldDescriptor in the global registry and any message created
+	// from that global registry.
```
JimLarson, line 108: Document somewhere the internal constraint that the cel library must not register anything in protoregistry.GlobalFiles as doing so might violate the isolation needed by ExprService.

JimLarson (approved):

common/types/pb/pb.go
TristonianJones: Done.

PR #413, opened 2021-02-08
Ensure that proto reflection falls back to dynamic lookup

TristonianJones:
When proto descriptors are not shared by CEL and an input proto message, make sure that the `GetFrom` and `IsSet` tests fallback to dynamic `FieldDescriptor` lookups rather than using cached values.

### DIFF
--- a/common/types/pb/type.go
+++ b/common/types/pb/type.go
@@ -202,7 +202,7 @@ func (fd *FieldDescription) IsSet(target interface{}) bool {
 		pbDesc := pbRef.Descriptor()
 		if pbDesc == fd.desc.ContainingMessage() {
 			// When the target protobuf shares the same message descriptor instance as the field
-			// descriptoruse the cached field descriptor value.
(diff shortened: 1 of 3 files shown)


common/types/pb/type.go
```diff
@@ -198,7 +198,16 @@ func (fd *FieldDescription) Descriptor() protoreflect.FieldDescriptor {
 func (fd *FieldDescription) IsSet(target interface{}) bool {
 	switch v := target.(type) {
 	case proto.Message:
-		return v.ProtoReflect().Has(fd.desc)
+		pbRef := v.ProtoReflect()
+		pbDesc := pbRef.Descriptor()
+		if pbDesc == fd.desc.ContainingMessage() {
+			// When the target protobuf shares the same message descriptor instance as the field
+			// descriptoruse the cached field descriptor value.
```
JimLarson: "descriptoruse" -> "descriptor, use"

JimLarson (approved):

PR #421, opened 2021-03-25
Return errors when signed or unsigned integer overflow is detected

jcking:
When integer arithmetic would result in overflow or underflow and error should be returned. This pull request brings CEL go runtime into compliance with the CEL spec on this topic.

### DIFF
--- a/common/types/int.go
+++ b/common/types/int.go
@@ -16,6 +16,7 @@ package types
 
 import (
 	"fmt"
+	"math"
 	"reflect"
 	"strconv"
 	"time"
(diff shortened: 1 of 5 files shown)


TristonianJones:
Would you mind also removing line 18 from `conformance/BUILD.bazel`:
https://github.com/google/cel-go/blob/master/conformance/BUILD.bazel#L18

jcking:
/gcbrun

EDIT: never mind cannot run it myself. Regardless removed line 18.

PR #422, opened 2021-03-26
Return errors when duration or timestamp overflow is detected

jcking:
Duration evaluation, if arithmetic on duration or timestamp causes overflow return an error. Additionally this pull request enables the timestamp and duration conformance tests and implements the functionality in the conformance server.

Timestamp range is defined per `google.protobuf.Timestamp` to be the range `0001-01-01T00:00:00Z` to `9999-12-31T23:59:59.999999999Z`.

### DIFF
--- a/common/types/int.go
+++ b/common/types/int.go
@@ -16,7 +16,6 @@ package types
 
 import (
 	"fmt"
-	"math"
 	"reflect"
 	"strconv"
 	"time"
@@ -64,10 +63,10 @@ func (i Int) Add(other ref.Val) ref.Val {
 	if !ok {
 		return ValOrErr(other, "no such overload")
 	}
-	if (otherInt > 0 && i > math.MaxInt64-otherInt) || (otherInt < 0 && i < math.MinInt64-otherInt) {
-		return NewErr("integer overflow")
+	if val, ok := addInt64Overflow(int64(i), int64(otherInt)); ok {
+		return Int(val)
 	}
-	return i + otherInt
+	return NewErr("integer overflow")
 }
 
 // Compare implements traits.Comparer.Compare.
@@ -166,8 +165,12 @@ func (i Int) ConvertToType(typeVal ref.Type) ref.Val {
 	case StringType:
 		return String(fmt.Sprintf("%d", int64(i)))
 	case TimestampType:
-		t := time.Unix(int64(i), 0).UTC()
-		return Timestamp{Time: t}
+		// The maximum positive value that can be passed to time.Unix is math.MaxInt64 minus the number
+		// of seconds between year 1 and year 1970. See comments on unixToInternal.
+		if int64(i) < minUnixTime || int64(i) > maxUnixTime {
+			return NewErr("timestamp overflow")
+		}
+		return timestampOf(time.Unix(int64(i), 0).UTC())
 	case TypeType:
 		return IntType
 	}
@@ -183,11 +186,10 @@ func (i Int) Divide(other ref.Val) ref.Val {
 	if otherInt == IntZero {
 		return NewErr("divide by zero")
 	}
-	// In twos complement, negating MinInt64 would result in a valid of MaxInt64+1.
-	if i == math.MinInt64 && otherInt == -1 {
-		return NewErr("integer overflow")
+	if val, ok := divideInt64Overflow(int64(i), int64(otherInt)); ok {
+		return Int(val)
 	}
-	return i / otherInt
+	return NewErr("integer overflow")
 }
 
 // Equal implements ref.Val.Equal.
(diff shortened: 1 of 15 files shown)


jcking:
I believe I need to address time zones in overflow detection, so please hold till I can verify.

server/server.go
```diff
@@ -33,7 +34,9 @@ import (
 	confpb "google.golang.org/genproto/googleapis/api/expr/conformance/v1alpha1"
 	exprpb "google.golang.org/genproto/googleapis/api/expr/v1alpha1"
 	rpc "google.golang.org/genproto/googleapis/rpc/status"
-	anypb "google.golang.org/protobuf/types/known/anypb"
+	"google.golang.org/protobuf/types/known/anypb"
```
TristonianJones: I know that OSS go files don't complain about a missing alias, but internally the code wants these aliases, so if you wouldn't mind, please preserve them and use the designators `anypb`, `dpb`, `tpb`.  The `rpc` one actually needs to be changed to `rpcpb`

common/types/uint.go
```diff
@@ -57,10 +57,10 @@ func (i Uint) Add(other ref.Val) ref.Val {
 	if !ok {
 		return ValOrErr(other, "no such overload")
 	}
-	if otherUint > 0 && i > math.MaxUint64-otherUint {
-		return NewErr("unsigned integer overflow")
+	if val, ok := addUint64Overflow(uint64(i), uint64(otherUint)); ok {
+		return Uint(val)
 	}
-	return i + otherUint
+	return NewErr("unsigned integer overflow")
```
TristonianJones: Could you actually add a constant error value (or equivalent) to `err.go`?

common/types/timestamp.go
```diff
@@ -37,6 +37,21 @@ type Timestamp struct {
...
 }
 
+func timestampOf(t time.Time) Timestamp {
+	// Note that this function does not valiate that time.Time is in our supported range.
+	return Timestamp{Time: t}
+}
+
+const (
+	// The number of seconds between year 1 and year 1970.
+	unixToInternal int64 = (1969*365 + 1969/4 - 1969/100 + 1969/400) * (60 * 60 * 24)
```
TristonianJones, line 48: Could you add a comment about the fact that these values come from https://golang.org/src/time/time.go?

common/types/int.go
```diff
@@ -64,10 +63,10 @@ func (i Int) Add(other ref.Val) ref.Val {
 	if !ok {
 		return ValOrErr(other, "no such overload")
 	}
-	if (otherInt > 0 && i > math.MaxInt64-otherInt) || (otherInt < 0 && i < math.MinInt64-otherInt) {
-		return NewErr("integer overflow")
+	if val, ok := addInt64Overflow(int64(i), int64(otherInt)); ok {
+		return Int(val)
 	}
-	return i + otherInt
+	return NewErr("integer overflow")
```
TristonianJones: Consider adding a new error constant value to `err.go`

common/types/overflow.go
```diff
@@ -0,0 +1,237 @@
...
+import (
+	"math"
+	"time"
+)
+
+const (
+	nanosPerSecond = 1000000000
+)
+
+func addInt64Overflow(x, y int64) (int64, bool) {
```
TristonianJones: Consider dropping the `Overflow` suffix from the method name. I'm torn between having a comment about the meaning of the `bool` return value or suggesting that you return a constant `error` value to make the contract more clear.

server/server.go
```diff
@@ -272,6 +275,28 @@ func RefValueToValue(res ref.Val) (*exprpb.Value, error) {
 	case types.UintType:
 		return &exprpb.Value{
 			Kind: &exprpb.Value_Uint64Value{Uint64Value: res.Value().(uint64)}}, nil
+	case types.DurationType:
```
TristonianJones, line 278: Good catch. Wonder why that didn't get tripped up before?

common/types/overflow.go
```diff
@@ -0,0 +1,237 @@
...
+	sec2 := int64(y) / nanosPerSecond  // Truncate to seconds.
+	nsec2 := int64(y) % nanosPerSecond // Get remainder.
+
+	// Add seconds first, detecting any overflow.
+	sec, ok := addInt64Overflow(sec1, sec2)
+	if !ok {
+		return time.Time{}, false
+	}
+
+	// Nanoseconds cannot overflow.
```
TristonianJones: Maybe say `cannot overflow as they are capped to 999,999,999 by the time.Time implementation`

common/types/overflow.go
```diff
@@ -0,0 +1,282 @@
...
+
+package types
+
+import (
+	"math"
+	"time"
+)
+
+const (
+	nanosPerSecond = 1000000000
```
TristonianJones: Consider using `int64(time.Second)` or simply using the `time.Second` value directly, deferring `Duration` -> `int64` conversions until absolutely necessary.

TristonianJones (approved):
One last item to address.

PR #426, opened 2021-04-01
Port optimized antlr.CharStream implementation to Go and use map for reserved IDs

jcking:


parser/input.go
```diff
@@ -0,0 +1,383 @@
...
+
+// asciiCodePointBuffer is an implementation for an array of code points that contain code points
+// only from the ASCII character set.
+type asciiCodePointBuffer struct {
+	arr []byte
+	pos int
+	src string
+}
+
+func (a *asciiCodePointBuffer) Consume() {
```
TristonianJones, line 33: Don't forget to add comments on exported methods.

parser/input.go
```diff
@@ -0,0 +1,383 @@
...
+	idx := 0
+	buf8 := make([]byte, 0, len(text))
+	for idx < len(text) {
+		r, s := utf8.DecodeRuneInString(text[idx:])
+		idx += s
+		if r < utf8.RuneSelf {
+			buf8 = append(buf8, byte(r))
+			continue
+		}
+		if r <= 0xffff {
```
TristonianJones, line 371: Rather than replicating the for-loops, consider using labels to jump to the appropriate loop.

```go
Copy16:
  buf16 = append(buf16, uint16(r))
  for idx < len(text) {
     r, s = utf8.DecodeRuneInString(text[idx:])
     idx += s				
     if r > 0xffff {
        buf32 := make([]rune, len(buf16), len(text))
        for i, v := range buf16 {
            buf32[i] = rune(uint32(v))
        }
        goto Copy32
     }
     buf16 = append(buf16, uint16(r))
     ...
  }
  return ...
Copy32:
  buf32 = append(buf32, r)
  for idx < len(text) {
  ...
  }
  return ...
```

TristonianJones (changes requested):

parser/input.go
jcking: Fixed.

parser/input.go
jcking: Done.

TristonianJones (approved):

PR #428, opened 2021-04-07
Introduce abstraction over source contents

jcking:
Introduces an abstraction for accessing a contiguous array of runes that facilitates deduplicating the source content within CEL and provides an optimized interface for interacting with the source contents.


TristonianJones (approved):
Nice!

PR #429, opened 2021-04-08
Export the built-in parser macros individually

TristonianJones:
Currently marcos are either custom, all, or none. This change exposes the builtin `parser` macros by name, making it easier to customize the macro set enabled for the application.

### DIFF
--- a/cel/macro.go
+++ b/cel/macro.go
@@ -1,0 +1,88 @@
+// Copyright 2022 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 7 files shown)


TristonianJones:
FYI @fernandolobato @rblank

jcking (dismissed):

TristonianJones:
PTAL @jcking

cel/macro.go
```diff
@@ -0,0 +1,88 @@
...
+// The MacroExpander accepts as arguments a MacroExprHelper as well as the arguments used in the function call
+// and produces as output an Expr ast node.
+//
+// Note: when the Macro.IsReceiverStyle() method returns true, the target argument will be nil.
+type MacroExpander = parser.MacroExpander
+
+// MacroExprHelper exposes helper methods for creating new expressions within a CEL abstract syntax tree.
+type MacroExprHelper = parser.ExprHelper
+
+var (
```
jcking: I get that using var is easier then re-declaring, but let's explicitly comment that re-assigning these is not supported.

TristonianJones: Done, wrapped into redeclarations with a note on variables which are simply being pushed upward into the `cel` package and a caution not to rename them.

jcking (approved):

Issue #420, opened 2021-03-22
Expose a Restriction for Compiling Expressions With Nested Macros


fernandolobato:
There are no linguistic limits on the amount of nesting permitting when using macros that allow arbitrary predicates such as the `all` macro. 

This can result in exponential runtime evaluation for expressions. The .all() macro is very useful for list variables and we would like to use it while limiting nested calls.

PR #430, opened 2021-04-08
Implement recursion limits and error recovery attempt limits in the parser

jcking:
This pull request ports the recursion limit and recovery attempt limit to the Go parser. This allows users to protect themselves from arbitrarily complex expressions that could cause stack limits to be exceeded or time complexity to drastically increase.

### DIFF
--- a/parser/options.go
+++ b/parser/options.go
@@ -1,0 +1,65 @@
+// Copyright 2021 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package parser
+
+import "fmt"
+
+type options struct {
+	maxRecursionDepth  int
+	errorRecoveryLimit int
+	macros             map[string]Macro
+}
+
+// Option configures the behavior of the parser.
+type Option func(*options) error
+
+// MaxRecursionDepth is an option which limits the maximum depth the parser will attempt to
+// parse the expression before giving up.
+func MaxRecursionDepth(maxRecursionDepth int) Option {
+	return func(opts *options) error {
+		if maxRecursionDepth < -1 {
+			return fmt.Errorf("max recursion depth must be greater than or equal to -1: %d", maxRecursionDepth)
+		}
+		opts.maxRecursionDepth = maxRecursionDepth
+		return nil
+	}
+}
+
+// ErrorRecoveryLimit is an option which limits the number of attempts the parser will perform
+// to recover from an error.
+func ErrorRecoveryLimit(errorRecoveryLimit int) Option {
+	return func(opts *options) error {
+		if errorRecoveryLimit < -1 {
+			return fmt.Errorf("error recovery limit must be greater than or equal to -1: %d", errorRecoveryLimit)
+		}
+		opts.errorRecoveryLimit = errorRecoveryLimit
+		return nil
+	}
+}
+
+// Macros adds the given macros to the parser.
+func Macros(macros ...Macro) Option {
+	return func(opts *options) error {
+		for _, m := range macros {
+			if m != nil {
+				if opts.macros == nil {
+					opts.macros = make(map[string]Macro)
+				}
+				opts.macros[m.MacroKey()] = m
+			}
+		}
+		return nil
+	}
+}
--- a/parser/parser.go
+++ b/parser/parser.go
@@ -33,6 +33,66 @@ import (
 	structpb "google.golang.org/protobuf/types/known/structpb"
 )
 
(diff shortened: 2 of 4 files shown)


parser/options.go
```diff
@@ -0,0 +1,72 @@
...
+	errorRecoveryLimit int
+	macros             map[string]Macro
+}
+
+// Option configures the behavior of the parser.
+type Option func(*options)
+
+// WithMaxRecursionDepth is an option which limits the maximum depth the parser will attempt to
+// parse the expression before giving up.
+func WithMaxRecursionDepth(maxRecursionDepth int) Option {
```
TristonianJones: nit: please drop the 'With` on the options.

parser/options.go
```diff
@@ -0,0 +1,72 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package parser
+
+type options struct {
+	maxRecursionDepth  int
+	errorRecoveryLimit int
+	macros             map[string]Macro
+}
```
TristonianJones, line 23: Consider just using the fields on the `parser` and switching the `Option` to a `ParserOption`

e.g. `type ParserOption func(*parser) (*parser, error)`

Since the `ParserOption` implementations refer to a private type `parser` that means that they must all be curated by us, so I don't think we need a separate `options` object.

parser/options.go
```diff
@@ -0,0 +1,72 @@
...
+					opts.macros = make(map[string]Macro)
+				}
+				opts.macros[m.MacroKey()] = m
+			}
+		}
+	}
+}
+
+func applyOptions(ropts *options, opts ...Option) {
+	for _, opt := range opts {
```
TristonianJones: I would just do this in place in the `newParser` call.

parser/parser.go
```diff
@@ -61,37 +110,107 @@ var reservedIds = map[string]struct{}{
...
+	rl.checkAttempts(recognizer)
+	return rl.DefaultErrorStrategy.RecoverInline(recognizer)
+}
+
+func (rl *recoveryLimitErrorStrategy) checkAttempts(recognizer antlr.Parser) {
+	if rl.attempts >= rl.maxAttempts {
+		rl.attempts++
+		msg := fmt.Sprintf("error recovery attempt limit exceeded: %d", rl.maxAttempts)
+		recognizer.NotifyErrorListeners(msg, nil, nil)
+		panic(&recoveryLimitError{
```
TristonianJones, line 209: I don't think we want to panic if this happens. We just want to abort the parse step.

TristonianJones (changes requested):

parser/options.go
jcking: The problem with parser is that it is not goroutine safe. I created Parser to have a re-usable interface which underneath creates new parser for each invocation, but doesn't need to build the map again.

I created options to keep the type private, instead of doing `func(*Parser)`.

Also ParserOption is repetative given that the package is parser.

parser/options.go
jcking: Done.

parser/options.go
jcking: Done.

parser/parser.go
TristonianJones: I see now that I missed the custom `recover` block below. Resolving.

parser/options.go
```diff
@@ -0,0 +1,54 @@
...
+package parser
+
+type options struct {
+	maxRecursionDepth  int
+	errorRecoveryLimit int
+	macros             map[string]Macro
+}
+
+// Option configures the behavior of the parser.
+type Option func(*options)
```
TristonianJones: Let's make sure this returns `error` and that we check the result when applying the options.

parser/options.go
```diff
@@ -0,0 +1,54 @@
...
+
+// Macros adds the given macros to the parser.
+func Macros(macros ...Macro) Option {
+	return func(opts *options) {
+		for _, m := range macros {
+			if m != nil {
+				if opts.macros == nil {
+					opts.macros = make(map[string]Macro)
+				}
+				opts.macros[m.MacroKey()] = m
```
TristonianJones, line 60: Consider erroring if the `MacroKey` is used more than once.

parser/options.go
TristonianJones: Fair point. I was trying to think about hoisting the options higher up. Ultimately, I'd like to have a `limits` package with all limits documented and configured in a single object. We can do this as a separate PR, but the idea would to address all compute / memory limits in a single place for the sake of getting a clear view of how to constrain CEL in a single place.  What do you think? I had #420 in mind as I wrote. I'm open to suggestions.

TristonianJones (changes requested):

parser/options.go
jcking: I avoided doing so in case a caller wanted to replace a builtin macro with their own, they could do something like `Macros(AllMacros...), Macros(MyHasMacro)`.

parser/options.go
jcking: Done.

parser/options.go
jcking: That seems fine to me.

parser/options.go
TristonianJones: I'm going to hoist the individual macros out of the `AllMacros` set and export them to the `cel` package, so it should be much easier to compose replacements. I can address the error semantics in my PR.

TristonianJones (approved):

PR #433, opened 2021-04-16
Add expression size limit to parser

jcking:


parser/options.go
```diff
@@ -49,6 +50,16 @@ func ErrorRecoveryLimit(errorRecoveryLimit int) Option {
 	}
 }
 
+func ExpressionSizeCodePointLimit(expressionSizeCodePointLimit int) Option {
+	return func(opts *options) error {
+		if expressionSizeCodePointLimit < -1 {
+			return fmt.Errorf("error recovery limit must be greater than or equal to -1: %d", expressionSizeCodePointLimit)
```
TristonianJones: Copy paste bug here, I think.

parser/parser.go
```diff
@@ -85,7 +91,14 @@ func (p *Parser) Parse(source common.Source) (*exprpb.ParsedExpr, *common.Errors
 	if !ok {
 		buf = runes.NewBuffer(source.Content())
 	}
-	e := impl.parse(buf, source.Description())
+	var e *exprpb.Expr
+	if buf.Len() > p.expressionSizeCodePointLimit {
+		e = impl.reportError(common.NoLocation,
+			"expression size exceeds limit: size: %d, limit %d",
```
TristonianJones: Maybe say `codepoint limit`? I think I missed this on the Java review.

parser/parser_test.go
```diff
@@ -1223,3 +1223,17 @@ func TestParse(t *testing.T) {
 		})
 	}
 }
+
+func TestExpressionSizeCodePointLimit(t *testing.T) {
```
TristonianJones, line 1227: Can you also add a test for a negative valued limit?

TristonianJones (changes requested):

parser/parser.go
jcking: Done.

parser/options.go
jcking: Yep, thanks for catching.

parser/parser_test.go
jcking: Done.

parser/parser_test.go
```diff
@@ -1223,3 +1223,21 @@ func TestParse(t *testing.T) {
...
+	p, err := NewParser(Macros(AllMacros...), ExpressionSizeCodePointLimit(-2))
+	if err == nil {
+		t.Fatal("expected NewParser to return an error when given an invalid expression size code point limit")
```
TristonianJones: Can you include the `error` in the test failure?

parser/parser_test.go
```diff
@@ -1223,3 +1223,21 @@ func TestParse(t *testing.T) {
...
+	}
+	p, err = NewParser(Macros(AllMacros...), ExpressionSizeCodePointLimit((2)))
+	if err != nil {
+		t.Fatal(err)
+	}
+	_, errs := p.Parse(common.NewTextSource("foo"))
+	if got, want := len(errs.GetErrors()), 1; got != want {
+		t.Fatalf("expected parsing to result in exactly %d errors, got %d errors", want, got)
```
TristonianJones: Usually, you want to phrase the error messages as `got X, wanted Y` rather than the other way around. Here and below.

TristonianJones (approved):
If you could just address the cosmetic changes in the test error messages, the rest looks ready to submit.

parser/parser_test.go
jcking: It's nil, as were are testing that an error did occur. Regardless changed the message.

parser/parser_test.go
jcking: Done.

PR #436, opened 2021-04-20
Update the `IsChecked` call to handle constant expressions.

TristonianJones:
Parsing and checking `true` will result in an empty reference map. The `cel.IsChecked()` logic was testing whether the reference map and type-map were non-nil, but the only requirement is technically that the type map contain at least one entry in order for an expression to have been checked.


TristonianJones:
@jcking PTAL

jcking (approved):

PR #437, opened 2021-05-19
Update text in errors_test.go to meaningless language.

chaopeng:
The updated text is meaningless in Chinese


chaopeng:
@TristonianJones

TristonianJones (approved):

PR #443, opened 2021-08-06
Change recursion checks to be rule based, better match the spec

TristonianJones:
The CEL spec indicates that the max recursion limit for any one grammar element
is effectively 32. At present, the current recursion limiting strategy employed by
the parser only checks for recursion at the top-level `expr` node, but recursion can
also occur within subrules like `calc`, `member`, etc. The new strategy enforces
the configured limit on a per-rule basis, rather than just for `expr` to account for
recursion in subrules.


parser/parser.go
```diff
@@ -158,30 +158,41 @@ func (re *recursionError) Error() string {
...
-			rl.depth++
-			panic(&recursionError{
-				message: fmt.Sprintf("expression recursion limit exceeded: %d", rl.maxDepth),
-			})
-		}
-		rl.depth++
+	if ctx == nil {
+		return
+	}
+	ruleIndex := ctx.GetRuleIndex()
```
jcking, line 173: Speed-wise, it might make more sense to use map[int]*int. That way you do not need to do do both a lookup and insert into a map each time.

jcking, line 173: I *think* this is safe. I remember having to put the increment after the recursion check. But looking at ANTLR Go, it appears that if EnterEveryRule panics, ExitEveryRule should not be called. Might want to double-check, otherwise things could go negative.

TristonianJones: Done. Added a defensive check since I'm using a pointer now as well.

parser/parser.go
```diff
@@ -158,30 +158,41 @@ func (re *recursionError) Error() string {
...
 }
 
 func (rl *recursionListener) ExitEveryRule(ctx antlr.ParserRuleContext) {
-	if ctx != nil && ctx.GetRuleIndex() == gen.CELParserRULE_expr {
-		rl.depth--
+	if ctx == nil {
+		return
 	}
+	ruleIndex := ctx.GetRuleIndex()
+	depth := rl.ruleTypeDepth[ruleIndex]
```
TristonianJones: I added a guard to make sure that the rule counter is found and > 0 before decrementing.

parser/parser.go
TristonianJones: Done

TristonianJones (commented):
Also, note that I added a check to validate that error recovery limits were being tested properly as well. I made a minor code change to prevent double-reporting of error recovery issues as a result of the test.

TristonianJones:
PTAL

jcking (approved):

PR #446, opened 2021-08-12
Helper DeclareContextProto to extend CEL environment

hickford:
Helper DeclareContextProto to extend CEL environment with declarations from the given context proto. Each field of the proto defines a variable of the same name in the environment (https://github.com/google/cel-spec/blob/master/doc/langdef.md#evaluation-environment).

TESTED=bazel test ...

### DIFF
--- a/cel/cel_test.go
+++ b/cel/cel_test.go
@@ -1263,3 +1263,19 @@ func TestResidualAst_Modified(t *testing.T) {
 		}
 	}
 }
+
+func TestDeclareContextProto(t *testing.T) {
+	descriptor := new(proto3pb.TestAllTypes).ProtoReflect().Descriptor()
+	option := DeclareContextProto(descriptor)
+	env, err := NewEnv(option)
+	if err != nil {
+		t.Fatalf("NewEnv(DeclareContextProto(%v)) failed: %s", descriptor, err)
+	}
+	expression := `single_int64 == 1 && single_double == 1.0 && single_bool == true && single_string == '' && single_nested_message == google.expr.proto3.test.TestAllTypes.NestedMessage{}
+	&& single_nested_enum == google.expr.proto3.test.TestAllTypes.NestedEnum.FOO && single_duration == duration('5s') && single_timestamp == timestamp('1972-01-01T10:00:20.021-05:00')
+	&& single_any == google.protobuf.Any{} && repeated_int32 == [1,2] && map_string_string == {'': ''} && map_int64_nested_type == {0 : google.expr.proto3.test.NestedTestAllTypes{}}`
+	_, iss := env.Compile(expression)
+	if iss.Err() != nil {
+		t.Fatalf("env.Compile(%s) failed: %s", expression, iss.Err())
+	}
+}
--- a/cel/options.go
+++ b/cel/options.go
@@ -17,6 +17,7 @@ package cel
 import (
 	"fmt"
 
+	"github.com/google/cel-go/checker/decls"
 	"github.com/google/cel-go/common/containers"
 	"github.com/google/cel-go/common/types/pb"
 	"github.com/google/cel-go/common/types/ref"
@@ -27,6 +28,7 @@ import (
 	"google.golang.org/protobuf/reflect/protodesc"
 	"google.golang.org/protobuf/reflect/protoreflect"
 	"google.golang.org/protobuf/reflect/protoregistry"
+	"google.golang.org/protobuf/types/dynamicpb"
 
 	exprpb "google.golang.org/genproto/googleapis/api/expr/v1alpha1"
 	descpb "google.golang.org/protobuf/types/descriptorpb"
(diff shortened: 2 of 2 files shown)


hickford:
@TristonianJones Here's the review discussed elsewhere

cel/options.go
```diff
@@ -86,7 +88,7 @@ func CustomTypeProvider(provider ref.TypeProvider) EnvOption {
 // for the environment. The NewEnv call builds on top of the standard CEL declarations. For a
 // purely custom set of declarations use NewCustomEnv.
 func Declarations(decls ...*exprpb.Decl) EnvOption {
-	// TODO: provide an alternative means of specifying declarations that doesn't refer
+	// TODO(b/192347333): provide an alternative means of specifying declarations that doesn't refer
```
TristonianJones: If you wouldn't mind, please remove the internal bug reference.

cel/options.go
```diff
@@ -376,3 +378,83 @@ func EvalOptions(opts ...EvalOption) ProgramOption {
 		return p, nil
 	}
 }
+
+func fieldToCELType(field protoreflect.FieldDescriptor) (*exprpb.Type, error) {
+	// https://github.com/google/cel-spec/blob/master/doc/langdef.md#protocol-buffer-data-conversion
+	switch (field).Kind() {
```
TristonianJones: Consider using `common/types/pb/checked.go` as the source of truth for type conversions and just group this into two chunks:

```go
if field.Kind() == protoreflect.MessageKind {
  msgName := field.Message().FullName()
  if type, found := pb.CheckedWellKnowns[msgName]; found {
    return type, nil
  }
  return decls.NewObjectType(msgName), nil
}
if type, found := pb.CheckedPrimitives(field.Kind()); found {
  return type, nil
}
return nil, <err>
```

cel/options.go
```diff
@@ -376,3 +378,83 @@ func EvalOptions(opts ...EvalOption) ProgramOption {
...
+	} else {
+		celType, err := fieldToCELType(field)
+		if err != nil {
+			return nil, err
+		}
+		return decls.NewVar(name, celType), nil
+	}
+}
+
+// DeclareContextProto returns an option to extend CEL environment with declarations from the given context proto (https://github.com/google/cel-spec/blob/master/doc/langdef.md#evaluation-environment). Each field of the proto defines a variable of the same name in the environment.
```
TristonianJones: Can you please break the comment up? The first line should probably end at 'given context proto' with the link being on the last line as a further reference.

cel/cel_test.go
```diff
@@ -1263,3 +1266,48 @@ func TestResidualAst_Modified(t *testing.T) {
...
+		decls.NewVar("single_any", decls.Any),
+		decls.NewVar("repeated_int32", decls.NewListType(decls.Int)),
+		decls.NewVar("map_int64_string", decls.NewMapType(decls.Int, decls.String)),
+	}
+
+	want, err := NewEnv(Declarations(wantDeclarations...))
+	if err != nil {
+		t.Fatalf("test setup failed: %s", err)
+	}
+	cmp.AllowUnexported()
```
TristonianJones: Maybe instead of testing the internal object structure of the `Env`, you could instead type-check and evaluate an expression which contains type tests these variables:

```cel
type(single_int64) == int && type(single_double) == double && ...
```

Note that `list` and `map` types are not parameterized at runtime and so `type(map_int64_string) == map` is as much information as the type-checker would validate.

Alternatively, you could attempt to test an expression which performs equality and then you only need to use the type-checker to validate behavior:

```cel
single_int64 == 1 && map_int64_string == {42: "answer to everything"} && ...
```

cel/options.go
hickford: Thanks, that worked (with an extra case for enum)

cel/cel_test.go
hickford: Simplified with second suggestion.

hickford:
Thanks for your comments. Ready for review again.

cel/cel_test.go
```diff
@@ -1263,3 +1264,21 @@ func TestResidualAst_Modified(t *testing.T) {
...
 }
+
+func TestDeclareContextProto(t *testing.T) {
+	descriptor := new(tpvpb.TestProtoValue).ProtoReflect().Descriptor()
+	option := DeclareContextProto(descriptor)
+	env, err := NewEnv(option)
+	if err != nil {
+		t.Fatalf("NewEnv(DeclareContextProto(%v)) failed: %s", descriptor, err)
+	}
+	expression := "single_int64 == 1 && single_double == 1.0 && single_bool == true && single_string == '' && single_nested_message == cel.testdata.TestProtoValue.NestedMessage{} && single_nested_enum == cel.testdata.TestProtoValue.NestedEnum.FOO && single_duration == duration('5s') && single_timestamp == timestamp('1972-01-01T10:00:20.021-05:00') && single_any == google.protobuf.Any{} && repeated_int32 == [1,2] && map_int64_string == {0: ''}"
```
TristonianJones: Could you please break up the expression onto multiple lines? I would recommend using the backtick ` for multiline string quoting

cel/cel_test.go
```diff
@@ -1263,3 +1264,21 @@ func TestResidualAst_Modified(t *testing.T) {
...
+	ast, iss := env.Parse(expression)
```
TristonianJones: Consider replacing the `Parse` and `Check` with `Compile`

cel/options.go
```diff
@@ -376,3 +378,71 @@ func EvalOptions(opts ...EvalOption) ProgramOption {
...
+		}
+		return decls.NewObjectType(msgName), nil
+	}
+	if primitiveType, found := pb.CheckedPrimitives[field.Kind()]; found {
+		return primitiveType, nil
+	}
+	if field.Kind() == protoreflect.EnumKind {
+		return decls.Int, nil
+	}
+	return nil, fmt.Errorf("field %s type %s not implemented", (field).FullName(), (field).Kind().String())
```
TristonianJones: nit: I don't think you need the parens around `(field)`, should be fine to leave them off.

cel/options.go
```diff
@@ -376,3 +378,71 @@ func EvalOptions(opts ...EvalOption) ProgramOption {
...
+}
+
+func fieldToDecl(field protoreflect.FieldDescriptor) (*exprpb.Decl, error) {
+	name := string(field.Name())
+	if (field).IsMap() {
```
TristonianJones: nit: Here and below please replace `(field)` with `field`

cel/options.go
```diff
@@ -376,3 +378,71 @@ func EvalOptions(opts ...EvalOption) ProgramOption {
...
+	}
+}
+
+// DeclareContextProto returns an option to extend CEL environment with declarations from the given context proto.
+// Each field of the proto defines a variable of the same name in the environment.
+// https://github.com/google/cel-spec/blob/master/doc/langdef.md#evaluation-environment
+func DeclareContextProto(descriptor protoreflect.MessageDescriptor) EnvOption {
+	return func(e *Env) (*Env, error) {
+		var decls []*exprpb.Decl
+		fields := (descriptor).Fields()
```
TristonianJones: nit: please remove the extra parens

cel/options.go
```diff
@@ -376,3 +378,71 @@ func EvalOptions(opts ...EvalOption) ProgramOption {
...
+		for i := 0; i < fields.Len(); i++ {
+			field := fields.Get(i)
+			decl, err := fieldToDecl(field)
+			if err != nil {
+				return nil, err
+			}
+			decls = append(decls, decl)
+		}
+		declarations := Declarations(decls...)
```
TristonianJones: Since you're not actually extending the environment, consider doing the following:
```go
var err error
e, err = Declarations(decls...)(e)
if (err) {
  return nil, err
}
return Types(...)(e)
```

cel/testdata/testprotovalue.proto
```diff
@@ -0,0 +1,32 @@
+syntax = "proto2";
```
TristonianJones: Instead of introducing a new protobuf, consider re-using `test/proto2pb/test_all_types.pb.go`. Doing so will change your test expression slightly, but hopefully not significantly..

cel/cel_test.go
hickford: Done.

cel/options.go
hickford: Done.

cel/options.go
hickford: Done.

cel/options.go
hickford: Done.

cel/testdata/testprotovalue.proto
hickford: Done.

cel/cel_test.go
hickford: Done.

cel/options.go
hickford: Done.

hickford:
Thanks again, ready.

TristonianJones (approved):

PR #448, opened 2021-09-02
Unify overflow checking in operators, field setting

TristonianJones:
This PR consolidates additional overflow checks into `overflow.go` and ensures they are applied consistently when setting object field values.

There are some extensive test refactors and an update to the latest version of cel-spec v0.6.0 as well to ensure that the new code paths are being exercised and that the changes adhere to the spec. As part of this change, a small set of conformance tests now check for value truncation (rather than rounding) when converting from floating point to integer values. Given that the fixes for the incoming spec behavior were small, the truncation change was also made as part of this PR.


common/types/overflow.go
```diff
@@ -222,57 +259,99 @@ func subtractTimeChecked(x, y time.Time) (time.Duration, bool) {
...
-		return time.Duration(0), false
+	sec, err := subtractInt64Checked(sec1, sec2)
+	if err != nil {
+		return time.Duration(0), err
 	}
 
 	// Nanoseconds cannot overflow as time.Time normalizes them to [0, 999999999].
 	nsec := nsec1 - nsec2
 
-	// We need to normalize nanoseconds to be positive and carry extra nanoseconds to seconds.
```
TristonianJones, line 233: Note, the nanos are properly normalized in this case, so the logic ported from `time.Unix` was redundant and removed.

common/types/double.go
```diff
@@ -127,17 +126,17 @@ func (d Double) ConvertToNative(typeDesc reflect.Type) (interface{}, error) {
 func (d Double) ConvertToType(typeVal ref.Type) ref.Val {
 	switch typeVal {
 	case IntType:
-		i := math.Round(float64(d))
```
TristonianJones, line 130: The removal of the rounding behavior on conversion is expected and part of a cel-spec update.

common/types/overflow.go
```diff
@@ -222,57 +259,99 @@ func subtractTimeChecked(x, y time.Time) (time.Duration, bool) {
...
+	if err != nil {
+		return time.Time{}, err
 	}
 	return addTimeDurationChecked(x, val)
 }
+
+// doubleToInt64Checked converts a double to an int64 value.
+//
+// If the conversion fails due to overflow the error return value will be non-nil.
+func doubleToInt64Checked(v float64) (int64, error) {
```
jcking, line 302: What about NaN?

common/types/overflow.go
```diff
@@ -222,57 +259,99 @@ func subtractTimeChecked(x, y time.Time) (time.Duration, bool) {
...
+	if math.IsInf(v, 0) || v <= float64(math.MinInt64) || v >= float64(math.MaxInt64) {
+		return 0, errIntOverflow
+	}
+	return int64(v), nil
+}
+
+// doubleToInt64Checked converts a double to a uint64 value.
+//
+// If the conversion fails due to overflow the error return value will be non-nil.
+func doubleToUint64Checked(v float64) (uint64, error) {
```
jcking, line 312: NaN as well.

common/types/overflow.go
```diff
@@ -222,57 +259,99 @@ func subtractTimeChecked(x, y time.Time) (time.Duration, bool) {
...
+	if math.IsInf(v, 0) || v < 0 || v >= doubleTwoTo64 {
```
jcking: Couldn't you just use math.MaxUint64? math.Ldexp seems less readable.

common/types/overflow.go
```diff
@@ -222,57 +259,99 @@ func subtractTimeChecked(x, y time.Time) (time.Duration, bool) {
...
+		return 0, errIntOverflow
+	}
+	return int32(v), nil
+}
+
+// uint64toUint32Checked converts a uint64 to a uint32 value.
+//
+// If the conversion fails due to overflow the error return value will be non-nil.
+func uint64ToUint32Checked(v uint64) (uint32, error) {
+	if v < 0 || v > math.MaxUint32 {
```
jcking: v is unsigned, so it is never less than 0.

common/types/overflow.go
TristonianJones: I've kept the same call pattern as is used in C++ for consistency (it uses `std::ldexp`)

common/types/overflow.go
TristonianJones: Yeah, not sure why I felt the need to add that. Removed.

common/types/overflow.go
TristonianJones: Done.

common/types/overflow.go
TristonianJones: Ah, right. The C++ libraries define `IsFinite` which accounts for NaN. Have to check separately here.

jcking (approved):

PR #451, opened 2021-09-15
Allow building Ast with textual contents

damz:
Unless I am missing something, it is currently not possible to build an `Ast` from an `exprpb.ParsedExpr` or `exprpb.CheckedExpr` with source contents. As a consequence, if an `Ast` built from `ParsedExprToAst` or `CheckedExprToAst` is checked or executed, the error messages will not show code snippets.

Introduce a `CheckedExprToAstWithSource` and `ParsedExprToAstWithSource` to allow providing the source contents.

### DIFF
--- a/cel/io.go
+++ b/cel/io.go
@@ -25,6 +25,18 @@ import (
 
 // CheckedExprToAst converts a checked expression proto message to an Ast.
 func CheckedExprToAst(checkedExpr *exprpb.CheckedExpr) *Ast {
+	return CheckedExprToAstWithSource(checkedExpr, nil)
+}
+
+// CheckedExprToAstWithSource converts a checked expression proto message to an Ast,
+// using the provided Source as the textual contents.
+//
+// In general the source is not necessary unless the AST has been modified between the
+// `Parse` and `Check` calls as an `Ast` created from the `Parse` step will carry the source
+// through future calls.
+//
+// Prefer CheckedExprToAst if loading expressions from storage.
+func CheckedExprToAstWithSource(checkedExpr *exprpb.CheckedExpr, src common.Source) *Ast {
 	refMap := checkedExpr.GetReferenceMap()
 	if refMap == nil {
 		refMap = map[int64]*exprpb.Reference{}
@@ -33,10 +45,17 @@ func CheckedExprToAst(checkedExpr *exprpb.CheckedExpr) *Ast {
 	if typeMap == nil {
 		typeMap = map[int64]*exprpb.Type{}
 	}
+	si := checkedExpr.GetSourceInfo()
+	if si == nil {
+		si = &exprpb.SourceInfo{}
+	}
+	if src == nil {
+		src = common.NewInfoSource(si)
+	}
 	return &Ast{
 		expr:    checkedExpr.GetExpr(),
-		info:    checkedExpr.GetSourceInfo(),
-		source:  common.NewInfoSource(checkedExpr.GetSourceInfo()),
+		info:    si,
+		source:  src,
 		refMap:  refMap,
 		typeMap: typeMap,
 	}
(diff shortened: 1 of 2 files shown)


TristonianJones:
@damz In general, the use case for converting to an `exprpb.ParsedExpr` or `exprpb.CheckedExpr` is storage since the parse / type-check is significantly more expensive than the evaluation. Is there a use case that you can share where you would need to re-check an expression? The code is sound, but I'm not sure I understand the benefit of adding the methods at the moment.

damz:
@TristonianJones We would like to parse an expression, and then rewrite it and split it into small parts that are checked and evaluated independently. Our use case is around building alerting conditions where some of the expression is actually a sub-aggregator.

An expression like this:

```
rate(a+b)+rate(c) > 1
```

Would be rewritten as:

```
@rate1+@rate2 > 1
```

And two aggregators:

```
@rate1 = rate(a+b)
@rate2 = rate(c)
```

This requires us to be able to check and evaluate the aggregator expressions separately, which works perfectly now.... except that we lose the source textual representation so error messages are less useful than they could be.

TristonianJones:
Thanks for the added context. I see the motivation for adding the method. Would you paraphrasing this explanation into the `CheckedExprToAstWithSource` with a recommendation to use `CheckedExprToAst` unless you're doing some additional analysis or massaging of the `ParsedExpr`?

TristonianJones (approved):

damz:
@TristonianJones I would welcome suggestions on the wording, as I am not convinced there are more use cases for one than for the other. Why would you want to load a pre-parsed `ParsedExpr` / `CheckedParsedExpr` from storage, and lose the code snippets at runtime?

In the meantime, I just added a "Use X if you do not have a source." comment to the two new functions.

TristonianJones:
In general, once you've type-checked an expression, you've removed almost all possibility of an error at runtime. There are some edge cases with dynamic types, but they are the exception rather than the rule.

If you have a checked expression and source stored separately, you just need the expression id to find the source offset offline. For most use cases carrying the original source doesn't have much value at runtime.

damz:
Do you think the comment I added is sufficient, or do you have something else in mind?

cel/io.go
```diff
@@ -25,6 +25,14 @@ import (
 
 // CheckedExprToAst converts a checked expression proto message to an Ast.
 func CheckedExprToAst(checkedExpr *exprpb.CheckedExpr) *Ast {
+	return CheckedExprToAstWithSource(checkedExpr, nil)
+}
+
+// CheckedExprToAstWithSource converts a checked expression proto message to an Ast,
+// using the provided Source as the textual contents.
+//
+// Use CheckedExprToAst if you do not have a source.
```
TristonianJones: Thanks for waiting. I had to think a bit about how this should be phrased. I would prefer the comment read as follows:

```
CheckedExprToAstWithSource converts a checked expression proto message to an Ast,
using the provided Source as the textual contents. 

In general the source is not necessary unless the AST has been modified between the
`Parse` and `Check` calls as an `Ast` created from the `Parse` step will carry the source
through future calls.

Prefer CheckedExprToAst if loading expressions from storage.
```

cel/io.go
```diff
@@ -59,14 +74,25 @@ func AstToCheckedExpr(a *Ast) (*exprpb.CheckedExpr, error) {
 
 // ParsedExprToAst converts a parsed expression proto message to an Ast.
 func ParsedExprToAst(parsedExpr *exprpb.ParsedExpr) *Ast {
+	return ParsedExprToAstWithSource(parsedExpr, nil)
+}
+
+// ParsedExprToAstWithSource converts a parsed expression proto message to an Ast,
```
TristonianJones, line 84: Please rephrase in a similar manner to the `CheckedExprToAstWithSource`. The use case on this method is actually really difficult for me to understand, but I understand the desire for functional parity.

Perhaps recommend only using this function if you need to re-check a previously type-checked expression?

TristonianJones:
@damz If you wouldn't mind updating the comments, I'm fine accepting the change. Let me know!

damz:
@TristonianJones Thanks for your guidance. I updated the PR.

My apologies for the delay, I was out of the office most of this week.

TristonianJones:
No problem @damz, just wanted you to know I hadn't forgotten about this. :)

PR #455, opened 2021-09-30
Populate macro calls

ZachSierp:
Populate the macroCalls map in sourceInfo with macro expressions while parsing. For macro expressions nested in other macros, the map stores the argument ID of the nested macro for self reference instead of the full expression.

Storing the ID instead of the full argument reduces expression duplication in macroCalls and reduces overall size.

While iterating the AST, expression IDs can be looked up in macroCalls and continue from there. Nested macros can also be looked up by ID while iterating a macro in macroCalls.

This change covers cases where a macro is the target so some other macro such as `has(a.b).filter(x, x > 0)`, and cases where macros are nested inside the arguments of subexpressions like `x.filter(y, y.exists(z, has(z.a)) && y.exists(z, has(z.b)))`

Fixes issue #292


parser/helper.go
```diff
@@ -207,6 +208,58 @@ func (p *parserHelper) getLocation(id int64) common.Location {
...
+		callExpr := &exprpb.Expr_CallExpr{
+			CallExpr: &exprpb.Expr_Call{
+				Function: expr.GetCallExpr().Function,
+			},
+		}
+		// Iterate the AST from `expr` recursively looking for macros. Because we are at most
+		// starting from the top level macro, this recursion is bounded by the size of the AST. This
+		// means that the depth check on the AST during parsing will catch recursion overflows
+		// before we get to here.
+		for _, arg := range expr.GetCallExpr().Args {
```
TristonianJones: Since you know the number of arguments, it's slightly better performance to allocate an exactly sized array and  update the indices rather than append. 

```
make([]*exprpb.Expr, len(expr.GetCallExpr().GetArgs()))
```

parser/helper.go
```diff
@@ -207,6 +208,58 @@ func (p *parserHelper) getLocation(id int64) common.Location {
...
+
+	if _, found := p.source.MacroCalls()[expr.Id]; found {
+		return resultExpr
+	}
+
+	switch expr.ExprKind.(type) {
+	case *exprpb.Expr_CallExpr:
+		callExpr := &exprpb.Expr_CallExpr{
+			CallExpr: &exprpb.Expr_Call{
+				Function: expr.GetCallExpr().Function,
```
TristonianJones: Make sure you're using the accessor functions rather than the fields directly: `GetFunction()` vs. `Function`

parser/helper.go
```diff
@@ -207,6 +208,58 @@ func (p *parserHelper) getLocation(id int64) common.Location {
...
+	if target != nil {
+		if _, found := p.source.MacroCalls()[target.Id]; found {
+			callExpr.CallExpr.Target = &exprpb.Expr{Id: target.Id}
+		} else {
+			callExpr.CallExpr.Target = target
+		}
+	}
+
+	for _, arg := range args {
+		callExpr.CallExpr.Args = append(callExpr.CallExpr.Args, p.buildMacroCallArg(arg))
```
TristonianJones: Prefer `callExpr.GetCallExpr().GetArgs()`

parser/options.go
```diff
@@ -92,3 +93,10 @@ func Macros(macros ...Macro) Option {
 		return nil
 	}
 }
+
```
TristonianJones, line 96: Make sure you add a comment of the following form for all exported methods: `// <CallName> <verb> <explanation>`

```
// PopulateMacroCalls ensures that the original call signatures replaced by expanded macros
// are preserved in the `SourceInfo` of parse result.

File: parser/parser_test.go [hunk 5]
```diff
@@ -1257,19 +1479,28 @@ type testInfo struct {
...
 type kindAndIDAdorner struct {
+	sourceInfo *exprpb.SourceInfo
 }
 
 func (k *kindAndIDAdorner) GetMetadata(elem interface{}) string {
 	switch elem.(type) {
 	case *exprpb.Expr:
 		e := elem.(*exprpb.Expr)
+		if k.sourceInfo != nil {
+			if val, found := k.sourceInfo.MacroCalls[e.Id]; found {
```
TristonianJones: `e.GetId()`

parser/parser_test.go
```diff
@@ -1257,19 +1479,28 @@ type testInfo struct {
...
+				return fmt.Sprintf("^#%d:%s#", e.Id, val.GetCallExpr().Function)
```
TristonianJones: `e.GetId()` and `GetFunction()`

parser/parser_test.go
```diff
@@ -1365,6 +1615,13 @@ func TestParse(t *testing.T) {
 					tt.Fatal(test.DiffMessage("location", actualWithLocation, tc.L))
 				}
 			}
+
+			if tc.M != "" {
+				actualAdornedMacroCalls := convertMacroCallsToString(expression.SourceInfo)
```
TristonianJones: If you're wondering why some of this code uses direct field access and some of it uses getters, it's because there was a change in the preferred style with last year's protobuf API upgrade. If you wouldn't mind fixing the other instances of direct field access (here and elsewhere). That would be appreciated

parser/parser_test.go
```diff
@@ -1365,6 +1615,13 @@ func TestParse(t *testing.T) {
...
+				if !test.Compare(actualAdornedMacroCalls, tc.M) {
+					tt.Fatal(test.DiffMessage("Macro Calls", actualAdornedMacroCalls, tc.M))
```
TristonianJones: Stylistically, you're supposed to say something like `Method(Args...): got <something>, wanted: <something else>`

The original code didn't follow this pattern, but that's only because I learned of the style preference a bit late. If you wouldn't mind updating the error messages to be of this format, I would appreciate it.

parser/helper.go
ZachSierp: Done.

parser/helper.go
ZachSierp: Done.

parser/helper.go
ZachSierp: Done.

parser/options.go
ZachSierp: Done here and in helper.go.

parser/parser_test.go
ZachSierp: Done.

parser/parser_test.go
ZachSierp: Done.

parser/parser_test.go
ZachSierp: Done. I checked other files as well but I might have missed something.

parser/parser_test.go
ZachSierp: Updated the DiffMessage function and where it is called in parser_test.go to follow this pattern.

TristonianJones (approved):

PR #464, opened 2021-10-19
Remove ineffective else clauses

TristonianJones:
Minor style update to reflect internal patches.


TristonianJones:
This reflects the internal code state which has already been reviewed.

PR #473, opened 2021-12-09
Numeric in/equality and comparisons across numeric types

TristonianJones:


common/types/compare.go
```diff
@@ -0,0 +1,95 @@
...
+	}
+	return compareDouble(d, Double(u))
+}
+
+func compareUintDouble(u Uint, d Double) Int {
+	return -compareDoubleUint(d, u)
+}
+
+func compareIntUint(i Int, u Uint) Int {
+	if i < 0 || u > math.MaxInt64 {
```
jcking, line 50: Doesn't this cause `compareIntUint(math.MaxInt64, math.MaxUint64)` to be -1? Is that intentional?

common/types/double.go
```diff
@@ -58,17 +59,19 @@ func (d Double) Add(other ref.Val) ref.Val {
...
-	otherDouble, ok := other.(Double)
-	if !ok {
+	switch ov := other.(type) {
+	case Double:
+		if math.IsNaN(float64(d)) || math.IsNaN(float64(ov)) {
+			return NewErr("NaN values cannot be ordered")
+		}
+		return compareDouble(d, ov)
+	case Int:
+		return compareDoubleInt(d, ov)
```
jcking, line 72: Shouldn't we check for d being NaN here as well?

common/types/double.go
```diff
@@ -58,17 +59,19 @@ func (d Double) Add(other ref.Val) ref.Val {
...
+	case Uint:
+		return compareDoubleUint(d, ov)
```
jcking, line 74: Same.

jcking (changes requested):

common/types/compare.go
TristonianJones: Yes, that's right. `int64` max is less than `uint64` max.

common/types/double.go
TristonianJones: Good point. Done.

common/types/double.go
TristonianJones: Done.

common/types/compare.go
jcking: Oh...wow. Ignore me my brain was not working. lol

jcking (approved):

PR #475, opened 2021-12-17
Support for accessing protobuf fields by their JSON names

charithe:
This change makes it possible to access a protobuf field by its field name (snake case) or the JSON field name (camel case). For example, given a protobuf definition of `message foo { string some_field = 1}`, the field can be accessed as either `fooInstance.some_field` or `fooInstance.someField`.

This is useful when end users normally deal with JSON representations of protobufs and expect the fields to be camel cased rather than snake cased. I was going to implement this as a custom `TypeProvider` in my own code but it would require duplicating a lot of the code that already exists in `cel-go`. Since it's only a 3-line change, I opted to create a PR directly. Please let me know if you prefer to discuss it further in an issue.

### DIFF
--- a/common/types/pb/type.go
+++ b/common/types/pb/type.go
@@ -47,7 +47,9 @@ func NewTypeDescription(typeName string, desc protoreflect.MessageDescriptor) *T
 	fields := desc.Fields()
 	for i := 0; i < fields.Len(); i++ {
 		f := fields.Get(i)
-		fieldMap[string(f.Name())] = NewFieldDescription(f)
+		fd := NewFieldDescription(f)
+		fieldMap[string(f.Name())] = fd
+		fieldMap[string(f.JSONName())] = fd
 	}
 	return &TypeDescription{
 		typeName:    typeName,
--- a/common/types/pb/type_test.go
+++ b/common/types/pb/type_test.go
@@ -156,23 +156,25 @@ func TestFieldDescriptionGetFrom(t *testing.T) {
 			"null": nil,
 		}),
 	}
-	for field, want := range expected {
-		f, found := td.FieldByName(field)
-		if !found {
-			t.Fatalf("td.FieldByName(%q) not found", field)
-		}
-		got, err := f.GetFrom(msg)
-		if err != nil {
-			t.Fatalf("field.GetFrom() failed: %v", err)
-		}
-		switch g := got.(type) {
-		case proto.Message:
-			if !proto.Equal(g, want.(proto.Message)) {
-				t.Errorf("got field %s value %v, wanted %v", field, g, want)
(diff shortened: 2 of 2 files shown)


TristonianJones:
My apologies for the delay in the review, I had been under the impression that you might wrap the type provider to attempt a lookup by snake case and a lookup by camel case (converted from snake case) to map the fields to the original protobuf types.

```go
struct JsonFieldProvider {
  ref.TypeProvider
}

func (p *JsonFieldProvider) FindFieldType(messageType string, fieldName string) (*ref.FieldType, bool) {
  if f, found := p.TypeProvider.FindFieldType(messageType, fieldName); found {
    return f, true
  }
  return p.TypeProvider.FindFieldType(messageType, camelCaseName(fieldName))
}
```

If you still want to pursue the option in the PR, I'd recommend making it possible to support exposing JSON field names _instead_ of snake case names. I think it's pretty rare to use both, and often it's one or the other. Thoughts?

charithe:
Thanks for taking a look at the PR. 

The expression would contain the camel case name and we will have to convert that to snake case because the type provider is only aware of the snake case protobuf field names. My thinking was that it would be better to use the protobuf descriptor to obtain the `JSONName` rather than assuming that it will always be camel case. 

In our particular case, we want users to be able to use either snake or camel case interchangeably. Hence why I didn't consider adding an option to choose one.

If you think this is too much of a special case to be supported by the core library, I'll close this PR.

TristonianJones:
It's more that I can't auto-enable this support for everyone, and generally I see this as an either-or scenario when it comes up internally. Supporting both style names seems error-prone. When running CEL across runtimes, the expectation would be that you're always using the snake_case format; however, if you're only executing type-checked within the cel-go environment it's safe to support alternative field names. 

How about this. If you make it an option to use either the JSON names or the snake_case names when you instantiate the registry, then you could, for your use case, combine the two registries:

```go
struct FlexibleTypeRegistry {
   pb ref.TypeProvider
   json ref.TypeProvider
}

func NewFlexibleTypeRegistry {
  return &FlexibleTypeRegistry{
    pb: types.NewRegistry(),
    json: types.NewJsonRegistry(),
  }
}

func (ft *FlexibleTypeRegistry) FindFieldType(...) {
   if f, found := pb.FindFieldType(...); found {
      return f, found
   }
   return json.FindFieldType(...)
}

// remaining ref.TypeProvider methods which proxy to pb and/or json registries. 
```

charithe:
Thank you for the suggestions. I see that this is probably going to needlessly complicate the library usage for other users -- most of whom probably don't need to care about it. For those who do, it's fairly easy to construct a custom type provider as you have demonstrated above so I'll close this PR.

PR #481, opened 2022-02-10
Updates to cel-go proto equality to mirror C++ MessageDifferencer 

TristonianJones:
This PR introduces a few small changes to how proto equality functions with CEL to ensure
that the same definition of equality is used consistently between cel-go, cel-cpp, and cel-java.

Additionally, the changes for proto equality have already been enabled in the latest version of the
cel-spec which can be observed from the WORKSPACE update.

Note, the `LICENSE` has been updated to indicate that the `proto.Equal` implementation was
derived within the `pb.go` file. This should have no material impact on how CEL is licensed though.

### DIFF
--- a/common/types/pb/equal.go
+++ b/common/types/pb/equal.go
@@ -1,0 +1,207 @@
+// Copyright 2022 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package pb
+
+import (
+	"bytes"
+	"math"
+	"reflect"
+
+	"google.golang.org/protobuf/encoding/protowire"
+	"google.golang.org/protobuf/proto"
+	"google.golang.org/protobuf/reflect/protoreflect"
+	"google.golang.org/protobuf/types/known/anypb"
+)
+
+// Equal returns whether two proto.Message instances are equal using the following criteria:
+//
+//   - Messages must share the same instance of the type descriptor
+//   - Known set fields are compared using semantics equality
+//     - Bytes are compared using bytes.Equal
+//     - Scalar values are compared with operator ==
+//     - List and map types are equal if they have the same length and all elements are equal
+//     - Messages are equal if they share the same descriptor and all set fields are equal
+//   - Unknown fields are compared using byte equality
+//   - NaN values are not equal to each other
+//   - google.protobuf.Any values are unpacked before comparison
+//   - If the type descriptor for a protobuf.Any cannot be found, byte equality is used rather than
+//     semantic equality.
+//
+// This method of proto equality mirrors the behavior of the C++ protobuf MessageDifferencer
+// whereas the golang proto.Equal implementation mirrors the Java protobuf equals() methods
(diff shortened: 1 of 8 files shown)


common/types/pb/file_test.go
```diff
@@ -1,3 +1,17 @@
+// Copyright 2018 Google LLC
```
jcking: 2022?

common/types/pb/equal.go
```diff
@@ -0,0 +1,207 @@
...
+		x, err := ax.UnmarshalNew()
+		if err != nil {
+			return false
+		}
+		mx = x.ProtoReflect()
+		y, err := ay.UnmarshalNew()
+		if err != nil {
+			return false
+		}
+		my = y.ProtoReflect()
```
jcking: What happens if this is google.protobuf.Any nested in google.protobuf.Any?

common/types/pb/equal.go
```diff
@@ -0,0 +1,207 @@
...
+	if x == nil || y == nil {
+		return x == nil && y == nil
+	}
+	xRef := x.ProtoReflect()
+	yRef := y.ProtoReflect()
+	return equalMessage(xRef, yRef)
+}
+
+func equalMessage(mx, my protoreflect.Message) bool {
+	if mx.IsValid() != my.IsValid() || mx.Descriptor() != my.Descriptor() {
```
jcking, line 58: Is it possible for them to both be invalid and have the same descriptor? What happens if they do? Not sure what the state of protoreflect.Message actually is.

common/types/pb/equal.go
```diff
@@ -0,0 +1,207 @@
...
+		protoreflect.Int64Kind, protoreflect.Sint64Kind,
+		protoreflect.Sfixed32Kind, protoreflect.Sfixed64Kind:
+		return x.Int() == y.Int()
+	case protoreflect.Uint32Kind, protoreflect.Uint64Kind,
+		protoreflect.Fixed32Kind, protoreflect.Fixed64Kind:
+		return x.Uint() == y.Uint()
+	case protoreflect.FloatKind, protoreflect.DoubleKind:
+		fx := x.Float()
+		fy := y.Float()
+		if math.IsNaN(fx) || math.IsNaN(fy) {
```
jcking: Go should be respecting IEEE 754 which says `NaN != NaN`, so shouldn't need to bother checking for NaN.

jcking (changes requested):

common/types/pb/equal.go
TristonianJones: I think you're not supposed to do that, but I understand it could happen.  Looks like I should recurse here for consistency. Currently, it will fallback to byte comparison.

common/types/pb/equal.go
TristonianJones: If the message isn't valid it means that it wasn't in a proper wire format. If the descriptors are found, but not the same instance, then they're not considered equal. I can document this if you want.

TristonianJones: My guess is that if both are invalid, they should fall through to byte equality in the unknown fields check.

common/types/pb/file_test.go
TristonianJones: This file was created in 2018, but didn't have the copyright. Should I just do 2022 anyway?

TristonianJones: Updated.

common/types/pb/equal.go
TristonianJones: Removed superfluous check.

common/types/pb/equal.go
TristonianJones: Updated. PTAL

common/types/pb/equal.go
TristonianJones: I added a comment about the expected behavior here and why the code is as it is.

jcking (approved):

PR #482, opened 2022-02-10
Heterogeneous equality

TristonianJones:
Enable heterogeneous equality within the runtime. 

The type-checker will raise errors when cross-type comparisons are made within strongly typed
expressions; however, for dynamically typed data such inferences are not possible meaning they
must be handled appropriately at runtime.

Heterogeneous equality is being introduced to ensure that CEL and proto equality are better 
aligned, that CEL's notion of runtime equality is consistent with Zeta SQLs, and to make it easier
to work with dynamically typed data: e.g.`json.string_value != 2`

This change is a semantic change which has not yet been reflected in the cel-spec but will be
soon. The change is generally regarded as safe as it shifts error states to non-error states, but
in very rare circumstances it is possible that an application may rely on this error to be produced.

### DIFF
--- a/interpreter/decorators.go
+++ b/interpreter/decorators.go
@@ -177,7 +177,6 @@ func maybeOptimizeSetMembership(i Interpretable, inlist InterpretableCall) (Inte
 		return NewConstValue(inlist.ID(), types.False), nil
 	}
 	it := list.Iterator()
-	var typ ref.Type
 	valueSet := make(map[ref.Val]ref.Val)
 	for it.HasNext() == types.True {
 		elem := it.Next()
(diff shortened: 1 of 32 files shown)


TristonianJones:
Note, this PR should only be submitted after #481 is merged.

TristonianJones:
PTAL

interpreter/decorators.go
```diff
@@ -177,25 +177,51 @@ func maybeOptimizeSetMembership(i Interpretable, inlist InterpretableCall) (Inte
...
 			// Note, non-primitive type are not yet supported.
 			return i, nil
 		}
-		if typ == nil {
-			typ = elem.Type()
-		} else if typ.TypeName() != elem.Type().TypeName() {
-			return i, nil
-		}
 		valueSet[elem] = types.True
+		switch ev := elem.(type) {
```
jcking, line 188: Hmmmm. Why are we only checking for lossless conversions for types.Double but not the others?

TristonianJones: The others are getting overflow checks as part of the `ConvertToType` call, but the double conversion to an int or uint in this block must only be permitted when the double is int-like; hence, the additional check.

jcking: Discussed offline.

jcking (approved):

PR #483, opened 2022-02-11
EstimateCost API

jpbetz:
Introduce a cost estimator that walks the parsed AST and calls a provided Estimator to estimate the lengths of variable length input data (strings, bytes, lists, maps) and the cost of extension function calls.

### DIFF
--- a/cel/cel_test.go
+++ b/cel/cel_test.go
@@ -1212,7 +1212,7 @@ func (tc testCostEstimator) EstimateCallCost(overloadId string, target *checker.
 	return nil
 }
 
-func TestCost(t *testing.T) {
+func TestEstimateCost(t *testing.T) {
 	allTypes := decls.NewObjectType("google.expr.proto3.test.TestAllTypes")
 	allList := decls.NewListType(allTypes)
 	intList := decls.NewListType(decls.Int)
@@ -1236,9 +1236,9 @@ func TestCost(t *testing.T) {
 		},
 		{
 			name:    "identity",
-			program: `"input"`,
+			program: `input`,
 			decls:   []*exprpb.Decl{decls.NewVar("input", intList)},
-			wanted:  zeroCost,
+			wanted:  checker.CostEstimate{Min: 1, Max: 1},
 		},
 		{
 			name:    "select: map",
@@ -1303,12 +1303,12 @@ func TestCost(t *testing.T) {
 			decls:   []*exprpb.Decl{decls.NewVar("input", str)},
 			hints:   map[*exprpb.Type]int64{str: 500},
 			program: `input.matches('[0-9]')`,
-			wanted:  checker.CostEstimate{Min: 1, Max: 6},
+			wanted:  checker.CostEstimate{Min: 1, Max: 101},
 		},
 		{
 			name:    "variable cost function with constant",
 			program: `'123'.matches('[0-9]')`,
-			wanted:  checker.CostEstimate{Min: 1, Max: 1},
+			wanted:  checker.CostEstimate{Min: 2, Max: 2},
 		},
 		{
 			name:    "or",
@@ -1348,7 +1348,7 @@ func TestCost(t *testing.T) {
 		{
 			name:    "in",
 			program: `2 in [1, 2, 3]`,
-			wanted:  checker.CostEstimate{Min: 11, Max: 11},
+			wanted:  checker.CostEstimate{Min: 13, Max: 13},
 		},
 		{
 			name:    "plus",
@@ -1390,14 +1390,14 @@ func TestCost(t *testing.T) {
 			decls:   []*exprpb.Decl{decls.NewVar("input", bytes)},
 			hints:   map[*exprpb.Type]int64{bytes: 500},
 			program: `string(input)`,
-			wanted:  checker.CostEstimate{Min: 1, Max: 6},
+			wanted:  checker.CostEstimate{Min: 1, Max: 51},
 		},
 		{
 			name:    "string to bytes conversion",
 			decls:   []*exprpb.Decl{decls.NewVar("input", str)},
 			hints:   map[*exprpb.Type]int64{str: 500},
 			program: `bytes(input)`,
-			wanted:  checker.CostEstimate{Min: 1, Max: 6},
+			wanted:  checker.CostEstimate{Min: 1, Max: 51},
 		},
 		{
 			name:    "int to string conversion",
@@ -1412,17 +1412,16 @@ func TestCost(t *testing.T) {
 				decls.NewVar("arg1", str),
 			},
 			hints:  map[*exprpb.Type]int64{str: 500},
-			wanted: checker.CostEstimate{Min: 2, Max: 27},
+			wanted: checker.CostEstimate{Min: 2, Max: 2502},
 		},
 		{
 			name:    "matches",
-			program: `input.matches(arg1)`,
+			program: `input.matches('\\d+a\\d+b')`,
 			decls: []*exprpb.Decl{
 				decls.NewVar("input", str),
-				decls.NewVar("arg1", str),
 			},
 			hints:  map[*exprpb.Type]int64{str: 500},
-			wanted: checker.CostEstimate{Min: 2, Max: 252},
+			wanted: checker.CostEstimate{Min: 1, Max: 101},
 		},
 		{
 			name:    "startsWith",
@@ -1432,7 +1431,7 @@ func TestCost(t *testing.T) {
 				decls.NewVar("arg1", str),
 			},
 			hints:  map[*exprpb.Type]int64{str: 500},
-			wanted: checker.CostEstimate{Min: 2, Max: 7},
+			wanted: checker.CostEstimate{Min: 2, Max: 52},
 		},
 		{
 			name:    "endsWith",
@@ -1442,7 +1441,7 @@ func TestCost(t *testing.T) {
 				decls.NewVar("arg1", str),
 			},
 			hints:  map[*exprpb.Type]int64{str: 500},
-			wanted: checker.CostEstimate{Min: 2, Max: 7},
+			wanted: checker.CostEstimate{Min: 2, Max: 52},
 		},
 		{
 			name:    "size receiver",
@@ -1450,7 +1449,6 @@ func TestCost(t *testing.T) {
 			decls: []*exprpb.Decl{
 				decls.NewVar("input", str),
 			},
-			hints:  map[*exprpb.Type]int64{str: 500},
 			wanted: checker.CostEstimate{Min: 2, Max: 2},
 		},
 		{
@@ -1459,7 +1457,6 @@ func TestCost(t *testing.T) {
 			decls: []*exprpb.Decl{
 				decls.NewVar("input", str),
 			},
-			hints:  map[*exprpb.Type]int64{str: 500},
 			wanted: checker.CostEstimate{Min: 2, Max: 2},
 		},
 	}
--- a/cel/cel_test.go
+++ b/cel/cel_test.go
@@ -1206,217 +1206,10 @@ func (tc testCostEstimator) EstimateSize(element checker.AstNode) *checker.SizeE
 
 func (tc testCostEstimator) EstimateCallCost(overloadId string, target *checker.AstNode, args []checker.AstNode) *checker.CostEstimate {
 	switch overloadId {
-	case "noop":
-		return &checker.CostEstimate{Min: 0, Max: 0}
+	case overloads.TimestampToYear:
+		return &checker.CostEstimate{Min: 7, Max: 7}
 	}
 	return nil
-}
-
-func BenchmarkCost(b *testing.B) {
-	cases := []struct {
-		name    string
-		program string
-		decls   []*exprpb.Decl
-		vars    interface{}
-	}{
-		{
-			name:    "const",
-			program: `"Hello World!"`,
-		},
-		{
-			name:    "identity",
-			program: `"input"`,
-			decls:   []*exprpb.Decl{decls.NewVar("input", decls.Int)},
-		},
-		{
-			name:    "select: map",
-			program: `"input['key']"`,
-			decls:   []*exprpb.Decl{decls.NewVar("input", decls.NewMapType(decls.String, decls.String))},
-			vars:    map[string]interface{}{"input": map[string]string{"key": "value"}},
-		},
-		{
-			name:    "select: field",
-			program: `"input.single_int32"`,
-			decls:   []*exprpb.Decl{decls.NewVar("input", decls.NewObjectType("google.expr.proto3.test.TestAllTypes"))},
-			vars:    map[string]interface{}{"input": map[string]*proto3pb.TestAllTypes{"key": {SingleInt32: 1}}},
-		},
-		{
-			name:    "select: field test only",
-			program: `"has(input.single_int32)"`,
-			decls:   []*exprpb.Decl{decls.NewVar("input", decls.NewObjectType("google.expr.proto3.test.TestAllTypes"))},
-			vars:    map[string]interface{}{"input": map[string]*proto3pb.TestAllTypes{"key": {SingleInt32: 1}}},
-		},
-		{
-			name:    "create list empty",
-			program: `[]`,
-		},
-		{
-			name:    "create list 10",
-			program: `[1, 2, 3, 4, 5, 6, 7, 8, 9, 10]`,
-		},
-		{
-			name:    "create struct empty",
-			program: `google.expr.proto3.test.TestAllTypes{}`,
-		},
-		{
-			name:    "create struct 3",
-			program: `google.expr.proto3.test.TestAllTypes{single_int32: 1, single_float: 3.14, single_string: 'str'}`,
-		},
-		{
-			name:    "create map empty",
-			program: `{}`,
-		},
-		{
-			name:    "create map 10",
-			program: `{'a': 1, 'b': 2, 'c': 3, 'd': 4, 'e': 5, 'f': 6, 'g': 7, 'h': 8, 'i': 9, 'j': 10}`,
-		},
-		{
-			name:    "all comprehension empty",
-			program: `input.all(e, true)`,
-			decls:   []*exprpb.Decl{decls.NewVar("input", decls.NewListType(decls.String))},
-			vars:    map[string]interface{}{"input": []string{}},
-		},
-		{
-			name:    "all comprehension 10",
-			program: `input.all(e, true)`,
-			decls:   []*exprpb.Decl{decls.NewVar("input", decls.NewListType(decls.String))},
-			vars:    map[string]interface{}{"input": []string{"a", "a", "a", "a", "a", "a", "a", "a", "a", "a"}},
-		},
-		{
-			name:    "contains empty",
-			program: `input.contains(arg1)`,
-			decls: []*exprpb.Decl{
-				decls.NewVar("input", decls.String),
-				decls.NewVar("arg1", decls.String),
-			},
-			vars: map[string]interface{}{"input": "", "arg1": "x"},
-		},
-		{
-			name:    "contains 10",
-			program: `input.contains(arg1)`,
-			decls: []*exprpb.Decl{
-				decls.NewVar("input", decls.String),
-				decls.NewVar("arg1", decls.String),
-			},
-			vars: map[string]interface{}{"input": "abcdefghij", "arg1": "abcdefghij"},
-		},
-		{
-			name:    "matches empty",
-			program: `input.matches(arg1)`,
-			decls: []*exprpb.Decl{
-				decls.NewVar("input", decls.String),
-				decls.NewVar("arg1", decls.String),
-			},
-			vars: map[string]interface{}{"input": "", "arg1": "[0-9]*"},
-		},
-		{
-			name:    "matches 10",
-			program: `input.matches(arg1)`,
-			decls: []*exprpb.Decl{
-				decls.NewVar("input", decls.String),
-				decls.NewVar("arg1", decls.String),
-			},
-			vars: map[string]interface{}{"input": "abcdefghi1", "arg1": "[0-9]*"},
-		},
-		{
-			name:    "or",
-			program: `true || false`,
-		},
-		{
-			name:    "and",
-			program: `true && false`,
-		},
-		{
-			name:    "lt",
-			program: `1 < 2`,
-		},
-		{
-			name:    "lte",
-			program: `1 <= 2`,
-		},
-		{
-			name:    "eq",
-			program: `1 == 2`,
-		},
-		{
-			name:    "gt",
-			program: `2 > 1`,
-		},
-		{
-			name:    "gte",
-			program: `2 >= 1`,
-		},
-		{
-			name:    "in empty list",
-			program: `2 in []`,
-		},
-		{
-			name:    "in list 10",
-			program: `2 in [1, 2, 3, 4, 5, 6, 7, 8, 9, 10]`,
-		},
-		{
-			name:    "plus",
-			program: `1 + 1`,
-		},
-		{
-			name:    "minus",
-			program: `1 - 1`,
-		},
-		{
-			name:    "divide",
-			program: `1 / 1`,
-		},
-		{
-			name:    "multiply",
-			program: `1 * 1`,
-		},
-		{
-			name:    "modulo",
-			program: `1 % 1`,
-		},
-		{
-			name:    "ternary",
-			program: `true ? 1 : 2`,
-		},
-		{
-			name:    "string size",
-			program: `size("123")`,
-		},
-		{
-			name:    "int to string conversion",
-			program: `string(1)`,
-		},
-	}
-	for _, tc := range cases {
-		b.Run(tc.name, func(bb *testing.B) {
-			vars := tc.vars
-			if vars == nil {
-				vars = map[string]interface{}{}
-			}
-			descriptor := new(proto3pb.TestAllTypes).ProtoReflect().Descriptor()
-			e, err := NewEnv(Declarations(tc.decls...), DeclareContextProto(descriptor))
-			if err != nil {
-				log.Fatalf("environment creation error: %s\n", err)
-			}
-			e, _ = e.Extend(CustomTypeAdapter(types.DefaultTypeAdapter))
-			ast, iss := e.Compile(tc.program)
-			if iss.Err() != nil {
-				log.Fatalln(iss.Err())
-			}
-			prg, err := e.Program(ast)
-			if err != nil {
-				log.Fatalf("program creation error: %s\n", err)
-			}
-			b.ResetTimer()
-			b.ReportAllocs()
-			for i := 0; i < bb.N; i++ {
-				_, _, err := prg.Eval(vars)
-				if err != nil {
-					b.Fatal(err)
-				}
-			}
-		})
-	}
 }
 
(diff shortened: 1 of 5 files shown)


checker/cost.go
```diff
@@ -0,0 +1,536 @@
...
+	EstimateSize(element Expr) *int64
+	// EstimateCallCost returns the estimated cost of an invocation, or nil if
+	// the estimator has no estimate to provide. EstimateCallCost is only called for functions
+	// that CEL does not already know the cost; EstimateCallCost is not called for standard operations
+	// or functions.
+	EstimateCallCost(name string, target *Expr, args []Expr) *CostEstimate
+}
+
+// Expr represents an AST node for the purpose of cost estimations.
+type Expr interface {
```
TristonianJones: Consider renaming this to `AstNode` as there are quite a few other `Expr` concepts.

checker/cost.go
```diff
@@ -0,0 +1,536 @@
...
+	"github.com/google/cel-go/common/containers"
+	"github.com/google/cel-go/common/types/ref"
+)
+
+// CostEstimator estimates the sizes of variable length input data and the costs of functions.
+type CostEstimator interface {
+	// EstimateSize returns the worst case size possible for the given Expr, or nil if
+	// the estimator has no estimate to provide. EstimateSize is only called for Exprs where
+	// CEL does not know the size; EstimateSize is not called for literals defined inline in CEL.
+	EstimateSize(element Expr) *int64
```
TristonianJones: Consider returning a `*SizeEstimate` here. though the size feels more like estimating the bounds. Perhaps, `EstimateBounds`?

Another option is to just reuse `CostEstimate` as the result from both calls, though the name might need to change if this were the case.

checker/cost.go
```diff
@@ -0,0 +1,536 @@
...
+	EstimateCallCost(name string, target *Expr, args []Expr) *CostEstimate
+}
+
+// Expr represents an AST node for the purpose of cost estimations.
+type Expr interface {
+	// Type returns the deduced type of the Expr.
+	Type() *exprpb.Type
+	// LiteralSize returns the size of the Expr if it is a literal defined inline in CEL, or
+	// nil if the Exprs size is not statically known.
+	LiteralSize() *int64
```
TristonianJones: Usually when I think of literals, I think of constants, but a list or struct literal would have very clear sizes. Perhaps, just calling this method `FixedSize()` and returning a pointer to an int would be sufficient.

checker/cost.go
```diff
@@ -0,0 +1,536 @@
...
+		switch ck := ek.ConstExpr.ConstantKind.(type) {
+		case *exprpb.Constant_StringValue:
+			v = int64(len(ck.StringValue))
+		case *exprpb.Constant_BytesValue:
+			v = int64(len(ck.BytesValue))
+		default:
+			return nil
+		}
+	case *exprpb.Expr_ListExpr:
+		v = int64(len(ek.ListExpr.Elements))
```
TristonianJones: It's worth a comment about how list size is used here. 

The list _could_ be the size of whatever elements it contains. For example a set membership test of `string in [a, b, c]` it might be relevant to know if either `a`, `b`, or `c` were quite large. It's very probable that you've accounted for this elsewhere, but in the event you haven't, then it's worth spending more time to hammer out.

checker/cost.go
```diff
@@ -0,0 +1,536 @@
...
+	case *exprpb.Expr_StructExpr:
+		if ek.StructExpr.MessageName == "" {
+			v = int64(len(ek.StructExpr.Entries))
```
TristonianJones: The entry set may be composed of expression keys and expression values. Do you want to just consider the entries as size or the sum of the sizes of nested entry key and value expressions?

checker/cost.go
```diff
@@ -0,0 +1,536 @@
...
+)
+
+// CostEstimate represents an estimated cost range.
+type CostEstimate struct {
+	Min, Max int64
+}
+
+// Add adds the costs and returns the sum.
+func (ce CostEstimate) Add(rhs CostEstimate) CostEstimate {
+	return CostEstimate{
```
TristonianJones, line 149: Any thoughts on whether to use a pointer instead? It's fine to use a value, just curious though since in the `Expr` interface it's not clear whether the pointer is being used to make the return "optional", or if the pointer is used to indicate that the value is longer lived somewhow.

checker/cost.go
```diff
@@ -0,0 +1,536 @@
...
+
+	// undeclared reference
+	c.setType(e, decls.Error)
+	return CostEstimate{}
+}
+
+func (c *coster) costSelect(e *exprpb.Expr) CostEstimate {
+	sel := e.GetSelectExpr()
+	var sum CostEstimate
+	// Before traversing down the tree, try to interpret as qualified name.
```
TristonianJones: I would refrain from copying the checker code here. All of the information you want should be in the `references` and `types` derived from the `CheckedExpr.reference_map` and `CheckedExpr.type_map` respectively.

checker/cost.go
```diff
@@ -0,0 +1,536 @@
...
+	case *exprpb.Expr_ComprehensionExpr:
+		cost = c.costComprehension(e)
+	default:
+		return CostEstimate{}
+	}
+	return cost
+}
+
+func (c *coster) costIdent(e *exprpb.Expr) CostEstimate {
+	identExpr := e.GetIdentExpr()
```
TristonianJones, line 259: Ident would be the top-level variable. Usually this will be an `identCost`. There should be no need to iterate through the identifiers in the environment since the type-checker will have already done this for you and rewritten the expression to ensure that only resolved identifiers appear in `IdentExpr` nodes.

checker/cost.go
```diff
@@ -0,0 +1,536 @@
...
+	// undeclared reference
+	c.setType(e, decls.Error)
+	return CostEstimate{}
+}
+
+func (c *coster) costSelect(e *exprpb.Expr) CostEstimate {
+	sel := e.GetSelectExpr()
+	var sum CostEstimate
+	// Before traversing down the tree, try to interpret as qualified name.
+	qname, found := containers.ToQualifiedName(e)
```
TristonianJones: Skip the qualified name resolution. It will have already been done for you.

checker/cost.go
```diff
@@ -0,0 +1,536 @@
...
+		if isDynOrError(targetType) {
+			resultType = decls.Dyn
+		} else {
+			// does not support field selection
+		}
+	}
+	if sel.TestOnly {
+		resultType = decls.Bool
+	}
+	c.setType(e, resultType)
```
TristonianJones: I would just make the `types` immutable (skip setting the type or computing the result type) since the information should already be filled in by the type-checker.

checker/cost.go
```diff
@@ -0,0 +1,536 @@
...
+		}
+	}
+
+	// Interpret as field selection, first traversing down the operand.
+	sum = sum.Add(c.cost(sel.Operand))
+	targetType := c.getType(sel.Operand)
+	// Assume error type by default as most types do not support field selection.
+	resultType := decls.Error
+	switch kindOf(targetType) {
+	case kindMap:
```
TristonianJones: I would combine the cases here:

```
switch kindOf(targetType) {
  case kindMap, kindObject, kindTypeParam:
    sum = sum.Add(selectCost)
}

File: checker/cost.go [hunk 12]
```diff
@@ -0,0 +1,536 @@
...
+		// Maps yield their value type as the selection result type.
+		mapType := targetType.GetMapType()
+		resultType = mapType.ValueType
+		sum.Add(selectCost)
```
TristonianJones: Don't forget to assign to sum or return here and below.

checker/cost.go
```diff
@@ -0,0 +1,536 @@
...
+			call.Function = fn.GetName()
+			return sum
+		}
+	}
+
+	// Regular instance call.
+	var targetType Expr
+	if call.Target != nil {
+		sum = sum.Add(c.cost(call.Target))
+		targetType = expr{t: c.types[call.Target.Id], expr: call.Target}
```
TristonianJones: Consider using `expr{t: c.getType(call.GetTarget()) ...}`

checker/cost.go
```diff
@@ -0,0 +1,536 @@
...
+	}
+	fn := c.env.LookupFunction(fnName)
+	// Function found, attempt overload resolution.
+	if fn != nil {
+		sum = sum.Add(c.functionCost(call.Function, &targetType, argTypes))
```
TristonianJones: The `reference_map` should indicate the specific overload id for the function being used. It's probably worthwhile to create a map of function overload id to `costFactor`

checker/cost.go
```diff
@@ -0,0 +1,536 @@
...
+	}
+	// Function name not declared.
+	return CostEstimate{}
+}
+
+func (c *coster) costCreateList(e *exprpb.Expr) CostEstimate {
+	create := e.GetListExpr()
+	var elemType *exprpb.Type
+	var sum CostEstimate
+	for _, e := range create.Elements {
```
TristonianJones: I feel like you need to either compute the cost in the aggregate or per-element. The per-element approach is likely to be more accurate in many cases as the use case for scaling expression cost by list or map size is usually when using a list or map variable whose size cannot be computed statically.

checker/cost.go
jpbetz: I've clarified that size is the same as what CEL size() computes (lengths of strings/bytes, # of list items, # of map entries).

checker/cost.go
jpbetz: A non-pointer seemed appropriate. It's just a composite scalar with operations that don't mutate state.

checker/cost.go
jpbetz: Ahh, this is what I needed. The duplication was really bothering me. This trims the code up substantially.

checker/cost.go
```diff
@@ -0,0 +1,645 @@
...
+func (ce CostEstimate) MultiplyByCostFactor(costPerUnit float64) CostEstimate {
+	return CostEstimate{
+		multiplyByCostFactor(ce.Min, costPerUnit),
+		multiplyByCostFactor(ce.Max, costPerUnit),
+	}
+}
+
+// addUint64NoOverflow adds non-negative ints. If the result is exceeds math.MaxUint64, math.MaxUint64
+// is returned.
+func addUint64NoOverflow(x, y uint64) uint64 {
```
TristonianJones, line 157: I'm okay with exporting the overflow.go calculations if we can just re-use them here.

checker/cost.go
jpbetz: Is createList only for when a list is declared inline in CEL, e.g. `[1, 2, 3 + 5 / 2]`? If so I think the per-element approach is right.

checker/cost.go
jpbetz: I was _almost_ able to make this work. I have one remaining setType that in costIdent that ensures that the type of identifiers is the exact same type reference that was provided by the typeProvider. I haven't isolated why this is needed yet. It might be better to look at this one together offline as it's a bit tough to explain.

checker/cost.go
jpbetz: I've clarified that size is the same as what CEL `size()` computes (lengths of strings/bytes, # of list items, # of map entries).

jpbetz:
Feedback applied. This is ready for another pass @TristonianJones

cel/env.go
```diff
@@ -405,6 +405,12 @@ func (e *Env) ResidualAst(a *Ast, details *EvalDetails) (*Ast, error) {
 	return checked, nil
 }
 
+// EstimateCost estimates the cost of the CEL expression using the length estimates of input data and extension functions
+// provided by estimator.
+func (e *Env) EstimateCost(ast *exprpb.CheckedExpr, estimator checker.CostEstimator) checker.CostEstimate {
```
jpbetz: @TristonianJones Should we include something like this? Would you expect it to automatically make sure the checker is run?

checker/cost.go
jpbetz: I agree. This should be removed. I hit a snag and added a TODO to try to explain. It might be worth syncing offline about this one. I'm not sure I understand the codebase well enough to resolve the problem.

cel/env.go
jpbetz: Nevermind, I think I figured out how to do this reasonably cleanly.

checker/cost.go
jpbetz: Great, I've incorporated overload ids.

checker/cost.go
```diff
@@ -0,0 +1,432 @@
...
+
+// SizeEstimate represents an estimated size of a variable length string, bytes, map or list.
+type SizeEstimate struct {
+	Min, Max uint64
+}
+
+// Multiply multiplies by another SizeEstimate and returns the sum.
+// If multiply would result in an uint64 overflow, the result is Maxuint64.
+func (ce SizeEstimate) Multiply(sizeEstimate SizeEstimate) CostEstimate {
+	return CostEstimate{
```
TristonianJones: Shouldn't this be a `SizeEstimate` here?

checker/cost.go
```diff
@@ -0,0 +1,432 @@
...
+}
+
+// SizeEstimate represents an estimated size of a variable length string, bytes, map or list.
+type SizeEstimate struct {
+	Min, Max uint64
+}
+
+// Multiply multiplies by another SizeEstimate and returns the sum.
+// If multiply would result in an uint64 overflow, the result is Maxuint64.
+func (ce SizeEstimate) Multiply(sizeEstimate SizeEstimate) CostEstimate {
```
TristonianJones: Prefer `se` or `e` for the `SizeEstimate`

checker/cost.go
```diff
@@ -0,0 +1,432 @@
...
+// Cost estimates the cost of the parsed CEL expression.
+// The parsedExpr is expected to have passed type checking. If it has not been
+// type checked, the results are undefined.
+func Cost(checker *exprpb.CheckedExpr,
+	env *Env, estimator CostEstimator) CostEstimate {
+	c := coster{
+		env:       env,
+		checker:   checker,
+		estimator: estimator,
+	}
```
TristonianJones, line 228: Save the `type_map` from the `CheckedExpr` here so that you can refer to the type by the `Expr.GetId()` rather than trying to compute it dynamically, e.g. 

```go
func (c *coster) getType(expr *exprpb.Expr) {
  return c.type_map[expr.GetId()]
}
```

checker/cost.go
TristonianJones: I added a comment about how to use the `CheckedExpr.type_map` near the `coster` struct definition. Hopefully that helps with removing the need for the `c.env.IdentLookup`

checker/cost.go
```diff
@@ -0,0 +1,432 @@
...
+		return *est
+	}
+	switch overloadId {
+	// O(n) functions
+	case overloads.StartsWithString, overloads.EndsWithString, overloads.StringToBytes, overloads.BytesToString:
+		if len(args) == 1 {
+			return c.sizeEstimate(args[0]).MultiplyByCostFactor(0.1)
+		}
+	case overloads.InList:
+		// TODO: account for optimization where lists are converted to sets?
```
TristonianJones: If the `CreateList` is composed entirely of constant values, then the optimization applies, but you'd need to get this information from the `CheckedExpr` args. It's probably fine the leave the cost computation as-is here.

checker/cost.go
```diff
@@ -0,0 +1,432 @@
...
+		// Set the range iteration variable to type DYN as well.
+		varType = decls.Dyn
+	default:
+		// not a valid range
+		varType = decls.Error
+	}
+
+	// Create a scope for the comprehension since it has a local accumulation variable.
+	// This scope will contain the accumulation variable used to compute the result.
+	c.env = c.env.enterScope()
```
TristonianJones: With the `type_map` change, I think you can remove the dependence on the `*Env` and the scope management used below.

checker/cost.go
jpbetz: Yes, good catch.

checker/cost.go
jpbetz: I'd really like to not have to define these. They're subtly different.  Any operation that would evaluate to a value greater than MaxInt instead evaluates to MaxInt.  I've also used uint64 throughout this code since cost should be non-negative and it seemed simpler to have that enforced by the type.

I'm on the fence but currently inclined to keep these locally defined due to the differences.

checker/cost.go
jpbetz: Okay, I've dug into this and figured out what's going wrong.

The problem I'm having is that I need to attach maxLength information to `exprpb.Type`. The way I'm doing that here is to build up a "hints" `map[exprpb.Type]CostEstimate`.  But Check substitutes the `exprpb.Type` in `CheckedExpr.TypeMap` (for good reason) so I lose the keys in the mapping.

Alternatives:
- Check returns an unsubstituted mapping (not a real option, this is unsafe)
- Check somehow returns access to its `exprpb.Type` -> `exprpb.Type` `mappings` so I can follow the substitutions
- I continue to reconstruct the `Expr`->`Type` mapping (a poor option, I agree)
- Some alternative I haven't thought of

(Note that two identical `exprpb.Type`s might have different CostEstimates, so I can't identify them structurally)

jpbetz: Per offline discussion, I've added field paths. They seem to work fine.

checker/cost.go
jpbetz: https://github.com/google/cel-go/pull/483/commits/49c0f67e25e528a1d7827caf6cd00d972e98de28#r809444558

checker/cost.go
jpbetz: For field paths I still need to do some scope tracking but I was able to drop all the type computations.

cel/cel_test.go
```diff
@@ -1190,38 +1194,296 @@ func TestCustomInterpreterDecorator(t *testing.T) {
...
+		t.Run(tc.name, func(t *testing.T) {
+			if tc.hints == nil {
+				tc.hints = map[string]int64{}
+			}
+			descriptor := new(proto3pb.TestAllTypes).ProtoReflect().Descriptor()
+			e, err := NewEnv(Declarations(tc.decls...), DeclareContextProto(descriptor))
+			if err != nil {
+				t.Fatalf("environment creation error: %s\n", err)
+			}
+			e, _ = e.Extend(CustomTypeAdapter(types.DefaultTypeAdapter))
```
TristonianJones: You can move up the `CustomTypeAdapter()` into the first `NewEnv()` call.

cel/cel_test.go
```diff
@@ -1190,38 +1194,296 @@ func TestCustomInterpreterDecorator(t *testing.T) {
...
-	if min != wantedCost[0] || max != wantedCost[1] {
-		log.Fatalf("Got cost interval [%v, %v], wanted %v",
-			min, max, wantedCost)
+
+	for _, tc := range cases {
+		t.Run(tc.name, func(t *testing.T) {
+			if tc.hints == nil {
+				tc.hints = map[string]int64{}
+			}
+			descriptor := new(proto3pb.TestAllTypes).ProtoReflect().Descriptor()
```
TristonianJones: I don't think you're actually using the `TestAllTypes` proto in this test, so you can remove this line and the `DeclareContextProto()` option below.

cel/env.go
```diff
@@ -405,6 +406,16 @@ func (e *Env) ResidualAst(a *Ast, details *EvalDetails) (*Ast, error) {
 	return checked, nil
 }
 
+// EstimateCost estimates the cost of a type checked CEL expression using the length estimates of input data and
+// extension functions provided by estimator.
+func (e *Env) EstimateCost(ast *Ast, estimator checker.CostEstimator) (checker.CostEstimate, error) {
+	if !ast.IsChecked() {
+		return checker.CostEstimate{}, fmt.Errorf("EsimateCost may only be called with a type checked Ast")
+	}
+	checked := &exprpb.CheckedExpr{Expr: ast.Expr(), ReferenceMap: ast.refMap, TypeMap: ast.typeMap, SourceInfo: ast.SourceInfo()}
```
TristonianJones: Prefer `checked := AstToCheckedExpr(ast)`

checker/cost.go
```diff
@@ -0,0 +1,430 @@
...
+	EstimateSize(element AstNode) *SizeEstimate
+	// EstimateCallCost returns the estimated cost of an invocation, or nil if
+	// the estimator has no estimate to provide.
+	EstimateCallCost(overloadId string, target *AstNode, args []AstNode) *CostEstimate
+}
+
+// AstNode represents an AST node for the purpose of cost estimations.
+type AstNode interface {
+	// Path returns a path to the AstNode. The first path element is a variable. All subsequent path elements are
+	// field selectors or '*' to indicate traversal into the elements or  of a map of list.
```
TristonianJones: nit: s/or of/of/

checker/cost.go
```diff
@@ -0,0 +1,430 @@
...
+		return CostEstimate{}
+	}
+	return cost
+}
+
+func (c *coster) costIdent(e *exprpb.Expr) CostEstimate {
+	identExpr := e.GetIdentExpr()
+
+	// resolve the field path
+	if ident := c.env.LookupIdent(identExpr.GetName()); ident != nil {
```
TristonianJones: The lookup here indicates whether something a variable or a type identifier, but by the time an expression is checked you can be guaranteed that the identifier will be found and that both have the same `identCost`.

Adding a reference to the `c.exprPath` for a type identifier won't do any harm, so I would remove the differentiation between variables and types with the `LookupIdent` call and just use the inner if-else block to differentiate between iter and non-iter identifiers.

checker/cost.go
```diff
@@ -0,0 +1,430 @@
...
+}
+
+func (c *coster) costCreateMap(e *exprpb.Expr) CostEstimate {
+	mapVal := e.GetStructExpr()
+	var sum CostEstimate
+	for _, ent := range mapVal.GetEntries() {
+		key := ent.GetMapKey()
+		sum = sum.Add(c.cost(key))
+
+		sum = sum.Add(c.cost(ent.Value))
```
TristonianJones: Try to consistently use the getter methods here and elsewhere on proto field accesses. I believe at some point only getters will be supported, e.g. `ent.GetValue()`

checker/cost.go
```diff
@@ -0,0 +1,430 @@
...
+	// EstimateCallCost returns the estimated cost of an invocation, or nil if
+	// the estimator has no estimate to provide.
+	EstimateCallCost(overloadId string, target *AstNode, args []AstNode) *CostEstimate
+}
+
+// AstNode represents an AST node for the purpose of cost estimations.
+type AstNode interface {
+	// Path returns a path to the AstNode. The first path element is a variable. All subsequent path elements are
+	// field selectors or '*' to indicate traversal into the elements or  of a map of list.
+	Path() []string
```
TristonianJones, line 43: It's possible you'll want to consider using the `AttributePattern` in the future as it will give you features like `input.*.field` which may be desirable for handling matches against element access while still retaining fine-grained estimates for sub-element access. 

Using `AttributePattern` would require this code be moved into another package though, probably something like `cel/internal` (new).

checker/cost.go
```diff
@@ -0,0 +1,430 @@
...
+
+func (c *coster) costComprehension(e *exprpb.Expr) CostEstimate {
+	comp := e.GetComprehensionExpr()
+	var sum CostEstimate
+	sum = sum.Add(c.cost(comp.IterRange))
+	sum = sum.Add(c.cost(comp.AccuInit))
+
+	c.env = c.env.enterScope()
+	c.env.Add(decls.NewVar(comp.AccuVar, nil)) // Don't need to track types
+	c.env = c.env.enterScope()
```
TristonianJones: I think you can remove the scope entry / exit based on the simplification suggestion related to simplifying identifier resolution.

checker/cost.go
jpbetz: I switched to this scheme for field path elements (after the var name): field name, '@items', '@keys', '@values'.

checker/cost.go
jpbetz: I've given it the college try :)

checker/cost.go
jpbetz: Sounds good. I've added some test around edge cases where I need to distinguish between the map key type and value type, so I landed on using "@items", "@keys" and "@values" instead of "*" in the path.

checker/cost.go
jpbetz: With scopes I lose 'items.@keys' field path for "k.contains" in this case:

```
input.all(k, input[k].all(k, true) && k.contains(k))
```

jpbetz: I'll use a really simple stack to get around this.

checker/cost.go
jpbetz: SGTM. I've tried to simplify this to match. Let me know if I misunderstood.

cel/cel_test.go
jpbetz: I'm using `TestAllTypes` to represent message types in the tests.

TristonianJones: Ah, right. Though I think the `DeclareContextProto` isn't doing what you expect. It should be publishing a bunch of declarations for all of the type-level fields in the proto.

checker/cost.go
```diff
@@ -0,0 +1,462 @@
...
+func (c *coster) costCall(e *exprpb.Expr) CostEstimate {
+	call := e.GetCallExpr()
+	target := call.GetTarget()
+	args := call.GetArgs()
+
+	var sum CostEstimate
+
+	argTypes := make([]AstNode, len(args))
+	for i, arg := range args {
+		// TODO: && || short circuit, so min cost should only include 1st arg eval
```
TristonianJones: Similar for ternary. Min cost should be `cond + min(a, b)` within `<cond> ? a : b`. Probably worth an additional comment.

jpbetz: +1, added a note about comprehension short circuiting as well. The minimums are not quite right in the current implementation.

checker/cost.go
```diff
@@ -0,0 +1,468 @@
...
+	}
+	return 0, false
+}
+
+// Cost estimates the cost of the parsed CEL expression.
+// The parsedExpr is expected to have passed type checking. If it has not been
+// type checked, the results are undefined.
+func Cost(checker *exprpb.CheckedExpr, estimator CostEstimator) CostEstimate {
+	c := coster{
+		checker:    checker,
```
TristonianJones: nit: prefer `expr` or `checkedExpr` since `checker` is the name of the package and the input variable is just data and not doing any computation.

checker/cost.go
```diff
@@ -0,0 +1,468 @@
...
+func (vs iterRangeScopes) peek(varName string) (int64, bool) {
+	varStack := vs[varName]
+	if len(varStack) > 0 {
+		return varStack[len(varStack)-1], true
+	}
+	return 0, false
+}
+
+// Cost estimates the cost of the parsed CEL expression.
+// The parsedExpr is expected to have passed type checking. If it has not been
```
TristonianJones: You probably don't need a note on `parsedExpr` here.

cel/cel_test.go
jpbetz: Oh, I get it. I've switched this to use `Types(...)`

checker/cost.go
jpbetz: Yeah, that's stale. Removing.

TristonianJones (approved):
I just have a couple of last minor nits, but otherwise looks good. Thanks @jpbetz!

checker/cost.go
jpbetz: checkedExpr SGTM. Updated.

PR #486, opened 2022-02-17
Context-based evaluation support

TristonianJones:
Allow CEL program evaluation to occur within a Golang `Context` to support the
cancellation and timeout requirements of latency critical applications.

### DIFF
--- a/cel/cel_test.go
+++ b/cel/cel_test.go
@@ -15,12 +15,14 @@
 package cel
 
 import (
+	"context"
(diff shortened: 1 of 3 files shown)


jpbetz:
LGTM

I'm interested in halting CEL execution (so it can't potentially run indefinitely in the background), but we can follow up on that.  This gets us the API and helps a ton for CEL programs that are taking too long and will terminate.

Issue #488, opened 2022-02-22
Poor performance of concatList Value() method


hzhaop:
hi, all,

expression: `data.map(item, item.id)`
input: 
`{"data": [{"id": "1"}, {"id": "2"}, ... , {"id": "1000"}]}`

I call the Value() method of the program.Eval() output, find it very slow to return the value.

I find out that it's because `data.map(item, item.id)` produce a concatList, which is very slow in Value() method. (about O(N^2) complexity)

Is there any workaround method that I could enhance this expression performance?

TristonianJones:
#491 should improve evaluation speed by 40x and ensure that the call to `Value()` is a constant time operation. The combination of the two effects should yield an even more significant performance boost for this use case.

PR #490, opened 2022-02-22
improve performance of concatList Value method  google/cel-go#488

hzhaop:
The time complexity of original Value() implementation of concatList is O(N^2), this PR improves it to O(N).

If prevList or nextList is type of concatList, there's no need to call Get method which is O(N) complexity.

Thanks.

### DIFF
--- a/common/types/list.go
+++ b/common/types/list.go
@@ -404,13 +404,31 @@ func (l *concatList) Type() ref.Type {
 func (l *concatList) Value() interface{} {
 	if l.value == nil {
 		merged := make([]interface{}, l.Size().(Int))
+
 		prevLen := l.prevList.Size().(Int)
-		for i := Int(0); i < prevLen; i++ {
-			merged[i] = l.prevList.Get(i).Value()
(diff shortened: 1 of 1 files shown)


common/types/list.go
```diff
@@ -404,13 +404,31 @@ func (l *concatList) Type() ref.Type {
 func (l *concatList) Value() interface{} {
 	if l.value == nil {
 		merged := make([]interface{}, l.Size().(Int))
+
 		prevLen := l.prevList.Size().(Int)
-		for i := Int(0); i < prevLen; i++ {
-			merged[i] = l.prevList.Get(i).Value()
+		switch l.prevList.(type) {
+		case *concatList:
+			newMerges := l.prevList.Value().([]interface{})
```
TristonianJones, line 411: There aren't any guarantees that the `Value()` call will return a `[]interface{}`, so this optimization won't always work.

TristonianJones:
I've decided to close this PR since the #491 should address most of what you were hoping to achieve, but should work for any CEL `Lister` implementation.

PR #491, opened 2022-02-22
List append optimizations for comprehension loops

### LINKED ISSUE #488: Poor performance of concatList Value() method

TristonianJones:
Introduce the list append optimizations used within C++ to ensure that comprehensions are able to treat internal
CEL variables as mutable for the purposes of list construction. Closes #488 

Before the optimization:
```
BenchmarkContextEval-8   	     250	   4680999 ns/op	   43948 B/op	    1323 allocs/op
```

After the optimization:
```
BenchmarkContextEval-8   	   10233	    116829 ns/op	   35516 B/op	    1038 allocs/op
```

The memory usage is not that different between the before and after, but the overall evaluation time (117us vs. 4.58ms) represents a 40x improvement in evaluation speed which will significantly help comprehension-heavy uses of CEL with even moderately sized inputs.

### DIFF
--- a/cel/cel_test.go
+++ b/cel/cel_test.go
@@ -857,7 +857,7 @@ func TestContextEval(t *testing.T) {
 		t.Errorf("ContextEval() got %v, wanted 75", out)
 	}
 
-	evalCtx, cancel := context.WithTimeout(ctx, 100*time.Microsecond)
+	evalCtx, cancel := context.WithTimeout(ctx, 50*time.Microsecond)
 	defer cancel()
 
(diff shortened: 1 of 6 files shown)


jnthntatum (approved):
LGTM

PR #492, opened 2022-02-23
Update ContextEval to halt CEL evaluation when cancelled

jpbetz:
Add a decorator that checks the context Done() channel every 100 InterpretableCall invocations to halt CEL evaluation when the context is cancelled.

### DIFF
--- a/cel/cel_test.go
+++ b/cel/cel_test.go
@@ -1017,6 +1017,17 @@ func TestContextEval(t *testing.T) {
 	if err != nil && err.Error() != "operation cancelled" {
 		t.Errorf("Got %v, wanted operation cancelled error", err)
 	}
+
+	//  Run CEL evaluation in the foreground (don't run it in a background go routine) and
+	// verify cancellation halts evaluation.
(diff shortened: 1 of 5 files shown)


jpbetz:
@TristonianJones Curious how you feel about this.

I don't like having to add interpretable nodes here, but that's the existing mechanism so I kept the pattern. Maybe an alternative in the future would be to have a program wide pre|post eval chain that we could put checks like this in.

TristonianJones:
@jpbetz In most cases, the halting checks should be isolated to comprehension loops where it's possible to evaluate for a significant period of time. My thought had been to provide a decorator which only affects the `evalFold` operation. The idea would be similar though. I'll look it over in more detail tomorrow.

jpbetz:
I've limited the decoration to the step eval of a fold, I discovered that this requires a panic to early exit because the fold continues to evaluate after the step op returns an error. I don't have a strong opinion on if we should avoid a panic here or not.

TristonianJones:
I'm going to close this in favor of #502

PR #494, opened 2022-02-23
Runtime cost calculation

cici37:
Introduce runtime cost calculation into CEL.

The work is part of this [issue](https://github.com/kubernetes/kubernetes/issues/107573)

The current PR includes:

- Add runtime cost calculation into evaluation path
- Test to verify runtime cost calculation
- Test to verify runtime cost fell into estimation cost range
- Test to allow cost calculation for extension function

Todo in following PR:

- Add cancellation
- Add more test for large input
- Add test coverage for aligning runtime cost with estimate cost


cel/program.go
```diff
@@ -125,20 +126,29 @@ func newCtxResult(val ref.Val, det *EvalDetails, err error) *ctxResult {
...
-	interpretable interpreter.Interpretable
-	attrFactory   interpreter.AttributeFactory
+	evalOpts          EvalOption
+	decorators        []interpreter.InterpretableDecorator
+	defaultVars       interpreter.Activation
+	dispatcher        interpreter.Dispatcher
+	interpreter       interpreter.Interpreter
+	interpretable     interpreter.Interpretable
+	attrFactory       interpreter.AttributeFactory
+	callCostEstimator interpreter.FunctionActualCostEstimator
```
jpbetz: nit: type name is long, maybe just "ActualCostEstimator" ?  I'm okay with keeping the "Estimator" suffix because for functions we're still making what is clearly an estimate.

cel/cel_test.go
```diff
@@ -784,6 +784,37 @@ func TestAstIsChecked(t *testing.T) {
 	}
 }
 
+func TestTrackCost(t *testing.T) {
```
jpbetz: Consider removing since other test is a superset.

cel/options.go
```diff
@@ -369,6 +370,8 @@ const (
 	// By itself, OptPartialEval does not change evaluation behavior unless the input to the
 	// Program Eval is an PartialVars.
 	OptPartialEval EvalOption = 1 << iota
+
+	OptTrackCost EvalOption = 1 << iota
```
jpbetz, line 381: +godoc

cel/options.go
```diff
@@ -381,6 +384,13 @@ func EvalOptions(opts ...EvalOption) ProgramOption {
 	}
 }
 
+func CallCostEstimator(costEstimator interpreter.FunctionActualCostEstimator) ProgramOption {
```
jpbetz: +godoc

cel/program.go
```diff
@@ -125,20 +126,29 @@ func newCtxResult(val ref.Val, det *EvalDetails, err error) *ctxResult {
 	}
 }
 
+func (ed *EvalDetails) ActualCost() *uint64 {
```
jpbetz, line 105: +godoc

cel/runtimecost_test.go
```diff
@@ -0,0 +1,433 @@
...
+)
+
+type testInfo struct {
+	env     *Env
+	in      interface{}
+	lhsExpr string
+	rhsExpr string
+}
+
+//func computeCosts(t *testing.T, info *testInfo) (lhsCost, rhsCost uint64) {
```
jpbetz: enable or remove commented out code

cel/runtimecost_test.go
```diff
@@ -0,0 +1,433 @@
...
+//	for i, testCase := range smallerCases {
+//		lhsCost, rhsCost := computeCosts(t, &testCase)
+//		if lhsCost >= rhsCost {
+//			t.Errorf(`Expected smaller cost case #%d, expect the cost of expression "%s" to be strictly smaller than "%s", but got %d vs. %d respectively`,
+//				i, testCase.lhsExpr, testCase.rhsExpr, lhsCost, rhsCost)
+//		}
+//	}
+//}
+
+func TestRuntimeCost(t *testing.T) {
```
jpbetz: I'd strongly prefer we don't duplicate all the tests cases from estimated cost.  Instead could we have the estimated cost test suite covers both estimated and actual cost (renaming it as appropriate)?

interpreter/interpretable.go
```diff
@@ -68,6 +68,8 @@ type InterpretableAttribute interface {
 
 	// Resolve returns the value of the Attribute given the current Activation.
 	Resolve(Activation) (interface{}, error)
+
+	Cost() (min, max int64)
```
jpbetz: remove? I don't think this is needed

interpreter/runtimecost.go
```diff
@@ -0,0 +1,178 @@
...
+	"github.com/google/cel-go/common/types/ref"
+	"math"
+	"reflect"
+)
+
+type FunctionActualCostEstimator interface {
+	CallCost(overloadId string, args []ref.Val) *uint64
+}
+
+type CostTracker struct {
```
jpbetz, line 87: +godoc

interpreter/runtimecost.go
```diff
@@ -0,0 +1,178 @@
...
+		// Note that the CEL bytes type is implemented with Go byte slices, therefore also supported by the following
+		// code.
+		case reflect.String, reflect.Array, reflect.Slice, reflect.Map:
+			return uint64(reflectV.Len())
+		}
+	}
+	return 0
+}
+
+func (c CostTracker) ActualCost() uint64 {
```
jpbetz, line 96: +godoc

interpreter/runtimecost.go
```diff
@@ -0,0 +1,178 @@
...
+	return val
+}
+
+type evalCostTrackerAttribute struct {
+	InterpretableAttribute
+	tracker *CostTracker
+}
+
+func (e *evalCostTrackerAttribute) Eval(ctx Activation) ref.Val {
+	// TODO: compute more accurate cost values for attributes.
```
jpbetz: Anything left TODO here? It looks reasonable

interpreter/runtimecost.go
```diff
@@ -0,0 +1,178 @@
...
+		for _, arg := range args {
+			if v, ok := e.tracker.state.Value(arg.ID()); ok {
+				argValues = append(argValues, v)
+			} else {
+				// FIXME: abort.
+			}
+		}
+		e.tracker.cost += *e.tracker.CallCostEstimator.CallCost(e.OverloadID(), argValues)
+	} else {
+		switch e.OverloadID() {
```
jpbetz: add bidirectional comments here and in the corresponding estimated function cost code referencing each other and reminding developers they must be kept in sync.

interpreter/runtimecost.go
```diff
@@ -0,0 +1,178 @@
...
+}
+
+type evalCostTrackerOp struct {
+	Interpretable
+	tracker *CostTracker
+}
+
+func (e *evalCostTrackerOp) Eval(ctx Activation) ref.Val {
+	val := e.Interpretable.Eval(ctx)
+	e.tracker.cost += 1
```
jpbetz: What's this for?

interpreter/runtimecost.go
```diff
@@ -0,0 +1,178 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Package interpreter provides functions to evaluate parsed expressions with
+// the option to augment the evaluation with inputs and functions supplied at
+// evaluation time.
+package interpreter
```
jpbetz, line 19: nit: empty line before package decl

jpbetz (changes requested):

jpbetz:
@cici37 Add any newly introduce go files to to the bazel files to appease cel-go-builder.

interpreter/interpretable.go
cici37: It is for calculating cost for the attribute. Do we wanna assume cost for const and getting attribute all 0?

jpbetz: CEL currently uses https://github.com/google/cel-go/blob/a29469fda231aa37802d476ef818e2224c25e157/interpreter/coster.go#L24 instead of adding Cost to individual interfaces. Let's continue with that approach. Note that you can use ` estimateCost` to cast and call.

interpreter/runtimecost.go
cici37: This is to calculate cost for operation not implemented InterpretableCall [here](https://github.com/google/cel-go/blob/13c5da6dfd824e1a44f5d5e710e1f0882cbade80/interpreter/runtimecost.go#L176). e.g [logic or ](https://github.com/google/cel-go/blob/a29469fda231aa37802d476ef818e2224c25e157/interpreter/interpretable.go#L180).

cici37:
> Cici Huang Add any newly introduce go files to to the bazel files to appease cel-go-builder.

Looks like I don't have permission to check cel-go-builder job detail. Is there any verification/script which I could use for checking?

interpreter/runtimecost.go
```diff
@@ -0,0 +1,184 @@
...
+	if e.tracker.CallCostEstimator != nil {
+		var argValues []ref.Val
+		for _, arg := range args {
+			if v, ok := e.tracker.state.Value(arg.ID()); ok {
+				argValues = append(argValues, v)
+			} else {
+				// FIXME: abort.
+			}
+		}
+		e.tracker.cost += *e.tracker.CallCostEstimator.CallCost(e.OverloadID(), argValues)
```
jpbetz: I was expecting CallCostEstimator to take precedence over the built in cost code, but to fall back to it if it returns nil. E.g.

```
cost := e.tracker.CallCostEstimator.CallCost(...)
if cost != nil {
  e.tracker.cost += *cost
  return val
}
// fallback to switch e.OverloadID()...
```

cel/options.go
```diff
@@ -374,6 +378,14 @@ func EvalOptions(opts ...EvalOption) ProgramOption {
 	}
 }
 
+// CallCostEstimator enables actual runtime cost estimator and overwrites callCostEstimator
```
jpbetz: Instead of "overwrites callCostEstimator" (which is an internal detail users are not expected to understand), how about "CallCostEstimator registers a ActualCostEstimator that can optionally provide a runtime cost estimate for any function calls. This enables runtime costs to be assigned to calls function library extensions." ?

interpreter/runtimecost.go
```diff
@@ -0,0 +1,184 @@
...
+
+import (
+	"github.com/google/cel-go/checker"
+	"github.com/google/cel-go/common/overloads"
+	"github.com/google/cel-go/common/types/ref"
+	"math"
+	"reflect"
+)
+
+// ActualCostEstimator is the interface to calculate cost for wired in function
```
jpbetz: `ActualCostEstimator provides function call cost estimations at runtime`

interpreter/runtimecost.go
```diff
@@ -0,0 +1,184 @@
...
+type ActualCostEstimator interface {
+	CallCost(overloadId string, args []ref.Val) *uint64
```
jpbetz, line 38: +godoc, maybe "CallCost returns an estimated cost for the function overload invocation with the given args, or nil if it has not estimate to provide. CEL attempts to provide reasonable estimates for it's standard function library, so CallCost should typically not need to provide an estimate for CELs standard function." ?

interpreter/runtimecost.go
jpbetz: :+1:

cel/options.go
```diff
@@ -362,6 +363,9 @@ const (
 	// By itself, OptPartialEval does not change evaluation behavior unless the input to the
 	// Program Eval is an PartialVars.
 	OptPartialEval EvalOption = 1 << iota
+
+	// OptTrackCost enables the cost calculation while validation and return cost within evalDetails
```
jpbetz: Clarify that the cost is made available via the "ActualCost()" function.

cel/options.go
```diff
@@ -374,6 +378,14 @@ func EvalOptions(opts ...EvalOption) ProgramOption {
 	}
 }
 
+// CallCostEstimator enables actual runtime cost estimator and overwrites callCostEstimator
+func CallCostEstimator(costEstimator interpreter.ActualCostEstimator) ProgramOption {
```
jpbetz: Need tests that use this

jpbetz:
This looks good to me. As future work I'd like to start reducing the number of decorators we will need, but I see that as an optimization.

@TristonianJones would you be willing to do a review pass?

interpreter/attributes.go
```diff
@@ -96,6 +96,9 @@ type Attribute interface {
 
 	// Resolve returns the value of the Attribute given the current Activation.
 	Resolve(Activation) (interface{}, error)
+
+	// RuntimeCost returns the runtime cost of the Attribute given the current Activation and EvalState.
+	RuntimeCost(ctx Activation, evalState EvalState) uint64
```
TristonianJones: I would move the `RuntimeCost` method next to the `Coster` interface as `RuntimeCoster` rather than include it on the `Attribute` and other existing interfaces.

interpreter/attributes.go
```diff
@@ -239,6 +242,19 @@ func (a *absoluteAttribute) Cost() (min, max int64) {
 	return
 }
 
+// RuntimeCost implements the RuntimeCost interface method.
+func (a *absoluteAttribute) RuntimeCost(ctx Activation, evalState EvalState) uint64 {
+	cost := uint64(len(a.namespaceNames))
+	if len(a.qualifiers) > 0 {
+		cost += 1
```
TristonianJones: Is the addition of an extra cost node for any qualifier set intentional? Seems like one qualifier should cost 1 and not 2.

interpreter/attributes.go
```diff
@@ -239,6 +242,19 @@ func (a *absoluteAttribute) Cost() (min, max int64) {
 	return
 }
 
+// RuntimeCost implements the RuntimeCost interface method.
+func (a *absoluteAttribute) RuntimeCost(ctx Activation, evalState EvalState) uint64 {
```
TristonianJones: I think something to consider is that the attribute value will have a size if it's a string, list, map, etc. So this code will need to track compute and the resulting size of the Attribute's leaf value for downstream calculations.

interpreter/interpretable.go
```diff
@@ -953,6 +991,24 @@ func (e *evalWatchConst) Cost() (min, max int64) {
 	return estimateCost(e.InterpretableConst)
 }
 
+// evalWatchCall describes a watcher of an instCall Interpretable.
+type evalWatchCall struct {
+	InterpretableCall
+	observer evalObserver
```
TristonianJones: This case should already be handled by `evalWatch`; however, it raises an interesting point.

I think it would be best to update the `evalObserver` interface with a third argument which may be either a `Qualifier` or `Interpretable` value:

```
func evalObserver(id int64, programStep interface{}, value ref.Val)
```

This would allow you to pass the contextual information needed to an observer implementation which could be used to update size without necessarily having to use the state tracking. At least not in the way it's currently done. 

You could in theory maintain a stack of values:

1. `InterpretableCall` values consume argument values off of the value stack, and push one in return.
2. `InterpretableConst` values push an argument value onto the value stack.
3. `Qualifier` values consume an argument value, and push one in return.

There is a missing instruction type, let's call it `InterpretableConstructor` which returns the type of what's being constructed and a list of all `Interpretable` nodes used to init the instance. 

If you have this information, then by the end of the computation you should have a value stack of size 1 which contains the actual cost. It should be slightly better because it allows you to decouple for the state tracking that's currently used, and also remove many of the `Cost` methods on the intepretable nodes since they can be modeled directly within the `Decorator` implementation.

interpreter/attributes.go
TristonianJones: Does the `RuntimeCost` need the `ctx Activation`? It doesn't seem to be used.

interpreter/attributes.go
cici37: Added the size for cost cal

interpreter/attributes.go
cici37: `ctx Activation` is used when calculating runtime cost for [maybeAttribute](https://github.com/google/cel-go/blob/d9052d5addb52468ef4f784c4bb1c296db0262b7/interpreter/attributes.go#L462). Let me know if my understanding of maybeAttribute is wrong

interpreter/interpretable.go
cici37: Thanks for the suggestion. It makes sense in terms of decoupling the state tracking. I am wondering if it is possible to involve this change into my next PR and get the current one merged to unblock the changes from k/k side first?

jpbetz: I'll prototype this real quick and report back.

jpbetz: I've committed a change to this PR to address this one.  

I've mostly followed @TristonianJones suggestions, augmented slightly with one of my own goals of being able to have multiple observers on `Interpretable` nodes using only the eval watch decorators. That helped set things up so that any combination of `OptTrackState`, `OptExhaustiveEval` and `OptTrackCost` can be used safely and with minimal overhead.

I added `InterpretableConstructor` like suggested and also added `InterpretableBooleanBinaryOp` to make it ergonomic to to handle `&&` and `||`.

Other than those changes, this is almost an entirely a simplification and consolidation of its logic to a single observer.

interpreter/attributes.go
jpbetz: All the RuntimeCost interface calls have been dropped, so this is obselete.

interpreter/interpretable.go
cici37: Thanks @jpbetz  for handling this. I have added a fix to avoid potential panic. It should be ready for the next round of review. @TristonianJones Would you mind to give it a second pass when you have time? Thank you

interpreter/interpreter.go
```diff
@@ -53,6 +53,35 @@ func Observe(observers ...EvalObserver) InterpretableDecorator {
 	return decObserveEval(observeFn)
 }
 
+// EvalCanceledError represents a cancelled program evaluation operation.
+type EvalCanceledError struct {
```
TristonianJones: nit: two 'l's in cancelled

cel/runtimecost_test.go
```diff
@@ -0,0 +1,613 @@
+// Copyright 2022 Google LLC
```
TristonianJones: nit: would you mind moving this next to the `interpreter/runtimecost.go` file. This makes it easier to spot deltas in code coverage over time.

interpreter/runtimecost.go
```diff
@@ -0,0 +1,185 @@
...
+		case InterpretableConstructor:
+			switch t.Type() {
+			case types.ListType:
+				tracker.cost += 10
+			case types.MapType:
+				tracker.cost += 30
+			default:
+				tracker.cost += 40
+			}
+		case InterpretableBooleanBinaryOp:
```
TristonianJones: Consider omitting the newly introduced `InterpretableBooleanBinaryOp`. It should be sufficient to add a `default` case which adds a single cost point.

interpreter/runtimecost.go
```diff
@@ -0,0 +1,185 @@
...
+func CostObserver(tracker *CostTracker) EvalObserver {
+	observer := func(id int64, programStep interface{}, val ref.Val) {
+		switch t := programStep.(type) {
+		case InterpretableConst:
+			// zero cost
+		case ConstantQualifier:
+			tracker.cost += 1
+		case InterpretableAttribute:
+			tracker.cost += 1
+		case Qualifier:
```
TristonianJones, line 60: I think `Qualifier` and `ConstantQualifier` should behave the same way with a pop on the value stack and a single increment.

cel/cel_test.go
```diff
@@ -1501,6 +1504,30 @@ func TestEstimateCost(t *testing.T) {
 				t.Fatalf("Got cost interval [%v, %v], wanted [%v, %v]",
 					est.Min, est.Max, tc.want.Min, tc.want.Max)
 			}
+
+			ctx := constructActivation(t, tc.in)
+			checked_ast, iss := e.Check(ast)
+			if iss.Err() != nil {
+				t.Fatalf(`Failed to check expression with error: %v`, iss.Err())
```
TristonianJones: nit: I haven't been very good about requesting this structure on error messages, but the internal style guide recommends the following format:

```go
t.Fatalf(`<Method>(<Args>) failed <to do what>: <error>`, ...)
```

Please update here and below.

cel/cel_test.go
```diff
@@ -1501,6 +1504,30 @@ func TestEstimateCost(t *testing.T) {
...
+			if err != nil {
+				t.Fatalf(`Failed to construct Program with error: %v`, err)
+			}
+			_, details, err := program.Eval(ctx)
+			if err != nil {
+				t.Fatalf(`Failed to evaluate expression with error: %v`, err)
+			}
+			actualCost := details.ActualCost()
+			if actualCost == nil {
+				t.Fatalf(`Null pointer returned for the cost of expression "%s"`, tc.expr)
```
TristonianJones: nit: Use the format `got %v, wanted %v` when describing a difference between the actual and expected values.

```go
t.Fatalf("details.ActualCost() got nil for expression %s, wanted ###", tc.expr)
```

cel/cel_test.go
```diff
@@ -1501,6 +1504,30 @@ func TestEstimateCost(t *testing.T) {
...
+			}
+
+			if est.Min > *actualCost || est.Max < *actualCost {
+				t.Fatalf("runtime cost %d is out of the range of estimate cost [%d, %d]", *actualCost,
```
TristonianJones: nit: consider making this call a `t.Errorf` since it's not a failed precondition for the test, with a format that's revised to discuss the error:  `"details.ActualCost() got %d, wanted value in the estimated cost range [%d, %d]"`

cel/options.go
```diff
@@ -387,6 +391,26 @@ func EvalOptions(opts ...EvalOption) ProgramOption {
 	}
 }
 
+// ActualCostTracking enables cost tracking and registers a ActualCostEstimator that can optionally provide a runtime cost estimate for any function calls.
+// This enables runtime costs to be assigned to calls function library extensions.
+func ActualCostTracking(costEstimator interpreter.ActualCostEstimator) ProgramOption {
```
TristonianJones: I think dropping the `Actual` and calling this `CostTracking` where `EvalDetails` exposes a `Cost` method is reasonable. Unless, of course, this approach collides with the legacy `Coster` methods somehow.

cel/options.go
```diff
@@ -387,6 +391,26 @@ func EvalOptions(opts ...EvalOption) ProgramOption {
...
+	return func(p *prog) (*prog, error) {
+		p.callCostEstimator = costEstimator
+		p.evalOpts |= OptTrackCost
+		return p, nil
+	}
+}
+
+// ActualCostLimit enables cost tracking and sets configures program evaluation to exit early with a
+// "runtime cost limit exceeded" error if the runtime cost exceeds the limit.
+func ActualCostLimit(limit uint64) ProgramOption {
```
TristonianJones: nit: Consider calling this `CostLimit` and be clear that the metric is in terms of CPU rather than memory ... or comment on how the value is an aggregate of both costs.

cel/options.go
```diff
@@ -387,6 +391,26 @@ func EvalOptions(opts ...EvalOption) ProgramOption {
...
+	return func(p *prog) (*prog, error) {
+		p.costLimit = &limit
```
TristonianJones: Is it possible to track the cost limit without setting a `CostEstimator`? Is it an error to use one option and not both?

interpreter/interpretable.go
```diff
@@ -88,6 +88,29 @@ type InterpretableCall interface {
...
+
+	// InitVals returns all the list elements, map key and values or struct field values.
+	InitVals() []Interpretable
+
+	// Type returns the type constructed.
+	Type() ref.Type
+}
+
+// InterpretableBooleanBinaryOp interface for inspecting "&&" and "||" booelan operations.
+type InterpretableBooleanBinaryOp interface {
```
TristonianJones: I'm not sure you need this interface at this time. I realize that it's a special case of an `InterpretableCall` because it can support short-circuiting and so its cost is a function of the sum(evaluation cost thus far) + 1.

interpreter/runtimecost.go
```diff
@@ -0,0 +1,185 @@
...
+	"github.com/google/cel-go/common/types"
+	"github.com/google/cel-go/common/types/ref"
+	"github.com/google/cel-go/common/types/traits"
+)
+
+// ActualCostEstimator provides function call cost estimations at runtime
+// CallCost returns an estimated cost for the function overload invocation with the given args, or nil if it has no
+// estimate to provide. CEL attempts to provide reasonable estimates for its standard function library, so CallCost
+// should typically not need to provide an estimate for CELs standard function.
+type ActualCostEstimator interface {
```
TristonianJones, line 37: nit: drop the `Actual` prefix.

interpreter/runtimecost.go
```diff
@@ -0,0 +1,185 @@
...
+		cost += 1
+
+	}
+	return cost
+}
+
+// actualSize returns the size of value
+func (c CostTracker) actualSize(value ref.Val) uint64 {
+	if sz, ok := value.(traits.Sizer); ok {
+		return uint64(sz.Size().Value().(int64))
```
TristonianJones: Consider returning: `uint64(sz.Size().(types.Int))`

interpreter/runtimecost.go
```diff
@@ -0,0 +1,185 @@
...
+		cost += strCost * regexCost
+	case overloads.ContainsString:
+		strCost := uint64(math.Ceil(float64(c.actualSize(argValues[0])) * 0.1))
+		substrCost := uint64(math.Ceil(float64(c.actualSize(argValues[1])) * 0.1))
+		cost += strCost * substrCost
+	default:
+		// The following operations are assumed to have O(1) complexity.
+		// 1. Concatenation of 2 lists: see the implementation of the concatList type.
+		// 2. Computing the size of strings, byte sequences, lists and maps: presumably, the length of each of these
+		//    data structures are cached and can be retrieved in constant time.
```
TristonianJones: I would add logical `&&` and logical `||` to this list.

interpreter/runtimecost.go
```diff
@@ -0,0 +1,185 @@
...
+		// in length.
+		regexCost := uint64(math.Ceil(float64(c.actualSize(argValues[1])) * 0.25))
+		cost += strCost * regexCost
+	case overloads.ContainsString:
+		strCost := uint64(math.Ceil(float64(c.actualSize(argValues[0])) * 0.1))
+		substrCost := uint64(math.Ceil(float64(c.actualSize(argValues[1])) * 0.1))
+		cost += strCost * substrCost
+	default:
+		// The following operations are assumed to have O(1) complexity.
+		// 1. Concatenation of 2 lists: see the implementation of the concatList type.
```
TristonianJones: For the concatenation of two lists this cost is true, but there's an O(c) multiplier on indexed lookups for concatenated lists where `c` is the number of concatenations performed. In most cases its simplest to consider concatenation as the creation of a new list containing all of the elements of each list. The one exception to this rule is when the right hand side argument only contains a single entry, in which case it's likely an append within a comprehension and should only cost 1. 

Kind of weird and tricky, but hopefully the above guidelines give you a better idea.

interpreter/runtimecost.go
```diff
@@ -0,0 +1,185 @@
...
+			cost += *callCost
+			return cost
+		}
+	}
+	// if user didn't specify, the default way of calculating runtime cost would be used.
+	// if user has their own implementation of ActualCostEstimator, make sure to cover the mapping between overloadId and cost calculation
+	switch call.OverloadID() {
+	// O(n) functions
+	case overloads.StartsWithString, overloads.EndsWithString, overloads.StringToBytes, overloads.BytesToString:
+		cost += uint64(math.Ceil(float64(c.actualSize(argValues[0])) * 0.1))
```
TristonianJones: For these magic multipliers, I think it would be good to create a map of overload id to cost multiplier that can be shared between the `checker/cost.go` and the `interpreter/runtimecost.go`. At the very least creating a static map within each file (if they're different for some reason) seems helpful.

interpreter/interpreter.go
```diff
@@ -38,22 +38,96 @@ type Interpreter interface {
...
+	// Type identifies the cause of the cancellation.
+	Cause CancelationCause
+}
+
+func (e EvalCanceledError) Error() string {
+	switch e.Cause {
+	case ContextCancelled:
+		return "operation cancelled: context was cancelled"
```
TristonianJones: Consider adding a string message to the `EvalCancelledError` rather than using a switch here, and then use a statically declared variable for the predefined error with cause where you need to.

interpreter/interpreter.go
```diff
@@ -38,22 +38,96 @@ type Interpreter interface {
 		decorators ...InterpretableDecorator) (Interpretable, error)
 }
 
+// EvalObserver is a functional interface that accepts an expression id and an observed value.
```
TristonianJones, line 41: Update the docs to indicate the relationship between the `programStep` and `value`. Also document what kinds of values are accepted for the `programStep` value (`Interpretable` and `Qualifier` values)

cel/runtimecost_test.go
```diff
@@ -0,0 +1,613 @@
...
+
+func computeCost(t *testing.T, env *Env, expr string, ctx *interpreter.Activation) uint64 {
+	t.Helper()
+
+	// Compile and check expression.
+	ast, iss := env.Compile(expr)
+	if iss.Err() != nil {
+		t.Fatalf(`Failed to compile expression "%s", error: %v`, expr, iss.Err())
+	}
+	checked_ast, iss := env.Check(ast)
```
TristonianJones: The compiled expression is already type-checked. No need to do this again.

cel/runtimecost_test.go
```diff
@@ -0,0 +1,613 @@
...
+			if err != nil {
+				t.Fatalf("environment creation error: %s\n", err)
+			}
+			ast, iss := e.Compile(tc.expr)
+			if iss.Err() != nil {
+				t.Fatal(iss.Err())
+			}
+
+			ctx := constructActivation(t, tc.in)
+			checkedAst, iss := e.Check(ast)
```
TristonianJones: No need to `Check` again, the `ast` value has all you need.

TristonianJones (changes requested):

cel/runtimecost_test.go
jpbetz: Ah, it's got to work with out access to the program API. I think I was able to figure it out.

interpreter/runtimecost.go
jpbetz: I'm okay with this. I'm going to leave a case explaining these mostly to document the cost expectations.

interpreter/runtimecost.go
jpbetz: SG. What's a test case for ConstantQualifier?

interpreter/runtimecost.go
jpbetz: I've broken them out as constants in a way that hopefully is reasonable.

cel/runtimecost_test.go
cici37: It would run into dependency loop if move under /interpreter

jpbetz: I've avoided the dep by restructuring the test

cel/options.go
jpbetz: Any combination of OptTrackCost, CostTracking and CostLimit can be used. If CostTracking or CostLimit is used (or both), OptTrackCost is automatically enabled.

interpreter/interpretable.go
jpbetz: Yes, this is true. I've dropped it.

interpreter/runtimecost.go
jpbetz: I follow the optimization. Thanks for explaining.  I've updated the cost to match (I don't track the index lookup cost-- it sounds like worst case becomes O(n) lengh t of the list, but I'm willing to wave my hands on that one right now).

interpreter/interpreter.go
```diff
@@ -38,22 +38,92 @@ type Interpreter interface {
...
 	}
-	return decObserveEval(observer)
 }
 
+// ExhaustiveEvalWrapper replaces operations that short-circuit with versions that evaluate
+// expressions and couples this behavior with the TrackState observer to provide
+// insight into the evaluation state of the entire expression. The EvalStateObserver must be
+// included in the underlying decorator. This decorator is not thread-safe, and the EvalState
+// must be reset between Eval() calls.
+func ExhaustiveEvalWrapper(underlying InterpretableDecorator) InterpretableDecorator {
```
TristonianJones: Since the observer signature is now exposed, I'd recommend just removing the `state` argument from `ExhaustiveEval` and moving the observer composition up to the `cel` package as you have. No need to deprecate and there may be other reasons for having exhaustive eval enabled besides state tracking.

```go
func ExhaustiveEval() { ... }
```

I think there are only 4 references to the existing signature.

cel/options.go
```diff
@@ -387,6 +391,28 @@ func EvalOptions(opts ...EvalOption) ProgramOption {
...
+}
+
+// CostLimit enables cost tracking and sets configures program evaluation to exit early with a
+// "runtime cost limit exceeded" error if the runtime cost exceeds the costLimit.
+// The CostLimit is a metric that corresponds to the number and estimated expense of operations
+// performed while evaluating an expression. It is indicative of CPU usage, not memory usage.
+func CostLimit(costLimit uint64) ProgramOption {
+	return func(p *prog) (*prog, error) {
+		p.costLimit = &costLimit
+		p.evalOpts |= OptTrackCost
```
TristonianJones, line 419: Is it an error to provide a CostLimit and not enable `CostTracking`? If so, I'd make sure that cost tracking is only enabled in one of the two functions and make sure that this option will have no effect unless CostTracking is also enabled.

interpreter/interpreter.go
```diff
@@ -38,22 +38,92 @@ type Interpreter interface {
...
+	CostLimitExceeded
+)
+
+// TODO: Replace all usages of TrackState with EvalStateObserver
+
 // TrackState decorates each expression node with an observer which records the value
 // associated with the given expression id. EvalState must be provided to the decorator.
 // This decorator is not thread-safe, and the EvalState must be reset between Eval()
 // calls.
+// DEPRECATED: Please use EvalStateObserver instead. It composes gracefully with additional observers.
```
TristonianJones, line 89: There are a few more calls to `TrackState` in various parts of the code. I'm okay with leaving this a TODO, but let's change the `ExhaustiveEval` option now since it's small and not at all related to observer correctness where effectively only one set of observers are supported since the decorators attempt to dedupe `watch` nodes which means that having separate decorators with observers won't work.

interpreter/interpreter.go
```diff
@@ -38,22 +38,92 @@ type Interpreter interface {
 		decorators ...InterpretableDecorator) (Interpretable, error)
 }
 
+// EvalObserver is a functional interface that accepts an expression id and an observed value.
+// The id identifies the expression that was evaluated, the programStep is the Interpretable or Qualifier that
+// was evaluated and value is the result of the evaluation.
+type EvalObserver func(id int64, programStep interface{}, value ref.Val)
+
+func Observe(observers ...EvalObserver) InterpretableDecorator {
```
TristonianJones, line 48: +godoc

TristonianJones (commented):
Almost there

cel/options.go
jpbetz: Any combination of CostLimit, CostTracking and OptTrackCost is valid and allowed. Just OptTrackCost will track cost using the default cost calculations, CostTracking allows the cost calculations to extended, and CostLimit enables a limit. Both CostTracking and CostLimit flip on OptTrackCost so that they behave as expected.

interpreter/interpreter.go
jpbetz: SG, I've applied the deprecation change for TrackState to the rest of the codebase.

interpreter/interpreter.go
jpbetz: Nice. This worked well.

interpreter/interpreter.go
```diff
@@ -38,32 +38,80 @@ type Interpreter interface {
...
+func (e EvalCancelledError) Error() string {
+	return e.Message
+}
+
+// CancellationCause enumerates the ways a program evaluation operation can be cancelled.
+type CancellationCause int
+
+const (
+	// ContextCancelled indicates that the operation was cancelled in response to a Golang context cancellation.
+	ContextCancelled CancellationCause = iota
```
TristonianJones, line 76: This is currently not used. Please do a follow up to plumb through the error or remove it from this context.

interpreter/interpreter.go
```diff
@@ -38,32 +38,80 @@ type Interpreter interface {
...
+// This decorator is not thread-safe, and the EvalState must be reset between Eval()
+// calls.
+func EvalStateObserver(state EvalState) EvalObserver {
+	return func(id int64, programStep interface{}, val ref.Val) {
 		state.SetValue(id, val)
 	}
-	return decObserveEval(observer)
 }
 
+// TODO: Replace all usages of ExhaustiveEval with ExhaustiveEvalWrapper
```
TristonianJones, line 104: I think you can remove this TODO since it's been replaced.

TristonianJones (approved):

PR #495, opened 2022-02-24
Add join to strings ext library

jpbetz:


ext/guards.go
```diff
@@ -246,3 +247,55 @@ func callInStrStrStrIntOutStr(fn func(string, string, string, int64) (string, er
...
+		vVal, ok := args1.(traits.Lister)
+		if !ok {
+			return types.MaybeNoSuchOverloadErr(args1)
+		}
+		strings := make([]string, vVal.Size().Value().(int64))
+		i := 0
+		for it := vVal.Iterator(); it.HasNext() == types.True; {
+			v, ok := it.Next().(types.String)
+			if !ok {
+				return types.MaybeNoSuchOverloadErr(v)
```
TristonianJones: You'll want to store a reference to the `it.Next()` result and pass that to the `MaybeNoSuchOverloadErr()` call here.

ext/guards.go
```diff
@@ -246,3 +247,55 @@ func callInStrStrStrIntOutStr(fn func(string, string, string, int64) (string, er
...
+		arg1Val, ok := args2.(types.String)
+		if !ok {
+			return types.MaybeNoSuchOverloadErr(args2)
+		}
+		strings := make([]string, vVal.Size().Value().(int64))
+		i := 0
+		for it := vVal.Iterator(); it.HasNext() == types.True; {
+			v, ok := it.Next().(types.String)
+			if !ok {
+				return types.MaybeNoSuchOverloadErr(v)
```
TristonianJones: Same comment here as above.

ext/strings.go
```diff
@@ -356,6 +377,11 @@ func (stringLib) ProgramOptions() []cel.ProgramOption {
 				Operator: "string_upper_ascii",
 				Unary:    callInStrOutStr(upperASCII),
 			},
+			&functions.Overload{
```
TristonianJones, line 383: You'll also want to create entries for the overload ids, similar to what's seen for `substring` relative to `string_substring_int` and `string_substring_int_int`

jpbetz: Good catch. I wasn't sure what the naming convention should be for a list. I went with: `list_join` for `<list<string>>.join()` and `list_join_string` for `<list<string>>.join(<string>)`.

jpbetz:
@TristonianJones thanks for the review! Feedback addressed.

ext/strings.go
TristonianJones: Usually I use the top-level types, so you picked what I would have here.

TristonianJones (approved):

TristonianJones:
You'll need to sync with HEAD to pick up a change that reduces flakiness in the context cancellation test. I made list append much faster in #490 and now the test is flaky.

TristonianJones:
Looks like you'll need to add a reference to the `traits` package to the `ext` BUILD.bazel

PR #497, opened 2022-02-26
Add option to compile regex constants at program creation time

jpbetz:
This provides a program option to inject an optimization (by way of a Interpretable factory) for any regex function.

A `matches` optimization is provided and enabled when `OptOptimize` is set.

unoptimized matches:

```
BenchmarkInterpreter/matches
BenchmarkInterpreter/matches-8  165296	      7135 ns/op	    5845 B/op	
```

optimized matches:
```
BenchmarkInterpreter/matches
BenchmarkInterpreter/matches-8  898917	      1413 ns/op	     210 B/op	
```

### DIFF
--- a/cel/cel_test.go
+++ b/cel/cel_test.go
@@ -20,6 +20,7 @@ import (
 	"io/ioutil"
 	"log"
 	"reflect"
+	"regexp"
 	"strings"
 	"sync"
 	"testing"
@@ -1603,5 +1604,112 @@ func TestDeclareContextProto(t *testing.T) {
 	_, iss := env.Compile(expression)
 	if iss.Err() != nil {
 		t.Fatalf("env.Compile(%s) failed: %s", expression, iss.Err())
+	}
+}
+
+type testRegexCall struct {
+	interpreter.InterpretableCall
+	compiledRegex *regexp.Regexp
+}
+
+func (r testRegexCall) Eval(ctx interpreter.Activation) ref.Val {
+	args := r.InterpretableCall.Args()
+	argVals := make([]ref.Val, len(args))
+	for i, arg := range args {
+		argVals[i] = arg.Eval(ctx)
+		if types.IsUnknownOrError(argVals[i]) {
+			return argVals[i]
+		}
+	}
+	return types.Bool(r.compiledRegex.MatchString(argVals[0].Value().(string)))
+}
+
+var testRegexOptimization = &interpreter.RegexOptimization{
+	Function:   "matches",
+	RegexIndex: 1,
+	Factory: func(call interpreter.InterpretableCall, regexIndex int, pattern ref.Val) (interpreter.Interpretable, error) {
+		compiledRegex, err := regexp.Compile(pattern.Value().(string))
+		if err != nil {
+			return nil, err
+		}
+		return &testRegexCall{call, compiledRegex}, nil
+	},
+}
+
+func TestRegexOptimizer(t *testing.T) {
+	var stringTests = []struct {
+		expr          string
+		optimizeRegex bool
+		progErr       string
+		err           string
+		parseOnly     bool
+	}{
+		{expr: `"123 abc 456".matches('[0-9]*')`},
+		{expr: `"123 abc 456".matches('[0-9]' + '*')`},
+		{expr: `"123 abc 456".matches('[0-9]*')`, optimizeRegex: true},
+		{expr: `"123 abc 456".matches('[0-9]' + '*')`, optimizeRegex: true},
+		{
+			expr: `"123 abc 456".matches(')[0-9]*')`, optimizeRegex: true,
+			progErr: "error parsing regexp: unexpected ): `)[0-9]*`",
+		},
+		{
+			expr: `"123 abc 456".matches(')[0-9]*')`,
+			err:  "error parsing regexp: unexpected ): `)[0-9]*`",
+		},
+	}
+
+	env, err := NewEnv()
+	if err != nil {
+		t.Fatal(err)
+	}
+	for i, tst := range stringTests {
+		tc := tst
+		t.Run(fmt.Sprintf("[%d]", i), func(tt *testing.T) {
+			var asts []*Ast
+			pAst, iss := env.Parse(tc.expr)
+			if iss.Err() != nil {
+				tt.Fatal(iss.Err())
+			}
+			asts = append(asts, pAst)
+			if !tc.parseOnly {
+				cAst, iss := env.Check(pAst)
+				if iss.Err() != nil {
+					tt.Fatal(iss.Err())
+				}
+				asts = append(asts, cAst)
+			}
+			for _, ast := range asts {
+				var opts []ProgramOption
+				if tc.optimizeRegex {
+					opts = append(opts, OptimizeRegex(testRegexOptimization))
+				}
+				prg, progErr := env.Program(ast, opts...)
+				if tc.progErr != "" {
+					if progErr == nil {
+						tt.Fatalf("wanted error %s for expr: %s", tc.progErr, tc.expr)
+					}
+					if tc.progErr != progErr.Error() {
+						tt.Errorf("got error %v, wanted error %s for expr: %s", progErr, tc.progErr, tc.expr)
+					}
+					continue
+				} else if progErr != nil {
+					tt.Fatal(progErr)
+				}
+				out, _, err := prg.Eval(NoVars())
+				if tc.err != "" {
+					if err == nil {
+						tt.Fatalf("got value %v, wanted error %s for expr: %s",
+							out.Value(), tc.err, tc.expr)
+					}
+					if tc.err != err.Error() {
+						tt.Errorf("got error %v, wanted error %s for expr: %s", err, tc.err, tc.expr)
+					}
+				} else if err != nil {
+					tt.Fatal(err)
+				} else if out.Value() != true {
+					tt.Errorf("got %v, wanted true for expr: %s", out.Value(), tc.expr)
+				}
+			}
+		})
 	}
 }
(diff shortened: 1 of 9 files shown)


jpbetz:
cc @TristonianJones @liggitt

jpbetz:
@TristonianJones I wasn't sure what API changes were best here. I'm open to suggestions alternatives to the `RegexPatternArg` functions.Override field and the `CompiledRegexStringConst` interface.  

Possible alternatives would be to have `RegexPatternArg` accept an enumeration of options instead of an `*int` (the regex pattern is always the 1st argument in all the usages I'm aware of).

Possible alternatives to `CompiledRegexStringConst` would be a trait.

interpreter/functions/functions.go
```diff
@@ -45,6 +45,17 @@ type Overload struct {
...
+	// RegexPatternArg optionally identifies one of the overload's arguments as being a regex pattern. RegexPatternArg
+	// identifies the argument by array index. E.g., 0 identifies the 1st argument.
+	// If the regex pattern at the call site is a constant value and OptCompileRegexConstants is enabled, the regex
+	// pattern will be pre-compiled and passed as a ref.Val implementing CompiledRegexStringConst to the overload
+	// implementation. If the regex pattern is a constant but OptCompileRegexConstants is disabled, a ref.Val
+	// implementing RegexStringConst will be passed in. If the regex pattern is not a constant, a normal ref.Val
+	// will be passed in. An overload with RegexPatternArg set MUST handle all these cases.
+	// If the argument implements CompiledRegexStringConst the compiled regex can be accessed by the overload function
+	// by calling CompiledRegex() on the CompiledRegexStringConst.
+	RegexPatternArg *int
```
TristonianJones: Instead of making this a first-class consideration of the function signature, I would prefer to have the decorator declare matching behaviors as part of its configuration

interpreter/decorators.go
```diff
@@ -118,6 +121,61 @@ func decOptimize() InterpretableDecorator {
 	}
 }
 
+// decCompileRegexConstants looks for regex pattern string constants, compiles any it finds and returns any
+// regex pattern compilation errors encountered.
+func decCompileRegexConstants() InterpretableDecorator {
```
TristonianJones: Consider instead a decorator which produces an alternative `InterpretableCall`:

```go
struct regexMatcher {
  function string
  regexIndex int
  factory func(call interpreter.InterpretableCall, regexIndex int, pattern ref.Val) (interepreter.Interpretable, error)
}

func regexOptimizer(matchers ...regexMatcher) interpreter.InterpretableDecorator {
       regexMatchMap := make(map[string]int)
       for _, m := range matchers {
          regexMatchMap[m.function] = regexMatcher
       }
       return func(i interpreter.Interpretable) (interpreter.Interpretable, error) {
            call, ok := i.(interpreter.InterpretableCall)
            if !ok {
              return i, nil
            }
            matcher, found := regexMatchMap[call.Function()]
            if !found || matcher.regexIndex >= len(call.Args()) {
              return i, nil
            }
            args := call.Args()
            regexArg := args[matcher.regexIndex]
            regexStr, isConst := regexArg.(interpreter.InterpretableConst)
            if !isConst {
              return i, nil
            }
            return matcher.factory(call, matcher.regexIndex, regexStr.Value())
        }
    }
}
```

I would then expect the specific optimizers to look something like the following:

```go
regexMatcher{
    function: "matches", 
    regexIndex: 0, 
    factory: func(call interpreter.InterpretableCall, regexIndex int, pattern ref.Val) (intepreter.Interpretable, error) {
        compiled, err := regexp.Compile(pattern)
	if err != nil {
	    return nil, fmt.Errorf("%v in '%s' function call", err, call.Function())
	}
        return interpreter.NewCall(call.ID(), call.Function(), call.OverloadID(), args[1:2], 
            func(params ...ref.Val) ref.Val {
                in, ok := params[0].(String)
                if !ok {
                  return types.MaybeNoSuchOverloadErr(params[0])
                }
                return types.Bool(compiled.Match(string(in)))
            }), nil
    },
}
```

I'm slightly oversimplifying, but not by much. The goal would be to minimize the top-level API changes. That said, you'd need to expose an `interpreter.NewInterpretableCall` (or similar) to produce a program plan step which would replace the original call and implementation.

It may not look very nice, but it lets you provide your own optimizations which should also be something you can enable without much impact on the exposed APIs

jpbetz: SGTM. I've restructured to match. This no longer provides any default optimizations, but there is an example of how to use it in the tests for "matches".

cel/cel_test.go
```diff
@@ -1857,3 +1858,110 @@ func TestDeclareContextProto(t *testing.T) {
 		t.Fatalf("env.Compile(%s) failed: %s", expression, iss.Err())
 	}
 }
+
+type testRegexCall struct {
```
TristonianJones: As a usability improvement, I'd suggest exposing a new call in the `interpretable.go` similar to `NewConstValue`, but for generating a new call node called, `NewCall` or `NewInterpretableCall` which under the covers instantiates an `interpretable.evalVarArgs{}` so that clients won't have to replicate the `IsErrorOrUnknown` checks.

cel/cel_test.go
```diff
@@ -1857,3 +1858,110 @@ func TestDeclareContextProto(t *testing.T) {
...
+	Factory: func(call interpreter.InterpretableCall, regexIndex int, pattern ref.Val) (interpreter.Interpretable, error) {
+		compiledRegex, err := regexp.Compile(pattern.Value().(string))
+		if err != nil {
+			return nil, err
+		}
+		return &testRegexCall{call, compiledRegex}, nil
+	},
+}
+
+func TestRegexOptimizer(t *testing.T) {
```
TristonianJones, line 1609: Consider adding a test to the `interpreter/interpreter_test.go` with a default implementation for `matches` that's enabled when the `OptOptimize` flag is set.

interpreter/interpreter.go
```diff
@@ -73,6 +73,22 @@ func Optimize() InterpretableDecorator {
...
 }
 
+// RegexOptimization provides a way to replace an InterpretableCall for a regex function if the
+// RegexIndex is a string constant argument. Typically, the Factory would compile the regex pattern at
+// RegexIndex and report any errors (at program creation time) and then use the compiled regex for
+// all regex function invocations.
+type RegexOptimization struct {
+	Function   string
+	RegexIndex int
+	Factory    func(call InterpretableCall, regexIndex int, pattern ref.Val) (Interpretable, error)
```
TristonianJones: Add a bit more information about the `Factory` contract and the relationship between `RegexIndex` and the `InterpretableCall.Args()` field.

cel/cel_test.go
jpbetz: That worked out to simplify things quite a bit. I've also simplified the `Factory` signature to make it easier to use as well.

cel/cel_test.go
jpbetz: Sounds good. I've tried to fix this together in a reasonable way.

interpreter/interpreter_test.go
```diff
@@ -1387,7 +1406,17 @@ func TestInterpreter(t *testing.T) {
 				"track":      TrackState(state),
 			}
 			for mode, opt := range opts {
-				prg, vars, err = program(t, &tc, opt)
+				opts := []InterpretableDecorator{opt}
+				if tc.extraOpts != nil {
```
TristonianJones, line 1410: Can you add the same block of code to the Benchmark as well?

interpreter/interpretable.go
```diff
@@ -533,6 +533,18 @@ type evalVarArgs struct {
 	impl     functions.FunctionOp
 }
 
+// NewCall creates a new call Interpretable.
+func NewCall(id int64, function, overload string, args []Interpretable, trait int, impl functions.FunctionOp) InterpretableCall {
```
TristonianJones: I'm debating whether to remove 'trait' from the argument set since that should only be for receiver functions and decorators that modify receiver functions seem error-prone. Thoughts?

interpreter/optimizations.go
```diff
@@ -0,0 +1,43 @@
...
+package interpreter
+
+import (
+	"regexp"
+
+	"github.com/google/cel-go/common/types"
+	"github.com/google/cel-go/common/types/ref"
+)
+
+var MatchesRegexOptimization = &RegexOptimization{
```
TristonianJones, line 27: +godoc

TristonianJones: You'll also need this in the BUILD.bazel file

interpreter/interpretable.go
jpbetz: I wanted to ditch it too, I just wasn't able to 100% convince myself it was okay. If you don't see any issues I'd love to remove it.

jpbetz: The alternative would be to provide a more use case specific function like `OverrideCallImpl(InterpretableCall, FunctionOp)`

edit: This might not work since there would be no way to guarantee the trait is copied over. I'll leave it as NewCall w/o trait.

interpreter/interpreter_test.go
jpbetz: Before:

```
BenchmarkInterpreter/matches
BenchmarkInterpreter/matches-8                                                      	  165296	      7135 ns/op	    5845 B/op	
```

After:
```
BenchmarkInterpreter/matches
BenchmarkInterpreter/matches-8                                                      	  898917	      1413 ns/op	     210 B/op	
```

interpreter/interpretable.go
TristonianJones: I think just removing `trait` is fine because ideally the functions provided from a decorator aren't relying on an intrinsic property of the first argument.

TristonianJones (approved):

PR #485, opened 2022-02-16
Improve test coverage for the 'cel' package

TristonianJones:
Minor improvements to code coverage for the `cel` package.

PR #498, opened 2022-03-01
Identify the issue in estimate cost ternary

cici37:
This PR is based on https://github.com/google/cel-go/pull/494.


TristonianJones:
@cici37 I've got a refactor of most of the cost tests into the `checker` package as part of #485. You might want to hold off on this change a bit. In general, I'd like to keep the tests with the package where the code is written so that the standard `go test` can easily generate coverage information.

cici37:
> Cici Huang I've got a refactor of most of the cost tests into the `checker` package as part of #485. You might want to hold off on this change a bit. In general, I'd like to keep the tests with the package where the code is written so that the standard `go test` can easily generate coverage information.

Thanks. Will rebase on the changes.

jpbetz:
Thanks @cici37! This has been incorporated into https://github.com/google/cel-go/pull/494.

PR #501, opened 2022-03-03
Adjust cost estimates to account for logical op short circuiting and …

jpbetz:
Cost model changes:

Ternery cost is now:  `cost(conditional) + {min: min(cost(ifTrue).min, cost(ifFalse).min), min: max(cost(ifTrue).max, cost(ifFalse).max)}`
Logical and, or is now: `{min: cost(lhs).min, max: cost(lhs).max + cost(rhs.max)}`

This bounds runtime costs much better.  It also reduces the computed estimated cost of comprehensions quite a bit lower since the fold uses logical operations, but this seems to be more accurate and justified.


jpbetz:
cc @cici37

TristonianJones (approved):

PR #502, opened 2022-03-05
Context Eval replacement for interruptable evaluation without coroutines

TristonianJones:
This change removes the goroutine-based approach to context cancellation and instead
supports early termination of comprehension loops at a configurable frequency.
Based on concepts from https://github.com/google/cel-go/pull/492

Additional changes:
* Fixes issue where unknown attribute code paths were not resolving the `PartialActivation`
from the `Activation` hierarchy. 
* Minor refactor of the `evalFold` to consolidate the various execution models within a single call.


jpbetz:
LGTM

PR #507, opened 2022-03-10
Add nil check and fix costTracker in contextEval 

cici37:
Return proper err for nil context when ContextEval be called.


cici37:
cc @jpbetz

TristonianJones (approved):

jpbetz:
LGTM

PR #517, opened 2022-03-22
Fix runtime cost ref.Val stack

jpbetz:
I discovered that the [stack](https://github.com/google/cel-go/blob/5e263d3f4e26e8fa248d52b2b02a41531ab89921/interpreter/runtimecost.go#L174) that I had introduced in https://github.com/google/cel-go/pull/494 as part of the introduction of runtime cost budgeting was not being managed correctly. Specifically, not all stack items pushed were being properly popped.

This PR fixes the stack by tracking the expression IDs associated with each stack value. The `pop` operations then specify the expression IDs of the stack items they wish to pop. If any spurious values are above the stack item being popped, they are removed from the stack as part of the `pop`.

This solves a bunch of problems and generally makes the stack much safer to use.

I have also improved the stack handling overall, which helps keep memory usage to a minimum. The only remaining case where I am certain that spurious stack items are left behind is the case where top-level variable lookup not being accounted for correctly.

### DIFF
--- a/interpreter/interpretable.go
+++ b/interpreter/interpretable.go
@@ -677,7 +677,15 @@ func (m *evalMap) Eval(ctx Activation) ref.Val {
 }
 
 func (m *evalMap) InitVals() []Interpretable {
-	return append(m.keys, m.vals...)
+	if len(m.keys) != len(m.vals) {
+		return nil
+	}
+	var result []Interpretable
+	for i, k := range m.keys {
+		v := m.vals[i]
+		result = append(result, k, v)
+	}
+	return result
 }
 
 func (m *evalMap) Type() ref.Type {
--- a/interpreter/interpretable.go
+++ b/interpreter/interpretable.go
@@ -680,10 +680,14 @@ func (m *evalMap) InitVals() []Interpretable {
 	if len(m.keys) != len(m.vals) {
 		return nil
 	}
-	var result []Interpretable
+	result := make([]Interpretable, len(m.keys)+len(m.vals))
+	idx := 0
 	for i, k := range m.keys {
 		v := m.vals[i]
-		result = append(result, k, v)
+		result[idx] = k
+		idx++
+		result[idx] = v
+		idx++
 	}
 	return result
 }
--- a/interpreter/runtimecost.go
+++ b/interpreter/runtimecost.go
@@ -72,11 +72,11 @@ func CostObserver(tracker *CostTracker) EvalObserver {
 		case Qualifier:
 			tracker.cost++
 		case InterpretableCall:
-			if argVals, ok := tracker.stack.popArgs(t.Args()); ok {
+			if argVals, ok := tracker.stack.dropArgs(t.Args()); ok {
 				tracker.cost += tracker.costCall(t, argVals, val)
 			}
 		case InterpretableConstructor:
-			tracker.stack.popArgs(t.InitVals())
+			tracker.stack.dropArgs(t.InitVals())
 			switch t.Type() {
 			case types.ListType:
 				tracker.cost += common.ListCreateBaseCost
@@ -199,7 +199,7 @@ func (s *refValStack) push(val ref.Val, id int64) {
 	*s = append(*s, value)
 }
 
-// TODO: Allowing drop and popArgs to remove stack items above the IDs they are provided is a workaround. drop and popArgs
+// TODO: Allowing drop and dropArgs to remove stack items above the IDs they are provided is a workaround. drop and dropArgs
 // should find and remove only the stack items matching the provided IDs once all attributes are properly pushed and popped from stack.
 
 // drop searches the stack for each ID and removes the ID and all stack items above it.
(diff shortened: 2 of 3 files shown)


interpreter/interpretable.go
```diff
@@ -677,7 +677,15 @@ func (m *evalMap) Eval(ctx Activation) ref.Val {
 }
 
 func (m *evalMap) InitVals() []Interpretable {
-	return append(m.keys, m.vals...)
+	if len(m.keys) != len(m.vals) {
+		return nil
+	}
+	var result []Interpretable
```
TristonianJones: You can also size this to be the same length as `result := make([]Interpretable, len(m.keys) + len(m.vals))` and then assign key and values to indices, but this is mostly cosmetic though it might slightly improve performance.

interpreter/runtimecost.go
```diff
@@ -170,19 +188,60 @@ func (c CostTracker) actualSize(value ref.Val) uint64 {
...
-func (s *refValStack) push(value ref.Val) {
+func (s *refValStack) push(value stackVal) {
 	*s = append(*s, value)
 }
 
-func (s *refValStack) pop(count int) ([]ref.Val, bool) {
-	if len(*s) < count {
-		return nil, false
+// drop1 removes the top item from the stack.
+func (s *refValStack) drop1() {
```
TristonianJones: If the end of the internal storage is considered the top, you can leave this as `pop`. I'm not sure why you wouldn't just use the IDs always though if you have them.

interpreter/runtimecost.go
```diff
@@ -170,19 +188,60 @@ func (c CostTracker) actualSize(value ref.Val) uint64 {
...
 	*s = (*s)[:idx]
-	return el, true
+}
+
+// TODO: Allowing drop and popArgs to remove stack items above the IDs they are provided is a workaround. drop and popArgs
+// should find and remove only the stack items matching the provided IDs once all attributes are properly pushed and popped from stack.
+
+// drop searches the stack for each ID and removes the ID and all stack items above it.
+// If none of the IDs are found, the stack is not modified.
+func (s *refValStack) drop(ids ...int64) {
```
TristonianJones, line 209: It's possible the ids are reported more than once with the way `Attribute` values are constructed and the way macros manipulate the AST. Macros should generally not do this, but there's no auditing to ensure they don't. It's worth updating the comment to indicate these are possible causes.

interpreter/interpretable.go
jpbetz: Can't hurt.

interpreter/runtimecost.go
jpbetz: Yeah, this section of code is not great.  This was the only remaining case where I didn't have easy access to an ID when popping.  I've decided to just leave the value on the stack and pop it, that way all pops are guaranteed safe.

interpreter/runtimecost.go
jpbetz: I've added a warning.I tried to reason through if this would have any problem implications and couldn't think of any.

jpbetz:
Feedback applied @TristonianJones

interpreter/runtimecost.go
```diff
@@ -41,26 +41,42 @@ func CostObserver(tracker *CostTracker) EvalObserver {
...
+				tracker.stack.drop(t.Attr().ID())
 				tracker.cost += common.SelectAndIdentCost
 			}
-		case *evalExhaustiveConditional, *evalOr, *evalAnd, *evalExhaustiveOr, *evalExhaustiveAnd:
+		case *evalExhaustiveConditional:
 			// Ternary has no direct cost. All cost is from the conditional and the true/false branch expressions.
+			tracker.stack.drop(t.attr.falsy.ID(), t.attr.truthy.ID(), t.attr.expr.ID())
+
+		// While the field names are identical, the boolean operation eval structs do not share an interface and so
+		// must be handled individually.
```
TristonianJones, line 61: I think these all share the `InterpretableCall` interface. Are you sure you need these explicit callouts for `evalOr`, `evalAnd`, etc.?

interpreter/runtimecost.go
```diff
@@ -170,19 +186,56 @@ func (c CostTracker) actualSize(value ref.Val) uint64 {
...
+	}
+}
+
+// popArgs searches the stack for all the args by their IDs, accumulates their associated ref.Vals and drops any
+// stack items above any of the arg IDs. If any of the IDs are not found the stack, false is returned.
+// Args are assumed to be found in the stack in reverse order, i.e. the last arg is expected to be found highest in
+// the stack.
+// WARNING: It is possible for multiple expressions with the same ID to exist (due to how macros are implemented) so it's
+// possible that a dropped ID will remain on the stack.  They should be removed when IDs on the stack are popped.
+func (s *refValStack) popArgs(args []Interpretable) ([]ref.Val, bool) {
```
TristonianJones: nit: consider calling this `dropArgs`

TristonianJones (changes requested):

interpreter/runtimecost.go
jpbetz: They're missing `Args()` (which is trivial to implement), `Function` and `Overload` receivers.

TristonianJones: Ah, I see. Then leaving this as-is is fine.

TristonianJones (approved):

PR #525, opened 2022-03-31
Add logic to allow functions to accept errors args

### LINKED ISSUE #523: Allow user definition of functions that avoid parameter error checking
**Feature request checklist**

**Change**

Allowing users to specify that this class of function in a limited way (only as global functions) would allow users to construct a form of error evasion/try evaluation on potentially failable expressions. This is important for expressions that have multiple parts where some can fail but an incompletely successful evaluation is still useful.

See https://groups.google.com/g/cel-go-discuss/c/YmQcarGB5V0/m/oS7qa1ppBgAJ for additional context.

**Example**

This would allow a user to define (bike-shed spelling) a "try" evaluation that lowers the significance of an error to a string message:

```
{
    "always_successful": safe_expression(par0),
    "sometimes_successful_1": try(less_safe_expression(par1)),
    "sometimes_successful_2": try(less_safe_expression(par2), "error message"),
}
```
which in the case that the evaluations of `less_safe_expression` failed might give a final result like:

```
{
    "always_successful": "yay!",
    "sometimes_successful_1": "par1 value evaluation failure result",
    "sometimes_successful_2": {"error message": "par2 value evaluation failure result"},
}
```
or on success
```
{
    "always_successful": "yay!",
    "sometimes_successful_1": "hurrah!",
    "sometimes_successful_2": "hurrah!",
}
```

**Alternatives considered**

I considered the option of adding a language operator that allows a similar outcome, but it becomes a language feature that is likely to be abused and would be less flexible.

I'm happy to do the work to implement this.

kortschak:
This adds additional logic to allow users to construct overloads that can accept errors as arguments.

Closes #523.

Please take a look.

### DIFF
--- a/interpreter/functions/functions.go
+++ b/interpreter/functions/functions.go
@@ -45,6 +45,9 @@ type Overload struct {
 	// Function defines the overload with a FunctionOp implementation. May be
 	// nil.
 	Function FunctionOp
+
+	// AllowError specifies whether the Overload will tolerate error arguments.
+	AllowError bool
 }
 
 // UnaryOp is a function that takes a single value and produces an output.
--- a/interpreter/functions/functions.go
+++ b/interpreter/functions/functions.go
@@ -46,8 +46,9 @@ type Overload struct {
 	// nil.
 	Function FunctionOp
 
-	// AllowError specifies whether the Overload will tolerate error arguments.
-	AllowError bool
+	// NonStrict specifies whether the Overload will tolerate arguments that
+	// are types.Err or types.Unknown.
+	NonStrict bool
 }
 
 // UnaryOp is a function that takes a single value and produces an output.
--- a/interpreter/interpretable.go
+++ b/interpreter/interpretable.go
@@ -421,12 +421,13 @@ func (zero *evalZeroArity) Args() []Interpretable {
 }
 
 type evalUnary struct {
-	id       int64
-	function string
-	overload string
-	arg      Interpretable
-	trait    int
-	impl     functions.UnaryOp
+	id         int64
+	function   string
+	overload   string
+	arg        Interpretable
+	trait      int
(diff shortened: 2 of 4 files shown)


interpreter/functions/functions.go
```diff
@@ -45,6 +45,9 @@ type Overload struct {
 	// Function defines the overload with a FunctionOp implementation. May be
 	// nil.
 	Function FunctionOp
+
+	// AllowError specifies whether the Overload will tolerate error arguments.
+	AllowError bool
```
TristonianJones: Let's prefer a boolean `NonStrict` flag here. For the use case you described I know that only permitting errors would be desirable, but from a CEL perspective functions are either strict or non-strict and the non-strict ones permit both error and unknown values as inputs. 

From an implementation perspective, it means that you could either return an `Unknown` value when it'd provided as input, or you can treat it as an error, or whatever you like. Let me know if this makes sense to you. Thanks for your patience on the review feedback.

kortschak: Yeah, that works for me — it's both simpler and richer.

interpreter/interpretable.go
```diff
@@ -439,14 +439,9 @@ func (un *evalUnary) ID() int64 {
...
 	// Early return if the argument to the function is unknown or error.
-	if un.allowError {
-		if types.IsUnknown(argVal) {
-			return argVal
-		}
-	} else {
-		if types.IsUnknownOrError(argVal) {
-			return argVal
-		}
+	strict := !un.nonStrict
```
kortschak, line 442: These assignments are for ease of reading; the SSA pass should resolve them to the same code as the reduced form.

kortschak:
PTAL

TristonianJones (approved):

PR #528, opened 2022-04-07
Prevent clobbering vars maps provided by previous libs

### LINKED ISSUE #527: Library-provided global vars are clobbered by later-loaded libraries with global vars
**Describe the bug**
Global vars are clobbered by later-loaded libraries if the later-loaded libraries have a `cel.Globals` program option.

**To Reproduce**

- [x]  parser
- [ ]  checker
- [ ]  interpreter

```cel
now
```

Test setup: (https://go.dev/play/p/dfLDqv6EcyX)
```go
package main

import (
	"fmt"
	"log"
	"time"

)

func main() {
	env, err := cel.NewEnv(
		libA(),

		// Comment out this line for the program to compile and run as
		// expected, or place it prior to the libA env option.
		libB(),
	)
	if err != nil {
		log.Fatalf("failed to create env: %v", err)
	}

	ast, iss := env.Compile(`now`)
	if iss.Err() != nil {
		log.Fatalf("failed compilation: %v", iss.Err())
	}

	prg, err := env.Program(ast)
	if err != nil {
		log.Fatalf("failed program instantiation: %v", err)
	}

	out, _, err := prg.Eval(interpreter.EmptyActivation())
	if err != nil {
		log.Fatalf("failed eval: %v", err)
	}
	fmt.Println(out)
}

func libA() cel.EnvOption {
	return cel.Lib(aLib{})
}

type aLib struct{}

func (aLib) CompileOptions() []cel.EnvOption {
	return []cel.EnvOption{
		cel.Declarations(
			decls.NewVar("now", decls.String),
		),
	}
}

func (aLib) ProgramOptions() []cel.ProgramOption {
	return []cel.ProgramOption{
		cel.Globals(map[string]interface{}{
			"now": func() interface{} { return time.Now().In(time.UTC) },
		}),
	}
}

func libB() cel.EnvOption {
	return cel.Lib(bLib{})
}

type bLib struct{}

func (bLib) CompileOptions() []cel.EnvOption {
	return []cel.EnvOption{
		cel.Declarations(
			decls.NewVar("yesterday", decls.String),
		),
	}
}

func (bLib) ProgramOptions() []cel.ProgramOption {
	return []cel.ProgramOption{
		cel.Globals(map[string]interface{}{
			"yesterday": func() interface{} { return time.Now().In(time.UTC).Add(-24 * time.Hour) },
		}),
	}
}
```

**Expected behavior**

The expression should be evaluated to the current time. This *is* the case when `libB()` is not executed, `libB()` is executed before `libA()` or `blib.ProgramOptions` does not have a `cel.Globals` program option.

**Actual behavior**

```
2022/04/07 16:21:46 failed eval: no such attribute: id: 1, names: [now]
```

kortschak:
Before this, when a lib is loaded with a Globals program option it would replace any existing global var maps, preventing the previously set maps from being available.

Happy to change docs wording if needed.

Closes #527.

Please take a look.

### DIFF
--- a/cel/cel_test.go
+++ b/cel/cel_test.go
@@ -667,9 +667,17 @@ func TestGlobalVars(t *testing.T) {
 			}})
 
 	// Global variables can be configured as a ProgramOption and optionally overridden on Eval.
-	prg, _ := e.Program(ast, funcs, Globals(map[string]interface{}{
-		"default": "third",
-	}))
+	// Add a previous globals map to confirm the order of shadowing and a final empty global
+	// map to show that globals are not clobbered.
+	prg, _ := e.Program(ast, funcs,
+		Globals(map[string]interface{}{
+			"default": "shadow me",
+		}),
+		Globals(map[string]interface{}{
+			"default": "third",
+		}),
+		Globals(map[string]interface{}{}),
+	)
 
 	t.Run("global_default", func(t *testing.T) {
 		vars := map[string]interface{}{
--- a/cel/options.go
+++ b/cel/options.go
@@ -342,7 +342,8 @@ func Functions(funcs ...*functions.Overload) ProgramOption {
 }
 
 // Globals sets the global variable values for a given program. These values may be shadowed by
-// variables with the same name provided to the Eval() call.
+// variables with the same name provided to the Eval() call. If Globals is used in a Library with
+// a Lib EnvOption, vars may shadow variables provided by previously added libraries.
 //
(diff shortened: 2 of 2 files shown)


kortschak:
I'm going to make this a draft for a bit. I'm working on something that would make use of this and I'm getting unexpected results, so I want to make sure it works before this goes forward.

...

All good now.

TristonianJones (approved):
Thanks for making the fix here!

PR #534, opened 2022-05-03
Set the error reporting limit to 100 for parity with C++

### LINKED ISSUE #531: OSS-Fuzz issue 39041
OSS-Fuzz has found a bug in this project. Please see https://oss-fuzz.com/testcase?key=6331542029336576 for details and reproducers.

This issue is mirrored from https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=39041 and will auto-close if the status changes there.

If you have trouble accessing this report, please file an issue at https://github.com/google/oss-fuzz/issues/new.

TristonianJones:
Error reporting was not capped meaning that a particularly errant input could generate a sizable output
from the parser. A simple cap of 100 error messages has been introduced which is at parity with C++.

Closes #531

### DIFF
--- a/common/errors.go
+++ b/common/errors.go
@@ -17,23 +17,32 @@ package common
 import (
 	"fmt"
 	"sort"
+	"strings"
 )
 
 // Errors type which contains a list of errors observed during parsing.
 type Errors struct {
-	errors []Error
-	source Source
+	errors            []Error
+	source            Source
+	numErrors         int
+	maxErrorsToReport int
 }
 
 // NewErrors creates a new instance of the Errors type.
 func NewErrors(source Source) *Errors {
 	return &Errors{
-		errors: []Error{},
-		source: source}
+		errors:            []Error{},
+		source:            source,
+		maxErrorsToReport: 100,
+	}
 }
 
 // ReportError records an error at a source location.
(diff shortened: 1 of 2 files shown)


common/errors.go
```diff
@@ -46,29 +56,40 @@ func (e *Errors) GetErrors() []Error {
...
+		numErrors:         e.numErrors + len(errs),
+		maxErrorsToReport: e.maxErrorsToReport,
 	}
 }
 
 // ToDisplayString returns the error set to a newline delimited string.
 func (e *Errors) ToDisplayString() string {
-	var result = ""
+	errorsTruncated := e.numErrors - e.maxErrorsToReport
+	errorsInString := int64(math.Min(float64(e.maxErrorsToReport), float64(e.numErrors)))
```
jcking: Well this is gross. Maybe just hand write it instead of casting to float and back?

jcking (changes requested):

common/errors.go
TristonianJones: Hah, yes it is. I've fixed it by just using if-else blocks instead.

TristonianJones:
PTAL

jcking (approved):

PR #539, opened 2022-05-12
Introduce a REPL for testing basic CEL expressions.

jnthntatum:
Initial implementation for issue/538. Adds a simple REPL CLI that
supports adding 'let' expressions to introduce variables into the
evaluation context and evaluate cel expressions using the default
environment.


repl/go.mod
```diff
@@ -0,0 +1,9 @@
+module github.com/google/cel-go/repl
+
+go 1.15
+
+require (
+	github.com/chzyer/readline v1.5.0
+	github.com/google/cel-go v0.11.3
+	google.golang.org/genproto v0.0.0-20220502173005-c8bf987b8c21
+)
```
TristonianJones, line 9: You can consider doing a local replace here as well: `replace github.com/google/cel-go => ./..`

This way you're making sure that the code is always building against the parent module.

repl/repl.go
```diff
@@ -0,0 +1,120 @@
...
+				fmt.Printf("Adding let failed:\n%v\n", err)
+			}
+		case "delete":
+			err = eval.DelLetVar(args[0])
+			if err != nil {
+				fmt.Printf("Deleting declaration failed:\n%v\n", err)
+			}
+		case "exit":
+			done = true
+		case "null":
```
TristonianJones: Any reason why null is special cased here?

repl/repl.go
```diff
@@ -0,0 +1,120 @@
...
+			if err != nil {
+				fmt.Printf("Adding let failed:\n%v\n", err)
+			}
+		case "delete":
+			err = eval.DelLetVar(args[0])
+			if err != nil {
+				fmt.Printf("Deleting declaration failed:\n%v\n", err)
+			}
+		case "exit":
+			done = true
```
TristonianJones: I think you can replace this with `break`

repl/evaluator_test.go
```diff
@@ -0,0 +1,122 @@
...
+package main
+
+import (
+	"testing"
+)
+
+func TestEvalSimple(t *testing.T) {
+	eval, err := NewEvaluator()
+	if err != nil {
+		panic(err)
```
TristonianJones: Prefer `t.Fatalf("NewEvaluator() failed: %v", err)` here and elsewhere

repl/evaluator_test.go
```diff
@@ -0,0 +1,122 @@
...
+	}
+
+	_, _, err = eval.Evaluate("[1, 2, 3]")
+
+	if err != nil {
+		t.Errorf("expected ok got: %v", err)
```
TristonianJones: For Go test assertions you typically state `got X, wanted Y` as the format, e.g.

```t.Errorf("eval.Evaluate() got %v, wanted non-error", err)```

File: repl/evaluator.go [hunk 6]
```diff
@@ -0,0 +1,221 @@
...
+func NewEvaluator() (*Evaluator, error) {
+	var e Evaluator
+	env, err := cel.NewEnv()
+	if err != nil {
+		return nil, err
+	}
+
+	e.env = env
+
+	return &e, nil
```
TristonianJones: Prefer instead writing this as :

```
return &Evaluator{env: env}
```

repl/commands_test.go
```diff
@@ -0,0 +1,118 @@
...
+		},
+		{
+			// not yet supported
+			commandLine: `%declare x: int64`,
+		},
+	}
+	for _, tc := range testCases {
+		_, _, _, err := Parse(tc.commandLine)
+		if err == nil {
+			t.Errorf("Expected errr for '%s' but got ok", tc.commandLine)
```
TristonianJones: nit: extra 'r' in 'err', and you'd want to write something closer to `t.Errorf("Parse(%s) failed: %v", tc.commandLine, err)`

TristonianJones (changes requested):

repl/go.mod
jnthntatum: done

repl/repl.go
jnthntatum: not really -- folded it into the default case

repl/repl.go
jnthntatum: done

repl/evaluator_test.go
jnthntatum: done

repl/evaluator_test.go
jnthntatum: done

repl/evaluator.go
jnthntatum: done

repl/commands_test.go
jnthntatum: done

TristonianJones (approved):
In a subsequent PR, would you mind adding a README.md?

PR #540, opened 2022-05-13
Fix the occurs check to prevent tagging types as substitutions for themselves

TristonianJones:
There are two locations where a type parameter can get accidentally tagged as a substitution for itself. The original code caught one location, and the code since https://github.com/google/cel-go/releases/tag/v0.11.1 caught a different one. 

This change catches the problematic tagging at both locations, thus avoiding a stack overflow when a type is marked
as a substitution of itself.

### DIFF
--- a/checker/types.go
+++ b/checker/types.go
@@ -238,9 +238,15 @@ func internalIsAssignable(m *mapping, t1 *exprpb.Type, t2 *exprpb.Type) bool {
 // - t2 has a type substitution (t2sub) assignable to t1
 // - t2 does not occur within t1.
 func isValidTypeSubstitution(m *mapping, t1, t2 *exprpb.Type) (valid, hasSub bool) {
+	// Early return if the t1 and t2 are the same instance.
+	kind1, kind2 := kindOf(t1), kindOf(t2)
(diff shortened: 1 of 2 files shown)


checker/types.go
```diff
@@ -238,9 +238,15 @@ func internalIsAssignable(m *mapping, t1 *exprpb.Type, t2 *exprpb.Type) bool {
...
+	kind1, kind2 := kindOf(t1), kindOf(t2)
+	if kind1 == kind2 && (t1 == t2 || proto.Equal(t1, t2)) {
+		return true, true
+	}
 	if t2Sub, found := m.find(t2); found {
-		kind1, kind2 := kindOf(t1), kindOf(t2)
-		if kind1 == kind2 && proto.Equal(t1, t2Sub) {
+		// Early return if t1 and t2Sub are the same instance as otherwise the mapping
+		// might mark a type as being a subtitution for itself.
+		if kind1 == kind2 && (t1 == t2Sub || proto.Equal(t1, t2Sub)) {
```
jcking: kind1 and kind2 are the same since the previous comparision above, so this expression is always true on the left side? Shouldn't kind2 be kindOf(t2Sub)?

TristonianJones: Done.

jcking (approved):

PR #544, opened 2022-05-20
Add basic print/parse support for types to REPL

### LINKED ISSUE #538: Add a basic Read Evaluate Print Loop Tool
**Feature request checklist**

**Change**
Introduce a simple REPL tool for testing expressions and updating an evaluation environment.

**Example**
Proposed use flow:
```
$ go run .
CEL REPL
%exit or EOF to quit.

cel-repl> %let x = 42
cel-repl> %let y = {'a': x, 'b': y}
Adding let failed:
Error updating y = {'a': x, 'b': y}
ERROR: <input>:1:15: undeclared reference to 'y' (in container '')
 | {'a': x, 'b': y}
 | ..............^
cel-repl> %let z = 41
cel-repl> %let y = {'a': x, 'b': z}
cel-repl> y.map(key, y[key]).filter(x, x > 41)
[42] (list_type:{elem_type:{primitive:INT64}})
```
**Alternatives considered**
Various playground tools have been implemented by clients that are tailored to their environment, but right now CEL doesn't provide a basic tool for exploring syntax and testing changes to the evaluation environment. Instead of providing a REPL tool, we could instead provide better guidance on debugging expressions in a given host environment. Providing a simple tool to learn syntax can help new clients understand how to use CEL so I think this is worth it (and doesn't preclude better debugging tooling in the future)

jnthntatum:
Add a simple grammar and pretty printer for the type mini-language:
e.g. map(int, list(string)).

Add basic type hinting for 'let' expressions and declaring (but not defining) variables:
e.g.

```
> %declare x : int
> %let y : string = 1
Adding let failed:
error updating y = 1
type mismatch got int expected string
> %let y = 10
> x < y
Expr failed:
no such attribute: id: 1, names: [x]
no such attribute: id: 1, names: [x] : bool
> x < y && y < 5
false : bool
```


repl/commands.go
```diff
@@ -23,8 +23,10 @@ import (
 var exprSubRe = `(?P<expr>.*)`
 var exprAssignRe = `(\s+=\s+)` + exprSubRe
 var identifier = `[_a-zA-Z][_a-zA-Z0-9]*`
+var typeSuffix = `(\s*:\s*(?P<type_hint>[^=]+))`
 var ws = regexp.MustCompile(`[ \t\r\n]+`)
-var letRe = regexp.MustCompile(`(?P<ident>` + identifier + `)(\s*:\s*(?P<type_hint>\w+))?` + exprAssignRe)
+var letRe = regexp.MustCompile(`(?P<ident>` + identifier + `)` + typeSuffix + `?` + exprAssignRe)
+var declRe = regexp.MustCompile(`(?P<ident>` + identifier + `)` + typeSuffix)
 var delRe = regexp.MustCompile(`(?P<ident>` + identifier + `)`)
```
TristonianJones: Would it make sense to use `delRe` as a common initial match against the input line? Maybe call it `cmdRe` since you can use it to determine the command name and variable id, then continue matching against the remainder of the line with the other regexes?

repl/commands_test.go
```diff
@@ -104,10 +110,6 @@ func TestParseErrors(t *testing.T) {
 			// no assignment
 			commandLine: "%let x: int64",
```
TristonianJones: Do you want to maybe use an invalid type name rather than one that looks close to right?

repl/evaluator.go
```diff
@@ -23,19 +23,22 @@ import (
 	"github.com/google/cel-go/checker/decls"
 	"github.com/google/cel-go/common/types/ref"
 	"github.com/google/cel-go/interpreter"
-	expr "google.golang.org/genproto/googleapis/api/expr/v1alpha1"
+
+	exprpb "google.golang.org/genproto/googleapis/api/expr/v1alpha1"
```
TristonianJones, line 27: Good catch!

repl/evaluator.go
```diff
@@ -174,29 +202,37 @@ func (e *Evaluator) DelLetVar(name string) error {
...
-
-		env = el.env
 	}
 
 	act, err := interpreter.NewActivation(vars)
 	if err != nil {
 		return nil, nil, err
 	}
 
+	var env *cel.Env
```
TristonianJones: The common idiom within go is to assign the default and then override:

```go
env := e.env
if len(e.ctx.letVars) > 0 {
  env = e.ctx.letVars[len(e.ctx.letVars)-1].env
}
```

repl/evaluator.go
```diff
@@ -174,29 +202,37 @@ func (e *Evaluator) DelLetVar(name string) error {
 }
 
 func (e *Evaluator) applyContext() (*cel.Env, interpreter.Activation, error) {
-	env := e.env
 	var vars = make(map[string]interface{})
 	for _, el := range e.ctx.letVars {
+		if el.prog == nil {
```
TristonianJones, line 217: Could you add a comment about why you ignore the let vars that haven't been initialized?

repl/evaluator_test.go
```diff
@@ -16,8 +16,18 @@ package main
 
 import (
 	"testing"
+
+	exprpb "google.golang.org/genproto/googleapis/api/expr/v1alpha1"
 )
 
+func MustParseType(t string) *exprpb.Type {
```
TristonianJones: No need to export this function since it's local to the test.

The preferred way to introduce helpers or asserts into tests is to make a test helper:

```go
func mustParseType(t testing.TB, typeName string) *exprpb.Type {
  t.Helper()
  ty, err := ParseType(typeName)
  if err != nil {
    t.Fatalf("ParseType(%s) failed: %v", typeName, err)
  }
  return ty
}
```

repl/parser/regen.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/sh
+
+ANTLR_JAR="/home/jdtatum/bin/antlr-4.10.1-complete.jar"
+antlr4="java -Xmx500M -cp \"$ANTLR_JAR:$CLASSPATH\" org.antlr.v4.Tool"
+
+$antlr4 \
+    -Dlanguage=Go \
+    -lib ../../parser/gen/ \
+    -visitor \
+    ./Types.g4 Types.g4
```
TristonianJones: I think you just need the one `./Types.g4` here. Also might want to make sure that `-package parser` is provided as a CLI arg.

repl/typefmt.go
```diff
@@ -0,0 +1,319 @@
...
+	case exprpb.Type_TIMESTAMP:
+		return "google.protobuf.Timestamp"
+	}
+	return "<unknown well-known type>"
+}
+
+// UnparseType pretty-prints a type for the REPL.
+//
+// TODO(issue/538): This is slightly different from core CEL's built-in formatter. Should
+// converge if possible.
```
TristonianJones, line 72: +1 That would be great to converge these two.

TristonianJones (changes requested):

repl/parser/regen.sh
jnthntatum: Done

repl/evaluator_test.go
jnthntatum: Done

repl/evaluator.go
jnthntatum: Done

repl/evaluator.go
jnthntatum: Done

repl/commands_test.go
jnthntatum: updated

repl/commands.go
jnthntatum: extracted the prefix part -- I'm going to swap to using antlr before making this anymore complicated in a follow up

repl/commands.go
```diff
@@ -84,16 +87,27 @@ func Parse(line string) (cmd string, args []string, expr string, err error) {
...
 			err = PromptError("invalid delete statement")
+			return
 		}
 		if i := subexpByName(delRe, m, "ident"); i != nil {
 			args = append(args, *i)
 		}
+	case "declare":
+		m := declRe.FindStringSubmatch(line)
+		if m == nil {
+			err = PromptError("invalid declare statement")
```
TristonianJones: Here and elsewhere: it's helpful to give a hint about a valid statement in addition to mentioning the current input is invalid.

repl/evaluator.go
```diff
@@ -122,22 +126,33 @@ func (e *Evaluator) updateContextPlans(ctx *EvaluationContext) error {
 	env := e.env
 	for i := range ctx.letVars {
 		el := &ctx.letVars[i]
-		if el.prog == nil {
+		if el.prog == nil && el.src != "" {
```
TristonianJones, line 134: A quick comment above the block to indicate what this condition indicates would be good. Does this condition relate in any way to the `applyContext()` condition used later on?

repl/evaluator_test.go
```diff
@@ -86,7 +97,24 @@ func TestLetError(t *testing.T) {
 		t.Fatalf("NewEvaluator() failed with: %v", err)
 	}
 
-	err = eval.AddLetVar("y", "z + 1")
+	err = eval.AddLetVar("y", "z + 1", nil)
+	if err == nil {
+		t.Errorf("eval.AddLetVar() got non-error, wanted error")
```
TristonianJones: Note the error here (and below) is formatted via `Errorf`, but does not provide any params. Did you mean to provide some additional state?

repl/evaluator_test.go
```diff
@@ -86,7 +97,24 @@ func TestLetError(t *testing.T) {
...
+	}
+
+	_, _, err = eval.Evaluate("y")
+	if err == nil {
+		t.Errorf("eval.Evaluate() got non-error, wanted error")
```
TristonianJones: You could capture `out` here and display what was returned in the error message.

TristonianJones (dismissed):

repl/commands.go
jnthntatum: Done

repl/evaluator_test.go
jnthntatum: done

repl/evaluator_test.go
jnthntatum: done

TristonianJones (approved):

PR #546, opened 2022-05-26
Add REPL support for "let" functions.

jnthntatum:
Add support for defining simple functions in the REPL.

Functions are defined with let commands e.g:

```
> %let myFun (x : int): int -> x + 41
> myFun(1)
42 : int
```

Current limitations:
- overloads aren't supported (identifier is used as the key for updating) 
- type agreement is based on proto equality instead of type check rules


repl/commands.go
```diff
@@ -16,117 +16,289 @@
 package main
 
 import (
+	"errors"
 	"fmt"
-	"regexp"
 	"strings"
+
+	"github.com/antlr/antlr4/runtime/Go/antlr"
+	"github.com/google/cel-go/repl/parser"
```
TristonianJones, line 23: nit: usually you group imports by domain/subdomain, so I think the following would be better style:

```
"github.com/google/cel-go/repl/parser"

"github.com/antlr/antlr4/runtime/Go/antlr"

```

repl/commands.go
```diff
@@ -16,117 +16,289 @@
...
+
+type simpleCmd struct {
+	cmd string
+}
+
+type evalCmd struct {
+	expr string
+}
+
+// Cmd interface
```
TristonianJones: Optional: For simple interfaces, a common naming convention is "<Noun>er", e.g. `Stringer` for converting an object to a `string`. Consider naming this `Commander`

Please fix: Indicate what the interface does.
```
<interface_name> provides an abstraction for accessing a CLI command name.
```

repl/commands.go
```diff
@@ -16,117 +16,289 @@
...
+type Cmd interface {
+	// simple name for the command for debugging if the client doesn't support a particular command kind
```
TristonianJones: nit: Comment style should be the `<MethodName> <verb> <description>`

repl/commands.go
TristonianJones: Optionally, consider embedding a `Cmd` object in your actions so the struct always has an exported `Cmd` name field.

repl/commands.go
```diff
@@ -16,117 +16,289 @@
...
+	case *letFnCmd:
+		cmd.src = expr
+	case *letVarCmd:
+		cmd.src = expr
+	default:
+		c.reportIssue(errors.New("unexpected CEL expression"))
+	}
+}
+
+func Parse(line string) (Cmd, error) {
```
TristonianJones: Don't forget to comment this method. You may want to group the other ANTLR method after the `Parse` with a general heading that indicates these are ANTLR listener methods.

repl/commands_test.go
```diff
@@ -15,86 +15,215 @@
...
+	t.Helper()
+
+	defer func() {
+		// catch type assertion errors
+		if v := recover(); v != nil {
+			result = false
+		}
+	}()
+
+	switch expected := expected.(type) {
```
TristonianJones: Consider separating the `expected` and `expected` variables here, e.g.:

```
switch want := expected.(type) {
   ...
}
```

repl/commands_test.go
```diff
@@ -15,86 +15,215 @@
...
+	case *evalCmd:
+		return got.(*evalCmd).expr == expected.expr
+	case *delCmd:
+		return got.(*delCmd).identifier == expected.identifier
+	case *simpleCmd:
+		return got.(*simpleCmd).cmd == expected.cmd
+	case *letFnCmd:
+		got := got.(*letFnCmd)
```
TristonianJones: Here and elsewhere, perhaps it would make the variable typing more clear if you were to change the assignment to `let := got.(*letFnCmd)` rather than to shadow the variable. When I was following along with the value types, I tracked the types wrong a couple times.

repl/commands_test.go
```diff
@@ -15,86 +15,215 @@
...
+		}
+		return true
+	case *letVarCmd:
+		got := got.(*letVarCmd)
+		return got.identifier == expected.identifier &&
+			proto.Equal(got.typeHint, expected.typeHint) &&
+			got.src == expected.src
 	}
-	return true
+
```
TristonianJones: super nit: extra blank line.

repl/commands_test.go
```diff
@@ -15,86 +15,215 @@
...
-		cmd, args, expr, err := Parse(tc.commandLine)
-		if err != nil {
-			t.Errorf("Parse(\"%s\") failed: %s", tc.commandLine, err)
-		}
-		if tc.wantCmd != cmd || !argsEqual(tc.wantArgs, args) || tc.wantExpr != expr {
-			t.Errorf("Parse('%s') got (%s, %v, %s) wanted (%s, %v, %s)", tc.commandLine,
-				cmd, args, expr, tc.wantCmd, tc.wantArgs, tc.wantExpr)
-		}
+		t.Run(tc.commandLine, func(t *testing.T) {
+			cmd, err := Parse(tc.commandLine)
```
TristonianJones, line 221: There's a common issue where not assigning the `tc` to a local variable in the inner loop will cause the tests to execute with a single `tc` instance value rather than each: https://jesseduffield.com/Golang-Gotcha-Of-The-Day/

TristonianJones (changes requested):

TristonianJones (approved):

PR #555, opened 2022-06-15
Add REPL support for loading protobuf descriptors

### LINKED ISSUE #538: Add a basic Read Evaluate Print Loop Tool

jnthntatum:
Adds the ability to load protobuf file descriptors from disk e.g.

`%load_descriptors "path/to/file_descriptor_set.textproto"`.

closes #538 -- with this the REPL tool has enough basic support for most use cases. We can address specific feature requests as they come up from use.

### DIFF
--- a/repl/evaluator.go
+++ b/repl/evaluator.go
@@ -18,6 +18,7 @@ package repl
 import (
 	"errors"
 	"fmt"
+	"io/ioutil"
 	"strings"
 
 	"github.com/google/cel-go/cel"
@@ -27,9 +28,11 @@ import (
 	"github.com/google/cel-go/interpreter"
 	"github.com/google/cel-go/interpreter/functions"
 
+	"google.golang.org/protobuf/encoding/prototext"
 	"google.golang.org/protobuf/proto"
 
 	exprpb "google.golang.org/genproto/googleapis/api/expr/v1alpha1"
+	descpb "google.golang.org/protobuf/types/descriptorpb"
 )
 
 // letVariable let variable representation
@@ -246,11 +249,20 @@ func (l *letVariable) clearPlan() {
 	l.prog = nil
 }
 
+// Optioner interface represents an option set on the base CEL environment used by
+// the evaluator.
+type Optioner interface {
+	// Option returns the cel.EnvOption that should be applied to the
+	// environment.
+	Option() cel.EnvOption
+}
+
 // EvaluationContext context for the repl.
 // Handles maintaining state for multiple let expressions.
 type EvaluationContext struct {
 	letVars []letVariable
 	letFns  []letFunction
+	options []Optioner
 }
 
 func (ctx *EvaluationContext) indexLetVar(name string) int {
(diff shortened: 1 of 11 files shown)


repl/evaluator.go
```diff
@@ -27,9 +28,11 @@ import (
 	"github.com/google/cel-go/interpreter"
 	"github.com/google/cel-go/interpreter/functions"
 
-	"google.golang.org/protobuf/proto"
-
 	exprpb "google.golang.org/genproto/googleapis/api/expr/v1alpha1"
+
+	"google.golang.org/protobuf/encoding/prototext"
+	"google.golang.org/protobuf/proto"
+	descpb "google.golang.org/protobuf/types/descriptorpb"
```
TristonianJones, line 35: nit: group the 'pb' suffixed imports together.

repl/evaluator.go
```diff
@@ -246,11 +249,22 @@ func (l *letVariable) clearPlan() {
 	l.prog = nil
 }
 
+// Optioner interface represents an option set on the base CEL environment used by
+// the evaluator.
+type Optioner interface {
+	fmt.Stringer // Implements Stringer for human readable representation.
```
TristonianJones: I don't think you need to embed the `Stringer` here since the `fmt` library checks to see if objects passed to `fmt.Sprintf` as a `%v` argument implement the `Stringer` interface.

repl/evaluator.go
```diff
@@ -586,15 +652,104 @@ func (e *Evaluator) applyContext() (*cel.Env, interpreter.Activation, error) {
...
+func (e *Evaluator) loadDescriptors(args []string) error {
+	if len(args) < 1 {
+		return errors.New("expected path for load descriptors")
+	}
+	flag := ""
+	if len(args) > 1 {
+		flag = args[0]
+	}
+	textfmt := true
+	if flag == "--textproto" {
```
TristonianJones: You can probably omit this flag check since the default will load the protos in `textFmt`. Instead just use the remaining `if flag == "--binarypb" {` to switch to reading binary format.

repl/evaluator_test.go
```diff
@@ -17,9 +17,12 @@ package repl
 import (
 	"testing"
 
+	"github.com/google/cel-go/cel"
 	exprpb "google.golang.org/genproto/googleapis/api/expr/v1alpha1"
```
TristonianJones, line 22: nit: leave a blank line between proto and non-proto imports.

TristonianJones (dismissed):

TristonianJones (approved):

Issue #493, opened 2022-02-23
Support non-overlapping overload dispatch using function


jpbetz:
cel-go allows for dispatch either by function name or overload ID. For example, given

```
cel.Declarations(
  decls.NewFunction("isSorted",
    decls.NewParameterizedInstanceOverload("list_a_is_sorted_bool",
      []*exprpb.Type{decls.NewListType(paramA)},
      decls.Bool, typeParamAList),
),
```

What works:

CEL will dispatch to `functions.Overload{Operator: "list_a_is_sorted_bool"...}` if it is declared in `ProgramOptions` otherwise it will dispatch to "isSorted".

What doesn't work:

If two `cel.Functions()` are registered that both declare "isSorted", CEL will report that the function name is already in use when the 2nd `Functions` are registered.

Proposal: If the two `Functions` declare non-overlapping overloads, allow them both to be registered using function name.

TristonianJones:
This has been updated for all extension functions, but has yet to be implemented for all of the standard operators, including type conversion methods which I think are essential for ensuring that non-standard types can be converted into standard ones.

PR #556, opened 2022-06-22
Dispatcher refactor to improve function declaration, definition, and extension

TristonianJones:
This change introduces a top-level `cel.Type` as well as functional options for configuring `cel.Variable` and `cel.Function` definitions using go-native APIs. 

This change is a shift away from the existing protobuf APIs which is also provides the following improvements:
- Automatic generation of function overload type-guards to provide basic invocation protections
- Automatic generation of dynamic dispatch functions for resolving overloads at runtime (for both parse-only expressions and dynamically typed data)
- Ability to declare and define functions within a single context.

The only drawback of this approach is that there is some minimal invocation overhead for functions declared/defined in this manner. Future optimizations may permit users to disable such type-guards if they deem them not to be necessary.

As a follow up PR, the `ext` function libraries and CEL standard operators will be shifted over to this new declaration style to improve the ease of extension of the core library implementations.

This is the basis for the feature requested in #493

### DIFF
--- a/cel/decls.go
+++ b/cel/decls.go
@@ -1,0 +1,1135 @@
+// Copyright 2022 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package cel
+
+import (
+	"fmt"
+	"strings"
+
+	"github.com/google/cel-go/checker/decls"
+	"github.com/google/cel-go/common/types"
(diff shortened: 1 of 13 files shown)


cel/decls.go
```diff
@@ -0,0 +1,1135 @@
...
+}
+
+// FunctionOpt defines a functional  option for configuring a function declaration.
+type FunctionOpt func(*functionDecl) (*functionDecl, error)
+
+// SingletonUnaryImpl creates a singleton function defintion to be used for all function overloads.
+//
+// Note, this approach works well if operand is expected to have a specific trait which it implements,
+// e.g. traits.ContainerType. Otherwise, prefer per-overload implementations.
+func SingletonUnaryImpl(fn functions.UnaryOp, traits ...int) FunctionOpt {
```
jcking: If its public, should it have `Impl` in the name?

jcking: Or is this to avoid collision?

TristonianJones: It's mostly just to avoid collision between `Function` as declaration and `Impl` as the binding. I could change it to `SingletonUnaryBinding` and so on if that sounds better.

jcking: `Binding` does sound less "internal details" then `Impl`, but I'll leave it up to you.

TristonianJones: Done.

jcking (approved):

PR #558, opened 2022-06-26
Support context propagation on overloads

### LINKED ISSUE #557: Support context propagation on overloads
**Feature request checklist**

**Change**

Currently, it is possible to use go contexts with the `ContextEval(context.Context, interface{})` method. However, this context is not usable by function overloads.

**Example**

```golang
fn := &functions.ContextOverload{
    Operator: "longRunningOperation",
    Unary: func(ctx context.Context, value ref.Val) {
        if err := longRunningOperation(ctx, value.Value()); err != nil {
            return types.Bool(false)
        }
        return types.Bool(true)
    }
}
```

nlachfr:
These changes introduce :
- a new interface `functions.Overloader` for keeping compatiblity with already defined `functions.Overload` structures
- a new structure `functions.ContextOverload` for defining context capable overloads
- the extension of the `interpreter.Activation` interface with `context.Context`
- a small test for checking that context cancellation can be catched with underlying overloads

The idea was to keep API compatibility while propagating the execution context to overload functions.

Resolves #557

### DIFF
--- a/cel/cel_test.go
+++ b/cel/cel_test.go
@@ -913,6 +913,58 @@ func TestContextEval(t *testing.T) {
 	}
 }
 
+func TestContextEvalPropagation(t *testing.T) {
+	env, err := NewEnv(
+		Declarations(
+			decls.NewFunction("sleep", decls.NewOverload(
+				"sleep", []*exprpb.Type{decls.Int}, decls.Null,
+			)),
+		),
+	)
+	if err != nil {
+		t.Fatalf("NewEnv() failed: %v", err)
+	}
+	ast, iss := env.Compile("sleep(20)")
+	if iss.Err() != nil {
+		t.Fatalf("env.Compile(expr) failed: %v", iss.Err())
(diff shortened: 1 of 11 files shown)


leaxoy:
This seems reasonable. The only question is, when can it be merged?

TristonianJones:
@leaxoy storing the Context as a member on a struct is an anti-pattern in Go, so I'd need to find another way to pass this information along. I had some thoughts about how to approach this with the async eval draft PR, but haven't had a chance to return to it.

nlachfr:
In #557, you mentioned the use of the callback channel directly. 

> Also, in Kubernetes it seems to be more common to store the callback channel.

Just propagating the done channel to function overloads would at least allow cancellation propagation (which was the initial objective of this MR), while removing the context embedding. We could then have a wrapper of the channel respecting `context.Context` interface for overloads.

I might be able to propose something for this if you're good with that @TristonianJones

PR #559, opened 2022-06-29
Support formatting options when unparsing ASTs to strings

### LINKED ISSUE #293: Support formatting options when unparsing ASTs to strings
**Feature request checklist**

**Change**
When unparsing ASTs, currently the tool will emit a single-line expression which is well-formatted, but maybe not the most kind on human eyes. Add options to support formatting rules, such as add newlines at `&&` and `||` when expressions are greater than 80 chars.

l46kok:
This change adds formatting capability to unparser by introducing new functional option`UnparseOption`. `Unparse` function will accept any number of `UnparseOption` augment the behavior of unparsing an AST.

This will allow support adding new lines for any desired non-unary operators when the length of a string exceeds the defined column limit.

Resolves #293

### DIFF
--- a/parser/unparser.go
+++ b/parser/unparser.go
@@ -37,9 +37,10 @@ import (
 // - Spacing around punctuation marks may be lost.
 // - Parentheses will only be applied when they affect operator precedence.
 //
-// This function optionally takes in one or more UnparserFormattingOption to dictate the formatting of the output.
-func Unparse(expr *exprpb.Expr, info *exprpb.SourceInfo, opts ...UnparserFormattingOption) (string, error) {
-	formattingOpts := &formattingOption{
+// This function optionally takes in one or more UnparserOption to alter the unparsing behavior, such as
+// performing word wrapping on operators.
+func Unparse(expr *exprpb.Expr, info *exprpb.SourceInfo, opts ...UnparserOption) (string, error) {
+	formattingOpts := &unparserOption{
 		wrapColumn: defaultWrapColumn,
 	}
 
@@ -66,7 +67,7 @@ func Unparse(expr *exprpb.Expr, info *exprpb.SourceInfo, opts ...UnparserFormatt
 type unparser struct {
 	str              strings.Builder
 	info             *exprpb.SourceInfo
-	options          *formattingOption
+	options          *unparserOption
 	lastWrappedIndex int
 }
 
@@ -472,7 +473,7 @@ func bytesToOctets(byteVal []byte) string {
 	return b.String()
 }
 
-// Takes in a function then inserts a newline based on the rules provided in formatting options
+// wrapForOperator inserts a newline after operators configured in the formatting options.
 func (un *unparser) wrapForOperator(fun string) bool {
 	_, wrapOperatorExists := un.options.operatorsToWrapOn[fun]
 	lineLength := un.str.Len() - un.lastWrappedIndex
@@ -489,18 +490,17 @@ var (
 	defaultWrapColumn = 80
 )
 
-// UnparserFormattingOption is a funcitonal option for configuring the output formatting
+// UnparserOption is a funcitonal option for configuring the output formatting
 // of the Unparse function.
-type UnparserFormattingOption func(*formattingOption) (*formattingOption, error)
+type UnparserOption func(*unparserOption) (*unparserOption, error)
 
-// Internal representation of the UnparserFormattingOption type
-type formattingOption struct {
+// Internal representation of the UnparserOption type
+type unparserOption struct {
 	wrapColumn        int // Defaults to 80
 	operatorsToWrapOn map[string]bool
 }
 
-// WrapOnColumn returns an UnparserFormattingOption with the column limit set.
-// Word wrapping will be performed when a line's length exceeds the limit
+// WrapOnColumn wraps the output expression when its string length exceeds a specified limit
 // for operators set by WrapOnOperators function.
 //
 // Example usage:
@@ -509,10 +509,14 @@ type formattingOption struct {
 //
 // This will insert a newline immediately after the logical AND operator for the below example input:
 //
-// Input: a && b
-// Output: a &&\nb
-func WrapColumn(col int) UnparserFormattingOption {
-	return func(opt *formattingOption) (*formattingOption, error) {
+// Input:
+// 'my-principal-group' in request.auth.claims && request.auth.claims.iat > now - duration('5m')
+//
+// Output:
+// 'my-principal-group' in request.auth.claims &&
+// request.auth.claims.iat > now - duration('5m')
+func WrapOnColumn(col int) UnparserOption {
+	return func(opt *unparserOption) (*unparserOption, error) {
 		if col < 1 {
 			return nil, fmt.Errorf("Invalid formatting option. Wrap column value must be greater than or equal to 1. Got %v instead", col)
 		}
@@ -525,8 +529,8 @@ func WrapColumn(col int) UnparserFormattingOption {
 // Word wrapping is supported on non-unary symbolic operators. Refer to operators.go for the full list
 //
 // This will replace any previously supplied operators instead of merging them.
-func WrapOnOperators(symbols ...string) UnparserFormattingOption {
-	return func(fmtOpt *formattingOption) (*formattingOption, error) {
+func WrapOnOperators(symbols ...string) UnparserOption {
+	return func(fmtOpt *unparserOption) (*unparserOption, error) {
 		fmtOpt.operatorsToWrapOn = make(map[string]bool)
 		for _, symbol := range symbols {
 			_, found := operators.FindReverse(symbol)
--- a/parser/unparser.go
+++ b/parser/unparser.go
@@ -43,6 +43,7 @@ func Unparse(expr *exprpb.Expr, info *exprpb.SourceInfo, opts ...UnparserOption)
 	formattingOpts := &unparserOption{
 		wrapColumn:           defaultWrapColumn,
 		wrapAfterColumnLimit: defaultWrapAfterColumnLimit,
+		operatorsToWrapOn:    defaultOperatorsToWrapOn,
 	}
 
 	var err error
@@ -500,9 +501,13 @@ func (un *unparser) writeOperatorWithWrapping(fun string, unmangled string) bool
 }
 
 // Defined defaults for the formatting options
-const (
+var (
 	defaultWrapColumn           = 80
 	defaultWrapAfterColumnLimit = true
+	defaultOperatorsToWrapOn    = map[string]bool{
+		operators.LogicalAnd: true,
+		operators.LogicalOr:  true,
+	}
 )
 
 // UnparserOption is a funcitonal option for configuring the output formatting
@@ -574,14 +579,15 @@ func WrapOnOperators(symbols ...string) UnparserOption {
 //
 //	Unparse(expr, sourceInfo, WrapColumn(40), WrapOnOperators(Operators.LogicalAnd), WrapAfterColumnLimit(false))
 //
-// This will insert a newline immediately before the logical AND operator for the below example input:
+// This will insert a newline immediately before the logical AND operator for the below example input, ensuring
+// that the length of a line never exceeds the specified column limit:
 //
 // Input:
 // 'my-principal-group' in request.auth.claims && request.auth.claims.iat > now - duration('5m')
 //
 // Output:
-// 'my-principal-group' in request.auth.claims &&
-// request.auth.claims.iat > now - duration('5m')
+// 'my-principal-group' in request.auth.claims
+// && request.auth.claims.iat > now - duration('5m')
 func WrapAfterColumnLimit(wrapAfter bool) UnparserOption {
 	return func(opt *unparserOption) (*unparserOption, error) {
 		opt.wrapAfterColumnLimit = wrapAfter
--- a/parser/unparser.go
+++ b/parser/unparser.go
@@ -38,7 +38,7 @@ import (
 // - Parentheses will only be applied when they affect operator precedence.
 //
 // This function optionally takes in one or more UnparserOption to alter the unparsing behavior, such as
-// performing word wrapping on operators.
+// performing word wrapping on expressions.
 func Unparse(expr *exprpb.Expr, info *exprpb.SourceInfo, opts ...UnparserOption) (string, error) {
 	formattingOpts := &unparserOption{
 		wrapColumn:           defaultWrapColumn,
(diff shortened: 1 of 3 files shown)


parser/unparser.go
```diff
@@ -37,7 +37,26 @@ import (
 // - Spacing around punctuation marks may be lost.
 // - Parentheses will only be applied when they affect operator precedence.
 func Unparse(expr *exprpb.Expr, info *exprpb.SourceInfo) (string, error) {
-	un := &unparser{info: info}
+	return UnparseFormatted(expr, info)
+}
+
+// TODO
+func UnparseFormatted(expr *exprpb.Expr, info *exprpb.SourceInfo, opts ...UnparserFormattingOption) (string, error) {
```
jnthntatum: Would it be reasonable to collapse the overloads?

parser/unparser_formatting_options.go
```diff
@@ -0,0 +1,52 @@
...
+	return func(opt *formattingOption) (*formattingOption, error) {
+		if col < 1 {
+			return nil, fmt.Errorf("Invalid formatting option. Wrap column value must be greater than or equal to 1. Got %v instead", col)
+		}
+		opt.wrapColumn = col
+		return opt, nil
+	}
+}
+
+// Support word wrapping on non-unary symbolic operators. Refer to operators.go for the full list
```
jnthntatum: nit: should read WrapOnOperators returns an UnparserFormattingOption...

parser/unparser_formatting_options.go
```diff
@@ -0,0 +1,52 @@
...
+)
+
+// Defined defaults for the formatting options
+var (
+	DefaultWrapColumn = 80
+)
+
+type UnparserFormattingOption func(*formattingOption) (*formattingOption, error)
+
+// Internal implementation of the UnparserFormattingOption interface
```
jnthntatum: internal representation? it's not implementing a go interface afaict

parser/unparser_formatting_options.go
```diff
@@ -0,0 +1,52 @@
...
+
+import (
+	"fmt"
+
+	"github.com/google/cel-go/common/operators"
+)
+
+// Defined defaults for the formatting options
+var (
+	DefaultWrapColumn = 80
```
jnthntatum: If you need to export this, make sure to document.

parser/unparser_formatting_options.go
```diff
@@ -0,0 +1,52 @@
+package parser
```
jnthntatum: Usually, it's good to define interfaces (or the functional option in this case) close to the API that consumes it -- unless you think this will grow to the point that it's a large component separate from the unparser (e.g. you would need to write tests specific to the apis in this file), you might consider just adding this to unparser.go

parser/unparser.go
l46kok: Yeah I think so. In this case, the change would be backwards compatible.

parser/unparser_formatting_options.go
l46kok: I'll move it to unparser. There's a possibility that it might grow, but I think YAGNI applies at this point. We can always pull it out into its own file later on.

parser/unparser_formatting_options.go
l46kok: Unexported the field

parser/unparser_formatting_options.go
l46kok: Done

parser/unparser_formatting_options.go
l46kok: Done

parser/unparser.go
```diff
@@ -444,3 +470,63 @@ func bytesToOctets(byteVal []byte) string {
...
+	}
+	return false
+}
+
+// Defined defaults for the formatting options
+var (
+	defaultWrapColumn = 80
+)
+
+type UnparserFormattingOption func(*formattingOption) (*formattingOption, error)
```
jnthntatum: exported type so you should document this (maybe very briefly)

parser/unparser_test.go
```diff
@@ -138,6 +140,182 @@ func TestUnparse(t *testing.T) {
...
+			name: "call_no_wrap_short_line",
+			in:   "a + b + c + d",
+			out:  "a + b + c + d",
+			formattingOptions: []UnparserFormattingOption{
+				WrapColumn(12),
+				WrapOnOperators(operators.Add),
+			},
+		},
+
+		// These expresisons will be formatted based on the formatting options provided
```
jnthntatum: nit -- typo

parser/unparser.go
```diff
@@ -444,3 +470,63 @@ func bytesToOctets(byteVal []byte) string {
...
+
+// Internal representation of the UnparserFormattingOption type
+type formattingOption struct {
+	wrapColumn        int // Defaults to 80
+	operatorsToWrapOn map[string]bool
+}
+
+// WrapOnColumn returns an UnparserFormattingOption with the column limit set.
+// Word wrapping will be performed when a line's length exceeds the limit
+// for operators set by WrapOnOperators function
```
jnthntatum: Can you provide a simple example in the documentation? This is a little confusing for me without looking at the expected behavior in tests.

parser/unparser.go
```diff
@@ -444,3 +470,63 @@ func bytesToOctets(byteVal []byte) string {
...
+		if col < 1 {
+			return nil, fmt.Errorf("Invalid formatting option. Wrap column value must be greater than or equal to 1. Got %v instead", col)
+		}
+		opt.wrapColumn = col
+		return opt, nil
+	}
+}
+
+// WrapOnOperators returns an UnparserFormattedOption with operators to perform word wrapping on.
+// Word wrapping is supported on non-unary symbolic operators. Refer to operators.go for the full list
```
jnthntatum, line 552: It looks like this replaces any previous option rather than merging. I don't think this is obvious so consider documenting it.

parser/unparser.go
```diff
@@ -36,8 +36,24 @@ import (
...
-	un := &unparser{info: info}
+//
+// This function optionally takes in one or more UnparserFormattingOption to dictate the formatting of the output.
+func Unparse(expr *exprpb.Expr, info *exprpb.SourceInfo, opts ...UnparserFormattingOption) (string, error) {
+	formattingOpts := &formattingOption{
+		wrapColumn: defaultWrapColumn,
+	}
+
+	for _, opt := range opts {
+		_, err := opt(formattingOpts)
```
jnthntatum: nit: consider explicitly updating formatting opts from the return value instead of assuming the argument is updated.

parser/unparser.go
l46kok: Done

parser/unparser_test.go
l46kok: Done

parser/unparser.go
l46kok: Done

parser/unparser.go
l46kok: Done

parser/unparser.go
l46kok: Done

parser/unparser.go
```diff
@@ -444,3 +471,76 @@ func bytesToOctets(byteVal []byte) string {
...
+}
+
+// Defined defaults for the formatting options
+var (
+	defaultWrapColumn = 80
+)
+
+// UnparserFormattingOption is a funcitonal option for configuring the output formatting
+// of the Unparse function.
+type UnparserFormattingOption func(*formattingOption) (*formattingOption, error)
```
TristonianJones: Consider renaming this to be more generic: `UnparserOption`. At the moment, the options are based on formatting, but at some point they might alter the `Unparse` behavior in another way. In terms of the signature, I would use an internal `unparserOptions` struct rather than `formattingOption`

parser/unparser.go
```diff
@@ -444,3 +471,76 @@ func bytesToOctets(byteVal []byte) string {
...
+
+// Internal representation of the UnparserFormattingOption type
+type formattingOption struct {
+	wrapColumn        int // Defaults to 80
+	operatorsToWrapOn map[string]bool
+}
+
+// WrapOnColumn returns an UnparserFormattingOption with the column limit set.
```
TristonianJones: nit: Consider renaming the function or updating the comment to match the function name, e.g. WrapColumn vs. WrapOnColumn.

parser/unparser.go
```diff
@@ -444,3 +471,76 @@ func bytesToOctets(byteVal []byte) string {
...
+// Word wrapping will be performed when a line's length exceeds the limit
```
TristonianJones: Consider updating the option comment to read more along the following lines:

`WrapOnColumn wraps the output expression when its string length exceeds a specified limit.`

In general you want `<Function> <verb>` as the format for comments.

parser/unparser.go
```diff
@@ -444,3 +471,76 @@ func bytesToOctets(byteVal []byte) string {
...
+// for operators set by WrapOnOperators function.
+//
+// Example usage:
+//
+//	Unparse(expr, sourceInfo, WrapColumn(3), WrapOnOperators(Operators.LogicalAnd))
```
TristonianJones: In this example is '3' meant to indicate number of characters? If so, you may want to consider a longer line limit to better convey what the behavior, e.g. `WrapColumn(40)` with an example expression like so:

 ```
'my-principal-group' in request.auth.claims &&\n
   request.auth.claims.iat > now - duration('5m')
```

Will the WrapColumn happen before or after the operator? Is this configurable? Does it need to be operator-specific, or can it just happen at a blank space after the column limit is reached?

File: parser/unparser.go [hunk 15]
```diff
@@ -444,3 +471,76 @@ func bytesToOctets(byteVal []byte) string {
...
+		un.lastWrappedIndex = un.str.Len()
+		un.str.WriteString("\n")
+		return true
+	}
+	return false
+}
+
+// Defined defaults for the formatting options
+var (
+	defaultWrapColumn = 80
```
TristonianJones: I think you can make this a constant, but I could be wrong.

parser/unparser.go
```diff
@@ -444,3 +471,76 @@ func bytesToOctets(byteVal []byte) string {
 	}
 	return b.String()
 }
+
+// Takes in a function then inserts a newline based on the rules provided in formatting options
```
TristonianJones: Even though this is a private function, use the same format for exported functions, e.g. `wrapForOperator inserts a newline after operators configured in the formatting options.`

parser/unparser_test.go
```diff
@@ -20,6 +20,7 @@ import (
 	"testing"
 
 	"github.com/google/cel-go/common"
+	"github.com/google/cel-go/common/operators"
 	"google.golang.org/protobuf/proto"
```
TristonianJones, line 25: nit: add a blank line between import directory groups.

parser/unparser_test.go
```diff
@@ -138,6 +140,191 @@ func TestUnparse(t *testing.T) {
...
+				WrapColumn(3),
+				WrapOnOperators(operators.Multiply),
+			},
+		},
+		{
+			name: "call_wrap_and_long_variables",
+			in:   "longVariableA && longVariableB && longVariableC",
+			out:  "longVariableA &&\nlongVariableB &&\nlongVariableC",
+			formattingOptions: []UnparserFormattingOption{
+				WrapColumn(3),
```
TristonianJones: Consider cases where the wrapColumn would not be applied, e.g. WrapColumn(10000) or WrapColumn(-1). Is there validation happening on the column value?

parser/unparser_test.go
```diff
@@ -138,6 +140,191 @@ func TestUnparse(t *testing.T) {
...
+				WrapOnOperators(operators.LogicalAnd),
```
TristonianJones, line 334: Showcase what happens when the `WrapOnOperators` is not specified. Wrapping on `&&` and `||` at 80+ chars might be a reasonable default behavior.

TristonianJones (changes requested):

parser/unparser_test.go
l46kok: The validation for invalid [column limit like -1 is done here](https://github.com/google/cel-go/pull/559/commits/604a4bd947dd3c724ca634f99d1e7a363fea1e74#diff-667600a8ae20f05557fb24c812a91e79c85833be4a084db2e1b05660d1eaf3c4R516-R520). The test triggers this [case at 0](https://github.com/google/cel-go/pull/559/commits/604a4bd947dd3c724ca634f99d1e7a363fea1e74#diff-2e0433ff11847f5508ec46127bc5ac5df75bd8001f0d324e67a01e722594c24eR445-R450). We could add another test for -1 for clarity if you'd like

l46kok: For `WrapColumn(10000)`, I thought about leaving it as is since this is technically a valid setting, albeit impractical. It might makes sense to set an upper limit that's reasonable for most screen sizes, say 1000.

parser/unparser.go
l46kok: Yeah 3 is the number of chars. Currently, WrapColumn happens after the operator but I think it'd be best to make it configurable. Thinking about this, it would make sense to support wrapping at a blank space since that's a pretty common use case.

parser/unparser_test.go
l46kok: [Per conversation here](https://github.com/google/cel-go/pull/559#discussion_r916352600), would supporting wrapping on blank spaces as default be a better route? It might be less surprising to the users.

parser/unparser.go
TristonianJones: In this case, I think I'm looking for an updated example. As a refinement, you could also support wrap at space as an option as well.

parser/unparser.go
TristonianJones: The comment and the method don't list the same name. Perhaps there's a commit that hasn't been pushed?

parser/unparser_test.go
TristonianJones: I mostly just thought it might be used to show that the line wrap doesn't happen at 1000 and also show that when the operator occurs _after_ the column specified how the code behaves.

parser/unparser_test.go
TristonianJones: It's an option, but one that sometimes leads to somewhat ugly / artificial line breaks. My main concern is what happens when you specify `WrapColumn` but don't specify any operators. Do you error, or pick default operators?

l46kok: I've went with default operators of `&&` and `||`. Let me know if this looks ok

parser/unparser_test.go
l46kok: Sounds good, added a test case for this

parser/unparser.go
l46kok: Updated the example. I'll make the wrap at space a separate PR to keep the diff small

parser/unparser.go
l46kok: This now include a map initialization for the operator defaults, so we have to leave it as `var`.

l46kok:
PTAL

TristonianJones (approved):

PR #560, opened 2022-06-30
Provide an option to ensure time operations are in UTC by default

### LINKED ISSUE #468: getHours() uses timezone of timestamp not UTC
According to the [CEL spec](https://github.com/google/cel-spec/blob/master/doc/langdef.md) `getHours()` should "get hours from the date in UTC, 0-23", but instead it uses the timezone of the timestamp object.

How to reproduce:

```
timestamp('2011-08-18T19:03:37.010+01:00').getHours() == timestamp('2011-08-18T19:03:37.010+01:00').getHours('UTC')
```

TristonianJones:
Currently, unary member functions on `timestamp` values in cel-go will use whatever the timezone is of the input object. However, this is not what is specified in the spec and can result in differences in evaluation behavior depending on where the call takes place. This fixes an issue with spec-conformance and should be enabled by default in future releases. To opt-in or opt-out see the usage below.

Usage: `cel.NewEnv(cel.DefaultUTCTimeZone(true))`

Closes #468 by ensuring that all time operations are in UTC by default.

### DIFF
--- a/cel/library.go
+++ b/cel/library.go
@@ -69,11 +69,10 @@ type stdLibrary struct{}
 
 // EnvOptions returns options for the standard CEL function declarations and macros.
 func (stdLibrary) CompileOptions() []EnvOption {
-	opts := []EnvOption{
+	return []EnvOption{
 		Declarations(checker.StandardDeclarations()...),
 		Macros(StandardMacros...),
 	}
-	return opts
 }
 
 // ProgramOptions returns function implementations for the standard CEL functions.
--- a/cel/library.go
+++ b/cel/library.go
@@ -303,7 +303,7 @@ func inTimeZone(ts, tz ref.Val) (time.Time, error) {
 	if err != nil {
 		return time.Time{}, err
 	}
-	min, err := strconv.Atoi(string(val[ind+1]))
+	min, err := strconv.Atoi(string(val[ind+1:]))
 	if err != nil {
 		return time.Time{}, err
 	}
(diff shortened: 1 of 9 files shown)


cel/library.go
```diff
@@ -75,3 +82,239 @@ func (stdLibrary) ProgramOptions() []ProgramOption {
...
+	if err != nil {
+		return time.Time{}, err
+	}
+	min, err := strconv.Atoi(string(val[ind+1]))
+	if err != nil {
+		return time.Time{}, err
+	}
+	var offset int
+	if string(val[0]) == "-" {
+		offset = hr*60 - min
```
jcking, line 312: Is this correct? Shouldn't both components be negative? -1:30 means -90 minutes, no?

TristonianJones: I think you're right. This bug has been here for a long time since I just copied it over from the `timestamp.go` object directly. I'll add tests and fix both locations.

TristonianJones: Actually, I checked the math. This works out. The `hr` is already negative but `min` is positive, so `hr * 60 - min`  for a timezone of `-01:30` would result in an offset of `-1 * 60 - 30` or `-90` as you would expect.

TristonianJones: Funnily enough, there was a different issue with the substring for the minutes offset, so the test caught a different issue. Thanks for asking me to double check here.

jcking (approved):

PR #564, opened 2022-07-12
Add an applied features map for one-time only features

TristonianJones:
Ensure one time only features are evaluated once per Env hierarchy, including within extensions created via the `Extend` call.


TristonianJones:
@cici37 This should fix the UTC fix enablement issue you were seeing.

cici37:
> Cici Huang This should fix the UTC fix enablement issue you were seeing.

Thank you for addressing this!

jcking (approved):

PR #565, opened 2022-07-13
Add the duration declarations in the new style when UTC is enabled

TristonianJones:
Since the UTC fix is isolated to timestamps only the timestamp overloads were provided in the new style using `Function`; however, this causes a problem with dynamic dispatch where the functions like `getHours` are split between timestamps and durations. The generated dynamic dispatch function was then receiving the call instead of the `Duration` object resulting in a `no such overload` error. 

This has been remedied by declaring duration functions that share the same function name with the timestamp functions using the new style when the UTC time fix is enabled.

### DIFF
--- a/cel/cel_test.go
+++ b/cel/cel_test.go
@@ -202,8 +202,10 @@ func TestCustomEnvError(t *testing.T) {
 }
 
 func TestCustomEnv(t *testing.T) {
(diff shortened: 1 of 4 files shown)


cici37:
Thanks for addressing this. Tested in k8s side and it works as expected

jcking (approved):

PR #570, opened 2022-08-03
Use proto safe traversal accessors

TristonianJones:
It seems that upcoming changes in the codegen for go proto messages seem to change some of the assumptions about safe-traversal. This change ensures that accessor methods are used which preserve the safe traversal promises of proto.


TristonianJones:
@jcking This seems to be why I was seeing different behavior between the public repo and the private mirror.

jcking (approved):

PR #571, opened 2022-08-10
Fix nested CallExpr cost estimation

DangerOnTheRanger:
This PR fixes instances where cost estimation is used on expressions where a `CallExpr` is nested inside of another `CallExpr` and `math.MaxUint64` is incorrectly returned. The test case included with this PR shows an example of where this can happen; `==` is treated as a function call/`CallExpr`, and will call `sizeEstimate` [on its left and right-hand expressions](https://github.com/google/cel-go/blob/25bb4c6da1759f0104ce8ee935431d32b71977b6/checker/cost.go#L557-L558). sizeEstimate will then try to [call AstNode.ComputedSize before trying EstimateSize and finally returning math.MaxUint64 for an upper limit](https://github.com/google/cel-go/blob/25bb4c6da1759f0104ce8ee935431d32b71977b6/checker/cost.go#L472-L480). 

That constant will get returned for operator-based `CallExpr`s inside `CallExpr`s whenever the inner one lacks an `EstimateSize`, such as with `list.size()`. Instead, this PR checks for `CallExpr` inside `SizeEstimate` and returns a min/max cost of 1.

The original issue was first reported in https://github.com/kubernetes/kubernetes/issues/111769.

### DIFF
--- a/checker/cost.go
+++ b/checker/cost.go
@@ -480,6 +480,9 @@ func (c *coster) sizeEstimate(t AstNode) SizeEstimate {
 	// lengths, since strings/bytes/more complex objects could be of
 	// variable length
 	if isScalar(t.Type()) {
+		// TODO: since the logic for size estimation is split between
+		// ComputedSize and isScalar, changing one will likely require changing
+		// the other, so they should be merged in the future if possible
 		return SizeEstimate{Min: 1, Max: 1}
 	}
 	return SizeEstimate{Min: 0, Max: math.MaxUint64}
--- a/checker/cost.go
+++ b/checker/cost.go
@@ -606,6 +606,9 @@ func (c *coster) newAstNode(e *exprpb.Expr) *astNode {
 	return &astNode{path: path, t: c.getType(e), expr: e, derivedSize: derivedSize}
 }
 
+// isScalar returns true if the given type is known to be of a constant size at
+// compile time. isScalar will return false for strings (they are variable-width)
+// in addition to protobuf.Any and protobuf.Value (their size is not knowable at compile time).
 func isScalar(t *exprpb.Type) bool {
 	switch kindOf(t) {
 	case kindPrimitive:
--- a/checker/cost.go
+++ b/checker/cost.go
@@ -476,6 +476,11 @@ func (c *coster) sizeEstimate(t AstNode) SizeEstimate {
 	if l := c.estimator.EstimateSize(t); l != nil {
 		return *l
 	}
+	// return a constant size if we're dealing with a CallExpr and EstimateSize
+	// came up empty; this code is reached when a CallExpr is nested inside
+	// another CallExpr and doesn't have an estimator that can handle it, such
+	// as `(1 == 2) == (3 == 4)` - without this block, that expression would
+	// return SizeEstimate{Min: 0, Max: math.MaxUint64}
 	if _, ok := t.Expr().ExprKind.(*exprpb.Expr_CallExpr); ok {
 		return SizeEstimate{Min: 1, Max: 1}
 	}
--- a/checker/cost.go
+++ b/checker/cost.go
@@ -482,7 +482,14 @@ func (c *coster) sizeEstimate(t AstNode) SizeEstimate {
 	// as `(1 == 2) == (3 == 4)` - without this block, that expression would
 	// return SizeEstimate{Min: 0, Max: math.MaxUint64}
 	if _, ok := t.Expr().ExprKind.(*exprpb.Expr_CallExpr); ok {
-		return SizeEstimate{Min: 1, Max: 1}
+		// ensure we only return an estimate of 1 for return types of set
+		// lengths, since strings/bytes/more complex objects could be of
+		// variable length
+		if kindOf(t.Type()) == kindPrimitive {
+			if t.Type().GetPrimitive() != exprpb.Type_STRING && t.Type().GetPrimitive() != exprpb.Type_BYTES {
+				return SizeEstimate{Min: 1, Max: 1}
+			}
+		}
 	}
 	return SizeEstimate{Min: 0, Max: math.MaxUint64}
 }
--- a/checker/cost.go
+++ b/checker/cost.go
@@ -619,19 +619,6 @@ func isScalar(t *exprpb.Type) bool {
 		if t.GetWellKnown() == exprpb.Type_DURATION || t.GetWellKnown() == exprpb.Type_TIMESTAMP {
 			return true
 		}
-	case kindObject:
-		switch t.GetMessageType() {
-		case "google.protobuf.Duration", "google.protobuf.Timestamp",
-			"google.protobuf.BoolValue", "google.protobuf.BytesValue",
-			"google.protobuf.DoubleValue", "google.protobuf.FloatValue",
-			"google.protobuf.Int32Value", "google.protobuf.Int64Value",
-			"google.protobuf.UInt32Value", "google.protobuf.UInt64Value":
-			return true
-		default:
-			return false
-		}
-	default:
-		return false
 	}
 	return false
 }
--- a/checker/cost.go
+++ b/checker/cost.go
@@ -485,10 +485,8 @@ func (c *coster) sizeEstimate(t AstNode) SizeEstimate {
 		// ensure we only return an estimate of 1 for return types of set
 		// lengths, since strings/bytes/more complex objects could be of
 		// variable length
-		if kindOf(t.Type()) == kindPrimitive {
-			if t.Type().GetPrimitive() != exprpb.Type_STRING && t.Type().GetPrimitive() != exprpb.Type_BYTES {
-				return SizeEstimate{Min: 1, Max: 1}
-			}
+		if isScalar(t.Type()) {
+			return SizeEstimate{Min: 1, Max: 1}
 		}
 	}
 	return SizeEstimate{Min: 0, Max: math.MaxUint64}
@@ -613,4 +611,13 @@ func (c *coster) newAstNode(e *exprpb.Expr) *astNode {
 		derivedSize = &size
 	}
 	return &astNode{path: path, t: c.getType(e), expr: e, derivedSize: derivedSize}
+}
+
+func isScalar(t *exprpb.Type) bool {
+	if kindOf(t) == kindPrimitive {
+		if t.GetPrimitive() != exprpb.Type_STRING && t.GetPrimitive() != exprpb.Type_BYTES {
+			return true
+		}
+	}
+	return false
 }
--- a/checker/cost.go
+++ b/checker/cost.go
@@ -476,18 +476,11 @@ func (c *coster) sizeEstimate(t AstNode) SizeEstimate {
 	if l := c.estimator.EstimateSize(t); l != nil {
 		return *l
 	}
-	// return a constant size if we're dealing with a CallExpr and EstimateSize
-	// came up empty; this code is reached when a CallExpr is nested inside
-	// another CallExpr and doesn't have an estimator that can handle it, such
-	// as `(1 == 2) == (3 == 4)` - without this block, that expression would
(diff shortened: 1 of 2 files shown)


checker/cost.go
```diff
@@ -476,6 +476,9 @@ func (c *coster) sizeEstimate(t AstNode) SizeEstimate {
 	if l := c.estimator.EstimateSize(t); l != nil {
 		return *l
 	}
+	if _, ok := t.Expr().ExprKind.(*exprpb.Expr_CallExpr); ok {
```
jpbetz: Add a comment explaining why this is correct?

checker/cost_test.go
```diff
@@ -354,6 +354,15 @@ func TestCost(t *testing.T) {
 			hints:  map[string]int64{"str1": 10, "str2": 10},
 			wanted: CostEstimate{Min: 2, Max: 6},
 		},
+		{
+			name: "list size comparison",
+			expr: `list1.size() == list2.size()`,
```
jpbetz, line 359: Can you expand the test coverage.  E.g. `x > y ? list1.size() : list2.size()`, and some functions that return a scalar other than `size()`?  Are there any other edge cases?

jpbetz (changes requested):

jpbetz:
Please add a link in the description to k8s issue where this was discovered.

jpbetz:
@TristonianJones I'd like to backport the fix to Kubernetes 1.24 and 1.23 if possible.

Once this is ready, would it be possible to backport to cel-go v0.10 and v0.9? These are the CEL versions those Kubernetes releases depend on (https://github.com/kubernetes/kubernetes/blob/release-1.24/go.mod#L267, https://github.com/kubernetes/kubernetes/blob/release-1.23/go.mod#L272)

checker/cost.go
alexzielenski: It appears to me that change causes the default size cost for all unhandled functions (not just built-in functions) to be `1`. If true I think it is potentially risky.

Consider the case of kubernetes or another developer adding a new function to their custom CEL library. If the developer forgets to implement the new case in their `EstimateSize` function, CEL will now cost their function at `1` instead of `MaxUint64` as before.

I think it might be safer to handle the builtins specifically, and to do so before calling `c.estimator.EstimateSize`, so that it is clear that unimplemented functions have large default cost.

jpbetz: A potential alternative would be to check the return type of the call expr here and set it to 1 if the result type is a scalar and MaxInt if it's some variable width type (map, list, string, types).

TristonianJones: In the worst-case, returning a default value of 1 would be an underestimate. The use of a max value was intended to signal that a function didn't have a cost associated. It seems like associating costs with the builtin operators would also be useful since even equality could require iteration over all list or map entries to determine whether the values are equal and that's certainly not a cost of 1.

DangerOnTheRanger: Gotcha, fixed

TristonianJones:
@jpbetz I'll cherry pick the squashed commit back into the previous releases with minor version bumps, e.g. v0.11.5, v0.10.2, v0.9.1. Sound good?

checker/cost.go
jpbetz: That's true. Fortunately the code we're working with here is estimating the (min, max) size of the result, not the cost of the function, so 1 is a safe size for a function returning a scalar.

That said, I agree that we're not estimating the cost of all functions right. We default to assuming cost 1 of functions unless we write a cost handler for them... Maybe better if we handle that as a separate issue though.

jpbetz:
> Joe Betz I'll cherry pick the squashed commit back into the previous releases with minor version bumps, e.g. v0.11.5, v0.10.2, v0.9.1. Sound good?

Thank you so much. I realize this is tedious.

checker/cost_test.go
DangerOnTheRanger: Marking this as resolved since coverage on overloaded operators as handled in `functionCost` should now be pretty good (outside of timestamps/durations as per offline discussion on some issues with those - I've created https://github.com/google/cel-go/issues/573 to track resolving that).

checker/cost_test.go
```diff
@@ -354,6 +354,40 @@ func TestCost(t *testing.T) {
...
+			decls: []*exprpb.Decl{
+				decls.NewVar("x", decls.Int),
+				decls.NewVar("y", decls.Int),
+				decls.NewVar("list1", decls.NewListType(decls.Int)),
+				decls.NewVar("list2", decls.NewListType(decls.Int)),
+			},
+			wanted: CostEstimate{Min: 5, Max: 5},
+		},
+		{
+			name: "str endsWith inequality",
```
jpbetz: why "inequality"?  It's a comparison but I don't see anything in the test that is verifying inequality

checker/cost.go
DangerOnTheRanger: Marking this as resolved now that the return type is checked.

checker/cost_test.go
DangerOnTheRanger: Whoops, you're right. I changed it to "equality" instead.

checker/cost.go
```diff
@@ -476,6 +476,21 @@ func (c *coster) sizeEstimate(t AstNode) SizeEstimate {
...
+	// came up empty; this code is reached when a CallExpr is nested inside
+	// another CallExpr and doesn't have an estimator that can handle it, such
+	// as `(1 == 2) == (3 == 4)` - without this block, that expression would
+	// return SizeEstimate{Min: 0, Max: math.MaxUint64}
+	if _, ok := t.Expr().ExprKind.(*exprpb.Expr_CallExpr); ok {
+		// ensure we only return an estimate of 1 for return types of set
+		// lengths, since strings/bytes/more complex objects could be of
+		// variable length
+		if kindOf(t.Type()) == kindPrimitive {
+			if t.Type().GetPrimitive() != exprpb.Type_STRING && t.Type().GetPrimitive() != exprpb.Type_BYTES {
```
jpbetz: suggestion only: put this into a small utility function?

DangerOnTheRanger: Makes sense - I've refactored that code into `isScalar`.

checker/cost.go
```diff
@@ -599,3 +612,12 @@ func (c *coster) newAstNode(e *exprpb.Expr) *astNode {
 	}
 	return &astNode{path: path, t: c.getType(e), expr: e, derivedSize: derivedSize}
 }
+
+func isScalar(t *exprpb.Type) bool {
+	if kindOf(t) == kindPrimitive {
+		if t.GetPrimitive() != exprpb.Type_STRING && t.GetPrimitive() != exprpb.Type_BYTES {
```
jpbetz, line 618: Should kind *exprpb.Type_WellKnown also be handled for the exprpb.Type_DURATION and exprpb.Type_TIMESTAMP cases somewhere in the return type checking?

jpbetz: I also see *exprpb.Type_MessageType has "google.protobuf.Duration" and "google.protobuf.Timestamp" cases. I don't know if wellknown or message checks would be needed here (or both? or neither?).  Do we need tests for these types?

TristonianJones: Within the `exprpb.Type`, if any of the following fields are set, it's probably reasonable to consider the type to have variable size:

- `message_type`* - you might want to check to see if the message_type is a well-known `google.protobuf` type name, such as `google.protobuf.Int64Value`, `google.protobuf.Duration`, etc.
- `abstract_type` - custom type whose contents we cannot inspect without accessor methods.
- `well_known`* - only the `ANY` type should be consider an object.
- `list_type` 
- `map_type`

I think that's the bulk of the type set that can be considered variable-size.

DangerOnTheRanger: I've changed `isScalar` to now handle other types besides `kindPrimitive`, so #573 can be closed if this PR merges/I'm marking this as resolved. On a sidenote, I came across [ExprTypeToType](https://github.com/google/cel-go/blob/25bb4c6da1759f0104ce8ee935431d32b71977b6/cel/decls.go#L993) today, which could simplify `isScalar` to something like:
```go
func isScalar(t *Type) bool {
    celType, err := ExprTypeToType(t)
    if err != nil {
        return false
    }
    _, hasIndexerTrait := celType.(traits.Indexer)
    return !hasIndexerTrait
}
```
Unfortunately importing it results in a circular import, and I'm not sure I see an easy way out (make `isScalar` public and move it to `cel/decls`? Move `ExprTypeToType`?), but I thought I'd put it out there just in case.

TristonianJones: I'll end up moving the implementation of `ExprTypeToType` down into common, but it's part of a much larger refactor. For now, I'd stick to a deliberate test against the `exprpb.Type` oneof

checker/cost.go
```diff
@@ -599,3 +605,30 @@ func (c *coster) newAstNode(e *exprpb.Expr) *astNode {
...
+		if t.GetWellKnown() == exprpb.Type_DURATION || t.GetWellKnown() == exprpb.Type_TIMESTAMP {
+			return true
+		}
+	case kindObject:
+		switch t.GetMessageType() {
+		case "google.protobuf.Duration", "google.protobuf.Timestamp",
+			"google.protobuf.BoolValue", "google.protobuf.BytesValue",
+			"google.protobuf.DoubleValue", "google.protobuf.FloatValue",
+			"google.protobuf.Int32Value", "google.protobuf.Int64Value",
+			"google.protobuf.UInt32Value", "google.protobuf.UInt64Value":
```
jpbetz: Do we have the test case for when this can happen?

checker/cost.go
```diff
@@ -599,3 +605,30 @@ func (c *coster) newAstNode(e *exprpb.Expr) *astNode {
...
+	switch kindOf(t) {
+	case kindPrimitive:
+		if t.GetPrimitive() != exprpb.Type_STRING && t.GetPrimitive() != exprpb.Type_BYTES {
+			return true
+		}
+	case kindWellKnown:
+		if t.GetWellKnown() == exprpb.Type_DURATION || t.GetWellKnown() == exprpb.Type_TIMESTAMP {
+			return true
+		}
+	case kindObject:
```
jpbetz: not clear under what conditions we expect to receive well known vs. message type. Can anyone explain? If both are possible, can we add tests cases for both?

checker/cost.go
```diff
@@ -599,3 +605,30 @@ func (c *coster) newAstNode(e *exprpb.Expr) *astNode {
 	}
 	return &astNode{path: path, t: c.getType(e), expr: e, derivedSize: derivedSize}
 }
+
+func isScalar(t *exprpb.Type) bool {
```
jpbetz, line 615: +godoc since many programmers consider strings scalars.

jpbetz: Also, when discussing with @TristonianJones it looks like `protobuf.Any` and `protobuf.Value` are not supportable since it's not possible to know if the value is a scalar till runtime. We should include a note about that in the godoc..

DangerOnTheRanger: Gotcha, I've added documentation with a note about both `Any` and `Value`.

checker/cost.go
DangerOnTheRanger: `kindObject`/`MessageType` seems to be for directly-created protobufs (i.e, an expression containing something like `google.protobuf.Int32Value{}` or `google.protobuf.Duration{}`). Since those cases were handled already by [this part](https://github.com/google/cel-go/blob/master/checker/cost.go#L107-L110) of `ComputedSize` I've removed the `kindObject` case from `isScalar` entirely.

checker/cost.go
DangerOnTheRanger: (marking this as resolved due to [this comment](https://github.com/google/cel-go/pull/571#discussion_r946080502))

jpbetz:
LGTM

TristonianJones (approved):

jpbetz (approved):

PR #574, opened 2022-08-15
Avoid exception for immutable list from folded expression

livebranch:
* Added check to avoid error thrown when returning a immutable list from a function called by a comprehension expression

This pull request fixes a regression introduced in commit 8cf150331f8f918625c0c4c83df500e6ce1fb82f. Specifically, an optimization was made which overrides an initial zero size list with a empty mutable list. The optimization makes an assumption that the result from the first step will always be mutable list, which is not always the case. Hopefully the code is self-explanatory, but I can add further details if required.

### DIFF
--- a/interpreter/interpretable.go
+++ b/interpreter/interpretable.go
@@ -835,7 +835,9 @@ func (fold *evalFold) Eval(ctx Activation) ref.Val {
 	varActivationPool.Put(accuCtx)
 	// Convert a mutable list to an immutable one, if the comprehension has generated a list as a result.
 	if !types.IsUnknownOrError(res) && buildingList {
-		res = res.(traits.MutableLister).ToImmutableList()
+		if _, ok := res.(traits.MutableLister); ok {
+			res = res.(traits.MutableLister).ToImmutableList()
+		}
 	}
 	return res
 }


TristonianJones (approved):

TristonianJones:
@livebranch Would you mind signing the CLA? Once you have I can merge the PR.

livebranch:
> @livebranch Would you mind signing the CLA? Once you have I can merge the PR.

Sorry, email was private so CLA didn't work. Now fixed.

TristonianJones:
@livebranch I've merged the change, but could you please attach a repro to the PR for future reference. Ideally, these kinds of repros are added into our tests as well. I ask mostly because the `buildingList` flag should not be `true` unless the list is mutable.

livebranch:
Thanks for the fast response in getting this merged!

Sorry, we don't currently have a concise repro to link to at this point. However, to address your point, consider this simple macro: 

```
func makeProblemMacro(eh parser.ExprHelper, target *expr.Expr, args []*expr.Expr) (*expr.Expr, *common.Error) {
	return eh.Fold(
		"iv",
		target,
		parser.AccumulatorName,
		eh.NewList(),
		eh.LiteralBool(true),
		eh.NewList(eh.LiteralInt(1)), // <---- problem is triggered when a non-mutable list is returned by this expr.
		eh.Ident(parser.AccumulatorName)), nil
}
```

The key here is the `step` expression. If it ever returns a non-mutable list, Eval throws the error `interface conversion: *types.baseList is not traits.MutableLister.`.

`buildingList` being set to `true` only implies the initial accumulator value is mutable, however, it's the eval result that's asserted of type `traits.MutableLister` (which it may or may not be, depending on the Eval result).

PR #579, opened 2022-08-25
ExprHelper.Copy to assist with unique id tagging of macro expanded ASTs

TristonianJones:
`ExprHelper.Copy()` to assist with unique id tagging of macro expanded ASTs.

Copy will preserve the original content of the `Expr` value, but renumber the nodes within
the AST and generate separate entries in the `SourceInfo` for the copied source locations.


TristonianJones:
Note, the `vendor` dependency was updated, so the bulk of the changes are not functionality related.

jnthntatum (approved):

PR #580, opened 2022-09-08
enable comparison of wrapped errors via errors.Is

awillis:
This change adds type.Err.Is, which enables comparisons of wrapped errors via errors.Is in the standard library.
Although this is of limited use within cel-go itself, it can help users of this library compare sentinel errors wrapped by types.NewErr


TristonianJones (approved):
Thanks for the change!

PR #581, opened 2022-09-09
Library for creating and working with optional values. 

TristonianJones:
This feature makes it possible to indicate to CEL that a variable may be optionally provided. It also makes it feasible to indicate that the expression result is a strongly typed optional value.

To enable this feature pass the `cel.OptionalTypes(true)` to the `cel.NewEnv()` call.

Optional field selection and message creation will follow in subsequent PRs.


cel/cel_test.go
```diff
@@ -1825,6 +1825,107 @@ func TestDynamicDispatch(t *testing.T) {
...
+		},
+		{
+			expr: `optional.ofNonZeroValue(z).or(optional.of(10)).value() == 42`,
+			in: map[string]interface{}{
+				"z": 42,
+			},
+			out: true,
+		},
+		{
+			expr: `(has(m.x) ? optional.of(m.x) : optional.none()).hasValue()`,
```
jnthntatum, line 1862: optional: may be worth noting that we're planning to add special syntax later to make expressions like this cleaner.

cel/cel_test.go
```diff
@@ -1825,6 +1825,107 @@ func TestDynamicDispatch(t *testing.T) {
...
+			in: map[string]interface{}{
+				"m": map[string]map[string]string{},
+			},
+			out: false,
+		},
+		{
+			expr: `optional.ofNonZeroValue('').or(optional.of(m.c['dashed-index'])).orValue('default value')`,
```
jnthntatum, line 1870: What's the intent of the test? that the middle optional gets applied? I think this might lead a reader to believe optional.of is doing some magic to check field presence.

cel/cel_test.go
```diff
@@ -1825,6 +1825,107 @@ func TestDynamicDispatch(t *testing.T) {
...
+			expr: `x.or(y).orValue(z)`,
+			in: map[string]interface{}{
+				"x": types.OptionalNone,
+				"y": types.OptionalNone,
+				"z": 42,
+			},
+			out: 42,
+		},
+		{
+			expr: `optional.ofNonZeroValue(z).or(optional.of(10)).value() == 42`,
```
jnthntatum, line 1854: could you add a test for unchecked / bad optional value access?

jnthntatum (commented):
Some questions about the evaluator tests (as documentation for behavior) -- otherwise lgtm

cel/cel_test.go
TristonianJones: Done.

cel/cel_test.go
TristonianJones: Done.

jcking (dismissed):

jcking (approved):

PR #585, opened 2022-09-15
Refinement on recursion depth checking

TristonianJones:
This change unnests common object graphs which resolve to simple expressions, and reduces the number of stack frames needed to build an AST from the parse graph. This change is believed to be safe as it expands the number of expressions which will pass max recursion depth test.

Additionally, all rules which may left recurse are only tracked if they contain a recursive parse graph thanks to the unnesting step.


parser/parser.go
```diff
@@ -904,3 +854,56 @@ func (p *parser) expandMacro(exprID int64, function string, target *exprpb.Expr,
...
+	if p.recursionDepth > p.maxRecursionDepth {
+		panic(&recursionError{message: "max recursion depth exceeded"})
+	}
+}
+
+func (p *parser) decrementRecursionDepth() {
+	p.recursionDepth--
+}
+
+func unnest(tree antlr.ParseTree) antlr.ParseTree {
```
l46kok, line 871: Could we document this method

parser/parser.go
```diff
@@ -904,3 +854,56 @@ func (p *parser) expandMacro(exprID int64, function string, target *exprpb.Expr,
...
+		case *gen.RelationContext:
+			if t.GetOp() != nil {
+				return t
+			}
+			tree = t.Calc()
+		case *gen.CalcContext:
+			if t.GetOp() != nil {
+				return t
+			}
+			tree = t.Unary()
```
l46kok, line 908: Are we ok not checking for the existence of unary here? Especially since we're removing the nil checks from the VisitCalc method. The same question applies for RelationContext above

parser/parser.go
```diff
@@ -506,21 +517,6 @@ func (p *parser) VisitUnary(ctx *gen.UnaryContext) interface{} {
...
-		return p.VisitPrimaryExpr(ctx.Member().(*gen.PrimaryExprContext))
-	case *gen.SelectOrCallContext:
-		return p.VisitSelectOrCall(ctx.Member().(*gen.SelectOrCallContext))
-	case *gen.IndexContext:
-		return p.VisitIndex(ctx.Member().(*gen.IndexContext))
-	case *gen.CreateMessageContext:
-		return p.VisitCreateMessage(ctx.Member().(*gen.CreateMessageContext))
-	}
-	return p.reportError(ctx, "unsupported simple expression")
-}
```
l46kok, line 522: This is more of an observation but with this change, there is no longer a 1-to-1 correspondence with the visit implementations [with respect to the grammar](https://github.com/google/cel-go/blob/master/parser/gen/CEL.g4#L53), so the readability here does drop a bit.

parser/parser.go
TristonianJones: I should probably add a condition in the for-loop that tests for `tree != nil` since that's the worst case scenario for traversing an incomplete parse graph that results from a syntax error.

The `unnest` ensures that the nested object graph is flattened down to leaves with interesting expression elements. Since all AST construction routes through the `Visit` method this behavior should be safe as it means that the primary switch statement is now oriented around compound parse nodes or simple leaves

parser/parser.go
TristonianJones: I think the docs on the `unnest` function will help a bit, but can understand how this degrades readability.

parser/parser.go
TristonianJones: Done.

l46kok (approved):

PR #592, opened 2022-09-29
Proto extension support

TristonianJones:
Extension to support the presence test and retrieval of proto2 `extensions`.

### DIFF
--- a/common/types/pb/file.go
+++ b/common/types/pb/file.go
@@ -20,17 +20,21 @@ import (
 	"google.golang.org/protobuf/reflect/protoreflect"
 )
 
-// NewFileDescription returns a FileDescription instance with a complete listing of all the message
+// newFileDescription returns a FileDescription instance with a complete listing of all the message
 // types and enum values declared within any scope in the file.
-func NewFileDescription(fileDesc protoreflect.FileDescriptor, pbdb *Db) *FileDescription {
+func newFileDescription(fileDesc protoreflect.FileDescriptor, pbdb *Db) *FileDescription {
 	metadata := collectFileMetadata(fileDesc)
 	enums := make(map[string]*EnumValueDescription)
 	for name, enumVal := range metadata.enumValues {
-		enums[name] = NewEnumValueDescription(name, enumVal)
+		enums[name] = newEnumValueDescription(name, enumVal)
 	}
 	types := make(map[string]*TypeDescription)
 	for name, msgType := range metadata.msgTypes {
-		types[name] = NewTypeDescription(name, msgType)
+		extensions, found := metadata.msgExtensionMap[name]
+		if !found {
+			extensions = []protoreflect.ExtensionDescriptor{}
+		}
+		types[name] = newTypeDescription(name, msgType, extensions)
 	}
 	return &FileDescription{
 		types: types,
@@ -95,6 +99,7 @@ type fileMetadata struct {
 	// enumValues maps from fully-qualified enum value to enum value descriptor.
 	enumValues map[string]protoreflect.EnumValueDescriptor
 	// TODO: support enum type definitions for use in future type-check enhancements.
+	msgExtensionMap map[string][]protoreflect.ExtensionDescriptor
 }
 
 // collectFileMetadata traverses the proto file object graph to collect message types and enum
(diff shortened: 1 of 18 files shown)


ext/protos.go
```diff
@@ -0,0 +1,135 @@
...
+// # Protos.GetExt
+//
+// Retrieves an extension field from the input proto2 syntax message. If the field
+// is not set, the default value for the extension field is returned.
+//
+//	proto.getExt(<msg>, <fully.qualified.extension.name>) -> <field-type>
+//
+// Examples:
+//
+//	proto.hasExt(msg, google.expr.proto2.test.int32_ext) // returns int value
```
jnthntatum: c/p error: getExt

ext/protos_test.go
```diff
@@ -0,0 +1,159 @@
...
+			asts = append(asts, pAst)
+			if !tc.parseOnly {
+				cAst, iss := env.Check(pAst)
+				if iss.Err() != nil {
+					t.Fatalf("env.Check(%v) failed: %v", tc.expr, iss.Err())
+				}
+				asts = append(asts, cAst)
+			}
+			msg := &proto2pb.ExampleType{
+				Name: proto.String("example0"),
```
jnthntatum: readability preference: since this doesn't vary per case you could extract the msg setup from the test loops closer to the test expression definitions to help a reader verify the expected results.

ext/protos_test.go
```diff
@@ -0,0 +1,159 @@
...
+			}
+			proto.SetExtension(msg, proto2pb.E_Int32Ext, int32(42))
+			proto.SetExtension(msg, proto2pb.E_Int32WrapperExt, wrapperspb.Int32(21))
+			proto.SetExtension(msg, proto2pb.E_NestedExample, &proto2pb.ExampleType{Name: proto.String("nested")})
+			proto.SetExtension(msg, proto2pb.E_ExtendedExampleType_EnumExt, proto2pb.GlobalEnum_GAZ)
+			proto.SetExtension(msg, proto2pb.E_ExtendedExampleType_ExtendedExamples, []string{"example1", "example2"})
+			for _, ast := range asts {
```
jnthntatum, line 135: preference: overly DRY style tests make them harder to read. it might be worth adding a separate test case for the parse only and parse / check tests.

ext/protos_test.go
```diff
@@ -0,0 +1,159 @@
...
+		{expr: `proto.getExt(msg, google.expr.proto2.test.ExtendedExampleType.enum_ext) == GlobalEnum.GAZ`},
+
+		// Negative test cases.
+		{
+			expr:      `proto.getExt(ExtendedExampleType{}, google.expr.proto2.test.ExtendedExampleType.enum_ext)`,
+			err:       "no such field 'google.expr.proto2.test.ExtendedExampleType.enum_ext'",
+			parseOnly: true,
+		},
+		{
+			expr: `proto.getExt(ExtendedExampleType{}, enum_ext)`,
```
jnthntatum: Does the extension id lookup follow container lookup rules if on is set or are only fully qualified names supported?

ext/protos.go
```diff
@@ -0,0 +1,135 @@
...
+//	proto.hasExt(msg, google.expr.proto2.test.int32_ext) // returns true || false
+func Protos() cel.EnvOption {
+	return cel.Lib(protoLib{})
+}
+
+type protoLib struct{}
+
+func (protoLib) CompileOptions() []cel.EnvOption {
+	return []cel.EnvOption{
+		cel.Macros(
```
jnthntatum, line 68: A quick implementation comment on the macro behavior might be helpful here (proto.hasExt(arg, com.ext.field) -> test only select expression with field set to the fully qualified extension name, right?)

ext/README.md
```diff
@@ -31,6 +31,32 @@ Example:
...
+### Protos.GetExt
+
+Retrieves an extension field from the input proto2 syntax message. If the field
+is not set, the default value for the extension field is returned.
+
+    proto.getExt(<msg>, <fully.qualified.extension.name>) -> <field-type>
+
+Example:
+
+	proto.hasExt(msg, google.expr.proto2.test.int32_ext) // returns int value
```
jnthntatum: getExt

jnthntatum (changes requested):
lgtm, just a few suggestions

common/types/pb/file.go
```diff
@@ -95,35 +99,46 @@ type fileMetadata struct {
 	// enumValues maps from fully-qualified enum value to enum value descriptor.
 	enumValues map[string]protoreflect.EnumValueDescriptor
 	// TODO: support enum type definitions for use in future type-check enhancements.
+	msgExtensionMap map[string][]protoreflect.ExtensionDescriptor
```
l46kok, line 103: nit: a comment on what the map maps to would be helpful (and to keep consistent with above)

common/types/pb/file.go
```diff
@@ -20,17 +20,21 @@ import (
 	"google.golang.org/protobuf/reflect/protoreflect"
 )
 
-// NewFileDescription returns a FileDescription instance with a complete listing of all the message
+// newFileDescription returns a FileDescription instance with a complete listing of all the message
 // types and enum values declared within any scope in the file.
```
l46kok: nit: Could we update comment to include extensions

ext/protos_test.go
```diff
@@ -0,0 +1,159 @@
...
+			expr: `proto.getExt(ExtendedExampleType{}, has(google.expr.proto2.test.int32_ext))`,
+			parseErr: `ERROR: <input>:1:40: invalid extension field
+			| proto.getExt(ExtendedExampleType{}, has(google.expr.proto2.test.int32_ext))
+			| .......................................^`,
+		},
+
+		// Even though 'getExt' is the macro, the call is left unexpanded is an identifier
+		// that's not named 'proto'
+		{
+			expr: `msg.getExt("google.expr.proto2.test.int32_ext", 0) == 42`,
```
l46kok, line 91: Could we have a similar test for `msg.hasExt` as well

ext/protos.go
```diff
@@ -0,0 +1,135 @@
...
+	if err != nil {
+		return nil, err
+	}
+	return meh.Select(args[0], extFieldName), nil
+}
+
+func isExtCall(meh cel.MacroExprHelper, fnName string, target *exprpb.Expr, args []*exprpb.Expr) (*exprpb.Expr, bool) {
+	switch target.GetExprKind().(type) {
+	case *exprpb.Expr_IdentExpr:
+		if target.GetIdentExpr().GetName() != "proto" {
```
l46kok: Maybe pull proto into a constant. I think we should also document that proto effectively becomes a keyword of a sort.

Actually, what happens if there's an existing expression containing a message variable named "proto"?

ext/protos_test.go
TristonianJones: I'll separate the parseErr tests into a separate group

ext/protos_test.go
TristonianJones: It doesn't, the extension name must be fully qualified. I'll add that to the docs.

ext/README.md
TristonianJones: Done.

ext/protos.go
TristonianJones: Done.

ext/protos.go
TristonianJones: If there is a variable called 'proto' then you'd have to use it in the exact same manner as the macro is expressed for it to cause a problem. The behavior is very similar to a namespaced function, but there's a potential for collision. I'll add it to the docs.

ext/protos.go
TristonianJones: Done. I've added comments to the macro expansion functions to indicate what they're doing.

ext/protos_test.go
TristonianJones: Done.

ext/protos_test.go
TristonianJones: Done.

common/types/pb/file.go
TristonianJones: Done.

l46kok (approved):

jnthntatum (approved):

PR #600, opened 2022-11-01
Shift proto extension tracking into the pb.Db

TristonianJones:
When a protobuf extension is not being used in the same file as where the `extendee` message is declared,
the protobuf extensions were getting dropped. Shifting tracking of extensions into a central location fixes the
issue.


test/proto2pb/test_all_types.proto
```diff
@@ -116,13 +116,6 @@ message ExampleType {
...
 }
 
-// Package scoped extensions.
-extend ExampleType {
-  optional ExampleType nested_example = 100;
-  optional int32 int32_ext = 101;
-  optional google.protobuf.Int32Value int32_wrapper_ext = 102;
-}
-
 // Message scoped extensions.
```
l46kok, line 119: Could we also move this to a separate file? That would [more accurately test the message scoped cases here](https://github.com/google/cel-go/blob/b8c669d93b21fc766a05c7199be531f439572fcc/common/types/pb/file_test.go#L48-L55)

l46kok (dismissed):

test/proto2pb/test_all_types.proto
TristonianJones: I intentionally split the extensions to cover the case where they're spread across the same file and a separate file. I can definitely add a message extension into the test_extensions proto though. Does that sound reasonable?

l46kok: I see, what you proposed sounds good to me.

TristonianJones: PTAL. It looked like the `protos.go` and `protos_test.go` files weren't included in the Bazel build and there were some additional issues there with how extensions were being resolved at runtime. Additionally, some issues cropped up with ensuring that extensions were hermetically sealed between `Db.Copy()` calls which have been addressed as well.

common/types/pb/pb.go
```diff
@@ -173,6 +195,18 @@ func (pbdb *Db) DescribeEnum(enumName string) (*EnumValueDescription, bool) {
 	return nil, false
 }
 
+// DescribeExtension returns a `FieldDescription` for a given type name and fully qualified extension
+// name if it exists in the `pb.Db`.
+func (pbdb *Db) DescribeExtension(typeName, extField string) (*FieldDescription, bool) {
```
l46kok: Can we remove this as it is no longer used?

common/types/pb/file.go
```diff
@@ -30,24 +32,51 @@ func newFileDescription(fileDesc protoreflect.FileDescriptor, pbdb *Db) *FileDes
...
+			extDesc := dynamicpb.NewExtensionType(ext).TypeDescriptor()
+			messageExtMap[string(ext.FullName())] = newFieldDescription(extDesc)
 		}
-		types[name] = newTypeDescription(name, msgType, extensions)
+		pbdb.extensions[typeName] = messageExtMap
 	}
 	return &FileDescription{
+		name:  fileDesc.Path(),
 		types: types,
 		enums: enums,
```
l46kok: This method now mutates the pbdb extension entries and also constructs a new FileDescription struct. I think it's easier to follow if it only did one of the two. Can we consider adding the extensions back into the struct and perform the registration to pb.Db in `RegisterDescriptor`?

common/types/provider.go
```diff
@@ -115,10 +115,7 @@ func (p *protoTypeRegistry) FindFieldType(messageType string,
 	}
 	field, found := msgType.FieldByName(fieldName)
 	if !found {
-		field, found = p.pbdb.DescribeExtension(messageType, fieldName)
-		if !found {
-			return nil, false
-		}
```
l46kok: Could you help me understand this change? I would've thought that we want to perform the extension lookup here.

TristonianJones: The field lookup now falls back to looking at extensions within the `TypeDescription`, so the use of `DescribeExtension` isn't necessary here.

common/types/pb/pb.go
TristonianJones: Done.

common/types/pb/file.go
TristonianJones: I've moved the merge up into the `pb.Db` by doing a multi-valued return from `newFileDescription` which includes an `extensionMap`. The result looks cleaner. Thanks for suggesting a change here.

TristonianJones:
@l46kok PTAL.

l46kok (approved):

PR #601, opened 2022-11-02
Parse and type-check support for optional field selection

TristonianJones:
This PR introduces support for the following expressions:

* `map[?optKey]` - optional map key, returns `optional.of(value)` if present, else `optional.none()`
* `list[?optIndex]` - optional list index, returns `optional.of(value)` if the index is between 0 and `list.size()`, else `optional.none()`
* `msg.?field` - optional field selection from a protobuf message type (or optional key selection within a map where the key is a valid identifier).

This change does alter a handful of existing error messages by exposing the existence of `?` as a lexical character in the grammar even when the feature is not enabled.


jcking (approved):

acynothia:
How about the syntax like javascript [optional chaining](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining)?

TristonianJones:
@mozartz see https://github.com/google/cel-spec/wiki/proposal-246 for the full details. The position of the `?` and `.` are inverted compared to JavaScript, but the operators can be chained and the `?` is synonymous with optional fields, map entries, and soon list elements in literal construction.

PR #604, opened 2022-11-04
Support assignment of a wrapper field to null

TristonianJones:
When a protobuf wrapper type field is assigned to null, the wrapper field should
be unset in the containing proto message as `null` indicates that the field was unset
to begin with. Thus, the following expression should roundtrip:

```
TestAllTypes{single_int32_wrapper: TestAllTypes{}.single_int32_wrapper}.single_int32_wrapper == null
```


jnthntatum (approved):

l46kok (approved):
This might be something good to add to conformance tests

TristonianJones:
@l46kok agreed. I think there are probably more proto-based tests which should cover edge cases like this one.

PR #605, opened 2022-11-04
Allow macros to return an empty Expr to indicate a non-expansion

TristonianJones:
Allow macro implementations to indicate a non-expansion by returning an empty `Expr` value.

In the cel-cpp project an empty `Expr` return value indicates a non-match, so this change ensures
that the capabilities of cel-go and cel-cpp are at parity.

### DIFF
--- a/parser/parser.go
+++ b/parser/parser.go
@@ -31,7 +31,6 @@ import (
 	"github.com/google/cel-go/parser/gen"
 
 	exprpb "google.golang.org/genproto/googleapis/api/expr/v1alpha1"
-	"google.golang.org/protobuf/proto"
 	structpb "google.golang.org/protobuf/types/known/structpb"
 )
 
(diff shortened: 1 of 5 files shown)


parser/parser.go
```diff
@@ -873,12 +874,18 @@ func (p *parser) expandMacro(exprID int64, function string, target *exprpb.Expr,
...
 	expr, err := macro.Expander()(eh, target, args)
+	// An error indicates that the macro was matched, but the arguments were not well-formed.
 	if err != nil {
 		if err.Location != nil {
 			return p.reportError(err.Location, err.Message), true
 		}
 		return p.reportError(p.helper.getLocation(exprID), err.Message), true
 	}
+	// An empty Expr value from the macro indicates that the macro implementation decided that
+	// an expansion should not be performed.
```
jnthntatum, line 884: Wouldn't returning nil be more idiomatic for not-present pointer value?

jnthntatum (changes requested):
LGTM but consider using nil instead of default expr message as a sentinel value.

parser/parser.go
TristonianJones: I naively copied the behavior from cel-cpp without thinking about what would be idiomatic. Updated.

TristonianJones:
PTAL.

jnthntatum (approved):

PR #608, opened 2022-11-08
Math extension library for supporting least and greatest macros

TristonianJones:
The `math` namespaced extension library will become a basis for several other functions, but the initial support only includes the vararg `least` and `greatest` macros.

### DIFF
--- a/ext/math_test.go
+++ b/ext/math_test.go
@@ -1,0 +1,15 @@
+// Copyright 2022 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package ext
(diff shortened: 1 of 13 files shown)


ext/math_test.go
```diff
@@ -0,0 +1,398 @@
...
+func TestMath(t *testing.T) {
+	mathTests := []struct {
+		expr string
+		in   any
+	}{
+		// Tests for math.least
+		{expr: "math.least(-0.5) == -0.5"},
+		{expr: "math.least(-1) == -1"},
+		{expr: "math.least(1u) == 1u"},
+		{expr: "math.least(42.0, -0.5) == -0.5"},
```
jnthntatum, line 34: It looks like you added overloads for mixed numeric types -- could you add a test for those cases (math.least(1, 1.1) etc)

jnthntatum (changes requested):

ext/math_test.go
TristonianJones: Certainly. Good call. There's some question whether I should just expose the underlying functions directly since they're more friendly to use with the type-checker. Thoughts?

jnthntatum: Do you mean in addition to the packaged library or instead of? I don't see any problem with exposing the extension function impl for flexibility, but I do prefer the consistency of providing the standard extension library.

TristonianJones: I'll keep the `@min`, `@max` definitions private then. Tests added.

TristonianJones:
@jnthntatum PTAL

jnthntatum (approved):

PR #612, opened 2022-11-16
Extension to support Golang structs as CEL types

### LINKED ISSUE #408: Question: Go structure value reflection?
Hello,

I couldn't find any way of binding bare Go values, especially Go structures, in CEL declarations. My understanding is that this is limited to protobufs only so far. Is there any way to have some "wrapper" value where I could use Go reflection?

A simple example of a use-case I have: accessing `*http.Request` values from CEL expressions. Ideally, I don't want to copy/convert it into another protobuf value, but rather use it in-place.

Thanks!

TristonianJones:
Initial support for Golang struct types within CEL. 

Given a struct defined as follows, you could use the type definition within CEL as follows:

```go
package auth;

type Account struct {
  Name string
  ID int64
}

env, err := cel.NewEnv(
    ext.NativeTypes(reflect.TypeOf(&auth.Account{})),
)
// ... check for errors ... 
ast, iss := env.Compile(`auth.Account{Name: 'test', ID: 1234}`)
// ... check for errors ... 
prg, err := env.Program(ast)
// ... check for errors ... 
out, _, err := prg.Eval(cel.NoVars())
reflect.DeepEqual(out.Value(), &auth.Account{Name: 'test', ID: 1234}) == true
```

The `ext.NativeTypes()` option supports safe traversal over struct types, as well as support for `has()` and the new [CEL optional value](https://github.com/google/cel-spec/wiki/proposal-246) feature introduced with `cel.OptionalTypes()`.

As a future refinement a new `cel` tag may be introduced in order to make it possible to preserve the snake_case field names convention between Golang and protobuf if working with both type systems.

Closes #408

### DIFF
--- a/ext/native.go
+++ b/ext/native.go
@@ -18,35 +18,232 @@ import (
 	"fmt"
 	"reflect"
 
+	"google.golang.org/protobuf/proto"
+	"google.golang.org/protobuf/reflect/protoreflect"
+
+	"github.com/google/cel-go/cel"
+	"github.com/google/cel-go/checker/decls"
 	"github.com/google/cel-go/common/types"
+	"github.com/google/cel-go/common/types/pb"
 	"github.com/google/cel-go/common/types/ref"
 	"github.com/google/cel-go/common/types/traits"
+
+	exprpb "google.golang.org/genproto/googleapis/api/expr/v1alpha1"
 )
 
 var (
 	nativeObjTraitMask = traits.FieldTesterType | traits.IndexerType
 )
 
-type nativeObj struct {
-	val      any
-	valType  *nativeType
-	refValue reflect.Value
+func NativeTypes(refTypes ...any) cel.EnvOption {
+	return func(env *cel.Env) (*cel.Env, error) {
+		tp, err := newNativeTypeProvider(env.TypeAdapter(), env.TypeProvider(), refTypes...)
+		if err != nil {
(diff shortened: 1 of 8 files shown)


TristonianJones:
@jcking Note, I had to fix a couple of bugs related to optional value function signatures and library publication as part of this change. Let me know if you want me to separate them out.

ext/native.go
```diff
@@ -0,0 +1,544 @@
...
+	nativeObjTraitMask = traits.FieldTesterType | traits.IndexerType
+)
+
+// NativeTypes creates a type provider which uses reflect.Type and reflect.Value instances
+// to produce type definitions that can be used within CEL.
+//
+// All struct types in Go are exposed to CEL via their simple package name and struct type name:
+//
+// ```go
+// package identity
```
kortschak, line 46: Match indentation?

ext/native.go
```diff
@@ -0,0 +1,544 @@
...
+//
+//	type Account struct {
+//	  ID int
+//	}
+//
+// ```
+//
+// The type `identity.Account` would be expored to CEL using the same qualified name, e.g.
```
kortschak: s/expored/exported/

kortschak (commented):
This looks like it will do what I needed. Thanks.

jcking (approved):

PR #615, opened 2022-12-08
Syntax, type-checking, and runtime support for optional list elements

TristonianJones:
As a refinement to optional values, this change introduces support for optional list elements during list literal construction:

```javascript
// If 'b' is present on 'a', then the output will be [a.b, 'hello', 'world']
// else ['hello', 'world']
[?a.?b, 'hello', 'world'] 
```

There are some minor changes to the grammar to split up the `.?` and `[?` tokens so that `?` appears as it's own token as otherwise the parser gets confused by `a[?b]` versus `[?a[?b]]`


checker/checker.go
```diff
@@ -406,16 +406,29 @@ func (c *checker) resolveOverload(
...
+	optionalIndices := create.GetOptionalIndices()
+	optionals := make(map[int32]bool, len(optionalIndices))
+	for _, optInd := range optionalIndices {
+		optionals[optInd] = true
+	}
+	for i, e := range create.GetElements() {
 		c.check(e)
-		elemType = c.joinTypes(c.location(e), elemType, c.getType(e))
+		elemType := c.getType(e)
+		if optionals[int32(i)] {
```
jcking, line 418: Hm. Been a while, does Go return the default value type if its not present in the map?

TristonianJones: In this case, the default value of false is returned, but I'll update the code to be more explicit about found and 'true'

jcking (approved):

checker/checker.go
jcking: Not necessary, up to you.

PR #616, opened 2022-12-08
Support for protobuf JSON serialization of native object types

### LINKED ISSUE #614: Using ext.NativeTypes prevents deserialisation of result when an array of byte string
**Describe the bug**

When the `ext.NativeTypes` extension is used, arrays of byte strings fail to be convertable.

**To Reproduce**

- [ ]  parser
- [ ]  checker
- [x]  interpreter (?)

```cel
[b'byte string']
```

Test setup:
```go
package main

import (
	"fmt"
	"reflect"

	"github.com/google/cel-go/ext"
	"google.golang.org/protobuf/encoding/protojson"
	"google.golang.org/protobuf/proto"
	"google.golang.org/protobuf/types/known/structpb"
)

func main() {
	const src = `[b'byte string']`

	fmt.Println(run(src))
	fmt.Println(run(src, ext.NativeTypes()))
	fmt.Println(run(src, ext.NativeTypes(reflect.TypeOf(&Native{}))))
}

type Native struct {
	ID int8
}

func run(src string, opts ...cel.EnvOption) (string, error) {
	env, err := cel.NewEnv(opts...)
	if err != nil {
		return "", fmt.Errorf("failed to create env: %v", err)
	}

	ast, iss := env.Compile(src)
	if iss.Err() != nil {
		return "", fmt.Errorf("failed compilation: %v", iss.Err())
	}

	prg, err := env.Program(ast)
	if err != nil {
		return "", fmt.Errorf("failed program instantiation: %v", err)
	}

	out, _, err := prg.Eval(interpreter.EmptyActivation())
	if err != nil {
		return "", fmt.Errorf("failed eval: %v", err)
	}

	v, err := out.ConvertToNative(reflect.TypeOf(&structpb.Value{}))
	if err != nil {
		return "", fmt.Errorf("failed proto conversion: %v", err)
	}
	b, err := protojson.MarshalOptions{Indent: "\t"}.Marshal(v.(proto.Message))
	if err != nil {
		return "", fmt.Errorf("failed native conversion: %v", err)
	}
	return string(b), nil
}
```

https://go.dev/play/p/ZnwBQwexvVR

**Expected behavior**

I expect three instances of the following to be printed.
```
[
	"Ynl0ZSBzdHJpbmc="
] <nil>
```
 Instead, I see
```
[
	"Ynl0ZSBzdHJpbmc="
] <nil>
 failed proto conversion: unsupported conversion to ref.Val: (uint8)98
 failed proto conversion: unsupported conversion to ref.Val: (uint8)98
```

If the expression is `['byte string']` the program behaves as expected, emitting three copies of
```
[
	"byte string"
] <nil>
```

**Additional context**

This may be due to changes that I need to make in the serialisation as I come out of the cel execution. If so, this is a question rather than a bug; what do I need to do with the code above to make the change work?

TristonianJones:
Support conversion of native object types to `google.protobuf.Value` instances.

Closes #614

### DIFF
--- a/ext/native_test.go
+++ b/ext/native_test.go
@@ -224,16 +224,21 @@ func TestNativeTypesJsonSerialization(t *testing.T) {
 		},
 		{
 			expr: `TestAllTypes{
-				NestedVal: TestNestedType{
-					NestedListVal: ["first", "second"],
-				},
 				BoolVal: true,
 				DurationVal: duration('5s'),
 				DoubleVal: 1.5,
 				FloatVal: 2.0,
 				Int32Val: 23,
 				Int64Val: 64,
-				StringVal: "string",
+				MapVal: {
+					'map-key': ext.TestAllTypes{
+						BoolVal: true
+					}
+				},
+				NestedVal: TestNestedType{
+					NestedListVal: ["first", "second"],
+				},
+				StringVal: "string"
 			}`,
 			out: `{
 				"BoolVal":  true,
@@ -242,6 +247,11 @@ func TestNativeTypesJsonSerialization(t *testing.T) {
 				"FloatVal":  2,
 				"Int32Val":  23,
 				"Int64Val":  64,
+				"MapVal": {
+	              "map-key": {
+    	            "BoolVal": true
+        	      }
+            	},
 				"NestedVal": {
 					"NestedListVal": [
 					  "first",
--- a/ext/native_test.go
+++ b/ext/native_test.go
@@ -21,7 +21,9 @@ import (
 	"testing"
 	"time"
 
+	"google.golang.org/protobuf/encoding/protojson"
 	"google.golang.org/protobuf/proto"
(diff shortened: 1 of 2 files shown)


ext/native_test.go
```diff
@@ -211,6 +213,73 @@ func TestNativeTypesStaticErrors(t *testing.T) {
...
+		out  string
+	}{
+		{
+			expr: `[b'string']`,
+			out:  `["c3RyaW5n"]`,
+		},
+		{
+			expr: `TestAllTypes{
+				NestedVal: TestNestedType{
+					NestedListVal: ["first", "second"],
```
l46kok: Can we also add a test for maps?

l46kok (dismissed):

ext/native_test.go
TristonianJones: Done.

l46kok (approved):

PR #617, opened 2022-12-19
Add string.format

DangerOnTheRanger:
This PR adds a `format` receiver function to CEL strings. The syntax is similar to the `printf` implementations found in various languages.  For example, `"%s and a number: %d".format(["a string", 42])` evaluates to `"a string and a number: 42"`.


common/types/string_test.go
```diff
@@ -253,3 +253,101 @@ func TestStringSize(t *testing.T) {
...
+		},
+		{
+			name:           "mid-string substitution",
+			format:         "str is {0} and some more",
+			formatArgs:     []ref.Val{String("filler")},
+			expectedOutput: "str is filler and some more",
+		},
+		{
+			name:           "bracket escaping",
+			format:         "{{ and also }}",
```
jpbetz: How about also testing:

- `{{{0}}}`
- `{0}}}`
- `{{{0}`

?

common/types/string_test.go
```diff
@@ -253,3 +253,101 @@ func TestStringSize(t *testing.T) {
...
+		},
+		{
+			name:   "unmatched left bracket",
+			format: "filler {",
+			err:    "error while parsing arg index: could not find matching '}'",
+		},
+		{
+			name:   "unmatched right bracket",
+			format: " } filler",
+			err:    "unmatched '}'",
```
jpbetz: Handle and test `unmatched '{'` case as well?

common/types/string.go
```diff
@@ -221,3 +228,101 @@ func stringStartsWith(s String, pre ref.Val) ref.Val {
...
+			i = newIndex
+		} else if formatStr[i] == '}' {
+			if i+1 < len(formatStr) && formatStr[i+1] == '}' {
+				err := builtStr.WriteByte('}')
+				if err != nil {
+					return NewErr("error writing format string: %s", err.Error())
+				}
+				i += 2
+				continue
+			} else {
```
jpbetz: nit: In above `formatStr[i] == '{'` condition, you don't use `if/else` because you short circuit with `continue` but here you both use `if/else` and a continue to short circuit. Be consistent?

common/types/string_test.go
```diff
@@ -253,3 +253,101 @@ func TestStringSize(t *testing.T) {
...
+			expectedOutput: "str is filler and some more",
+		},
+		{
+			name:           "bracket escaping",
+			format:         "{{ and also }}",
+			expectedOutput: "{ and also }",
+		},
+		{
+			name:           "multiple substitutions",
+			format:         "{0} {1} {2}, {3} {4} {5}",
```
jpbetz: Test a positional argument that is >1 digit in length, e.g. `10` ?

common/types/string.go
```diff
@@ -221,3 +228,101 @@ func stringStartsWith(s String, pre ref.Val) ref.Val {
...
+				return -1, -1, errors.New("no argument index given (\"{}\" is not valid)")
+			}
+			break
+		} else if !unicode.IsDigit(rune(formatStr[i])) {
+			return -1, -1, fmt.Errorf("unrecognized character '%c' found in argument index", formatStr[i])
+		}
+		indexStr.WriteByte(formatStr[i])
+		i++
+	}
+	argIndex, err := strconv.Atoi(indexStr.String())
```
jpbetz: Is `strconv.Atoi` compatible with all the digit representations `unicode.IsDigit` supports? Do you want to support just ASCII digits?

jpbetz (changes requested):
Not a full review pass, but noticed a few things.

common/types/string.go
DangerOnTheRanger: `Atoi` only works on ASCII digits (so does `ParseInt`, consequently), so at least as written it just works on ASCII digits. For what it's worth, Python handles Bengali here, I think:

```
>> "{৩}".format(1, 2, 3, 4, 5)
'4'
>>> "{১}".format(1, 2, 3, 4, 5)
'2'
>>> "{১৪}".format(1, 2, 3, 4,)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
IndexError: Replacement index 14 out of range for positional args tuple
```

It's worth pointing out that I picked other systems from [this list](https://en.wikipedia.org/wiki/List_of_numeral_systems#By_culture_/_time_period) and tried them as well, but couldn't get any to work.

I also tried a similar snippet with C#:

```c#
decimal temp = 20.4m;
string s = String.Format("The temperature is {০}°C.", temp);
Console.WriteLine(s);
```

But that gave me an error:

```
System.FormatException: Input string was not in a correct format.
```

(If the index was parsed, it would have resulted in `System.FormatException: Index (zero based) must be greater than or equal to zero and less than the size of the argument list.`)

Based on all this, I think it'd be safest to keep CEL support to just ASCII digits (that seems to be the common denominator across languages/string formatters here), but I'm curious what others think.

*edit* - I did some quick peeking around the Python codebase to see what they did: the [pure-Python string implementation](https://github.com/python/cpython/blob/a2bb3b7f9d8d15c81b724726454d68357fb31d1c/Lib/string.py#L297) delegates index parsing to the [C side of the interpreter](https://github.com/python/cpython/blob/a2bb3b7f9d8d15c81b724726454d68357fb31d1c/Objects/stringlib/unicode_format.h#L319), which after a [`get_integer` call](https://github.com/python/cpython/blob/a2bb3b7f9d8d15c81b724726454d68357fb31d1c/Objects/stringlib/unicode_format.h#L111) and a [`Py_UNICODE_TODECIMAL` expansion](https://github.com/python/cpython/blob/a2bb3b7f9d8d15c81b724726454d68357fb31d1c/Include/cpython/unicodeobject.h#L921) ends up simply using [`_PyUnicode_ToDecimalDigit`](https://github.com/python/cpython/blob/a2bb3b7f9d8d15c81b724726454d68357fb31d1c/Objects/unicodectype.c#L104) to look up the corresponding [Unicode digit property](https://en.wikipedia.org/wiki/Unicode#General_Category_property).  We could in theory do the same thing, though Golang's unicode module doesn't support doing the decimal mask trick that Python uses to extract the digit value/we have other CEL implementations to worry about.

common/types/string_test.go
DangerOnTheRanger: Gotcha, fixed

common/types/string_test.go
DangerOnTheRanger: Right, fixed

common/types/string_test.go
DangerOnTheRanger: Yeah, I think those are interesting corner cases to add. Fixed

common/types/string.go
DangerOnTheRanger: If I'm understanding correctly here you're referring to how the `else if formatStr[i] == '}' ` check has a `return NewErr("unmatched '}'")`, but the `if formatStr[i] == '{'` check doesn't have a similar `NewErr` line, right?   For the `}` case there's no valid reason to have a lone `}` in the string, so we know we can immediately return an error.

For the `{` case, if the next `if` isn't true (`if i+1 < len(formatStr) && formatStr[i+1] == '{'`) we're either at the end of the string or the next character isn't also `{`, so we call into `parseFormatArg` since it handles all possible cases (end of string, `{<n>}` substitution, and an unmatched `{`) and returns errors as appropriate. I feel like it's reasonable to have the resulting asymmetry since the formatting grammar is a bit different for `{` and `}` here.

common/types/string.go
jpbetz: Oh, I'm all in favor of ascii only. My point was that you use Unicode.IsDigit above and maybe should swap that out with something that only matches ascii digits?

common/types/string.go
jpbetz: I'm suggesting something much simpler. You have two code flows both in for loops, one like `if ... { <do something> continue } <do something else>` and another like `if ... { <do something> continue } else { <do something else }`.  Both are equivalent. Pick a style throughout the code?

TristonianJones:
@DangerOnTheRanger Please move this functionality to `ext/strings.go` since string formatting would be an extension function that's optionally enabled rather than a part of the CEL standard library.

TristonianJones:
@DangerOnTheRanger I have some questions based on what I've seen. Can you tell me the intended behavior for the following calls:

```js
// error based on list literal size or simply a reuse of the first argument in the string?
"{0} < {0}".format([1, 2]) 

// How are brackets escaped?
"{{0}: {1}}".format([key, value])
```

It sounds like you and Joe have adequately answered whether the digits are Ascii-only, and I'm fine with that as well.

DangerOnTheRanger:
> Kermit Alexander II I have some questions based on what I've seen. Can you tell me the intended behavior for the following calls:
> 
> ```js
> // error based on list literal size or simply a reuse of the first argument in the string?
> "{0} < {0}".format([1, 2]) 
> 
> // How are brackets escaped?
> "{{0}: {1}}".format([key, value])
> ```
> 
> It sounds like you and Joe have adequately answered whether the digits are Ascii-only, and I'm fine with that as well.

For the first one, as currently written the implementation would accept this (the first argument would be reused) but after offline discussion with @jpbetz I think this should result in a compile-time error if the formatting string is a constant.

For the second, that would result in an error (`{{` is treated as an escape sequence so the first `}` is unmatched), which feels reasonable to me. Admittedly the "correct form" of the formatting string (`{{{0}: {1}}}`) does look very verbose, but for use cases where the user wants to use the same characters that are used for formatting, (like `{` and `}` in this case) I'm not sure there's a way around that.

common/types/string.go
DangerOnTheRanger: Right, I'll mark this as resolved once I do.

common/types/string.go
DangerOnTheRanger: Oh, I misunderstood, sorry. I'll refactor and mark as resolved.

common/types/string.go
```diff
@@ -497,3 +523,225 @@ func parseClausePrecision(lastStrIndex int, formatStr string) (int, int, error)
...
+	tag, _ := language.MatchStrings(langMatcher, locale)
+	return tag
+}
+
+func initTags() []language.Tag {
+	// this is used for initializing locales whose behavior does not match
+	// the default (en_US) - if this function is used for other purposes
+	// and/or localization is used for datatypes besides floating-point
+	// numbers, this list may need to be updated
+	// golang's default catalog/matcher is empty, so we need to build our own
```
jpbetz: Use a function name that hints at the limits? Move this comment up to the function's godoc?

common/types/string.go
```diff
@@ -497,3 +523,225 @@ func parseClausePrecision(lastStrIndex int, formatStr string) (int, int, error)
 func isASCIIDigit(r rune) bool {
 	return r <= unicode.MaxASCII && unicode.IsDigit(r)
 }
+
+func matchLanguage(locale string) language.Tag {
+	tag, _ := language.MatchStrings(langMatcher, locale)
+	return tag
+}
```
jpbetz: Do we need this function? If we omitted it and instead required callers provide a language tag, could we avoid having to maintain the `initTags()` implementation?

common/types/string.go
```diff
@@ -230,11 +242,21 @@ func stringStartsWith(s String, pre ref.Val) ref.Val {
 }
 
 func stringFormat(s String, args ref.Val) ref.Val {
+	return stringFormatWithLocale(s, args, String(defaultLocale))
```
jpbetz: Instead of a hard coded `en_US` `defaultLocale` here, should CEL offer a way to set the locale in the `Env`? That seems a lot more flexible to me. (I'd be okay with that locale defaulting to `en_US` if unset).

DangerOnTheRanger: Right, that makes sense to me, I'll go ahead and make that change and mark this as resolved when I do.

EDIT:  I might have spoken too soon - is this possible/supported? Maybe I'm not looking in the right places, but I don't think I see any examples of an env setting being retrieved during eval time.

common/types/string.go
DangerOnTheRanger: The [definition](https://github.com/golang/text/blob/ec5565b1b747ce5ca569aeefc09e737b479a12ac/internal/language/language.go#L35) of `language.Tag` is fairly intricate, so I'm not sure it'd be good to leave supplying it up to expression writers.

As far as using `message.MatchLanguage` directly goes (intuitively it'd work instead of `initTags`, and was the first thing I tried, unfortunately): If the Go standard library's [default catalog wasn't empty by default](https://github.com/golang/text/blob/e328d63cff14134669501e0e154e4f141c784322/message/catalog.go#L26), then I don't think `initTags` would be necessary, but trying to use the default catalog as-is results in [no locale support](https://go.dev/play/p/CyqWfmrG7U1) (the number should use a comma decimal separator given the locale).

As far as I can tell this is the intended behavior; I think the intention is for programs to explicitly add the languages they support.

common/types/string.go
DangerOnTheRanger: Makes sense - I've moved the comment to the godoc and chosen a more descriptive name.

common/types/string.go
```diff
@@ -221,3 +240,447 @@ func stringStartsWith(s String, pre ref.Val) ref.Val {
 	}
 	return Bool(strings.HasPrefix(string(s), string(preStr)))
 }
+
+func stringFormat(s String, args ref.Val) ref.Val {
```
jpbetz: The format syntax needs to be documented _somewhere_.

cc @TristonianJones for ideas on where that should be.

common/types/string.go
jpbetz: > As far as I can tell this is the intended behavior; I think the intention is for programs to explicitly add the languages they support.

This is my exact point. Let the application programmer provide the tag. In the trivial case, I would expect the application programmer to pass in `language.MustParse("locale I care about")`. If the application programmer has a localized app, it is likely they have already performed a language match and have the appropriate tag.

common/types/string.go
```diff
@@ -221,3 +240,447 @@ func stringStartsWith(s String, pre ref.Val) ref.Val {
...
+		}
+		// locale is intentionally unused as integers formatted as binary
+		// strings are locale-independent
+		return fmt.Sprintf("%b", argInt), nil
+	}, nil
+}
+
+func formatHex(lastStrIndex int, formatStr string, useUpper bool) (int, clauseImpl, error) {
+	return lastStrIndex + 1, func(arg ref.Val, locale string) (string, error) {
+		argIntVal := arg.ConvertToType(IntType)
```
jpbetz: I was expecting Hex to also support strings and byte sequences.

common/types/string_test.go
```diff
@@ -253,3 +253,190 @@ func TestStringSize(t *testing.T) {
...
+		{
+			name:           "int substitution",
+			format:         "int is %s",
+			formatArgs:     []ref.Val{Int(1)},
+			expectedOutput: "int is 1",
+		},
+		{
+			name:           "double substitution",
+			format:         "double is %s",
+			formatArgs:     []ref.Val{Double(0.5)},
```
jpbetz: I was not expecting this to be allowed.

common/types/string_test.go
```diff
@@ -253,3 +253,190 @@ func TestStringSize(t *testing.T) {
...
+			formatArgs: []ref.Val{
+				String("percent"),
+			},
+			expectedOutput: "%escaped percent%",
+		},
+		{
+			name:   "fixed point formatting clause",
+			format: "%.3f",
+			formatArgs: []ref.Val{
+				Double(1.2345),
```
jpbetz: NaN, -Inf, +Inf?

common/types/string.go
```diff
@@ -221,3 +240,447 @@ func stringStartsWith(s String, pre ref.Val) ref.Val {
...
+		return argStr, nil
+	}, nil
+}
+
+func formatFixed(lastStrIndex int, formatStr string, precision *int) (int, clauseImpl, error) {
+	if precision == nil {
+		return -1, nil, errors.New("precision required for fixed-point output")
+	}
+	return lastStrIndex + 1, func(arg ref.Val, locale string) (string, error) {
+		argFloatVal := arg.ConvertToType(DoubleType)
```
jpbetz: Does this mean that we can pass in any type that converts to the target type of the formatter? Is that what we want?

We do need some way to CEL's timestamp and duration types (BTW we should add tests for these). So _maybe_ this is the right way to format those types? But right now, if you "%h" format a "123" string, won't this result in the the hex value of int(123) when what most people would expect is the hex values of the individual chars of the string?

common/types/string_test.go
```diff
@@ -253,3 +253,190 @@ func TestStringSize(t *testing.T) {
...
+			formatArgs: []ref.Val{
+				Int(1),
+			},
+			err: "could not parse formatting clause: unrecognized formatting clause \"a\"",
+		},
+		{
+			name:       "out of bounds arg index",
+			format:     "%s %s %s",
+			formatArgs: []ref.Val{Int(0), Int(1)},
+			err:        "index 2 out of range",
```
jpbetz: I was expecting a lot more err cases. E.g. Illegal arg types (arrays, maps, structs, unsupported scalars for the particular formatter), all the invalid arg cardinality cases, lots of malformed format clauses.

common/types/string.go
```diff
@@ -221,3 +240,447 @@ func stringStartsWith(s String, pre ref.Val) ref.Val {
...
+		argStr, ok := argStrVal.Value().(string)
+		if !ok {
+			return "", fmt.Errorf("could not convert argument %q to string", argStrVal)
+		}
+		return argStr, nil
+	}, nil
+}
+
+func formatFixed(lastStrIndex int, formatStr string, precision *int) (int, clauseImpl, error) {
+	if precision == nil {
```
jpbetz: Why not use a default if a precision is not provided? (not being able to just do %f is surprising to me)

common/types/string_test.go
DangerOnTheRanger: Right - I'll add a full test suite tonight. I think most of the happy path cases are covered decently enough; most of the coverage is missing in the error cases.

TristonianJones: @DangerOnTheRanger before you get too much farther, try to move this functionally into the `ext/strings.go` file since the intent is that this functionally is purely optional, even if it's well defined. If you do need to make some changes to the string type, that's okay, but they should hopefully be minimal.

DangerOnTheRanger: Whoops, you're right. I'll also take care of that tonight as well.

common/types/string_test.go
DangerOnTheRanger: Good catch for corner cases, I'll add those.

common/types/string.go
DangerOnTheRanger: Hmm, due to Go supporting this? (I wasn't aware that was possible before, very good to know) I think in terms of least-surprise I think it might be safer to keep it to just integers - I'm not aware of the hexadecimal formatting option doing this with other languages, but I could be wrong.

common/types/string_test.go
DangerOnTheRanger: Yes, I think it'd make more sense with `printf` syntax to disallow this - I'll fix it.

common/types/string.go
DangerOnTheRanger: I think the current behavior might be a little too forgiving, the more I think about it. I'll make `%s` only work for specific types (and not numbers) + add tests to make sure those guardrails work, and the same for `%e` and `%f` (I'll hold off on the `%h` one since we were still working out the behavior there). 

I think `%s` for timestamp and duration would work well, I'll add a test in for those.

common/types/string.go
DangerOnTheRanger: Right, I think for other developers it'd be unusual. I'll set the default to 6 for now, but I can change it if others have thoughts.

common/types/string.go
DangerOnTheRanger: How about we make a language tag list with only one member (the user-given language tag), so that if we have a CEL expression:

```
"%f".format([3.14], "fr_FR")
```

We then make a tag list containing only the French locale, and then match against that? I don't think there's much overhead in building this list every time. If an invalid locale is passed, the default behavior should be in line with `en_US` (or if it isn't, we can build a two-member list every time with `en_US` as the fallback option).

(Sorry if that's what you were suggesting - I had interpreted it to read as passing those tags in on the Golang side.)

common/types/string.go
jpbetz: Go does support this. Having a way to safely output byte strings is biggest motivator here. Developers rarely need to output a single int in hex, but it far more common that they need to safely output bytes.

common/types/string.go
jpbetz: For the case of a `format(..., locale)` would you even need a tag list? Couldn't you just call `language.Parse(locale)`?

Also, I'm skeptical of the value of a `format(..., locale)` function. I expect the vast majority of users to want to call format without locale and instead configure the `Env` with the locale.

common/types/string_test.go
jpbetz: This also makes it clear where to put the documentation as `ext/strings.go` has clear documentation conventions.

common/types/string.go
jpbetz: > is this possible/supported?

Yes, the env settings guide runtime behavior today. The codepaths are windy but are worth exploring. If you have trouble navigating them after a few hours raise the alarm and we work together on this.

common/types/string.go
DangerOnTheRanger: I'll add the docs to the string extension library documentation that already exists, which seems like a good starting point.

common/types/string.go
DangerOnTheRanger: Right - I'll remove the `...locale` overload and add it into the `env` and mark this as resolved when I do so.

common/types/string.go
DangerOnTheRanger: (marking as resolved since `formatHex` supports strings/byte buffers now)

ext/strings.go
```diff
@@ -485,6 +521,509 @@ func join(strs []string) (string, error) {
...
+				for _, c := range arg.Value().(string) {
+					_, err := buffer.WriteString(fmt.Sprintf(fmtStr, c))
+					if err != nil {
+						return "", fmt.Errorf("error formatting string as hex: %w", err)
+					}
+				}
+			}
+			return buffer.String(), nil
+		}
+		if arg.Type() != types.IntType {
```
jpbetz: Instead use a switch statement and a default for the unsupported type case?

ext/strings.go
```diff
@@ -485,6 +521,509 @@ func join(strs []string) (string, error) {
...
+
+func formatHex(lastStrIndex int, formatStr string, useUpper bool) (int, clauseImpl, error) {
+	return lastStrIndex + 1, func(arg ref.Val, locale string) (string, error) {
+		// special case to print out strings/byte arrays
+		if arg.Type() == types.StringType || arg.Type() == types.BytesType {
+			var buffer strings.Builder
+			fmtStr := "%x"
+			if useUpper {
+				fmtStr = "%X"
+			}
```
jpbetz: Move out of if and share with int case as well?

ext/strings.go
```diff
@@ -485,6 +521,509 @@ func join(strs []string) (string, error) {
...
+			if arg.Type() == types.BytesType {
+				for _, b := range arg.Value().([]byte) {
```
jpbetz: Why iterate over bytes and strings in this function instead of passing []byte directly into fmt.SPrintf("%x", ...)?

DangerOnTheRanger: Whoops, you're right. I've refactored both cases.

ext/strings_test.go
```diff
@@ -329,3 +330,229 @@ func TestStringsWithExtension(t *testing.T) {
...
+			expectedOutput: "1e is 20 in hexadecimal",
+		},
+		{
+			name:           "uppercase hexadecimal formatting clause",
+			format:         "%H is 20 in hexadecimal",
+			formatArgs:     "30",
+			expectedOutput: "1E is 20 in hexadecimal",
+		},
+		{
+			name:           "string support with hexadecimal formatting clause",
```
jpbetz, line 452: I was expecting cases for both `%h` and `%H` for both string and []byte.

ext/strings_test.go
```diff
@@ -329,3 +330,229 @@ func TestStringsWithExtension(t *testing.T) {
...
+			name:           "default precision for fixed-point clause",
+			format:         "%f",
+			formatArgs:     "2.71828",
+			expectedOutput: "2.718280",
+		},
+		{
+			name:           "default precision for scientific notation",
+			format:         "%e",
+			formatArgs:     "2.71828",
+			expectedOutput: "2.718280\u202f×\u202f10⁰⁰",
```
jpbetz: What's going on with this output?  I was not expecting to see any escape sequences. Go just outputs `2.718280e+00`

ext/strings_test.go
```diff
@@ -329,3 +330,229 @@ func TestStringsWithExtension(t *testing.T) {
...
+			name:           "NaN support for fixed-point",
+			format:         "%f",
+			formatArgs:     `"NaN"`,
+			expectedOutput: "NaN",
+		},
+		{
+			name:           "positive infinity support for fixed-point",
+			format:         "%f",
+			formatArgs:     `"Infinity"`,
+			expectedOutput: "∞",
```
jpbetz, line 521: Neat. Did go output this? Is this what we want? I'll admit "Inf" would be more mundane, but do we want to get into unicode output here?

DangerOnTheRanger: This [and this](https://github.com/google/cel-go/pull/617#discussion_r1095219394) are both due to how `text/message`'s `Sprintf` implementation works (see [here](https://go.dev/play/p/Q8P1gQTTh76) for a playground example).

I'm not particularly a fan of the Unicode output, but I'm also concerned that the alternative would be extremely difficult - building our own `%e` printing implementation to avoid Unicode would require us to handle `,` versus `.`, likely resulting in some kind of table similar to what `initTags` has for now.

ext/strings_test.go
DangerOnTheRanger: Fixed - added the additional cases.

ext/strings_test.go
jpbetz: Swap "\u202f" in these examples with the unicode equivalent? Right now this output has a mix of escaped unicode and actual unicode.

ext/strings_test.go
jpbetz: Ok, this is a bit fancier than I had hoped for, especially for "default" locale. Sadly I don't see anything in https://github.com/golang/text/blob/71a9c9afc4cd710b9412f7f99f0d8e35b10e488a/number/option.go that helps.  If no one else has better ideas I think I can live with this.

ext/strings_test.go
```diff
@@ -329,3 +330,247 @@ func TestStringsWithExtension(t *testing.T) {
 		t.Fatalf("env.Extend(Strings()) failed: %v", err)
 	}
 }
+
+func TestStringFormat(t *testing.T) {
```
jpbetz, line 339: Support+tests for CEL's date, time and duration types are still needed, I believe.

ext/strings_test.go
DangerOnTheRanger: I've swapped it over, though I think it's worth considering that `\u202f` is a [narrow space](https://codepoints.net/U+202F) and could possibly get confused with a normal/ASCII space inside the unit tests.

jpbetz: Good point. Maybe show a test case of both? It seems beneficial to be able to see both how it will _look_ and what it _really is_ in the tests.

DangerOnTheRanger: Right, I've reverted some of the test cases to use the escape sequence. It's also worth pointing out that that is not a regular ASCII "x" in there, but a [multiplication sign](https://codepoints.net/U+00D7?lang=en). Github's font makes it relatively clear, but I thought I'd point it out just in case. I've also added a comment making that fact about the multiplication sign clear, as well.

jpbetz: In the unicode escaped case, escape this too, and the superscripts at end as well. Then the side-by-side of the "unicode escaped" and "with unicode" test cases will highlight all the unicode a bit better?

DangerOnTheRanger: (marking this as resolved since the supercripts/multiplication signs have both been escaped in the unit tests with escaped spaces)

common/types/string.go
DangerOnTheRanger: (marking as resolved)

common/types/string_test.go
DangerOnTheRanger: (marking as resolved, tracking test suite in https://github.com/google/cel-go/pull/617#discussion_r1095261962 instead)

ext/strings.go
```diff
@@ -44,6 +60,40 @@ import (
...
+// this can be adjusted. The strings `Infinity`, `-Infinity`, and `NaN` are also valid input
+// for this clause.
+// `%e` - substitutes a double in scientific notation. The default precision is 6, but this
+// can be adjusted.
+// `%E` - same as a above, but with an uppercase E.
+// `%b` - substitutes an integer with its equivalent binary string.
+// `%h` - substitutes an integer with its equivalent in hexadecimal, or if given a string, will
+// output each character's equivalent in hexadecimal.
+// `%H` - same as above, but with A-F capitalized.
+// `%o` - substitutes an integer with its equivalent in octal.
```
jpbetz, line 78: Also document %h, %H for bytes and CEL date, time and duration support?

ext/strings_test.go
DangerOnTheRanger: Fixed - I've added support/tests for `Timestamp` and `Duration`.

common/types/string.go
DangerOnTheRanger: (marking as resolved since `%s` is used for timestamp/duration now)

ext/strings.go
DangerOnTheRanger: Right, I've added documentation for both - fixed

ext/strings_test.go
```diff
@@ -329,3 +330,265 @@ func TestStringsWithExtension(t *testing.T) {
...
+			name:           "timestamp support for string",
+			format:         "%s",
+			formatArgs:     `timestamp("2023-02-03T23:31:20+00:00")`,
+			expectedOutput: "2023-02-03T23:31:20Z",
+		},
+		{
+			name:           "duration support for string",
+			format:         "%s",
+			formatArgs:     `duration("1h45m47s")`,
+			expectedOutput: "6347s",
```
jpbetz, line 565: I was expecting the opposite. For `duration("6347s")` I was expecting `"1h45m47s"` to be output.

DangerOnTheRanger: Right - `Duration` now uses the human-readable string instead.

common/types/string.go
DangerOnTheRanger: (marking as resolved with the addition of `LocaleOptionBinding`)

cel/env.go
```diff
@@ -639,6 +641,14 @@ func getStdEnv() (*Env, error) {
 	return stdEnv, stdEnvErr
 }
 
+// ConfigureLocaleOption configures the Env with the given locale tag.
+func ConfigureLocaleOption(locale string) EnvOption {
```
jpbetz: Put this in options.go? Name it `Locale`? (the rest of the options don't have a 'configure' prefix or 'option' suffix.

DangerOnTheRanger: Sorry (wasn't aware about `options.go`, should've checked) - fixed

ext/strings.go
```diff
@@ -44,6 +56,42 @@ import (
 //	'hello'.charAt(5)  // return ''
 //	'hello'.charAt(-1) // error
 //
+// # Format
+//
+// Returns a new string with substitutions being performed, printf-style.
```
jpbetz, line 62: Any other types from https://github.com/google/cel-spec/blob/master/doc/langdef.md#values that should be supported? `boolean`? `type`?

ext/strings.go
```diff
@@ -44,6 +56,42 @@ import (
...
+// The valid formatting clauses are:
+//
+// `%s` - substitutes a string. This can also be used on Duration and Timestamp.
```
jpbetz: Are we OK with conflating duration and timestamp with string?  This locks us out of ever using directives to determine how a duration or timestamp is formatted, but it would still be possible to pre-format the duration or timestamp as a string before passing it to format.

EDIT: I think I might be OK with what we have, but am curious what others think. Right now, `%s` is behaving somewhat like`%v` in Go.

cc @TristonianJones

ext/strings.go
```diff
@@ -485,6 +545,293 @@ func join(strs []string) (string, error) {
...
+
+func matchLanguage(locale string) language.Tag {
+	tag, _ := language.MatchStrings(makeMatcher(locale), locale)
+	return tag
+}
+
+func makeMatcher(locale string) language.Matcher {
+	tags := make([]language.Tag, 0)
+	tag, err := language.Parse(locale)
+	// silenly fall back if the locale tag is invalid
```
jpbetz: If the developer mistypes a tag the result is the default tag instead of an error? Is this what we want?

ext/strings_test.go
```diff
@@ -329,3 +330,266 @@ func TestStringsWithExtension(t *testing.T) {
...
+			name:           "uppercase scientific notation formatting clause",
+			format:         "%.6E",
+			formatArgs:     "1052.032911275",
+			expectedOutput: "1.052033\u202f\u00d7\u202f10\u2070\u00b3",
+		},
+		{
+			name:           "locale support",
+			format:         "%.3f",
+			formatArgs:     "3.14",
+			locale:         "fr_FR",
```
jpbetz, line 486: Shouldn't most (all?) tests have a variant that uses the non-default locale?

ext/strings_test.go
```diff
@@ -329,3 +330,266 @@ func TestStringsWithExtension(t *testing.T) {
...
+		},
+		{
+			name:           "substitution with one escaped percent sign on the left",
+			format:         "%%%s",
+			formatArgs:     `"percent on the left"`,
+			expectedOutput: "%percent on the left",
+		},
+		{
+			name:           "multiple substitutions",
+			format:         "%d %d %d, %s %s %s, %d %d %d, %s %s %s",
```
jpbetz, line 386: Test `uint(<some integer>)` somewhere?

ext/strings.go
```diff
@@ -485,6 +545,293 @@ func join(strs []string) (string, error) {
...
+			return argStr, nil
+		default:
+			return "", fmt.Errorf("string clause can only be used on strings, durations, and timestamps, was given %s", arg.Type().TypeName())
+		}
+	}, nil
+}
+
+func formatDecimal(lastStrIndex int, formatStr string) (int, clauseImpl, error) {
+	return lastStrIndex + 1, func(arg ref.Val, locale string) (string, error) {
+		if arg.Type() != types.IntType {
```
jpbetz: What about uint?

ext/strings_test.go
DangerOnTheRanger: Right, I'll add a uint test and mark this [and this](https://github.com/google/cel-go/pull/617#discussion_r1097646454) as resolved once I do so.

ext/strings.go
DangerOnTheRanger: Fixed - I've changed it so that the tag is checked during env setup.

ext/strings.go
DangerOnTheRanger: `list` and `map` should be added into `%s`. Go has its own formatting clause for booleans, and I'd be comfortable doing the same here. `type` seems like something else `%s` could handle.

ext/strings_test.go
DangerOnTheRanger: Yes,  though it's worth pointing out that only the output of `%e` and `%f` are locale-dependent. It'd still be good to ensure other clauses remain unchanged regardless of locale.

jpbetz: > It'd still be good to ensure other clauses remain unchanged regardless of locale.

This is the primary motivation for my statement. It should be easy to test.

Alfus:
It seems like c++ is following python and others and using "{}" instead of "%". I don't know the backstory about why this transition is being made. I would guess it has to do with the ability to reference arguments by position or name **and** that the arguments are strongly typed, so do not need to be redundantly (e.g. bug-prone) placed in the format string as well. I am curious if this style was considered. See:
https://en.cppreference.com/w/cpp/utility/format/format
https://fmt.dev/latest/index.html

TristonianJones:
The use of % and the printf-style formatting controls are
well-supported across C++, Java, and Go. I'm placing a premium on
familiarity across the supported languages so it's easy to recognize and,
in most cases, copy+paste from one language to another with similar
results. If at some point we choose to support alternative formats, we can
always package up another extension.

-Tristan

ext/strings_test.go
DangerOnTheRanger: (marking as resolved now that i18n-independent tests have an extra check to ensure they don't change with locale)

ext/strings_test.go
DangerOnTheRanger: (marking as resolved, seems everyone is fine with this)

ext/strings.go
DangerOnTheRanger: (marking as resolved, seems everyone thinks this is fine)

ext/strings.go
TristonianJones: The standard for hex in `printf`-like format specifiers is `%x` and `%X`. What's the motivation for `%h/H`?

ext/strings.go
```diff
@@ -44,6 +57,45 @@ import (
...
+// `%o` - substitutes an integer with its equivalent in octal.
+// `%t` - substitutes a boolean with the strings "true" or "false".
+//
+// <string>.format(<list>)` -> <string>
+//
+// Examples:
+//
+//  "this is a string: %s\nand an integer: %d".format(["str", 42]) // returns "this is a string: str\nand an integer: 42"
+// "timestamp: %s".format([timestamp("2023-02-03T23:31:20+00:00")]) // returns "timestamp: 2023-02-03T23:31:20Z"
+// "duration: %s".format([duration("1h45m47s")]) // returns "1h45m47s"
```
TristonianJones: The expression `string(duration("1h45m47s"))`  will emit a string with duration converted to seconds without any other formatting. The reason for this is that conversion from a duration value to a JSON string value in protobuf will render the duration as a value with an `s` suffix:

https://protobuf.dev/programming-guides/proto3/#json

I think it might make sense to follow the same formatting rules for `%s` for all types, otherwise you'll have a different output when using `format` versus when building the string using concatenation and appropriate type conversion.

ext/strings.go
```diff
@@ -44,6 +57,45 @@ import (
...
+//  "%f".format([3.14]) // returns "3.140000"
+//  "scientific notation: %e".format([2.71828]) // returns "scientific notation: 2.718280\u202f\u00d7\u202f10\u2070\u2070"
+//  "5 in binary: %b".format([5]), // returns "5 in binary; 101"
+//  "26 in hex: %h".format([26]), // returns "26 in hex: 1a"
+//  "26 in hex (uppercase): %H".format([26]) // returns "26 in hex (uppercase): 1A"
+//  "30 in octal: %o".format([30]) // returns "30 in octal: 36"
+// "a map inside a list: %s".format([[1, 2, 3, {"a": "x", "b": "y", "c": "z"}]]) // returns "a map inside a list: [1, 2, 3, {a:x, b:y, c:d}]"
```
TristonianJones: We should update the formatting of map / list values that are output when calling `String()` so that they're consistent with the proto3 -> JSON conversion rules where possible. Originally the `Stringer` implementation were just intended to spit out something human readable for the repl tool, but we can do better here.

TristonianJones (changes requested):

ext/strings.go
DangerOnTheRanger: Gotcha, I'll re-work formatting for `%s` according to the proto-to-json rules. If I understand the rules around map keys correctly, the keys should have their contents quoted after being converted to strings - maybe that can land as part of https://github.com/google/cel-go/pull/639.

ext/strings.go
DangerOnTheRanger: (marking as resolved now that `%s` handles lists and maps and `%t` handles booleans)

ext/strings.go
```diff
@@ -485,6 +548,468 @@ func join(strs []string) (string, error) {
...
+	r := rune(formatStr[i])
+	i++
+	switch r {
+	case 's':
+		return i, formatString, nil
+	case 'd':
+		return i, formatDecimal, nil
+	case 'f':
+		return i, formatFixed(precision), nil
+	case 'e', 'E':
```
DangerOnTheRanger: I think it's worth pointing out that `%E` does not differ from `%e` at present since `text/message`'s underlying `%E` implementation never uses the actual E character when doing scientific notation. We could remove `%E` since it seems like this is intended behavior in upstream.

(Hex strings are not passed through the i18n printer, so this isn't a concern for `%x`/`%X`.)

ext/strings.go
DangerOnTheRanger: (marking as resolved now that formatting of map/lists is compatible with https://protobuf.dev/programming-guides/proto3/#json)

ext/strings.go
DangerOnTheRanger: Marking as resolved - `%s` handles strings in addition to `Duration` and `Timestamp`, and both of the latter two follow the proto-to-JSON rules now. Some strings should be safe if double quotes are put around them, but string quoting  will ultimately be solved by #639; bytes are also base64-encoded when formatted as a part of lists/maps, also as per the standard.

ext/strings_test.go
jpbetz: The expected output looks wrong here?

jpbetz: nvm, I see the other comment thread about %s handling.

ext/strings_test.go
```diff
@@ -329,3 +331,338 @@ func TestStringsWithExtension(t *testing.T) {
...
+			formatArgs:     `["abc", 3.14, [9, 8, 7, 6]]`,
+			expectedOutput: `["abc", "3.140000", ["9", "8", "7", "6"]]`,
+			locale:         "en_US",
+		},
+		{
+			name:           "map support for string",
+			format:         "%s",
+			formatArgs:     `{"key1": b"xyz", "key2": duration("2h"), "key4": true, "key3": 2.71828}`,
+			locale:         "nl_NL",
+			expectedOutput: `{"key1":"eHl6", "key2":"7200s", "key3":"2,718280", "key4":true}`,
```
jpbetz: Apologies if I missed this being discussed elsewhere, but what is the rationale for quoting and localizing numbers (rather than representing them as a json number) in map/list output while representing booleans with their proper json type?

EDIT: If https://protobuf.dev/programming-guides/proto3/#json is the north star here, I might be interpreting it slightly differently. I certainly wouldn't expect numbers to be localized. Quoted numbers do seem be be _allowed_, but it is not clear that they are preferred. Curious if we considered dropping quotes from numbers, since this would be less surprising to the vast majority of JSON users.

ext/strings.go
jpbetz: Was this resolved out-of-band? I'm curious what the rationale to keep %h was.

ext/strings_test.go
DangerOnTheRanger: > Curious if we considered dropping quotes from numbers, since this would be less surprising to the vast majority of JSON users.

I thought about it, though at least for floats/doubles `"Nan"` and `"Infinity"` means they'll be strings at least sometime, so I thought it would be less surprising if they were consistently strings. Integers don't have that issue, so that doesn't necessarily apply to them, though I thought it would be weird if one set of numbers were always strings and another set weren't.

RE: locale/surprise, I guess it depends on the intended use case. (Can protobuf deal with comma decimal separators?) My own opinion is that the option of least surprise would be that the locale applies to all numbers (since the user has to explicitly pass it as an option to the env), but if others disagree, I don't think I'd push back too much.

jpbetz: Is the goal to conform to https://protobuf.dev/programming-guides/proto3/#json? This seems reasonable since other languages might then be able to implement this by leveraging existing proto3->json conversion utilities.   If so, I don't think localizing numbers is an option.

For numbers, I'm not convinced. When I see a data dump that is obviously JSON, it is super surprising for me to see types mucked with. When I first saw numbers quoted I was very confused, then I found https://protobuf.dev/programming-guides/proto3/#json and realized it was allowed, but I still don't prefer it. I realize this is part of a a formatting library so there is a temptation to try to do special formatting things, but this is clearly the "data dump" part of that library, and if the data dump format is JSON. Given that I would prefer it be as compliant with JSON as possible. Ideally, I could copy paste the JSON into other tools and it would work as expected. That is not the case once we type swap numbers to strings.

"NaN" and "Infinity" are corner cases that we have to cope with, but for the vast majority of use cases, they are not encountered, so I'm uncomfortable saying that we should make the normal behavior less convenient to accommodate the case. I would just handle those two cases as strings and leave all the other cases as number.

TristonianJones: @jpbetz @DangerOnTheRanger I was off the mark. I saw the unquoted string keys in the map and immediately went to "standard encodings for maps" and thus json encoding by way of proto3; however, this isn't ideal.

CEL value types are a super set of JSON types. The proto3 -> JSON conversion addresses questions about how to handle byte strings, unsigned integers, and numeric map keys in JSON, but there's no reason that the string representation of a map or object couldn't just be however CEL would represent the type as a literal. If the CEL object is valid JSON, then the output would look like JSON and if it's valid CEL, it would look like a CEL literal. 

Thoughts?

jpbetz: Oh, I like the idea of CEL literals.  That would be my preference.

I don't mind if JSON in CEL is formatted as JSON (k8s wouldn't see it, I don't think), I don't have a good sense on if that creates any edge cases.

DangerOnTheRanger: That makes sense to me as well. I'll go ahead and adjust the output such that it's valid CEL.

DangerOnTheRanger: (marking as resolved now that `%s` prints out maps/lists as CEL literals and test cases are included that check that the output is valid CEL)

cel/options.go
```diff
@@ -328,6 +329,18 @@ func registerFiles(reg ref.TypeRegistry, files *protoregistry.Files) error {
 	return err
 }
 
+// Locale configures the Env with the given locale tag.
```
jpbetz: Document what happens if Locale is not explicitly set?

ext/strings.go
```diff
@@ -44,6 +57,46 @@ import (
...
+// `%f` - substitutes a double with fixed-point precision. The default precision is 6, but
+// this can be adjusted. The strings `Infinity`, `-Infinity`, and `NaN` are also valid input
+// for this clause.
+// `%e` - substitutes a double in scientific notation. The default precision is 6, but this
+// can be adjusted.
+// `%E` - same as a above, but with an uppercase E.
+// `%b` - substitutes an integer with its equivalent binary string.
+// `%x` - substitutes an integer with its equivalent in hexadecimal, or if given a string or
+// bytes, will output each character's equivalent in hexadecimal.
+// `%x` - same as above, but with A-F capitalized.
```
jpbetz:
```suggestion
// `%X` - same as above, but with A-F capitalized.
```

ext/strings.go
```diff
@@ -485,6 +550,490 @@ func join(strs []string) (string, error) {
...
+			if err != nil {
+				return "", fmt.Errorf("error writing format string: %w", err)
+			}
+			i++
+		}
+	}
+	return builtStr.String(), nil
+}
+
+func parseFormatArg(lastStrIndex, argIndex int, formatStr string, args traits.Lister, locale string) (int, string, error) {
```
jpbetz, line 623: Somehow document return values? (named args? godoc?)

ext/strings_test.go
```diff
@@ -329,3 +331,398 @@ func TestStringsWithExtension(t *testing.T) {
...
+			out, _, err := program.Eval(cel.NoVars())
+			return out, err
+		}
+		t.Run(tt.name, func(t *testing.T) {
+			expr := fmt.Sprintf(`"%%s".format([%s])`, tt.formatLiteral)
+			literalVal, err := parseAndEval(expr, t)
+			if err != nil {
+				t.Fatalf("program.Eval failed: %v", err)
+			}
+			out, err := parseAndEval(literalVal.Value().(string), t)
```
jpbetz, line 987: Also validate that `literalVal.Value().(string) + " == " + formatLiteral`? This detects a lot of potential problems not detected by checking the return type.

ext/strings_test.go
```diff
@@ -329,3 +331,398 @@ func TestStringsWithExtension(t *testing.T) {
...
+
+func TestLiteralOutput(t *testing.T) {
+	tests := []struct {
+		name          string
+		formatLiteral string
+		expectedType  string
+	}{
+		{
+			name:          "map literal support",
+			formatLiteral: `{"key1": b"xyz", "key2": duration("2h"), "key4": true, "key3": 2.71828}`,
```
jpbetz: Some types are missing here, same for below list case.

ext/strings.go
```diff
@@ -44,6 +57,46 @@ import (
 //	'hello'.charAt(5)  // return ''
 //	'hello'.charAt(-1) // error
 //
+// # Format
+//
+// Returns a new string with substitutions being performed, printf-style.
+// The valid formatting clauses are:
+//
+// `%s` - substitutes a string. This can also be used on lists, maps, bytes, Duration and Timestamp.
```
jpbetz: Are enums supported? (I don't see tests either way)

ext/strings.go
```diff
@@ -44,6 +57,46 @@ import (
...
+// can be adjusted.
+// `%E` - same as a above, but with an uppercase E.
+// `%b` - substitutes an integer with its equivalent binary string.
+// `%x` - substitutes an integer with its equivalent in hexadecimal, or if given a string or
+// bytes, will output each character's equivalent in hexadecimal.
+// `%x` - same as above, but with A-F capitalized.
+// `%o` - substitutes an integer with its equivalent in octal.
+// `%t` - substitutes a boolean with the strings "true" or "false".
+//
+// <string>.format(<list>)` -> <string>
```
jpbetz, line 80: Is it possible to print a type? (e.g. `%T` in Go)

Have we made a clear decision not to print objects? Can we add test case showing that case is an error with `%s` if we are not supporting it?

cc @TristonianJones

ext/strings.go
TristonianJones: If `%E` is non-functional, then it's better to remove it.

ext/strings.go
TristonianJones: Since enums get mapped to integers, I'm not sure that we would expect anything different for enum vs. `int`. This makes rendering enums as human-readable strings difficult without introducing support for strong enums, at least at runtime.

ext/strings.go
TristonianJones: Good question. I think `string(type(<value>))` or `type(<value>)`  as the argument to the format string would be the canonical CEL translation for `%T`, so it's possible for us to support; however, it's not a common convention in C++ or Java. In fact, in Java the `%t`, `%T` refers time-formats.  I'm inclined to leave it out for now. If it's a common use case, we can come back to it.

@jpbetz In terms of objects, you would need a way to iterate over the fields in the object based on the type. It is possible to do this, but not without significant support from the type providers or an overhaul of the `ref.Type`. Perhaps, due to complexity we leave this out, or introduce a new interface that K8s can implement which simplifies iterating over set fields in the object such that it works with the extensions here?

jpbetz: `string(type(<value>))` SGTM

I think I'm OK without any object print support. I agree we can add in utilities to k8s if we want that. A test that verifies that attempting to print an object results in a sane error (and not surprising behavior) would be nice.

ext/strings.go
jpbetz: @DangerOnTheRanger Maybe just a test here showing that enums are not supported?

DangerOnTheRanger: Right, I'll add a test for that.

ext/strings_test.go
DangerOnTheRanger: Fixed, type coverage should be good now.

ext/strings_test.go
DangerOnTheRanger: Fixed - I've added the equality check expression.

ext/strings.go
DangerOnTheRanger: (marking this as resolved, added godoc)

ext/strings.go
DangerOnTheRanger: Whoops, thanks for catching this, fixed

ext/strings.go
DangerOnTheRanger: Makes sense - I've removed %E.

cel/options.go
DangerOnTheRanger: Right, I've added more documentation to `Locale` that explains default behavior.

ext/strings.go
DangerOnTheRanger: Right, I've added a test that shows printing an arbitrary object is unsupported.

ext/strings.go
```diff
@@ -44,6 +57,45 @@ import (
...
+// this can be adjusted. The strings `Infinity`, `-Infinity`, and `NaN` are also valid input
+// for this clause.
+// `%e` - substitutes a double in scientific notation. The default precision is 6, but this
+// can be adjusted.
+// `%b` - substitutes an integer with its equivalent binary string.
+// `%x` - substitutes an integer with its equivalent in hexadecimal, or if given a string or
+// bytes, will output each character's equivalent in hexadecimal.
+// `%X` - same as above, but with A-F capitalized.
+// `%o` - substitutes an integer with its equivalent in octal.
+// `%t` - substitutes a boolean with the strings "true" or "false".
```
TristonianJones: I think I mentioned the `%t` and `%T` could collide with Java `String.format` formatting clauses for time. Wouldn't the boolean case here be covered by our definition of `%s` and `%b`?

ext/strings.go
```diff
@@ -485,6 +549,490 @@ func join(strs []string) (string, error) {
...
+		if err != nil {
+			return "", fmt.Errorf("error matching locale: %w", err)
+		}
+		fmtStr := fmt.Sprintf("%%%de", *precision)
+		return message.NewPrinter(matchedLocale).Sprintf(fmtStr, argFloat), nil
+	}
+}
+
+func formatBinary(arg ref.Val, locale string) (string, error) {
+	argIntVal := arg.ConvertToType(types.IntType)
```
TristonianJones: What about unsigned integers and boolean values?

ext/strings.go
```diff
@@ -485,6 +549,490 @@ func join(strs []string) (string, error) {
...
+		}
+		switch arg.Type() {
+		case types.StringType, types.BytesType:
+			if arg.Type() == types.BytesType {
+				return fmt.Sprintf(fmtStr, arg.Value().([]byte)), nil
+			} else {
+				return fmt.Sprintf(fmtStr, arg.Value().(string)), nil
+			}
+		case types.IntType:
+			argInt, ok := arg.Value().(int64)
```
TristonianJones, line 962: How about unsigned integer values?

ext/strings.go
```diff
@@ -485,6 +549,490 @@ func join(strs []string) (string, error) {
...
+			return fmt.Sprintf("%q", unquotedValue)
+		}
+		return unquotedValue
+	default:
+		return unquotedValue
+	}
+}
+
+func formatString(arg ref.Val, locale string) (string, error) {
+	switch arg.Type() {
```
TristonianJones, line 826: Note: all of the simple types in CEL have conversions to string, so it's possible you could attempt conversion to string for any simple `ref.Val` value.

ext/strings.go
```diff
@@ -485,6 +549,490 @@ func join(strs []string) (string, error) {
...
+	case types.ListType:
+		return formatList(arg, locale)
+	case types.MapType:
+		return formatMap(arg, locale)
+	case types.DurationType:
+		duration := arg.(types.Duration)
+		durationStr := duration.ConvertToType(types.StringType).Value().(string)
+		return durationStr, nil
+	case types.StringType, types.TimestampType, types.BytesType:
```
TristonianJones: You can combine the types.DurationType check into this list without any change in functionality. Also consider supporting `types.BoolType` conversion to string here.

TristonianJones (changes requested):

ext/strings.go
DangerOnTheRanger: That's definitely true, though wouldn't we lose out on potential type-safety (i.e, passing a number to `%d` or `%x` instead)? @jpbetz had wanted (and I was going to implement as a follow-up PR) compile-time checks to make sure the types passed to `string.format` were correct when the expression allows for doing so - I think we talked about that here somewhere.

ext/strings.go
DangerOnTheRanger: Ah, you're right, that case is redundant. I've also merged `%t`'s functionality into `%s` here.

ext/strings.go
TristonianJones: That's a fair point, but if you provide a `list` to `%s` and it contains all numeric values, what's the expected behavior?

You can either go toward restricting the types to make sure they're used appropriately (true for all numeric formatting clauses) or more permissive and allowing as much as possible which can be converted to that formatting clause into a human-readable string. For `%s` I'm inclined to support everything, but can see preferring to direct users toward the numeric formatting controls in those cases since they offer better fidelity / clarity to the person reviewing the code.

ext/strings.go
DangerOnTheRanger: (marking as resolved now that `%s` and `%b` both handle bools)

ext/strings.go
DangerOnTheRanger: Right - I've added support/test cases to cover this, fixed

ext/strings.go
DangerOnTheRanger: Ah, good point - fixed

ext/strings.go
DangerOnTheRanger: When passing a list to `%s` and there's only numeric values, `%d` is used to treat the integers and `%f` is used to treat doubles; I think that's a decent middle ground - for type-checking it's tricky to be as thorough as with scalar types since they can contain more than one type as an element. We could have a dedicated list/map clause, but I think that might be surprising coming from other languages. 

It might follow that adding a dedicated bool clause/adding `%t` back would also make sense, though I think the behavior with `%s` and `%t` right now is fine (though I wouldn't push back if others wanted to go back the `%t` route), and the other types that `%s` supports (like duration/timestamp) are probably too specific to warrant their own formatting clause.

jpbetz: >  For %s I'm inclined to support everything.

The current state of `%s` has been bothering me for a while and I should have flagged it earlier too. My complaint is that it supports selected types, but if you asked a casual developer which types it supports and why, I don't see how they could figure out the rhyme or reason of the types we've shoehorned into to `%s`.

Let's either restrict `%s` to only string, or allow "everything" for `%s`. If we do restrict `%s` to only string, then at the very least need to provide a way to "data dump" lists and maps. This could be provided either with a `%v` type directive or CEL utility functions. If we allow everything, we mostly just need to document the behavior (for most types, I would expect that the behavior matches a direct `string(whatever)` conversion).

ext/strings_test.go
```diff
@@ -329,3 +333,467 @@ func TestStringsWithExtension(t *testing.T) {
...
+			format:     "object is %s",
+			formatArgs: `ext.TestAllTypes{PbVal: test.TestAllTypes{}}`,
+			err:        "error during formatting: string clause can only be used on strings, bools, bytes, maps, lists, durations, and timestamps, was given ext.TestAllTypes",
+		},
+		{
+			name:       "enum is treated as int",
+			format:     "enum is %s",
+			formatArgs: `ext.TestAllTypes{PbVal: test.TestAllTypes{standalone_enum: 1}}.PbVal.standalone_enum`,
+			err:        "error during formatting: string clause can only be used on strings, bools, bytes, maps, lists, durations, and timestamps, was given int",
+		},
```
jpbetz, line 838: Apologies for noticing this so late in the review process, but I just realized we have no tests for `dyn` types. Would you please add? I'd like to be sure that if we don't have static type information that runtime types are still handled well.

ext/strings.go
DangerOnTheRanger: Right - I've added numerical support to `%s`, which is equivalent to everything (it was already used for `Timestamp`/`Duration` before, etc.).

ext/strings_test.go
DangerOnTheRanger: Gotcha, I've added equivalent coverage for `dyn` where applicable.

ext/strings_test.go
```diff
@@ -329,3 +335,673 @@ func TestStringsWithExtension(t *testing.T) {
...
+			name:       "type() not allowed",
+			format:     "type is %s",
+			formatArgs: `type("test string")`,
+			err:        "error during formatting: string clause can only be used on strings, bools, bytes, ints, doubles, maps, lists, durations, and timestamps, was given type",
+		},
+		{
+			name:       "null not allowed for %s",
+			format:     "null: %s",
+			formatArgs: "null",
+			err:        "error during formatting: string clause can only be used on strings, bools, bytes, ints, doubles, maps, lists, durations, and timestamps, was given null_type",
```
jpbetz: Allowing null for %s seems _very_ useful, particularly for optional fields.

ext/strings_test.go
```diff
@@ -329,3 +335,673 @@ func TestStringsWithExtension(t *testing.T) {
...
+			err:        `could not parse formatting clause: unrecognized formatting clause "E"`,
+		},
+		{
+			name:       "object not allowed",
+			format:     "object is %s",
+			formatArgs: `ext.TestAllTypes{PbVal: test.TestAllTypes{}}`,
+			err:        "error during formatting: string clause can only be used on strings, bools, bytes, ints, doubles, maps, lists, durations, and timestamps, was given ext.TestAllTypes",
+		},
+		{
+			name:       "type() not allowed",
```
jpbetz: Add comment here explaining how `string()` can be used to print a type if needed?

ext/strings_test.go
```diff
@@ -329,3 +335,673 @@ func TestStringsWithExtension(t *testing.T) {
...
+		{
+			name:       "uppercase not allowed for scientific clause",
+			format:     "double is %E",
+			formatArgs: "0.5",
+			err:        `could not parse formatting clause: unrecognized formatting clause "E"`,
+		},
+		{
+			name:       "object not allowed",
+			format:     "object is %s",
+			formatArgs: `ext.TestAllTypes{PbVal: test.TestAllTypes{}}`,
```
jpbetz, line 782: What happens if you have a list of objects or a map containing objects?

DangerOnTheRanger: It fails with an error, which I feel is the correct behavior - I've added a couple of tests to make sure this is made clear though.

ext/strings_test.go
DangerOnTheRanger: Gotcha, I've added a comment.

ext/strings_test.go
DangerOnTheRanger: Right, `null` can be used with `%s` now. (Documented this as well.)

jpbetz:
/gcbrun

LGTM

DangerOnTheRanger:
(force push to fix merge conflict)

cel/decls.go
```diff
@@ -564,6 +564,32 @@ func FunctionBinding(binding functions.FunctionOp) OverloadOpt {
 	}
 }
 
+// LocaleOptionBinding conditionally adds a new binding based on locale, passing newOp an additional final
+// argument that represents the configured locale as a string. If no locale was configured, LocaleOptionBinding
+// will do nothing. If no name/overloadID match is made, then LocaleOptionBinding will return an error.
+func LocaleOptionBinding(name, overloadID string, newOp functions.FunctionOp) EnvOption {
```
TristonianJones: Do you need this to be part of the top-level CEL API exposed to K8s? If not, it's probably fine to just have an unexported helper that does this within `ext/strings.go`

TristonianJones: Also, the use is limited enough that you could just declare the three argument signature and proxy to a common implementation.

ext/strings.go
```diff
@@ -485,6 +551,498 @@ func join(strs []string) (string, error) {
...
+	case types.IntType, types.UintType, types.DoubleType,
+		types.BoolType, types.StringType, types.TimestampType, types.BytesType, types.DurationType, types.TypeType:
+		argStrVal := arg.ConvertToType(types.StringType)
+		argStr, ok := argStrVal.Value().(string)
+		if !ok {
+			return "", fmt.Errorf("could not convert argument %q to string", argStrVal)
+		}
+		return argStr, nil
+	case types.NullType:
+		return "<null>", nil
```
TristonianJones: Note, the conversion of null to a string should probably just be `null` since that's how it would be represented as a CEL literal.

jpbetz: +1, I missed this.

cel/decls.go
DangerOnTheRanger: `LocaleOptionBinding` interacts with the unexported `Env.locale` field, so I don't think we could move it to another package without making `locale` exported. We could definitely do that if you think it'd be better to turn it into `Env.Locale` though.

ext/strings.go
DangerOnTheRanger: Right, makes sense - I've changed it to `null`.

cel/decls.go
TristonianJones: I see, I suppose the other option would be to move the Locale to an option passed to `ext.Strings()` which would then allow you to store the parsed locale in to the library instance returned by `ext.Strings()` and make it clear that `Locale` only applies within the `Strings` library for now.

ext/strings_test.go
```diff
@@ -879,15 +878,17 @@ func TestStringFormat(t *testing.T) {
 			),
 		}
 		if locale != "" {
-			opts = append([]cel.EnvOption{cel.Locale(locale)}, opts...)
+			opts = append([]cel.EnvOption{Strings(StringsLocale(locale))}, opts...)
```
TristonianJones: I think you can always set `StringsLocale()` since an empty value for `locale` will just leave the string formatting unaltered.

ext/strings.go
```diff
@@ -250,19 +250,62 @@ const (
...
 // CompileOptions implements the Library interface method.
-func (stringLib) CompileOptions() []cel.EnvOption {
+func (sl stringLib) CompileOptions() []cel.EnvOption {
+	formatOverload := cel.MemberOverload("string_format", []*cel.Type{cel.StringType, cel.ListType(cel.DynType)}, cel.StringType,
+		cel.FunctionBinding(func(args ...ref.Val) ref.Val {
+			s := args[0].(types.String).Value().(string)
+			formatArgs := args[1].(traits.Lister)
+			return stringOrError(stringFormat(s, formatArgs))
+		}))
+	if sl.locale != "" {
```
TristonianJones, line 286: You could consider using a local variable assigned to the defaultLocale, and then testing if the optional locale is set, parsing it, and returning an error or updating the local variable ... this would allow you to provide the local variable to a single `stringFormatWithLocale` overload and preserve the existing behavior.

TristonianJones (changes requested):

ext/strings.go
DangerOnTheRanger: Gotcha, fixed

ext/strings_test.go
DangerOnTheRanger: Right, I've removed that logical branch now.

cel/decls.go
DangerOnTheRanger: (marking as resolved, `LocaleOptionBinding` has been removed and the locale option has been moved into `ext/strings.go`)

ext/strings.go
```diff
@@ -298,12 +299,7 @@ func (sl stringLib) CompileOptions() []cel.EnvOption {
 				},
 			}
 		}
-		formatOverload = cel.MemberOverload("string_format", []*cel.Type{cel.StringType, cel.ListType(cel.DynType)}, cel.StringType,
-			cel.FunctionBinding(func(args ...ref.Val) ref.Val {
-				s := args[0].(types.String).Value().(string)
-				formatArgs := args[1].(traits.Lister)
-				return stringOrError(stringFormatWithLocale(s, formatArgs, sl.locale))
-			}))
+		formatLocale = sl.locale
```
TristonianJones, line 296: This is probably me being paranoid, but could you update the local `formatLocale` immediately after declaration?

TristonianJones (dismissed):

ext/strings.go
DangerOnTheRanger: Gotcha, I'll swap the order of the `if sl.locale != ""` block and the `formatOverload` declaration.

TristonianJones (dismissed):

TristonianJones (dismissed):

TristonianJones (approved):

cici37:
@jpbetz Would you mind checking if the request changes addressed? Looks like the change request is blocking the merging process on this one. Thank you :)

jpbetz (approved):

PR #619, opened 2022-12-28
Ensure that macro call tracking is updated during AST pruning

TristonianJones:
Ensure that macro call tracking is updated during AST pruning as otherwise the pruned expression cannot be successfully rendered back into a human-readable string.

### DIFF
--- a/interpreter/prune.go
+++ b/interpreter/prune.go
@@ -26,6 +26,7 @@ import (
 
 type astPruner struct {
 	expr       *exprpb.Expr
+	macroCalls map[int64]*exprpb.Expr
 	state      EvalState
 	nextExprID int64
 }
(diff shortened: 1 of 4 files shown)


interpreter/prune.go
```diff
@@ -65,13 +66,17 @@ type astPruner struct {
 // compiled and constant folded expressions, but is not willing to constant
 // fold(and thus cache results of) some external calls, then they can prepare
 // the overloads accordingly.
-func PruneAst(expr *exprpb.Expr, state EvalState) *exprpb.Expr {
+func PruneAst(expr *exprpb.Expr, macroCalls map[int64]*exprpb.Expr, state EvalState) *exprpb.ParsedExpr {
```
l46kok, line 69: The return type change here might break consumers, but looks like this is a required modification (probably worth mentioning in the release notes).

l46kok (approved):

interpreter/prune.go
TristonianJones: Agreed. this is a breaking change if anyone is currently using this interface. I'll make sure it's in the release notes.

PR #624, opened 2023-01-05
Fix nested CallExpr cost estimation (#571)

TristonianJones:
* Fix nested CallExpr cost estimation.
* Add TODO about merging isScalar/ComputedSize.

### DIFF
--- a/checker/cost.go
+++ b/checker/cost.go
@@ -476,6 +476,15 @@ func (c *coster) sizeEstimate(t AstNode) SizeEstimate {
 	if l := c.estimator.EstimateSize(t); l != nil {
 		return *l
 	}
+	// return an estimate of 1 for return types of set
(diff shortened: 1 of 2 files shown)


checker/cost.go
```diff
@@ -599,3 +608,20 @@ func (c *coster) newAstNode(e *exprpb.Expr) *astNode {
 	}
 	return &astNode{path: path, t: c.getType(e), expr: e, derivedSize: derivedSize}
 }
+
+// isScalar returns true if the given type is known to be of a constant size at
```
jcking, line 612: Do we need to care about the scalar wrapper types?

jcking (approved):

checker/cost.go
TristonianJones: This is a backport of a fix currently in v0.10.2, but not on v0.10.3 ... Happy to revisit the question.

PR #626, opened 2023-01-09
Potential index out of bounds fix

TristonianJones:
For manually constructed `Expr` values, it's possible to provide invalid indices in the optional indices list which can potentially trigger an index out of bounds error.


interpreter/planner.go
```diff
@@ -519,6 +519,9 @@ func (p *planner) planCreateList(expr *exprpb.Expr) (Interpretable, error) {
 	elements := list.GetElements()
 	optionals := make([]bool, len(elements))
 	for _, index := range optionalIndices {
+		if index > int32(len(elements)) {
```
l46kok: Should this be `>=`?

TristonianJones: Good catch. Yes.

l46kok (approved):

PR #639, opened 2023-01-31
Add strings.quote

DangerOnTheRanger:
This PR adds `strings.quote`, which is intended to quote escape sequences mentioned in the [CEL spec](https://github.com/google/cel-spec/blob/master/doc/langdef.md) and allow the output to be safely copy-pasted into a CEL expression.


ext/strings.go
```diff
@@ -327,6 +327,11 @@ func (stringLib) CompileOptions() []cel.EnvOption {
 					d := delim.(types.String)
 					return stringOrError(joinSeparator(l.([]string), string(d)))
 				}))),
+		cel.Function("escape", cel.MemberOverload("string_escape", []*cel.Type{cel.StringType}, cel.StringType,
```
TristonianJones: Let's call this `quote` since `escape` depends on the context in which you want to escape the string. Also, we should probably follow the convention for encoding that we did for `base64.encode` by making it a globally namespaced `strings.quote` since we might want to introduce `strings.unquote` at a later point as well.

ext/strings.go
```diff
@@ -485,6 +490,35 @@ func join(strs []string) (string, error) {
...
+		case '\f':
+			escapedStrBuilder.WriteString("\\f")
+		case '\n':
+			escapedStrBuilder.WriteString("\\n")
+		case '\r':
+			escapedStrBuilder.WriteString("\\r")
+		case '\t':
+			escapedStrBuilder.WriteString("\\t")
+		case '\v':
+			escapedStrBuilder.WriteString("\\v")
```
TristonianJones: I'm guessing you'd need to escape `\` as well.

ext/strings.go
DangerOnTheRanger: Right, I'll rename it and make it a globally-namespaced function instead of a receiver function and mark this as resolved when I do.

ext/strings.go
DangerOnTheRanger: Right, I'll add a branch and test case for `\`.

ext/strings_test.go
```diff
@@ -99,6 +99,14 @@ var stringTests = []struct {
...
 	{expr: `[].join() == ''`},
 	{expr: `[].join('-') == ''`},
+	// Escaping tests.
+	{expr: `"first\nsecond".escape() == "\"first\\nsecond\""`},
+	{expr: `"bell\a".escape() == "\"bell\\a\""`},
+	{expr: `"\bbackspace".escape() == "\"\\bbackspace\""`},
+	{expr: `"\fform feed".escape() == "\"\\fform feed\""`},
+	{expr: `"carriage \r return".escape() == "\"carriage \\r return\""`},
+	{expr: `"horizontal tab\t".escape() == "\"horizontal tab\\t\""`},
+	{expr: `"vertical \v tab".escape() == "\"vertical \\v tab\""`},
```
cici37: Missing test cases for:
1. string has `\\`
2. string has more than two `\` together like `\a\n`
3. string has `\x` which does not exist
4. string ends with `\`
5. string does not have `\`
You might be able to come up with more :)

ext/strings.go
DangerOnTheRanger: (marking as resolved now that branch/test case have been added)

ext/strings.go
DangerOnTheRanger: (marking as resolved, it's now renamed to `quote` and no longer a receiver function)

ext/strings_test.go
DangerOnTheRanger: Right, I've added all but the 3rd one (non-existent escape sequence), since those are caught at parse-time - for instance, modifying `{expr: `strings.quote("printable unicode😀") == "\"printable unicode😀\""`},` to include a `\j` in the middle gives:

```
--- FAIL: TestStrings (0.04s)
    --- FAIL: TestStrings/[72] (0.00s)
        strings_test.go:302: env.Parse(strings.quote("printable \j unicode😀") == "\"printable unicode😀\"") failed: ERROR: <input>:1:15: Syntax error: token recognition error at: '"printable \j'
             | strings.quote("printable \j unicode😀") == "\"printable unicode😀\""
             | ..............^
            ERROR: <input>:1:36: Syntax error: token recognition error at: '😀'
             | strings.quote("printable \j unicode😀") == "\"printable unicode😀\""
             | ...................................＾
            ERROR: <input>:1:37: Syntax error: mismatched input '") == "' expecting ')'
             | strings.quote("printable \j unicode😀") == "\"printable unicode😀\""
             | ...................................．^
            ERROR: <input>:1:44: Syntax error: token recognition error at: '\'
             | strings.quote("printable \j unicode😀") == "\"printable unicode😀\""
             | ...................................．.......^
FAIL
exit status 1
FAIL    github.com/google/cel-go/ext    0.081s
```

DangerOnTheRanger: (marking as resolved)

ext/strings_test.go
```diff
@@ -105,6 +105,20 @@ var stringTests = []struct {
 	{expr: `['x', 'y'].join('-') == 'x-y'`},
 	{expr: `[].join() == ''`},
 	{expr: `[].join('-') == ''`},
+	// Escaping tests.
```
jpbetz, line 109: This looks right to me, but if you want to be extra sure, you could implement "unquote" and then add a fuzzer test. The fuzzer test could run fuzzed inputs through a quote->unquote to make sure they round trip. I've had good luck finding subtle bugs with this on other escaping utilities.

TristonianJones: Tests that roundtrip in both directions, quoted <-> unquoted, would probably be useful. 

It's also useful to look at src/strconv/quote.go to determine how to handle unprintable characters which are written as unicode escape sequences. I'm not sure we want to support everything that is supported by that method and whether the question is whether we're only handling quoting only Ascii or also all of unicode?

Can you add tests which include quotes mid-string, e.g. `strings.quote('quote"int(42)"')`

ext/strings_test.go
```diff
@@ -105,6 +107,21 @@ var stringTests = []struct {
...
+	{expr: `strings.quote("carriage \r return") == "\"carriage \\r return\""`},
+	{expr: `strings.quote("horizontal tab\t") == "\"horizontal tab\\t\""`},
+	{expr: `strings.quote("vertical \v tab") == "\"vertical \\v tab\""`},
+	{expr: `strings.quote("double \\\\ slash") == "\"double \\\\\\\\ slash\""`},
+	{expr: `strings.quote("two escape sequences \a\n") == "\"two escape sequences \\a\\n\""`},
+	{expr: `strings.quote("verbatim") == "\"verbatim\""`},
+	{expr: `strings.quote("ends with \\") == "\"ends with \\\\\""`},
+	{expr: `strings.quote("\\ starts with") == "\"\\\\ starts with\""`},
+	{expr: `strings.quote("printable unicode😀") == "\"printable unicode😀\""`},
+	{expr: `strings.quote("mid string \" quote") == "\"mid string \\\" quote\""`},
```
TristonianJones, line 123: Could you add a test case where the string is in single quotes and it contains a double-quote? `'single-quote with "double quote"'`

Also, could you add test cases for unicode escape sequences `\U<code>`. I believe the cel-spec has some example strings that might be interesting to use which were just added by @l46kok.

ext/strings_test.go
DangerOnTheRanger: (forgot to mark this as resolved - resolving)

ext/strings.go
```diff
@@ -1084,6 +1090,43 @@ func makeMatcher(locale string) (language.Matcher, error) {
 	return language.NewMatcher(tags), nil
 }
 
+// quote implements a string quoting function. The string will be wrapped in
+// double quotes, and all valid CEL escape sequences will be escaped to show up
+// literally if printed. If the input is not valid UTF-8, quote will return with
+// an error.
+func quote(s string) (string, error) {
+	var quotedStrBuilder strings.Builder
+	if !utf8.ValidString(s) {
```
DangerOnTheRanger: After thinking it over a bit more I'm not so sure this is the right approach. Golang [replaces any invalid runes when quoting](https://go.dev/src/strconv/quote.go), that would probably be better.

ext/strings_test.go
DangerOnTheRanger: Gotcha, I can add that test case.

> Also, could you add test cases for unicode escape sequences \Ucode. I believe the cel-spec has some example strings that might be interesting to use which were just added by @l46kok.

Sure thing, I can do that as well. You're referring to the unicode tests in [string.textproto](https://github.com/google/cel-spec/blob/5564acd4f3ad79cd86a2e96fc4fb4a0981922ad3/tests/simple/testdata/string.textproto), right? I'll port those over and try to add some additional cases on top of those.

ext/strings.go
DangerOnTheRanger: (marking as resolved, invalid UTF-8 characters are now replaced)

ext/strings_test.go
DangerOnTheRanger: (marking as resolved, requested test cases have been added)

TristonianJones (approved):

jpbetz:
Please version gate this and add version test case.

ext/strings.go
```diff
@@ -433,6 +434,11 @@ func (sl *stringLib) CompileOptions() []cel.EnvOption {
 					d := delim.(types.String)
 					return stringOrError(joinSeparator(l.([]string), string(d)))
 				}))),
+		cel.Function("strings.quote", cel.Overload("strings_quote", []*cel.Type{cel.StringType}, cel.StringType,
```
TristonianJones: Also, don't forget to update the doc comment and associated README.md with the same usage information. Thank you!

TristonianJones (changes requested):

ext/strings_test.go
```diff
@@ -1079,3 +1101,217 @@ func mustParseDuration(s string) time.Duration {
...
+func unquote(s string) (string, error) {
+	r := []rune(sanitize(s))
+	if r[0] != '"' || r[len(r)-1] != '"' {
+		return "", fmt.Errorf("expected given string to be enclosed in double quotes: %q", r)
+	}
+	var unquotedStrBuilder strings.Builder
+	noQuotes := r[1 : len(r)-1]
+	for i := 0; i < len(noQuotes); {
+		c := noQuotes[i]
+		hasNext := i < len(noQuotes)
```
jpbetz: This is always 'false'.

This is a good example of a corner cases where you should first add the failing test case, then fix it.

DangerOnTheRanger: Gotcha - I've added a test and a fix.

ext/strings.go
DangerOnTheRanger: Gotcha, godoc at the top of the file and README have both been amended.

jpbetz:
Please add runtime and estimated costs.

ext/strings_test.go
```diff
@@ -105,6 +106,27 @@ var stringTests = []struct {
 	{expr: `['x', 'y'].join('-') == 'x-y'`},
 	{expr: `[].join() == ''`},
 	{expr: `[].join('-') == ''`},
+	// Escaping tests.
+	{expr: `strings.quote("first\nsecond") == "\"first\\nsecond\""`},
```
jpbetz, line 110: Please add an empty string test case to ensure this function is never broken for that edge case.

TristonianJones (approved):

ext/strings_test.go
DangerOnTheRanger: Gotcha, fixed

PR #645, opened 2023-02-21
Add compile-time string.format checking

DangerOnTheRanger:
This PR adds compile-time checking for `string.format` calls that checks the syntax of the format string, the cardinality of the arguments against the format string, and the formatting clauses used against the types of the arguments given, if the format string and argument list are both constants.


cel/program.go
```diff
@@ -207,6 +207,55 @@ func newProgram(e *Env, ast *Ast, opts []ProgramOption) (Program, error) {
...
+			t, err := ExprTypeToType(ast.typeMap[id])
+			if err != nil {
+				return false, err
+			}
+			if t.kind == DynKind {
+				return true, nil
+			}
+			for _, vt := range validTypes {
+				var vtKind Kind
+				switch vt {
```
jpbetz: Can we put this mapping in a shared location? E.g. https://github.com/google/cel-go/blob/master/cel/decls.go ?

interpreter/interpreter.go
```diff
@@ -151,6 +156,185 @@ func CompileRegexConstants(regexOptimizations ...*RegexOptimization) Interpretab
...
+		fmtStrInter, ok := args[0].(InterpretableConst)
+		if !ok {
+			return inter, nil
+		}
+		fmtArgsInter, ok := args[1].(InterpretableConstructor)
+		if !ok {
+			return inter, nil
+		}
+		if fmtArgsInter.Type() != types.ListType {
+			return nil, fmt.Errorf("expected list as second argument to string.format (got %s", fmtArgsInter.Type().TypeName())
```
jpbetz: `return iter, nil` here to tolerate `Dyn` values which may, at runtime, turn out to be valid lists?

interpreter/interpreter.go
```diff
@@ -151,6 +156,185 @@ func CompileRegexConstants(regexOptimizations ...*RegexOptimization) Interpretab
 	return decRegexOptimizer(regexOptimizations...)
 }
 
+type typeVerifier func(int64, ...*types.TypeValue) (bool, error)
+
+// CheckStringFormat checks the syntax and cardinality of any string.format calls present in the expression and reports
+// any errors at compile time.
+func CheckStringFormat(enableCheckArgTypes bool, verifier typeVerifier) InterpretableDecorator {
```
jpbetz: Split out some utility functions here? The size and depth of nesting in this function harms readability.

interpreter/interpreter.go
```diff
@@ -151,6 +156,185 @@ func CompileRegexConstants(regexOptimizations ...*RegexOptimization) Interpretab
...
+			return inter, nil
+		}
+		if fmtArgsInter.Type() != types.ListType {
+			return nil, fmt.Errorf("expected list as second argument to string.format (got %s", fmtArgsInter.Type().TypeName())
+		}
+		formatStr := fmtStrInter.Value().Value().(string)
+		formatArgs := fmtArgsInter.InitVals()
+		curArgIndex := 0
+		i := 0
+		for i < len(formatStr) {
```
jpbetz: It looks like this is a duplication implementation of the lexer? Can a single lexer implementation be shared for both runtime evaluation and compile time checking?  I'm imagining a function that invokes a callback for each directive it encounters that could then be used in both places.

jpbetz (changes requested):

interpreter/interpreter.go
jpbetz: Note that even if a CEL expression is type checked, `Dyn` types can still be provided. You'll want to add a test for this. It would probably be best to write the test that fails first so you know you wrote the test right, then fix the bug.

cel/program.go
DangerOnTheRanger: Right, I've added it to a separate function in `decls`.

interpreter/interpreter.go
DangerOnTheRanger: Fixed

interpreter/interpreter.go
DangerOnTheRanger: I've moved out the datatype-specific checks into their own separate functions, so the layout now is fairly close to what `string.format` has. After the duplicated lexer functionality is removed it should be a bit cleaner, but I'll wait to resolve this until after that refactor is pushed.

interpreter/interpreter.go
DangerOnTheRanger: Fixed, I've refactored so there's only one lexer now.

cici37:
Note for having runtime cost calculated correctly for string.format.

ext/strings.go
```diff
@@ -1062,6 +934,49 @@ func makeMatcher(locale string) (language.Matcher, error) {
 	return language.NewMatcher(tags), nil
 }
 
+type callbackImpl struct {
```
jpbetz: nit: The idiomatic way to state that a struct implemented a interface is `var _ typeName = InterfaceName{}`. Typically the struct is given it's own descriptive name. Maybe `stringFormatter` ?

ext/strings.go
```diff
@@ -1062,6 +934,49 @@ func makeMatcher(locale string) (language.Matcher, error) {
...
+	args traits.Lister
+}
+
+func (c *callbackImpl) String(arg ref.Val, locale string) (string, error) {
+	return formatString(arg, locale)
```
jpbetz: is there and advantage to calling out to a separate function here or can this be inlined?  (same for all the below functions)

ext/strings_test.go
```diff
@@ -338,256 +338,296 @@ func TestStringsWithExtension(t *testing.T) {
...
-		err            string
-		expectedOutput string
+		name               string
+		format             string
+		dynArgs            map[string]any
+		formatArgs         string
+		locale             string
+		err                string
+		expectedOutput     string
+		enableCompileCheck bool
```
jpbetz: nit: If almost all tests should do a compile check, invert this boolean (e.g. `skipCompileCheck`) so it doesn't need to be set repeatedly on almost all the tests?

interpreter/functions/formatting.go
```diff
@@ -0,0 +1,164 @@
...
+	"strings"
+	"unicode"
+
+	"github.com/google/cel-go/common/types/ref"
+)
+
+// FormatCallback is an interface that allows user-defined behavior
+// for formatting clause implementations, as well as argument retrieval.
+type FormatCallback interface {
+	String(ref.Val, string) (string, error)
```
jpbetz: Explain somewhere here what these functions return? From this interface declaration it is not obvious to me. If all the functions have uniform semantics, I'd be fine with it being stated on the type declaration.

interpreter/functions/formatting.go
```diff
@@ -0,0 +1,164 @@
...
+	Decimal(ref.Val, string) (string, error)
+	Fixed(*int) func(ref.Val, string) (string, error)
+	Scientific(*int) func(ref.Val, string) (string, error)
+	Binary(ref.Val, string) (string, error)
+	Hex(bool) func(ref.Val, string) (string, error)
+	Octal(ref.Val, string) (string, error)
+	Arg(int64) (ref.Val, error)
+	ArgSize() int64
```
jpbetz: These are not like the others. What do they do?

interpreter/interpreter.go
```diff
@@ -151,6 +156,229 @@ func CompileRegexConstants(regexOptimizations ...*RegexOptimization) Interpretab
...
+
+func (c *formatCheck) ArgSize() int64 {
+	return int64(len(c.formatArgs))
+}
+
+func isASCIIDigit(r rune) bool {
+	return r <= unicode.MaxASCII && unicode.IsDigit(r)
+}
+
+func verifyRecursive(sub Interpretable, verifier typeVerifier) (bool, error) {
```
jpbetz: nit: instead of encoding in the function name that the function is recursive, use the name to describe what the function does (i.e. what it verifies). If the caller needs to care that the function is recursive, put it in the godoc. If only implementation readers need to know the function is recursive (and it is not obvious), consider using comments in the code.

interpreter/interpreter.go
```diff
@@ -151,6 +156,229 @@ func CompileRegexConstants(regexOptimizations ...*RegexOptimization) Interpretab
...
+		}
+		// use a placeholder locale, since locale doesn't affect syntax
+		_, err := functions.ParseFormatString(formatStr, formatCallback, "en_US")
+		if err != nil {
+			return nil, err
+		}
+		seenArgs := formatCallback.argsRequested
+		if len(formatArgs) > seenArgs {
+			return nil, fmt.Errorf("too many arguments supplied to string.format (expected %d, got %d)", seenArgs, len(formatArgs))
+		}
```
jpbetz: Are excess format args in the arg list ignored? Is that documented?

interpreter/functions/formatting.go
```diff
@@ -0,0 +1,164 @@
...
+}
+
+type clauseImpl func(ref.Val, string) (string, error)
+
+// ParseFormatString takes a format string, FormatCallback implementation, and locale as input,
+// and returns a formatted string if formatting succeeds, or an error otherwise.
```
jpbetz: nit: Instead of repeating information already in the function signature (what the args are), focus more on describing what the function does?

interpreter/functions/formatting.go
jpbetz: And should these be in the same interface? All the format directive handler functions clearly belong together in an interface. But `Arg` and `ArgSize` seem to serve a very different purpose.

cel/program.go
```diff
@@ -207,6 +207,30 @@ func newProgram(e *Env, ast *Ast, opts []ProgramOption) (Program, error) {
 	if len(p.regexOptimizations) > 0 {
 		decorators = append(decorators, interpreter.CompileRegexConstants(p.regexOptimizations...))
 	}
+	// Enable compile-time checking of syntax/cardinality for string.format calls.
+	if p.evalOpts&OptCheckStringFormat == OptCheckStringFormat {
+		enableCheckArgTypes := ast.IsChecked()
+		isValidType := func(id int64, validTypes ...*types.TypeValue) (bool, error) {
```
jpbetz: Does this need to be declared inline? Is there a benefit? Would the code be simpler if it were defined elsewhere?

interpreter/functions/formatting.go
DangerOnTheRanger: It's used during parsing (`ArgSize` here: https://github.com/google/cel-go/pull/645/files#diff-3eaeae7b0bf9e5fba3c8585d57a53a450b0476fedc18fefece60812d54a38c66R90-R92 and `Arg` here: https://github.com/google/cel-go/pull/645/files#diff-3eaeae7b0bf9e5fba3c8585d57a53a450b0476fedc18fefece60812d54a38c66R101) - I agree a separate interface would probably be better (call it `FormatList` or something).

interpreter/functions/formatting.go
```diff
@@ -0,0 +1,164 @@
...
+}
+
+func parsePrecision(lastStrIndex int, formatStr string) (int, *int, error) {
+	i := lastStrIndex
+	if formatStr[i] != '.' {
+		return i, nil, nil
+	}
+	i++
+	var buffer strings.Builder
+	for true {
```
jpbetz:
```suggestion
	for {
```

interpreter/functions/formatting.go
```diff
@@ -0,0 +1,164 @@
...
+	}
+	return builtStr.String(), nil
+}
+
+// parseFormatArg takes the most recently-parsed string index, the next argument index to be used,
+// the complete formatting string,, a FormatCallback implementation, and the locale (or an
+// empty string if no locale was set). It returns the next index in the formatting string that
+// should be parsed, as well as a string containing the formatted argument, or an error if a problem
+// was encountered.
+func parseFormatArg(lastStrIndex, argIndex int, formatStr string, callback FormatCallback, locale string) (int, string, error) {
```
jpbetz: I found this function difficult to understand at first. I believe it could be improved with some combination of:

- Godoc that says more about what the function does.
- A name that is slightly more descriptive (I found the use of the term "arg" confusing-- I think of the args as the values being interpolated into the string, but this seems to be more about the format clauses)
- Reduced coupling with `ParseFormatString`, particularly `lastStrIndex` and `argIndex`. Maybe:
  -  Eliminate `lastStrIndex` by using Go slices with something like `formatStr[i:]` to pass just the substring needed from `ParseFormatString` to `parseFormatArg`? Note that `parseFormatArg` will be returning a count of how many chars were consumed which `ParseFormatString` will then need to use as an offset.
  - Instead of passing in `argIndex`, figure out the right arg in `ParseFormatString` and then just pass in the arg?

jpbetz (commented):
Added some comments, mostly focused on readability and code organization.

interpreter/functions/formatting.go
jpbetz: Separate interface + more documentation SGTM

ext/strings.go
DangerOnTheRanger: Makes sense, I've renamed it to `stringFormatter`.

ext/strings.go
DangerOnTheRanger: Some are used elsewhere (`formatString`/`formatDecimal` are reused for map/list formatting), but I've inlined all the ones that are not.

ext/strings_test.go
DangerOnTheRanger: Makes sense, I've flipped the bool and renamed the field to `skipCompileCheck`.

interpreter/functions/formatting.go
DangerOnTheRanger: Right, I've added docs.

interpreter/functions/formatting.go
DangerOnTheRanger: Gotcha, fixed.

interpreter/interpreter.go
DangerOnTheRanger: Makes sense, I've renamed it to `verifyString`.

interpreter/interpreter.go
DangerOnTheRanger: If I'm understanding you correctly (is something like `"%d %d".format([1, 2, 3])` supported), that's what this case is meant to handle - `len(formatArgs)` is `len([1, 2, 3])` in that example. If there were more formatting clauses (say, 4 total uses of `%d` in that example), then that would be caught in `parseFormatArg`.

It is true that this specific `len` check is not done if compile-time checking is disabled/unavailable - I'll document that.

interpreter/functions/formatting.go
DangerOnTheRanger: Right, I've adjusted it more to describe what it does.

cel/program.go
DangerOnTheRanger: It uses `ast` - we could use a closure in order to continue to use `ast` and define it elsewhere, but I'm not quite sure that's more readable.

interpreter/functions/formatting.go
DangerOnTheRanger: Whoops, fixed

interpreter/functions/formatting.go
DangerOnTheRanger: Makes sense, I've implemented each of those changes.

interpreter/interpreter.go
jpbetz: Gotcha. Should it be `seenClauses` ? I'm seeing a mix of the terms "args" and "clauses" and it can be tedious trying to untangle which means what.

DangerOnTheRanger: You're right, that mix is confusing. I'll rename and do a pass on the PR to see if there's any similar refactors I should perform.

interpreter/functions/formatting.go
```diff
@@ -0,0 +1,184 @@
...
+			}
+			i++
+		}
+	}
+	return builtStr.String(), nil
+}
+
+// subFormatClause parses the format clause at the start of the given string with val, and returns
+// how many characters were consumed and the substituted string form of val, or an error if one occurred.
+func subFormatClause(formatStr string, val ref.Val, callback FormatCallback, list FormatList, locale string) (int, string, error) {
```
jpbetz: nit: `subFormatClause` sounds like a noun to me (unless "sub format" is meant as a verb, but I couldn't figure out what that might mean), maybe something like `parseAndFormatClause`?

interpreter/functions/formatting.go
```diff
@@ -0,0 +1,184 @@
...
+	Arg(int64) (ref.Val, error)
+	// ArgSize returns the length of the argument list.
+	ArgSize() int64
+}
+
+type clauseImpl func(ref.Val, string) (string, error)
+
+// ParseFormatString formats a string according to the string.format syntax, taking the clause implementations
+// from the provided FormatCallback and the args from the given FormatList.
+func ParseFormatString(formatStr string, callback FormatCallback, list FormatList, locale string) (string, error) {
```
jpbetz: `ParseAndInterpolateString` ?

jpbetz: or maybe `InterpolateFormattedString` ?

interpreter/functions/formatting.go
```diff
@@ -0,0 +1,184 @@
...
+
+// FormatCallback is an interface that allows user-defined behavior
+// for formatting clause implementations, as well as argument retrieval.
+// The formatting functions (String, Decimal, etc.) take the a ref.Val
+// and the current locale identifier as input and return the string/substituted
+// form, or an error if one occurred/the given Val is incompatible with the function
+// used. For functions with a differing interface (Fixed/Scientific/etc.), they
+// are detailed individually. Each function is expected to support the appropriate
+// types as laid out in the string.format documentation.
+type FormatCallback interface {
```
jpbetz: `FormatStringInterpolator` ?

interpreter/functions/formatting.go
DangerOnTheRanger: `parseAndFormatClause` sounds good to me, I'll use that.

interpreter/functions/formatting.go
DangerOnTheRanger: Fixed, I went with `InterpolateFormattedString`.

interpreter/functions/formatting.go
DangerOnTheRanger: Makes sense, fixed

interpreter/interpreter.go
DangerOnTheRanger: (marking as resolved, didn't notice anything else)

cel/program.go
```diff
@@ -207,6 +207,30 @@ func newProgram(e *Env, ast *Ast, opts []ProgramOption) (Program, error) {
 	if len(p.regexOptimizations) > 0 {
 		decorators = append(decorators, interpreter.CompileRegexConstants(p.regexOptimizations...))
 	}
+	// Enable compile-time checking of syntax/cardinality for string.format calls.
+	if p.evalOpts&OptCheckStringFormat == OptCheckStringFormat {
+		enableCheckArgTypes := ast.IsChecked()
```
TristonianJones: Instead of passing a flag to indicate whether to check, just use a function that returns trivially true for parsed-only expressions.

interpreter/interpreter.go
```diff
@@ -151,6 +156,230 @@ func CompileRegexConstants(regexOptimizations ...*RegexOptimization) Interpretab
 	return decRegexOptimizer(regexOptimizations...)
 }
 
+type typeVerifier func(int64, ...*types.TypeValue) (bool, error)
```
TristonianJones: Is there any reason we couldn't move this up into the `ProgramOptions()` for the `ext.Strings()` library?

interpreter/interpreter.go
```diff
@@ -151,6 +156,230 @@ func CompileRegexConstants(regexOptimizations ...*RegexOptimization) Interpretab
...
+			return nil, fmt.Errorf("wrong number of arguments to string.format (expected 2, got %d)", len(args))
+		}
+		fmtStrInter, ok := args[0].(InterpretableConst)
+		if !ok {
+			return inter, nil
+		}
+		var fmtArgsInter InterpretableConstructor
+		fmtArgsInter, ok = args[1].(InterpretableConstructor)
+		if !ok {
+			// check for an evalWatch wrapper, since evalWatch implements
```
TristonianJones: This seems like a short-coming of the watcher code. Could you instead create an `evalWatchConstructor` case in `decorators.go` that exposes the right interface for `InterpretableConstructor`? If you do that, then you can move the whole function and associated supporting code into `ext.Strings()` 

Moving the implementation up also allows you to use the cel/decls.go objects in your code rather than have to do multiple conversions between runtime type, expr protobuf type, and cel type.

TristonianJones (changes requested):

interpreter/interpreter.go
DangerOnTheRanger: Wouldn't there be an import cycle even after adding an `evalWatchConstructor` if we moved `InterpolateFormattedString` into `ext/strings.go`? The extension library already imports `"github.com/google/cel-go/cel"`, and if we moved `InterpolateFormattedString`, the flag check would read something like:

```go
        // Enable compile-time checking of syntax/cardinality for string.format calls.
	if p.evalOpts&OptCheckStringFormat == OptCheckStringFormat {
		...
		decorators = append(decorators, strings.InterpolateFormattedString(enableCheckArgTypes, isValidType))
	}
```

Definitely agree on implementing `evalWatchConstructor` though, I'll do that shortly.

interpreter/interpreter.go
DangerOnTheRanger: It's a bit of a hack meant to break a dependency cycle as opposed to a user-configurable option; if the only `typeVerifier` implementation (which is currently in `cel/program.go`) was moved into `interpreter/interpreter.go`, there would be a dependency cycle between `cel`, `interpreter`, and `ext/strings`. I'm for removing it if another option is possible, though (maybe documenting that it's a hack would work?).

cel/program.go
DangerOnTheRanger: I'm not quite sure I understand here - how would we go about doing that? I don't think there's any function on `ref.Val` that would support doing something like that.

interpreter/interpreter.go
TristonianJones: I believe the function `InterpolateFormattedString` doesn't have any internal deps on unexported interpreter functions and types now, so it should be possible to move it into `ext/strings`

cel/program.go
TristonianJones: A `typeVerifier` which returns trivially valid for expressions that haven't been type-checked (`!ast.IsChecked()`) would allow you to remove checks for whether to type-check args later on.

interpreter/interpreter.go
TristonianJones: Ah, I see what you're getting at. You need the AST reference, or some way to resolve type information given an id ... let me think about it. It's very close to being something that can be self contained. Ideally, the verification would happen right after type-check as a new `EnvOption`, but that's an exercise for another time.

interpreter/interpreter.go
TristonianJones: Ideally, you'd be able to return the decorator as part of the `stringLib.ProgramOptions()`; however, you need the ast reference, so you have to defer it until here. This is something I'll have to address later then, but let me see if I can think of any convenient ways of approaching this sooner.

interpreter/decorators.go
```diff
@@ -42,6 +42,11 @@ func decObserveEval(observer EvalObserver) InterpretableDecorator {
 				InterpretableConst: inst,
 				observer:           observer,
 			}, nil
+		case InterpretableConstructor:
+			return &evalWatchConstructor{
```
TristonianJones, line 46: Don't forget to add `*evalWatchConstructor` to the list of early returns at the top of this switch statement.

cel/program.go
DangerOnTheRanger: Oh, makes sense, so something like:

```golang
func typeVerifier(...) bool, error {
    if !ast.IsChecked() {
        return true, nil
    }
     // tests in case we're dealing with a type-checked expression  
}
```

I think the tricky part would be doing this and lifting `typeVerifier` out as you mentioned, since we need AST access for that - I don't have a good answer for that either at the moment unfortunately, but I'll see if I can come up with one.

TristonianJones: More like you could either skip creating this decorator at all if the ast is not checked, or you could create an `isValidType` function that just returns `true, nil`

```
isValidType := func(int64, ...*types.TypeValue) (bool, error) {
   return true, nil
}
if ast.IsChecked() {
  isValidType = ...
}
```

interpreter/decorators.go
DangerOnTheRanger: Right, fixed

cel/program.go
DangerOnTheRanger: (marking as resolved, `checkEnableArgTypes` is now removed in favor of changing the body of `isValidType`)

interpreter/interpreter.go
DangerOnTheRanger: Think this is fine to mark as resolved since `evalWatchConstructor` is implemented. I'll track moving `InterpolateFormattedString` in  https://github.com/google/cel-go/pull/645#discussion_r1122454806.

interpreter/interpreter.go
DangerOnTheRanger: (marking as resolved due to offline discussion to leave the code organized as-is for now, but definitely feel free to comment/unresolve)

DangerOnTheRanger:
(force push to fix merge conflict)

cel/decls.go
```diff
@@ -1204,3 +1204,34 @@ func collectParamNames(paramNames map[string]struct{}, arg *Type) {
 		collectParamNames(paramNames, param)
 	}
 }
+
+func typeValueToKind(tv *types.TypeValue) (Kind, error) {
```
TristonianJones, line 1208: Perhaps you could either map this type to kind mapping complete and adjust the logic later on in program to skip types which cannot be formatted, or rename it and move it closer to where it's referenced in `program.go`. My preference would be to make this more general and update the logic in `program.go`

common/overloads/overloads.go
```diff
@@ -154,6 +154,7 @@ const (
 	EndsWithString   = "ends_with_string"
 	MatchesString    = "matches_string"
 	StartsWithString = "starts_with_string"
+	FormatString     = "string_format"
```
TristonianJones: This file is really only intended for the common set of overloads used in the standard environment, but since you're referring to the name in `interpreter` as well for runtime cost calculations, maybe let's make it clear this is an extension: `ExtFormatString` with a comment that indicates why it's here, e.g. `Extension function overloads with complex behaviors that need to be referenced in runtime and static analysis cost computations`

interpreter/functions/formatting.go
```diff
@@ -0,0 +1,183 @@
...
+	"errors"
+	"fmt"
+	"strconv"
+	"strings"
+	"unicode"
+
+	"github.com/google/cel-go/common/types/ref"
+)
+
+// FormatStringInterpolator is an interface that allows user-defined behavior
```
TristonianJones: Consider consolidating all of the formatting related code into `interpreter/formatting.go`. Since it's an extension which needs to exist here due to a deficit in the cel-go design related to argument validation, I'd like to have most of this functionality in one place to make sure it's simple to hoist into `ext` once the design issues have been addressed.

TristonianJones (changes requested):
I just have a few last code organization things to make updates simpler in the future, and I think we're good to go.

interpreter/functions/formatting.go
DangerOnTheRanger: Gotcha, `functions/formatting.go` has been folded into `interpreter/formatting.go`.

common/overloads/overloads.go
DangerOnTheRanger: Right, I've renamed `FormatString` to `ExtFormatString`, and added a comment as well.

cel/decls.go
DangerOnTheRanger: What did you have in mind in terms of making `typeValueToKind` more general? I think all types with both a defined `Kind` and `TypeValue` singleton are already supported - there's `DynKind` but no equivalent `DynType` singleton for instance.

TristonianJones: I don't believe that `DynType`, `ObjectType`, or `OptionalType` and a few others are supported in this list because they're not supported as string formatting arguments, so it's worth either supporting all type to kind mappings here, or moving the function to a place where this mapping of type to kind is suitable for only those types which support string formatting.

DangerOnTheRanger: It didn't occur to me to use `TypeValue.TypeName()` to attempt to map `TypeValue` to `Kind` in cases where a singleton isn't available, though this still leaves `StructKind` and `TypeParamKind` without a mapping since as far as I can gather, both of those have caller-set values for their names. Is there some other solution for those two `Kind` cases?

ext/strings.go
```diff
@@ -1000,8 +1005,9 @@ func formatHex(useUpper bool) clauseImpl {
 		case types.StringType, types.BytesType:
 			if arg.Type() == types.BytesType {
 				return fmt.Sprintf(fmtStr, arg.Value().([]byte)), nil
+			} else {
```
TristonianJones: nit: please revert to the de-dented return below

interpreter/formatting.go
```diff
@@ -0,0 +1,372 @@
...
+	// so precision can be set without needing an additional function call/configuration.
+	Fixed(*int) func(ref.Val, string) (string, error)
+	// Scientific functions identically to Fixed, except the string returned from the closure
+	// is expected to be in scientific notation.
+	Scientific(*int) func(ref.Val, string) (string, error)
+	Binary(ref.Val, string) (string, error)
+	// Hex takes a boolean that, if true, indicates the hex string output by the returned
+	// closure should use uppercase letters for A-F.
+	Hex(bool) func(ref.Val, string) (string, error)
+	Octal(ref.Val, string) (string, error)
```
TristonianJones, line 249: Would you mind documenting the String / Decimal / Binary / Octal methods below and place a newline between each method.

interpreter/formatting.go
```diff
@@ -0,0 +1,372 @@
...
+	i := 0
+	if formatStr[i] != '.' {
+		return i, nil, nil
+	}
+	i++
+	var buffer strings.Builder
+	for {
+		if i >= len(formatStr) {
+			return -1, nil, errors.New("could not find end of precision specifier")
+		} else if !isASCIIDigit(rune(formatStr[i])) {
```
TristonianJones: The previous line returns, consider de-denting

ext/strings.go
DangerOnTheRanger: Gotcha, I've de-dented this block.

interpreter/formatting.go
DangerOnTheRanger: Gotcha, I've de-dented this as well (so it's just `if !isASCIIDigit(rune(formatStr[i]))...` now)

interpreter/formatting.go
DangerOnTheRanger: Gotcha, I've added documentation for the missing functions and also formatted the interface definition with the appropriate newlines.

TristonianJones (dismissed):

jpbetz:
Is estimated cost implemented for this function? I was expecting to see a test case in `https://github.com/google/cel-go/blob/master/checker/cost_test.go`, but I don't see on in this PR or on `master`.

jpbetz (approved):
LGTM

/gcbrun

PR #649, opened 2023-02-28
Revert go-genproto api update that is incompatible with Kubernetes

TristonianJones:
Revert the go-genproto api update which is incompatible with Kubernetes builds since it brings in a denylisted dependency. The change has no effect the main CEL module, and dependent submodules which care about the update will receive it via minimum version selection.


TristonianJones:
@DangerOnTheRanger FYI

jpbetz (approved):
LGTM

DangerOnTheRanger:
LGTM as well.

cici37:
Thank you!
/lgtm

PR #650, opened 2023-03-01
Add version option to Strings extension library

jpbetz:
Adds a `StringsVersion()` option for the string extension library, a convention for declaring which version new functions are introduced at, and a way for exhaustively testing the versions of a library for function support.

### DIFF
--- a/ext/strings.go
+++ b/ext/strings.go
@@ -59,12 +59,16 @@ const (
 //
 // # Format
 //
+// Introduced at version: 1
+//
 // Returns a new string with substitutions being performed, printf-style.
 // The valid formatting clauses are:
 //
-// `%s` - substitutes a string. This can also be used on bools, lists, maps, bytes, Duration and Timestamp, in addition to all numerical types (int, uint, and double).
-//	Note that the dot/period decimal separator will always be used when printing a list or map that contains a double,
-// and that null can be passed (which results in the string "null") in addition to types.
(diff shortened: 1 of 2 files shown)


ext/strings.go
```diff
@@ -251,15 +257,16 @@ const (
 //	'TacoCat'.upperAscii()      // returns 'TACOCAT'
 //	'TacoCÆt Xii'.upperAscii()  // returns 'TACOCÆT XII'
 func Strings(options ...StringsOption) cel.EnvOption {
-	s := &stringLib{}
+	s := &stringLib{version: math.MaxUint32}
```
TristonianJones, line 258: I didn't notice it before, but could you update all references to member functions on the `stringLib` to use `*stringLib` instead?

ext/strings.go
```diff
@@ -280,6 +287,20 @@ func StringsLocale(locale string) StringsOption {
 	}
 }
 
+// Version configures the version of the string library. The version limits which
```
TristonianJones: Could you update this to `StringsVersion` since other extensions in the `ext` package will likely need to use similar techniques going forward.

ext/strings.go
jpbetz: Done!

ext/strings.go
jpbetz: Done

TristonianJones (approved):

PR #651, opened 2023-03-02
Fix regression of cost behavior for has()

DangerOnTheRanger:
This PR adds a hack that records the presence of `fieldType` on `evalTestOnly` in order to report a different cost when `fieldType` is present.

### DIFF
--- a/interpreter/interpretable.go
+++ b/interpreter/interpretable.go
@@ -110,11 +110,10 @@ type InterpretableConstructor interface {
 // Core Interpretable implementations used during the program planning phase.
 
 type evalTestOnly struct {
-	id           int64
-	attr         InterpretableAttribute
-	qual         Qualifier
-	field        types.String
-	isProtoField bool
+	id    int64
+	attr  InterpretableAttribute
+	qual  Qualifier
+	field types.String
 }
 
 // ID implements the Interpretable interface method.
--- a/interpreter/interpretable.go
+++ b/interpreter/interpretable.go
@@ -110,10 +110,11 @@ type InterpretableConstructor interface {
 // Core Interpretable implementations used during the program planning phase.
 
 type evalTestOnly struct {
-	id    int64
-	attr  InterpretableAttribute
-	qual  Qualifier
-	field types.String
+	id           int64
+	attr         InterpretableAttribute
+	qual         Qualifier
+	field        types.String
+	isProtoField bool
 }
 
 // ID implements the Interpretable interface method.
--- a/interpreter/runtimecost.go
+++ b/interpreter/runtimecost.go
@@ -38,6 +38,12 @@ type ActualCostEstimator interface {
 func CostObserver(tracker *CostTracker) EvalObserver {
 	observer := func(id int64, programStep any, val ref.Val) {
 		switch t := programStep.(type) {
+		case *evalTestOnly:
+			// Protobuf has() checks do not incur runtime cost due to not using Eval()
+			// in older versions of CEL; this keeps the runtime cost behavior consistent
+			if !t.isProtoField {
+				tracker.cost++
+			}
 		case ConstantQualifier:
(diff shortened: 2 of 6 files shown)


interpreter/interpretable.go
```diff
@@ -1277,6 +1279,38 @@ func (a *evalAttr) Resolve(ctx Activation) (any, error) {
 	return a.attr.Resolve(ctx)
 }
 
+type evalWatchConstructor struct {
```
jpbetz: Where is this allocated and used?

interpreter/interpretable.go
```diff
@@ -114,6 +114,8 @@ type evalTestOnly struct {
 	attr  InterpretableAttribute
 	qual  Qualifier
 	field types.String
+	// hack attribute to preserve runtime cost behavior (see CostObserver for more)
+	protoHack bool
```
jpbetz: Maybe don't call this out as a hack in the name of the field? Doing so does not aid the readability of the code, and unless we have some plan to eliminate this hack later, I'd prefer to make this edit as clean as possible. The purpose of this field is to identify that the `field` field is a protobuf field. Maybe something like `isProtoField` ?

interpreter/runtimecost.go
```diff
@@ -38,6 +38,12 @@ type ActualCostEstimator interface {
 func CostObserver(tracker *CostTracker) EvalObserver {
 	observer := func(id int64, programStep any, val ref.Val) {
 		switch t := programStep.(type) {
+		case *evalTestOnly:
+			// Hack to avoid incrementing runtime cost when using protobufs
```
jpbetz: "Protobuf has() checks to not incur runtime cost because {some amazing reason}" ?

interpreter/interpretable.go
DangerOnTheRanger: Makes sense - I've renamed it to `isProtoField`.

interpreter/interpretable.go
DangerOnTheRanger: Terribly sorry, that somehow came from https://github.com/google/cel-go/pull/645. I've double-checked to make sure there's no other code from other branches.

interpreter/runtimecost.go
```diff
@@ -38,6 +38,12 @@ type ActualCostEstimator interface {
 func CostObserver(tracker *CostTracker) EvalObserver {
 	observer := func(id int64, programStep any, val ref.Val) {
 		switch t := programStep.(type) {
+		case *evalTestOnly:
+			// Protobuf has() checks do not incur runtime cost due to not using Eval()
```
TristonianJones: The `evalTestOnly` instruction should contain an `InterpretableAttribute` and a qualifier. The `InterpretableAttribute` in the operand will have had it's cost tracked already, but the presence test check should probably still count the same as `common.SelectAndIdentCost` and not differentiate based on whether it's a proto field or not.

interpreter/runtimecost_test.go
```diff
@@ -306,7 +306,7 @@ func TestRuntimeCost(t *testing.T) {
 			name:  "select: field test only",
```
jpbetz, line 306: Please also add test case for behavior that regressed to prevent future regressions.

interpreter/runtimecost.go
DangerOnTheRanger: (resolving since this is outdated)

interpreter/runtimecost.go
DangerOnTheRanger: Gotcha, I've used `common.SelectAndIdentCost` and updated the relevant test cases, should be fixed now

TristonianJones (dismissed):

interpreter/runtimecost_test.go
DangerOnTheRanger: Gotcha, I've added an additional test - fixed

jpbetz (approved):
LGTM

PR #653, opened 2023-03-08
Enable extensions for REPL

l46kok:
This allows the usage of [all extensions in REPL](https://github.com/google/cel-go/blob/master/ext/README.md).

Example:

```
cel-repl> %option --extension 'strings'
cel-repl> 'hello world'.substring(6)
world : string
```

Available extension options:

```
%option --extension 'strings'
%option --extension 'math'
%option --extension 'protos'
%option --extension 'encoders'
%option --extension 'all' (adds all 4 above)
```

### DIFF
--- a/repl/evaluator.go
+++ b/repl/evaluator.go
@@ -720,23 +720,20 @@ func (e *Evaluator) setOption(args []string) error {
 	for idx := 0; idx < len(args); {
 		arg := args[idx]
 		idx++
-		if arg == "--container" {
-			if idx >= len(args) {
-				issues = append(issues, "not enough args for container")
-			}
-			container := args[idx]
+		switch arg {
+		case "--container":
+			err := e.loadContainerOption(idx, args)
 			idx++
-			err := e.AddOption(&containerOption{container: container})
 			if err != nil {
 				issues = append(issues, fmt.Sprintf("container: %v", err))
 			}
-		} else if arg == "--extension" {
+		case "--extension":
 			err := e.loadExtensionOption(idx, args)
 			idx++
 			if err != nil {
 				issues = append(issues, fmt.Sprintf("extension: %v", err))
 			}
-		} else {
+		default:
 			issues = append(issues, fmt.Sprintf("unsupported option '%s'", arg))
 		}
 	}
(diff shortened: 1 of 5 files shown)


repl/evaluator.go
```diff
@@ -430,7 +431,12 @@ type Evaluator struct {
 
 // NewEvaluator returns an inialized evaluator
 func NewEvaluator() (*Evaluator, error) {
-	env, err := cel.NewEnv()
+	env, err := cel.NewEnv(
+		ext.Strings(),
```
jnthntatum: It might be better to make this configurable since most clients are going to start with the standard env. Ok to defer, but update the README to note that these are enabled by default

jnthntatum:
can you check on the cloud build failure? maybe the build files need to be fixed.

repl/evaluator.go
```diff
@@ -430,7 +431,12 @@ type Evaluator struct {
...
+		ext.Protos(),
+		ext.Maths(),
```
jnthntatum: add a simple test that the base environment includes these extensions

repl/evaluator.go
l46kok: Done

repl/evaluator.go
l46kok: Done

l46kok:
PTAL

jnthntatum (dismissed):

TristonianJones:
@l46kok as a convenience could you add `--extension 'all'`

Also, is it possible to disable the extensions once set?

l46kok:
> @l46kok as a convenience could you add `--extension 'all'`
> 
> Also, is it possible to disable the extensions once set?

Done

The extensions can be disabled by `%reset`, which resets the environment to initial state.

repl/evaluator.go
```diff
@@ -697,6 +730,12 @@ func (e *Evaluator) setOption(args []string) error {
 			if err != nil {
 				issues = append(issues, fmt.Sprintf("container: %v", err))
 			}
+		} else if arg == "--extension" {
```
TristonianJones: nit: You could consider making this a `switch`

repl/evaluator.go
```diff
@@ -707,6 +746,41 @@ func (e *Evaluator) setOption(args []string) error {
 	return nil
 }
 
+func (e *Evaluator) loadExtensionOption(idx int, args []string) error {
+	if idx >= len(args) {
```
TristonianJones, line 747: Consider a helper function which tests the input argument count against the index and returns an error. You'll want to use the same function between this call and the `--container` check since it looks like there's a bug where not enough args are detected for the container, but there's still an array access on the invalid index.

If it's always the case that an option will be a pair of args, then consider running this check at the beginning of the for-loop whether the option name is supported or not.

repl/evaluator.go
l46kok: Done

repl/evaluator.go
l46kok: Done. I've left the check inside the loop to allow future options to accept any number of arguments.

TristonianJones (approved):

PR #664, opened 2023-03-22
Add estimated cost support to strings.quote

DangerOnTheRanger:
The initial PR that implemented strings.quote (#639) did not add estimated cost support for the function, which is fixed in this PR. This PR also changes the string constant case for `astNode.ComputedSize` to use runes instead of bytes, which was needed to make the estimated cost for `strings.quote` equivalent with the runtime cost (which already used runes). This change didn't break any other tests, so it should be safe.


jpbetz:
LGTM

/gcbrun

jpbetz (approved):

TristonianJones (approved):

PR #671, opened 2023-03-27
Additional coverage for static and runtime cost computations

TristonianJones:
With the possibility of variadic logical operator ASTs, improve coverage for and/or branches which have an associated cost.


jpbetz (approved):
LGTM

This looks correct

PR #678, opened 2023-04-06
Fix for type guard on optimized set membership step

TristonianJones:
There's a missing check for error / unknown before accessing the pre-computed value map


jnthntatum (changes requested):
Could you add a motivating test case for coverage?

TristonianJones:
Thanks! PTAL.

jnthntatum (approved):

PR #684, opened 2023-04-19
Add ProtoToMap utility to complement DeclareContextProto.

dstephan1:
DeclareContextProto() exists, but there is no obvious way to actually apply a proto to it.

I'm not sure where exactly to put this.

### DIFF
--- a/cel/io.go
+++ b/cel/io.go
@@ -278,3 +278,18 @@ func ValueToRefValue(adapter ref.TypeAdapter, v *exprpb.Value) (ref.Val, error) 
 	}
 	return nil, errors.New("unknown value")
 }
+
+// ProtoToMap creates an activation for a program Eval() based on a context
+// proto. Useful in conjunction with DeclareContextProto.
+func ProtoToMap(m proto.Message) map[string]any {
+	fields := make(map[string]any)
+	if !m.ProtoReflect().IsValid() {
+		return nil
+	}
+	fds := m.ProtoReflect().Descriptor().Fields()
+	for i := 0; i < fds.Len(); i++ {
+		fd := fds.Get(i)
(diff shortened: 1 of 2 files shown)


cel/io.go
```diff
@@ -278,3 +278,18 @@ func ValueToRefValue(adapter ref.TypeAdapter, v *exprpb.Value) (ref.Val, error)
 	}
 	return nil, errors.New("unknown value")
 }
+
+// ProtoToMap creates an activation for a program Eval() based on a context
+// proto. Useful in conjunction with DeclareContextProto.
+func ProtoToMap(m proto.Message) map[string]any {
```
TristonianJones, line 284: Sorry for the delayed review. For consistency, let's use `BindProtoToActivation` and return an (`interpreter.Activation`, error) as the results.

cel/io.go
```diff
@@ -278,3 +278,18 @@ func ValueToRefValue(adapter ref.TypeAdapter, v *exprpb.Value) (ref.Val, error)
...
+	fields := make(map[string]any)
+	if !m.ProtoReflect().IsValid() {
+		return nil
```
TristonianJones, line 287: Consider erroring here instead.

cel/io.go
```diff
@@ -278,3 +278,18 @@ func ValueToRefValue(adapter ref.TypeAdapter, v *exprpb.Value) (ref.Val, error)
...
+	}
+	fds := m.ProtoReflect().Descriptor().Fields()
+	for i := 0; i < fds.Len(); i++ {
```
TristonianJones, line 290: I think in C++ there's an option to omit unset fields. You may want to provide a similar option here.

TristonianJones (changes requested):

PR #692, opened 2023-04-29
Update the ConvertToNative API documentation and strings.join

### LINKED ISSUE #688: Unexpected behavior: string.join relies on ConvertToNative returning a []string
**Describe the bug**

In k8s we encountered https://github.com/kubernetes/kubernetes/issues/117590.

We had [expected](https://github.com/kubernetes/kubernetes/blob/1d116c437c2ad531a36bd9aabef165f5efa0f0e9/staging/src/k8s.io/apiserver/pkg/cel/common/values.go#L425-L431) that if `ConvertToNative` is asked for a `reflect.Type` of `Slice` that `[]interface{}` would be an appropriate return type, although the godoc doesn't spell this out:

https://github.com/google/cel-go/blob/377a0bba20d07926e0583b4e604509ca7f3583b7/common/types/ref/reference.go#L38-L40

**To Reproduce**

https://github.com/kubernetes/kubernetes/issues/117590 provides a reproduction.

**Expected behavior**

- `ConvertToNative()` expects `[]interface{}` not `[]string`, or that `ConvertToNative()` defines more clearly the expected behavior in the godoc.
- `string.join()` does not rely on `ConvertToNative()` and instead operates on `ref.Val`.

**Additional context**

I'll put a workaround in k8s.  I'm open to discussion on what the right way to fix this is.  It would be great if we can establish clear rules on:

- What exactly `ConvertToNative` is expected to return in all cases
- If functions should be call `ConvertToNative()`

TristonianJones:
Update the `ConvertToNative` API docs to clarify the call contract. 

The change also introduces a refactor of the `strings.join` method which
does not rely on `ConvertToNative`, but rather uses the `traits.Lister` with
an expectation that all contained elements will be `types.String` values.

This change can be taken advantage of using `ext.StringsVersion(2)` if
you are using extension versioning. If you are not using extension versioning,
the change will be default enabled.

The functionality should be equivalent with the prior behavior, though there
are some limited cases where a `ConvertToNative` might succeed when
the new approach might fail.

Closes #688

### DIFF
--- a/common/types/ref/reference.go
+++ b/common/types/ref/reference.go
@@ -37,9 +37,18 @@ type Type interface {
 type Val interface {
 	// ConvertToNative converts the Value to a native Go struct according to the
 	// reflected type description, or error if the conversion is not feasible.
+	//
+	// The ConvertToNative method is intended to be used to support conversion between CEL types
+	// and native types during object creation expressions or by clients who need to adapt the,
+	// returned CEL value into an equivalent Go value instance.
+	//
+	// When implementing or using ConvertToNative, the following guidelines apply:
+	// - Use ConvertToNative when marshalling CEL evaluation results to native types.
+	// - Do not use ConvertToNative within CEL extension functions.
+	// - Document whether your implementation supports non-CEL field types, such as Go or Protobuf.
 	ConvertToNative(typeDesc reflect.Type) (any, error)
 
-	// ConvertToType supports type conversions between value types supported by the expression language.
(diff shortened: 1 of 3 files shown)


TristonianJones:
@jpbetz let me know if this addresses your use case, and also whether the updated functionality will work with your object types.

jpbetz (approved):
This looks great. Thanks @TristonianJones !

PR #694, opened 2023-04-29
Additional string.format() test for message field accesses

### LINKED ISSUE #690: OptCheckStringFormat causes Program to fail generation if format argument has field access
**Describe the bug**

With `ext.Strings`, setting the `OptCheckStringFormat` option when generating the `Program` fails if the argument to the format string refers to a field on a message.

**To Reproduce**

- [x]  parser
- [x]  checker
- [ ]  interpreter

Test setup:
```go
func Test(t *testing.T) {
	// message Msg {
	//   double fld = 1;
	// }
	msg := &Msg{Fld: 1.23}

	expr := "'%s'.format([msg.fld])"

	env, err := cel.NewEnv(
		ext.Strings(),
		cel.Types(msg),
		cel.Variable("msg",
			cel.ObjectType(string(msg.ProtoReflect().Descriptor().FullName()))),
	)
	require.NoError(t, err)

	ast, issues := env.Compile(expr)
	require.NoError(t, issues.Err())

	prog, err := env.Program(ast, cel.EvalOptions(
		cel.OptOptimize,
		cel.OptCheckStringFormat,
	))
	require.NoError(t, err)

	val, _, err := prog.Eval(map[string]any{"msg": msg})
	require.NoError(t, err)
	t.Logf("%#v", val)
}
```

**Actual behavior**

```
=== RUN   Test
    validator_test.go:59: 
        	Error:      	Received unexpected error:
        	            	error during formatting: string clause can only be used on strings, bools, bytes, ints, doubles, maps, lists, types, durations, and timestamps
--- FAIL: Test (0.00s)
```

**Expected behavior**

`"1.23"`

**Additional context**

Test passes with either modification:

- Remove `cel.OptCheckStringFormat`
- Replace `msg` type with `cel.DynType`

TristonianJones:
Refined the test approach in `ext/strings_test.go` to better cover type-based validation of format args.

Fixed an issue where attribute ids were referring to the variable id rather than the leaf value of the attribute.
This has a minor impact on attribute state tracking which should not affect semantic meaning, but which
does have a real impact on whether or not the attribute expression corresponds to the type identifier in
the `checked.proto` `type_map` field.

Closes #690

### DIFF
--- a/ext/strings_test.go
+++ b/ext/strings_test.go
@@ -942,6 +942,19 @@ func TestStringFormat(t *testing.T) {
 			expectedEstimatedCost: checker.CostEstimate{Min: 13, Max: 13},
 		},
 		{
+			name:       "message field support",
+			format:     "message field msg.single_int32: %d, msg.single_double: %.1f",
+			formatArgs: `msg.single_int32, msg.single_double`,
+			dynArgs: map[string]any{
+				"msg": &proto3pb.TestAllTypes{
+					SingleInt32:  2,
+					SingleDouble: 1.0,
+				},
+			},
+			locale:         "en_US",
+			expectedOutput: `message field msg.single_int32: 2, msg.single_double: 1.0`,
+		},
+		{
 			name:             "unrecognized formatting clause",
 			format:           "%a",
 			formatArgs:       "1",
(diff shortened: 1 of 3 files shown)


TristonianJones:
FYI @jpbetz

jpbetz (approved):
This looks right to me.

I'm far from an expert on the attributes aspect, but it looks sane.

Thanks for finding and fixing!

PR #703, opened 2023-05-12
AST scan to establish a non-conflicting expression ID for pruning

### LINKED ISSUE #699: PruneAST overwrites existing nodes and produces an invalid AST
**Describe the bug**
Calling `interpreter.PruneAST` function produces an incorrect AST if it decides to replace an  expression with a constant. This is because the ID for the new constant node clashes with an existing node ID and ends up overwriting it.

For example, consider the expression `foo == "bar" && R.attr.loc in ["GB", "US"]`. If I set the variable `foo` to "bar" and partially evaluate the expression, I'd expect to get back `R.attr.loc in ["GB", "US"]`. However, with the change introduced in #677 and released in version 0.15.0, the output is `"US".loc in ["GB", "US"]`. 

**To Reproduce**

- [ ]  parser
- [ ]  checker
- [x]  interpreter

```cel
foo == "bar" && R.attr.loc in ["GB", "US"]
```

Test setup:
```go
package main

import (
	"fmt"
	"log"

	expr "google.golang.org/genproto/googleapis/api/expr/v1alpha1"
)

func main() {
	env, err := cel.NewEnv(cel.Declarations(
		decls.NewVar("R", decls.NewMapType(decls.String, decls.Dyn)),
		decls.NewVar("foo", decls.String),
	))
	exitOnErr(err)

	pvars, err := cel.PartialVars(map[string]any{"foo": "bar"}, cel.AttributePattern("R").QualString("attr").Wildcard())
	exitOnErr(err)

	ast, iss := env.Compile(`foo == "bar" && R.attr.loc in ["GB", "US"]`)
	exitOnErr(iss.Err())

	prg, err := env.Program(ast, cel.EvalOptions(cel.OptTrackState, cel.OptPartialEval))
	exitOnErr(err)

	_, details, err := prg.Eval(pvars)
	exitOnErr(err)

	pruned := interpreter.PruneAst(ast.Expr(), ast.SourceInfo().GetMacroCalls(), details.State()).Expr
	prunedAST := cel.ParsedExprToAst(&expr.ParsedExpr{Expr: pruned})
	fmt.Println(cel.AstToString(prunedAST))

}

func exitOnErr(err error) {
	if err != nil {
		log.Fatalf("Error: %v", err)
	}
}

```

**Expected behavior**

The output should be: 

```cel
R.attr.loc in ["GB", "US"]
```

**Additional context**

I believe the problem stems from the fact that `nextExprId` is set to 1 in the `astPruner`.

https://github.com/google/cel-go/blob/75ed853e7e1f4db40fd499ee360b9c23e2e571af/interpreter/prune.go#L75-L79

If I apply the following patch, everything works as expected.

```patch
diff --git a/interpreter/prune.go b/interpreter/prune.go
index 24c7e79..342de6e 100644
--- a/interpreter/prune.go
+++ b/interpreter/prune.go
@@ -68,15 +68,19 @@ type astPruner struct {
 // the overloads accordingly.
 func PruneAst(expr *exprpb.Expr, macroCalls map[int64]*exprpb.Expr, state EvalState) *exprpb.ParsedExpr {
 	pruneState := NewEvalState()
+	var maxID int64 = 0
 	for _, id := range state.IDs() {
 		v, _ := state.Value(id)
 		pruneState.SetValue(id, v)
+		if id > maxID {
+			maxID = id
+		}
 	}
 	pruner := &astPruner{
 		expr:       expr,
 		macroCalls: macroCalls,
 		state:      pruneState,
-		nextExprID: 1}
+		nextExprID: maxID}
 	newExpr, _ := pruner.maybePrune(expr)
 	return &exprpb.ParsedExpr{
 		Expr:       newExpr,

```

TristonianJones:
Thanks @charithe for the repro cases. I've done a non-recursive scan, but the tradeoff is memory. Open to suggestions.

Closes #699

### DIFF
--- a/interpreter/prune.go
+++ b/interpreter/prune.go
@@ -76,6 +76,46 @@ func PruneAst(expr *exprpb.Expr, macroCalls map[int64]*exprpb.Expr, state EvalSt
 			maxID = id + 1
 		}
 	}
+	exprs := []*exprpb.Expr{expr}
+	for len(exprs) != 0 {
+		e := exprs[0]
+		if e.GetId() >= maxID {
+			maxID = e.GetId() + 1
+		}
+		exprs = exprs[1:]
+		switch e.GetExprKind().(type) {
+		case *exprpb.Expr_SelectExpr:
+			exprs = append(exprs, e.GetSelectExpr().GetOperand())
+		case *exprpb.Expr_CallExpr:
+			call := e.GetCallExpr()
+			if call.GetTarget() != nil {
+				exprs = append(exprs, call.GetTarget())
+			}
+			exprs = append(exprs, call.GetArgs()...)
+		case *exprpb.Expr_ComprehensionExpr:
+			compre := e.GetComprehensionExpr()
+			exprs = append(exprs,
+				compre.GetIterRange(),
+				compre.GetAccuInit(),
+				compre.GetLoopCondition(),
+				compre.GetLoopStep(),
+				compre.GetResult())
+		case *exprpb.Expr_ListExpr:
+			list := e.GetListExpr()
+			exprs = append(exprs, list.GetElements()...)
+		case *exprpb.Expr_StructExpr:
+			for _, entry := range expr.GetStructExpr().GetEntries() {
+				if entry.GetMapKey() != nil {
+					exprs = append(exprs, entry.GetMapKey())
+				}
+				exprs = append(exprs, entry.GetValue())
+				if entry.GetId() >= maxID {
+					maxID = entry.GetId() + 1
+				}
+			}
+		}
+	}
+
 	pruner := &astPruner{
 		expr:       expr,
 		macroCalls: macroCalls,
--- a/interpreter/prune.go
+++ b/interpreter/prune.go
@@ -68,59 +68,15 @@ type astPruner struct {
 // the overloads accordingly.
 func PruneAst(expr *exprpb.Expr, macroCalls map[int64]*exprpb.Expr, state EvalState) *exprpb.ParsedExpr {
 	pruneState := NewEvalState()
-	maxID := int64(1)
 	for _, id := range state.IDs() {
 		v, _ := state.Value(id)
 		pruneState.SetValue(id, v)
-		if id > maxID {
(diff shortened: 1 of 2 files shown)


interpreter/prune.go
```diff
@@ -76,6 +76,46 @@ func PruneAst(expr *exprpb.Expr, macroCalls map[int64]*exprpb.Expr, state EvalSt
 			maxID = id + 1
```
l46kok: Since we're traversing the entire tree, can this be removed?

interpreter/prune.go
```diff
@@ -76,6 +76,46 @@ func PruneAst(expr *exprpb.Expr, macroCalls map[int64]*exprpb.Expr, state EvalSt
...
 		}
 	}
+	exprs := []*exprpb.Expr{expr}
+	for len(exprs) != 0 {
```
l46kok: Just an idea -- can we find the max from SourcePositions? Presently, I would think its keys contain all of the Expr IDs so iterating over that would be far cheaper

interpreter/prune.go
```diff
@@ -76,6 +76,46 @@ func PruneAst(expr *exprpb.Expr, macroCalls map[int64]*exprpb.Expr, state EvalSt
 			maxID = id + 1
 		}
 	}
+	exprs := []*exprpb.Expr{expr}
```
l46kok: Consider extracting work into a function e.g: `getMaxID`

interpreter/prune.go
TristonianJones: If we have an optimized AST, I'm not sure we could rely on the source positions; however, we should be able to rely on updated type metadata. That said, the prune step works with both checked and parsed expressions, so I think this is the only real common denominator between each format.

interpreter/prune.go
TristonianJones: SGTM

interpreter/prune.go
TristonianJones: Done.

TristonianJones:
PTAL

l46kok (approved):

PR #705, opened 2023-05-16
Support for pruning macro call nodes with limited optional support

TristonianJones:
Ensure that when macro call arguments have been successfully evaluated,
that the pruned and unparsed result reflects the evaluation state rather than
errors.

Additional support for generating duration literals and for optional values
within lists and maps.

### DIFF
--- a/interpreter/prune.go
+++ b/interpreter/prune.go
@@ -16,6 +16,7 @@ package interpreter
 
 import (
 	"github.com/google/cel-go/common/operators"
+	"github.com/google/cel-go/common/overloads"
 	"github.com/google/cel-go/common/types"
 	"github.com/google/cel-go/common/types/ref"
 	"github.com/google/cel-go/common/types/traits"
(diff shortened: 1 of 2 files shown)


interpreter/prune.go
```diff
@@ -196,29 +212,37 @@ func (p *astPruner) maybeCreateLiteral(id int64, val ref.Val) (*exprpb.Expr, boo
...
-		return nil, false
+func (p *astPruner) maybePruneOptional(elem *exprpb.Expr) (*exprpb.Expr, bool) {
+	elemVal, found := p.value(elem.GetId())
+	if found && elemVal.Type() == types.OptionalType {
+		opt := elemVal.(*types.Optional)
+		if !opt.HasValue() {
+			return nil, true
+		}
+		if newElem, pruned := p.maybeCreateLiteral(elem.GetId(), opt.GetValue()); pruned {
+			return newElem, true
```
l46kok, line 223: Could we return `pruned` instead?

interpreter/prune.go
```diff
@@ -460,14 +485,37 @@ func (p *astPruner) nextID() int64 {
...
+	return astVisitor{
+		visitExpr: func(e *exprpb.Expr) {
+			if e.GetId() >= *maxID {
+				*maxID = e.GetId() + 1
+			}
+		},
+		visitEntry: func(e *exprpb.Expr_CreateStruct_Entry) {
+			if e.GetId() >= *maxID {
+				*maxID = e.GetId() + 1
+			}
```
l46kok, line 511: I think the map key should also be visited here

interpreter/prune.go
```diff
@@ -491,15 +539,12 @@ func getMaxID(expr *exprpb.Expr) int64 {
 			exprs = append(exprs, list.GetElements()...)
 		case *exprpb.Expr_StructExpr:
 			for _, entry := range expr.GetStructExpr().GetEntries() {
```
l46kok:
```suggestion
			for _, entry := range e.GetStructExpr().GetEntries() {
```

We may want to add a test to ensure pruning of nested maps terminates: 

e.g: `{{'nested_key': 'hello'}.nested_key: 5}`

interpreter/prune_test.go
```diff
@@ -194,7 +312,21 @@ func TestPrune(t *testing.T) {
 		state := NewEvalState()
 		reg := newTestRegistry(t)
 		attrs := NewPartialAttributeFactory(containers.DefaultContainer, reg, reg)
-		interp := NewStandardInterpreter(containers.DefaultContainer, reg, reg, attrs)
+		dispatcher := NewDispatcher()
+		dispatcher.Add(functions.StandardOverloads()...)
+		dispatcher.Add(&functions.Overload{
```
l46kok: Similar to `StandardOverloads` above, could we group the optional overloads?

l46kok (changes requested):

interpreter/prune.go
TristonianJones: Map keys are visited as part of the `visitExpr` call. I can add some comments to clarify.

interpreter/prune_test.go
TristonianJones: Ah, these are not standard overloads (since they're technically defined in a much higher package), but I can make the available for testing via a convenient name.

interpreter/prune.go
TristonianJones: For consistency with the other parts of the code, I'm going to leave this as `true`

interpreter/prune.go
TristonianJones: Comments added.

interpreter/prune.go
TristonianJones: Good catch. There was a bug using `expr` when I should have used `e`

interpreter/prune_test.go
TristonianJones: Test method added.

l46kok (approved):

PR #707, opened 2023-05-19
Fix the sets overload IDs to mirror best practices.

TristonianJones:
This change updates set overload ids to match the best practice for naming overloads.

This is a breaking change for those depending on v0.15.0+ who are already using `ext.Sets()`


TristonianJones:
On second thought, I don't want to risk breaking existing users for this.

PR #712, opened 2023-05-23
Fix for macro call tracking with optional list elements

TristonianJones:


TristonianJones:
FYI @l46kok this is probably a step that needs updating on cel-java as well.

l46kok (approved):

PR #718, opened 2023-05-27
Fix overload collision checks by ignoring type param names

TristonianJones:
When the top-level `cel/decls.go` computes the set of type param names, it may alter the order in which they are specified causing overloads with more than one type parameter to appear as though they're colliding.


TristonianJones:
FYI @jpbetz

jpbetz (approved):
LGTM. Thanks!

PR #720, opened 2023-05-30
Expose ExpressionSizeCodePointLimit as env option.

notyourusualaccountname2:
Currently, the parser options cannot be customized when creating a new environment. With this change, one can override the max. number of code points the parser accepts via an env option. This is analogous to ParserRecursionLimit.

### DIFF
--- a/cel/options.go
+++ b/cel/options.go
@@ -568,3 +568,12 @@ func ParserRecursionLimit(limit int) EnvOption {
 		return e, nil
 	}
 }
+
+// ParserExpressionSizeLimit adjusts the number of code points the expression parser is allowed to parse.
+// Defaults defined in the parser package.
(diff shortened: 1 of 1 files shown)


cel/options.go
```diff
@@ -568,3 +568,13 @@ func ParserRecursionLimit(limit int) EnvOption {
 		return e, nil
 	}
 }
+
+// ParserExpressionSizeLimit adjusts the max. number of code points the expression parser is allowed
```
TristonianJones: nit: remove 'max.' and keep the first sentence on a single line.

Thanks!

notyourusualaccountname2: Done.

TristonianJones (approved):

PR #721, opened 2023-05-30
Presence test cost tracking options

TristonianJones:
In #711, presence tests began getting tracked with the same cost as a typical select expression, but this may break some users. A new set of functional options have been added to both the `checker` and `interpreter` packages to allow users to opt-out of this change.

Also, note this change breaks the current API contract for `interpreter.CostTracker` and `checker.Cost`. Prefer using `interpreter.NewCostTracker()` which accepts a set of functional arguments for configuring behavior and which may return an error. Likewise, the `checker.Cost` will return `CostEstimate, error` whereas before it did not. Note, the top-level `cel.EstimateCost()` function signature remains unchanged.


jpbetz (approved):
LGTM

On by default makes sense to me for this. Appreciate the inclusion of a flag so we can disable and get the legacy behavior.

PR #724, opened 2023-05-31
Upgrade to latest version of ANTLR

marvinkite:
Upgrade to latest version of v4.13.0 ANTLR 

The main reason is to change the name `github.com/antlr4-go/antlr/v4` for this [reason](https://github.com/antlr4-go/antlr/#why) which cause problem to keep project dependency updated. 

The [ parser test ](https://github.com/marvinkite/cel-go/blob/update-antlr/parser/parser_test.go#L1869)unfortunately became unstable as randomly some test fail or succeed entirely. That problem must be resolved but unfortunately I'm not skilled enough to find the root cause.


TristonianJones:
@marvinkite we haven't been observing any flakiness with the ANTLR version used with CEL before. Could you describe what you're seeing in terms of flakiness? Or could you point us to a reference?

marvinkite:
The ANTLR was updated to latest version v4.12.0 in this [commit](https://github.com/google/cel-go/commit/af2e9432d89762934a75772bc0d84fd8fdbe2c6d) from v4.10.0. This commit updates to recently released ANTLR v4.13.0. The v4.13.0 is a big release, hence the minor version bump. Despite the performance gain stated in [release note](https://github.com/antlr/antlr4/releases/tag/4.13.0) there are few fixes.  This is the reason to version bump.

The other reason is the change in the package name to github.com/antlr4-go/antlr/v4 for this [reason](https://github.com/antlr4-go/antlr/#why) which cause problem to keep project dependency updated.

When the project depends on this package name:

```
require (
	github.com/antlr/antlr4/runtime/Go/antlr/v4 v4.0.0-20230305170008-8188dc5388df
)
```

the project which require go-cel which require as above the `go get` and other tools like Dependabot or Renovate has hard time to update or resolve dependency and fails with 

```
Command failed: go get -d -t ./...
go: github.com/antlr/antlr4/runtime/Go/antlr/v4@20230305170008-8188dc5388df: parsing go.mod:
	module declares its path as: github.com/antlr4-go/antlr/v4
	        but was required as: github.com/antlr/antlr4/runtime/Go/antlr/v4
```

These are the two reasons. The v4.12.0 was also moved to new repository so there could be an alternative fix where only the package name is changed and the version is not bumped. I thought due to the improvement in the new version it worth to consider the upgrade. 

Everything looked good and all test passes but as I wrote I noticed one problem. If I run 10 times the [parser test](https://github.com/marvinkite/cel-go/blob/update-antlr/parser/parser_test.go#L1869) it fails 2-3 times. Exactly as you can see the builder failed. 

Due to this problem this PR MUST NOT BE accepted as is but I hope someone sees the benefit and can take it over and fix the last bits. If I have some more time I will investigate and may be able to fix it.

TristonianJones:
@marvinkite what you're describing sounds like an issue with concurrency safety in v4.13.0. I'll investigate it, but I don't think we can upgrade CEL to the latest ANTLR based on what you're describing.

TristonianJones:
Refreshed in #857

PR #725, opened 2023-06-02
Include expression id on parser and checker errors

TristonianJones:


common/errors.go
```diff
@@ -31,32 +31,38 @@ type Errors struct {
...
+
+// ReportErrorAtID records an error at a source location and expression id.
+func (e *Errors) ReportErrorAtID(id int64, l Location, format string, args ...any) {
 	e.numErrors++
 	if e.numErrors > e.maxErrorsToReport {
 		return
 	}
-	err := Error{
+	err := &Error{
+		ExprID:   id,
```
l46kok, line 52: Could we add a test that shows the expr ID exists on parsing/type-checking errors?

TristonianJones: PTAL

l46kok (approved):

PR #726, opened 2023-06-02
Variadic logical operator AST support

TristonianJones:
Support for variadic logical operators to assist with reducing expression recursion depth.

Note, this change is not yet compatible with the cel-cpp or cel-java runtimes, so enable it carefully.

### DIFF
--- a/interpreter/interpretable.go
+++ b/interpreter/interpretable.go
@@ -223,21 +223,14 @@ func (or *evalOr) Eval(ctx Activation) ref.Val {
 			return types.True
 		}
 		if !ok {
-			if types.IsUnknown(val) {
-				if unk == nil {
-					unk = types.Unknown{}
-				}
-				unk = append(unk, val.(types.Unknown)...)
-				continue
-			}
-			if types.IsError(val) {
-				if unk == nil && err == nil {
+			isUnk := false
+			unk, isUnk = types.MaybeMergeUnknowns(val, unk)
+			if !isUnk && err == nil {
+				if types.IsError(val) {
 					err = val
+				} else {
+					err = types.MaybeNoSuchOverloadErr(val)
 				}
-				continue
-			}
-			if unk == nil {
-				err = types.MaybeNoSuchOverloadErr(val)
 			}
 		}
 	}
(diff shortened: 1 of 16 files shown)


parser/options.go
```diff
@@ -125,3 +126,15 @@ func EnableOptionalSyntax(optionalSyntax bool) Option {
...
 }
+
+// EnableVariadicOperatorASTs enables a compact representation of chained like-kind commutative
+// operators. e.g. `a || b || c || d` -> `call(op='||', args=[a, b, c, d])`
+//
+// The benefit of enabling variadic operators ASTs is a more compact representation deeply nested
+// logic graphs.
+func EnableVariadicOperatorASTs(varArgASTs bool) Option {
+	return func(opts *options) error {
+		opts.enableVariadicOperatorASTs = varArgASTs
```
l46kok, line 137: Could we bump the `syntax_version` and additionally document that this is an incompatible change for runtimes that do not yet support variadic operators?

parser/parser.go
```diff
@@ -480,33 +482,33 @@ func (p *parser) VisitExpr(ctx *gen.ExprContext) any {
...
 	result := p.Visit(ctx.GetE()).(*exprpb.Expr)
-	b := newBalancer(p.helper, operators.LogicalOr, result)
+	l := p.newLogicManager(operators.LogicalOr, result)
 	rest := ctx.GetE1()
 	for i, op := range ctx.GetOps() {
 		if i >= len(rest) {
 			return p.reportError(ctx, "unexpected character, wanted '||'")
 		}
 		next := p.Visit(rest[i]).(*exprpb.Expr)
 		opID := p.helper.id(op)
```
l46kok, line 493: Consider pulling this outside the loop. I think the operator ID only needs to be generated once to avoid creating holes between the arguments' Expr IDs. (i.e: `false || false || false || false` generating 1, 2, 4, 6)

interpreter/interpretable.go
```diff
@@ -214,41 +213,46 @@ func (or *evalOr) ID() int64 {
...
+			}
+			if types.IsError(val) {
+				if unk == nil && err == nil {
+					err = val
+				}
+				continue
+			}
+			if unk == nil {
+				err = types.MaybeNoSuchOverloadErr(val)
+			}
```
l46kok: We might be getting to the point where we should extract the unknown and error handling into a function.

parser/parser.go
TristonianJones: In the current format, the IDs will be preserved when the feature is not enabled. As for holes, they'll appear where the operators do, so the non-operator IDs will be consistent.

parser/options.go
TristonianJones: I'm not sure this is a syntax change. It does feel like a runtime version change though; however, that only appears on the checked.proto. We could move up the runtime version into source_info?

l46kok: I was originally thinking of a minor version bump, but what you suggest sounds better. I do think we need a way of signaling the AST change for the runtime.

parser/parser.go
l46kok: Got it, we can leave this as is.

parser/options.go
TristonianJones: Since the runtimes can be updated to support either format equally well, I think I'll just make it impossible to enable until we've made the update elsewhere.

interpreter/interpretable.go
TristonianJones: I've added a helper in types/unknown.go and simplified the logic here. PTAL

l46kok (approved):

PR #734, opened 2023-06-09
Split types and declarations into separate files

TristonianJones:
Split out types from the `decls.go` file. Also, added a missing test and missing `go_default_test` bazel rule.


jnthntatum (approved):
lgtm

PR #740, opened 2023-06-13
Clear iteration variable data during expression pruning

### LINKED ISSUE #735: interpreter.PruneAst does not prune `IterRange` expression
**Describe the bug**
interpreter.PruneAst function ignores the `IterRange` field of a comprehension expression.

**To Reproduce**

- [ ]  parser
- [ ]  checker
- [x]  interpreter

`users` is a list of const strings, and `R` is an unknown attribute pattern.

```cel
users.filter(u, u.startsWith(R.attr.prefix))
```

Test setup:
```go
package main

import (
	"fmt"
	"log"

	expr "google.golang.org/genproto/googleapis/api/expr/v1alpha1"
)

func main() {
	env, err := cel.NewEnv(cel.Declarations(
		decls.NewVar("R", decls.NewMapType(decls.String, decls.Dyn)),
		decls.NewVar("users", decls.NewListType(decls.String)),
	))
	exitOnErr(err)

	ast, iss := env.Compile(`users.filter(u, u.startsWith(R.attr.prefix))`)
	exitOnErr(iss.Err())

	prg, err := env.Program(ast, cel.EvalOptions(cel.OptTrackState, cel.OptPartialEval))
	exitOnErr(err)

	in := map[string]any{
		"users": []string{"alice", "bob"},
	}
	vars, err := cel.PartialVars(in, cel.AttributePattern("R").QualString("attr").Wildcard())
	exitOnErr(err)

	_, details, err := prg.Eval(vars)
	exitOnErr(err)

	pruned := interpreter.PruneAst(ast.Expr(), ast.SourceInfo().GetMacroCalls(), details.State()).Expr
	ce := pruned.ExprKind.(*expr.Expr_ComprehensionExpr).ComprehensionExpr
	printExpr("IterRange:", ce.IterRange)
}

func printExpr(name string, e *expr.Expr) {
	prunedAST := cel.ParsedExprToAst(&expr.ParsedExpr{Expr: e})
	s, err := cel.AstToString(prunedAST)
	exitOnErr(err)
	fmt.Println(name, s)
}

func exitOnErr(err error) {
	if err != nil {
		log.Fatalf("Error: %v", err)
	}
}

```

**Expected behavior**
The output should be:

```
IterRange: ["alice", "bob"]
```

instead of
```
IterRange: users
```

**Additional context**
This works as expected in v0.15.2, but in v0.15.3, the pruner [was modified](https://github.com/google/cel-go/pull/705/files#diff-531f45e46499da01fa105d485bd3d5765123ee5624c4128cad21545ce18bce25L411-L431) to skip the `IterRange` expression. Is this a breaking change?

TristonianJones:
This PR clears iteration variable data from expression tracking state during macro pruning.
The clearing of the iteration variable is based on the iteration variable name which could
potentially shadow an existing, but in the worst case, this would just cause a duplicate name
to be cleared more than once in a nested macro scope.

Closes #735

### DIFF
--- a/interpreter/prune.go
+++ b/interpreter/prune.go
@@ -341,6 +341,12 @@ func (p *astPruner) prune(node *exprpb.Expr) (*exprpb.Expr, bool) {
 		}
 	}
 	if macro, found := p.macroCalls[node.GetId()]; found {
+		// Ensure that intermediate values for the comprehension are cleared during pruning
+		compre := node.GetComprehensionExpr()
+		if compre != nil {
+			visit(macro, clearIterVarVisitor(compre.IterVar, p.state))
+			visit(macro, clearIterVarVisitor(compre.AccuVar, p.state))
+		}
 		// prune the expression in terms of the macro call instead of the expanded form.
 		if newMacro, pruned := p.prune(macro); pruned {
 			p.macroCalls[node.GetId()] = newMacro
@@ -524,6 +530,17 @@ func getMaxID(expr *exprpb.Expr) int64 {
 	return maxID
 }
 
+func clearIterVarVisitor(varName string, state EvalState) astVisitor {
+	return astVisitor{
+		visitExpr: func(e *exprpb.Expr) {
+			ident := e.GetIdentExpr()
+			if ident != nil && ident.GetName() == varName {
+				state.SetValue(e.GetId(), nil)
+			}
+		},
+	}
+}
+
 func maxIDVisitor(maxID *int64) astVisitor {
 	return astVisitor{
 		visitExpr: func(e *exprpb.Expr) {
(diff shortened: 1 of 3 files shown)


interpreter/prune.go
```diff
@@ -341,6 +341,12 @@ func (p *astPruner) prune(node *exprpb.Expr) (*exprpb.Expr, bool) {
 		}
 	}
 	if macro, found := p.macroCalls[node.GetId()]; found {
+		// Ensure that intermediate values for the comprehension are cleared during pruning
+		compre := node.GetComprehensionExpr()
+		if compre != nil {
+			visit(macro, clearIterVarVisitor(compre.IterVar, p.state))
+			visit(macro, clearIterVarVisitor(compre.AccuVar, p.state))
```
l46kok: Does the accumulation variable actually need to be cleared here? If so, could we add a test for this

TristonianJones: Oh, no, you're right. We can skip the accumulation variable clearing since it never appears in the macro call AST

TristonianJones:
PTAL

l46kok (approved):

PR #744, opened 2023-06-15
Introduce optFlatMap and OptionalTypes documentation

TristonianJones:
The `optFlatMap` macro allows for the use of optional syntax within mapping
expressions which are applied if a value exists. 

This feature exists at `OptionalTypesVersion(1)` and above.

### DIFF
--- a/cel/library.go
+++ b/cel/library.go
@@ -172,9 +172,10 @@ func (stdLibrary) ProgramOptions() []ProgramOption {
 // will ensure the type on the right-hand side must be optional(T) where T
 // is the type of the field or key-value.
 //
-//		// returns a map with the key expression set only if the subfield is
-//	 // present otherwise an empty map is created.
-//		{?key: obj.?field.subfield}
+// The following returns a map with the key expression set only if the
+// subfield is present, otherwise an empty map is created:
+//
+//	{?key: obj.?field.subfield}
 //
 // ## Optional Element Setting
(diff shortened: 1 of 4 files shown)


ext/strings.go
```diff
@@ -57,7 +57,7 @@ const (
 //	'hello'.charAt(5)  // return ''
 //	'hello'.charAt(-1) // error
 //
-// # Format
+// # Strings.Format
```
l46kok: nit: Does this need to be prefixed with a namespace? It's a receiver call on an existing string

cel/library.go
```diff
@@ -132,29 +134,182 @@ func (stdLibrary) ProgramOptions() []ProgramOption {
...
+//	map[?key]
+//
+// ## Optional Field Setting
+//
+// When creating map or message literals, if a field may be optionally set
+// based on its presence, then placing a `?` before the field name or key
+// will ensure the type on the right-hand side must be optional(T) where T
+// is the type of the field or key-value.
+//
+//		// returns a map with the key expression set only if the subfield is
```
l46kok: nit: odd indentation

ext/strings.go
TristonianJones: My mistake. Undone.

TristonianJones:
PTAL

l46kok (approved):

Issue #751, opened 2023-06-22
List helper functions


chancez:
**Change**
I'm looking at whether or not CEL could be used as a DSL for an existing gRPC based query API. For a query API, we may need to operate on a list of objects and perform various operations. Some macros like `filter`, `exists`, and `map` already exist for some use-cases, and I would like to see if we could add additional functions that operate on lists.

**Example**
- `myList.slice(start, end)` could slice a list, much like in Go or Python via: `mylist[start:end]`.
- `myList.first(n)` could take the first elements from a list. 
- `myList.last(n)` could take the last elements from a list. 

Any many more could be considered, but at the moment, these 3 are the first ones I'm thinking about. 

**Alternatives considered**

This could be handled outside of CEL, but being able to manipulate lists would be very useful for using CEL in a query language. 

I wrote a quick MVP of `list.slice` here: https://github.com/google/cel-go/compare/master...chancez:cel-go:pr/chancez/slice_function

TristonianJones:
This issue was resolved with the introduction of slice() in the `lists` extension and the introduction of `first` and `last` in the optional extension.

PR #753, opened 2023-06-23
ext: Add list library with slice method

chancez:
Introduce a slice extension function for slicing lists (#751)

### DIFF
--- a/ext/lists.go
+++ b/ext/lists.go
@@ -1,0 +1,94 @@
+// Copyright 2023 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package ext
+
+import (
+	"fmt"
+
+	"github.com/google/cel-go/cel"
+	"github.com/google/cel-go/common/types"
+	"github.com/google/cel-go/common/types/ref"
+	"github.com/google/cel-go/common/types/traits"
+)
+
+// Lists returns a cel.EnvOption to configure extended functions for list manipulation.
+// As a general note, all indices are zero-based.
+// # Slice
+//
+// Returns a new sub-list using the indexes provided.
+//
+//	<list>.slice(<int>, <int>) -> <list>
+//
+// Examples:
+//
+//	[1,2,3,4].slice(1, 3) // return [2, 3]
+//	[1,2,3,4].slice(2, 4) // return [3 ,4]
+func Lists() cel.EnvOption {
+	return cel.Lib(listsLib{})
+}
+
+type listsLib struct{}
+
+// LibraryName implements the SingletonLibrary interface method.
+func (listsLib) LibraryName() string {
+	return "cel.lib.ext.lists"
+}
+
+// CompileOptions implements the Library interface method.
+func (listsLib) CompileOptions() []cel.EnvOption {
+	listType := cel.ListType(cel.TypeParamType("T"))
+	return []cel.EnvOption{
+		cel.Function("slice",
+			cel.MemberOverload("list_slice",
+				[]*cel.Type{listType, cel.IntType, cel.IntType}, listType,
+				cel.FunctionBinding(func(args ...ref.Val) ref.Val {
+					list := args[0].(traits.Lister)
+					start := args[1].(types.Int)
+					end := args[2].(types.Int)
+					result, err := slice(list, start, end)
+					if err != nil {
+						return types.WrapErr(err)
+					}
+					return result
+				}),
+			),
+		),
+	}
+}
+
+// ProgramOptions implements the Library interface method.
+func (listsLib) ProgramOptions() []cel.ProgramOption {
+	return []cel.ProgramOption{}
+}
+
+func slice(list traits.Lister, start, end types.Int) (ref.Val, error) {
+	listLength := list.Size().(types.Int)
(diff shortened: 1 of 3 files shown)


ext/lists.go
```diff
@@ -0,0 +1,74 @@
...
+
+import (
+	"fmt"
+
+	"github.com/google/cel-go/cel"
+	"github.com/google/cel-go/common/types"
+	"github.com/google/cel-go/common/types/ref"
+	"github.com/google/cel-go/common/types/traits"
+)
+
```
TristonianJones, line 25: nit: please add a comment for the `Lists` method.

ext/lists.go
```diff
@@ -0,0 +1,74 @@
...
+	if int64(listLength) < end {
+		return nil, fmt.Errorf("cannot slice(%d, %d), list is length %d", start, end, listLength)
+	}
+	if start > end {
+		return nil, fmt.Errorf("cannot slice(%d, %d), start index must be less than end index", start, end)
+	}
+
+	var newList []any
+	for i := types.Int(start); int64(i) < end; i++ {
+		newList = append(newList, list.Get(i).Value())
```
TristonianJones: Remove the `.Value()` call after `list.Get(i)` as this will ensure that the type-adapted elements won't need to be re-adapted in the `NativeToValue` call later.

ext/lists.go
```diff
@@ -0,0 +1,74 @@
...
+func slice(list traits.Lister, start, end int64) ([]any, error) {
+	listLength := list.Size().(types.Int)
+	if int64(listLength) < end {
+		return nil, fmt.Errorf("cannot slice(%d, %d), list is length %d", start, end, listLength)
+	}
+	if start > end {
+		return nil, fmt.Errorf("cannot slice(%d, %d), start index must be less than end index", start, end)
+	}
+
+	var newList []any
```
TristonianJones: Use the type `var newList []ref.Val` to avoid additional type conversions later.

ext/lists.go
```diff
@@ -0,0 +1,74 @@
...
+	for i := types.Int(start); int64(i) < end; i++ {
+		newList = append(newList, list.Get(i).Value())
+	}
+	return newList, nil
```
TristonianJones: Consider calling `types.DefaultTypeAdapter.NativeToValue` here instead of creating a helper in the `guards.go`.

ext/lists.go
```diff
@@ -0,0 +1,74 @@
...
+	listType := cel.ListType(cel.TypeParamType("T"))
+	return []cel.EnvOption{
+		cel.Function("slice",
+			cel.MemberOverload("list_slice",
+				[]*cel.Type{listType, cel.IntType, cel.IntType}, listType,
+				cel.FunctionBinding(func(args ...ref.Val) ref.Val {
+					list := args[0].(traits.Lister)
+					start := args[1].(types.Int)
+					end := args[2].(types.Int)
+					return listOrError(slice(list, int64(start), int64(end)))
```
TristonianJones: Prefer the following:

```go
result, err := slice(list, start, end)
if err != nil {
  return types.WrapErr(err)
}
return result
```

Note, the `start` and `end` are left as `types.Int`

ext/lists.go
TristonianJones: It's possible that `i < 0`. Please check whether the result of the list `Get` call is `types.IsError` and return it as the result

TristonianJones (changes requested):

ext/lists.go
chancez: So `slice()` should take `types.Int` instead of `int64`?

ext/lists.go
chancez: I can check `i < 0` at the top of the function, do I still need the error check on `Get()` then? If so: do I return it  as `ref.Val` or `error`?

ext/lists.go
TristonianJones: I would recommend it since you're having to convert between `int64` and `types.Int` later in the code.

ext/lists.go
TristonianJones: It's probably fine to just check for `i < 0` at the beginning, but the general contract for `Get` is that it will perform the same array index range checks as you're doing here. 

You could do any of the following:
1. Check up front (knowing it'll get checked later as well)
1. Switch to an iterator which skips the elements `< start` and stops iterating after `end - start` iterations (cleanest)
1. Check the result of `Get` to see if its a `types.IsError` and early return

I think a combination of 1 & 2 is probably best since the iterator could be optimized under the covers later.

chancez: Okay, I'll take a look at that, I didn't realize `Get` ended up doing those checks. I started with an iterator, but ended up switching to this approach since I could put the iteration conditions into the `for`.

ext/lists.go
```diff
@@ -0,0 +1,94 @@
...
+// # Slice
+//
+// Returns a new sub-list using the indexes provided.
+//
+//	<list>.slice(<int>, <int>) -> <list>
+//
+// Examples:
+//
+//	[1,2,3,4].slice(1, 3) // return [2, 3]
+//	[1,2,3,4].slice(2, 4) // return [3 ,4]
```
TristonianJones, line 37: Don't forget to include similar content into the `ext/README.md` as well

ext/lists.go
TristonianJones: Fair enough, the code contract is clearer with the `for` loop.

TristonianJones:
@chancez Looking good! Just update the `ext/README.md` and we'll be good to submit the PR. Thanks for the contribution!

ext/lists.go
chancez: Okay, I updated the readme, let me know if that's enough, it's a bit terse.

TristonianJones: @chancez perhaps a push is needed? I don't see the updates for the ext/README.md in this PR yet. Thanks!

chancez: 🤦 I pushed but it failed because I used the wrong remote, sorry about that 😂

TristonianJones: No worries! Thanks again for the PR

TristonianJones (approved):

Issue #568, opened 2022-07-14
Replace `exprpb.Type` references with `cel.Type`


TristonianJones:
**Feature request checklist**

- [X] There are no issues that match the desired change
- [X] The change is large enough it can't be addressed with a simple Pull Request
- [ ] If this is a bug, please file a [Bug Report](./ISSUE_TEMPLATE).

**Change**
The top-level API currently emphasizes using `cel.Type` values rather than `exprpb.Type`
values in an effort to reduce the protobuf surface area visible on the cel-go APIs. However,
the `ref.TypeProvider` currently exposes a method to `FindType` and `FindFieldType` which
return an `exprpb.Type`. This interface is used by a handful of customers to extend the
CEL type-system, and creates a disconnect between the top-level API and the extensions.

The change should introduce a new `TypeProvider` which can wrap the old style and produce
`cel.Type` values from these methods for use within the type-checker and runtimes.

**Alternatives considered**
Provide new methods on the existing `TypeProvider` interface.
* Functional, but would be a breaking API change for existing users.
* Introduces a more complex migration burden of having to support both the old and new-style methods.

PR #759, opened 2023-06-26
Migrate type-checker to native types and declarations

TristonianJones:
Migrate the type-checker to native types and declarations

This change simplifies the internals of the type-checker considerably by relying
on function overload collision checks to happen within the `decls.FunctionDecl`
instances on `Merge`. Additionally, the native types have replaced the protobuf
type declarations used throughout the checker previously.

The files with the most non-trivial changes are the `checker/checker.go` and
`checker/types.go`. When reviewing the code, please consider these changes
carefully.

Only one location continues to use the proto-based `exprbp.Type`, and that's
the `ref.TypeProvider` which will be updated in a follow-up PR to complete
#568

### DIFF
--- a/checker/types.go
+++ b/checker/types.go
@@ -15,154 +15,54 @@
 package checker
 
 import (
-	"fmt"
-	"strings"
-
-	"github.com/google/cel-go/checker/decls"
-
-	"google.golang.org/protobuf/proto"
-
-	exprpb "google.golang.org/genproto/googleapis/api/expr/v1alpha1"
+	"github.com/google/cel-go/common/types"
 )
-
-const (
-	kindUnknown = iota + 1
-	kindError
-	kindFunction
-	kindDyn
-	kindPrimitive
-	kindWellKnown
-	kindWrapper
-	kindNull
-	kindAbstract
-	kindType
-	kindList
-	kindMap
-	kindObject
-	kindTypeParam
-)
-
-// FormatCheckedType converts a type message into a string representation.
-func FormatCheckedType(t *exprpb.Type) string {
-	switch kindOf(t) {
-	case kindDyn:
-		return "dyn"
-	case kindFunction:
-		return formatFunction(t.GetFunction().GetResultType(),
-			t.GetFunction().GetArgTypes(),
-			false)
-	case kindList:
-		return fmt.Sprintf("list(%s)", FormatCheckedType(t.GetListType().GetElemType()))
-	case kindObject:
-		return t.GetMessageType()
-	case kindMap:
-		return fmt.Sprintf("map(%s, %s)",
-			FormatCheckedType(t.GetMapType().GetKeyType()),
-			FormatCheckedType(t.GetMapType().GetValueType()))
-	case kindNull:
-		return "null"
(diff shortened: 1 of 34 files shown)


jcking (approved):

jcking (changes requested):
protobuf `Get` accessors protect against `nil` dereferencing by returning default values. Here we switch to direct accessors. Hopefully there are not places where we accidently accepted `nil` and got a default string value, where now we will dereference. We should probably add more exhaustive checks to ensure `nil` doesn't get placed into the environment accidently.

checker/types.go
```diff
@@ -15,154 +15,54 @@
...
-		for i, p := range params {
-			paramStrs[i] = FormatCheckedType(p)
-		}
-		return fmt.Sprintf("%s(%s)", at.GetName(), strings.Join(paramStrs, ", "))
-	}
-	return t.String()
-}
-
 // isDyn returns true if the input t is either type DYN or a well-known ANY message.
-func isDyn(t *exprpb.Type) bool {
```
jcking, line 106: kindOf used to handle `nil` gracefully here. In the new code we will SIGSEGV on `nil`, that safe?

checker/types.go
```diff
@@ -15,154 +15,54 @@
...
-		case *exprpb.Constant_StringValue:
-			return literal.GetStringValue(), true
-		}
-	}
-	return "", false
-}
-
 // isEqualOrLessSpecific checks whether one type is equal or less specific than the other one.
 // A type is less specific if it matches the other type using the DYN type.
-func isEqualOrLessSpecific(t1 *exprpb.Type, t2 *exprpb.Type) bool {
```
jcking, line 158: Same as other, kindOf gracefully handled `nil`. Now we do not. Not sure if safe.

checker/types.go
```diff
@@ -173,38 +73,34 @@ func isEqualOrLessSpecific(t1 *exprpb.Type, t2 *exprpb.Type) bool {
...
-	case kindList:
-		return isEqualOrLessSpecific(t1.GetListType().GetElemType(), t2.GetListType().GetElemType())
-	case kindMap:
-		m1 := t1.GetMapType()
-		m2 := t2.GetMapType()
-		return isEqualOrLessSpecific(m1.GetKeyType(), m2.GetKeyType()) &&
-			isEqualOrLessSpecific(m1.GetValueType(), m2.GetValueType())
-	case kindType:
+	case types.ListKind:
+		return isEqualOrLessSpecific(t1.Parameters[0], t2.Parameters[0])
```
jcking: How sure are we about the length of `x.Parameters()`? Previously we could get `nil` and `kindOf` logic would handle it. Now I am not sure.

checker/types.go
TristonianJones: I've updated `*types.Type`, `*decls.VariableDecl`, `*decls.FunctionDecl`, and `*decls.OverloadDecl` to be nil-safe. While this won't prevent nil dereferences in all cases, the cases which still exist would have been present with the proto-based representations.

jcking (approved):

Issue #655, opened 2023-03-09
string.format() cannot be used as intended with HomogeneousAggregateLiterals


jpbetz:
We encountered this with Kubernetes, where we have `HomogeneousAggregateLiterals` enabled.

`string.format()` takes a list of values to interpolate in a formatted string, e.g. `"%s - %d".format(['a', 1])`.

When `HomogeneousAggregateLiterals` is enabled, CEL enforces that the items of a literal list be the same type.

The result is that expressions like `"%s - %d".format(['a', 1])` are not allowed when `HomogeneousAggregateLiterals` is enabled.

jpbetz:
Workarounds like `"%s - %d".format([string('a'), string(1)])` or `"%s - %d".format([dyn('a'), dyn(1)])`, exist, but are messy and I don't recommend them.

TristonianJones:
Thank you, Joe.

jpbetz:
Some short term questions:

- Should we temporarily take `string.format()` out of the strings extension library until we have a plan? Would doing this help keep our design options open? (or has that ship sailed)
- Any next steps we should explore?

TristonianJones:
I don't think taking `string.format()` out really changes anything. In terms of design, even if you were to use a variadic argument macro (not recommended), it would generate a mixed type list under the covers and suffer the same issue. 

As a next step, I would like to shift the `HomogeneousAggregateLiteral` option and the `string.format()` argument checking into something more general purpose that can be run post type-check and still provide the same feedback and levels of correctness so we can address issues #359, #420, and #496 more effectively. By making it possible to introduce these individual checks, it should also allow for some customization of when mixed type list / map literals are permitted.

Issue #496, opened 2022-02-25
Type check and pre-compile fixed regex strings


jpbetz:
**Change**

Currently regex patterns are compiled at runtime each time a matches() invocation is evaluated by the interpreter.

Instead, if the regex patterns is a string constant (e.g. `x.matches('[0-9]+')`)
- Check the regex at compile time and return a type check error if the regex string is malformed (similar in spirit to https://github.com/google/cel-go/issues/359)
- Pre-compile regex at CEL compile time (at least for programs where `OptOptimize` is enabled) and reuse it at runtime instead of recompiling it for each invocation

**Example***

`x.matches('[0-9]+') // regex pattern is fixed, so type check it and pre-compile it at CEL compile time

`x.matches(foo.pattern)` // regex pattern is from a variable input so continue to compile it at runtime for each matches() invocation

**Alternatives considered**

I'm open to ideas here.

Based on what little I know of CEL, the approach that came to mind is:
- Annotate Function Overloads to identify arguments that are regex pattern strings
- Modify the type checker to look for regex pattern arguments where the arg expression is a const, compile the regex and report any errors
- Modify the planner to drop in a new type of Interpretable for regex pattern argument exprs that holds the compiled regex. Also add a new ref.Val type that the new Interpretable can return that holds a reference to the compile regex in addition to the string constant the ref.Val represents.  matches() can then inspect the ref.Val to see if it has a pre-compiled regex, and use if it is does (otherwise it can compile as usual)

jpbetz:
xref: https://github.com/kubernetes/kubernetes/pull/108312#discussion_r814461567

PR #762, opened 2023-06-30
Validators using navigable expr pattern

TristonianJones:
Expression Navigation utilities for querying the expression graph

#359, #420, #496, #655

### DIFF
--- a/common/ast/ast_test.go
+++ b/common/ast/ast_test.go
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-package ast
+package ast_test
 
 import (
 	"reflect"
@@ -22,6 +22,7 @@ import (
 	"google.golang.org/protobuf/proto"
 
 	chkdecls "github.com/google/cel-go/checker/decls"
+	"github.com/google/cel-go/common/ast"
 	"github.com/google/cel-go/common/overloads"
 	"github.com/google/cel-go/common/types"
 	"github.com/google/cel-go/common/types/ref"
@@ -30,16 +31,16 @@ import (
 )
 
 func TestConvertAST(t *testing.T) {
-	ast := &CheckedAST{
+	goAST := &ast.CheckedAST{
 		Expr:       &exprpb.Expr{},
 		SourceInfo: &exprpb.SourceInfo{},
 		TypeMap: map[int64]*types.Type{
 			1: types.BoolType,
 			2: types.DynType,
 		},
-		ReferenceMap: map[int64]*ReferenceInfo{
-			1: NewFunctionReference(overloads.LogicalNot),
-			2: NewIdentReference("TRUE", types.True),
+		ReferenceMap: map[int64]*ast.ReferenceInfo{
+			1: ast.NewFunctionReference(overloads.LogicalNot),
+			2: ast.NewIdentReference("TRUE", types.True),
 		},
 	}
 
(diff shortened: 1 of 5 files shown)


common/ast/ast_test.go
```diff
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-package ast
```
l46kok, line 15: Curious, why the package naming change?

common/ast/expr.go
```diff
@@ -0,0 +1,692 @@
+// Copyright 2023 Google LLC
```
l46kok, line 1: It maybe worth renaming this file to `naviagable_expr.go`, especially if we plan on having a native representation of `exprpb.Expr` in the future.

common/ast/expr.go
```diff
@@ -0,0 +1,692 @@
...
+	}
+}
+
+type childFactory func(*navigableExprImpl) []NavigableExpr
+
+func noopFactory(*navigableExprImpl) []NavigableExpr {
+	return nil
+}
+
+func selectFatory(nav *navigableExprImpl) []NavigableExpr {
```
l46kok: nit: typo `selectFactory`

common/ast/expr.go
```diff
@@ -0,0 +1,692 @@
...
+	SelectKind
+	CallKind
+	ListKind
+	MapKind
+	StructKind
+	ComprehensionKind
+)
+
+// NavigateExpr converts a protobuf Expr node and a map of expression types by id to a NavigableExpr
+func NavigateExpr(expr *exprpb.Expr, typeMap map[int64]*types.Type) NavigableExpr {
```
l46kok: Thoughts on accepting an Ast here instead and verify that it is type-checked inside? The validator should already be working with a type-checked AST, and the reference map would be needed for optimizer in the future.

common/ast/expr.go
```diff
@@ -0,0 +1,692 @@
...
+// Such a result would work well with subsequent MatchList calls.
+func AllMatcher() ExprMatcher {
+	return func(NavigableExpr) bool {
+		return true
+	}
+}
+
+// MatchDescendents takes a NavigableExpr and ExprMatcher and produces a list of NavigableExpr values of the
+// descendents which match.
+func MatchDescendents(expr NavigableExpr, matcher ExprMatcher) []NavigableExpr {
```
l46kok: super nit: I think `Descendants` is more commonly used. Not opposed to keeping it -- I'll just make the Java side consistent.

common/ast/expr.go
```diff
@@ -0,0 +1,692 @@
...
+}
+
+// ExprMatcher takes a NavigableExpr in and indicates whether the value is a match.
+//
+// This function type should be use with the `Match` and `MatchList` calls.
+type ExprMatcher func(NavigableExpr) bool
+
+// ConstantValueMatcher returns an ExprMatcher which will return true if the input NavigableExpr
+// is comprised of all constant values, such as a simple literal or even list and map literal.
+func ConstantValueMatcher() ExprMatcher {
```
l46kok, line 51: Do you have a use case in mind for this?

l46kok (commented):
LGTM, just a few minor things

common/ast/ast_test.go
TristonianJones: I needed to avoid a cyclic dependency in the `ast` package since the test relies on the `checker` and the `checker` relies on `ast`.

common/ast/expr.go
TristonianJones: I was thinking of having the `Expr` representation present in this file in the future. I imagine the delta between `Expr` and `NavigableExpr` will be small, mostly a refactor of `Expr` into a base interface. I've added the `is_navigable()` method to ensure that no one outside the CEL package tries to implement the interface which should allow for composition refactors in the future.

common/ast/expr.go
TristonianJones: It's a typo on my side. I'll fix it. Thanks.

common/ast/expr.go
TristonianJones: I was thinking of using this for constant folding in the future to sanity check whether function calls have all constant inputs using the `MatchExpr` (note, I think `MatchExpr` needs a different name, maybe `MatchSubset`?)

common/ast/expr.go
TristonianJones: Sounds reasonable to me. I may have started this way because the `CheckedAST` was still out for review. Happy to update it.

TristonianJones:
PTAL

l46kok (approved):

PR #765, opened 2023-07-06
Validators for common literals and use cases

### LINKED ISSUE #359: Check the timestamp format at compile time

### LINKED ISSUE #496: Type check and pre-compile fixed regex strings

### LINKED ISSUE #655: string.format() cannot be used as intended with HomogeneousAggregateLiterals

TristonianJones:
Introduce validators for duration, timestamp, and regex string literals as well as support for homogeneous aggregate literals with a carve-out for string format calls.

Closes #359 
Closes #496 
Closes #655


TristonianJones (commented):
@DangerOnTheRanger @jpbetz @cici37 What do you think of the validator support being added here? I 'll update the string formatting checks to happen immediately after type-check in a subsequent PR, but I wanted to showcase what could be done with the latest improvements.  I'm also tempted to update the checker cost model to use the new NavigableExpr type since it's more thorough than the decorators and easier to manage. Thoughts?

This PR will be introduced after #762 is reviewed and submitted.

cel/env.go
```diff
@@ -207,12 +209,22 @@ func (e *Env) Check(ast *Ast) (*Ast, *Issues) {
...
-	return &Ast{
+	ast = &Ast{
 		source:  ast.Source(),
 		expr:    res.Expr,
 		info:    res.SourceInfo,
 		refMap:  res.ReferenceMap,
-		typeMap: res.TypeMap}, nil
+		typeMap: res.TypeMap}
+
+	// Apply additional validators on the type-checked result.
```
l46kok, line 223: Could we also update the method documentation around this change?

cel/validator.go
```diff
@@ -0,0 +1,282 @@
...
+// ValidateRegexLiterals ensures that regex patterns are validated after type-check.
+func ValidateRegexLiterals() ASTValidator {
+	return newFormatValidator(overloads.Matches, 0, compileRegex)
+}
+
+// ValidateHomogeneousAggregateLiterals checks that all list and map literals entries have the same types, i.e.
+// no mixed list element types or mixed map key or map value types.
+//
+// Note: the string format call relies on a mixed element type list for ease of use, so this check skips all
+// literals which occure within string format calls.
```
l46kok: nit: 

```suggestion
// literals which occur within string format calls.
```

cel/validator.go
```diff
@@ -0,0 +1,282 @@
...
+
+type formatValidator struct {
+	funcName string
+	argNum   int
+	check    argChecker
+}
+
+// Name returns the unique name of this function format validator.
+func (v formatValidator) Name() string {
+	return fmt.Sprintf("cel.lib.std.functions.%s", v.funcName)
```
l46kok: Thoughts on including "validator" somewhere in the name? `cel.lib.std.functions.timestamp` for example sounds like a generic standard function.

cel/validator.go
```diff
@@ -0,0 +1,282 @@
...
+				v.typeMismatch(errs, key.ID(), keyType, kt)
+			}
+			if !valType.IsEquivalentType(vt) {
+				v.typeMismatch(errs, val.ID(), valType, vt)
+			}
+		}
+	}
+}
+
+func hasStringFormatAncestor(e ast.NavigableExpr) bool {
```
l46kok, line 224: I'm a bit torn on this. IMO, I think this type of check is too specific for homogeneous literal validation to be concerned of, especially considering that `string.format` is not part of the standard library. I think I would've preferred an approach where the string formatter validator composes homogenous literal validator then decide how to validate.

cel/validator.go
```diff
@@ -0,0 +1,282 @@
...
+type errorReporter struct {
+	iss  *Issues
+	info *exprpb.SourceInfo
+}
+
+func (er *errorReporter) reportErrorAtID(id int64, message string, args ...any) {
+	er.iss.errs.ReportErrorAtID(id, locationByID(id, er.info), message, args...)
+}
+
+func locationByID(id int64, sourceInfo *exprpb.SourceInfo) common.Location {
```
l46kok, line 260: Some of the functions here, such as this seems useful enough to offer in a common helper.

TristonianJones: I ended up moving most of this functionality up into the `Issues` object in a later PR. See https://github.com/google/cel-go/pull/768 cel/env.go line 693. I can probably move that change into its own PR if you want.

cel/validator.go
TristonianJones: Since I don't know when the validators will be applied (or if they're applied), I think I have to have a carve-out here for format checking. I can make it more dynamic in the sense that it can check for library enablement to determine what to do, but I'm not sure I can do much better than that. 

If you switch the order of `ext.Strings()` and `cel.HomogeneousAggregateLiterals()` in the `NewEnv()` call, I think there's a risk that the format validator might not see the aggregate literal validator or vice-versa. The only other thing I can do is allow for replacement of the validator, but that's a bit risky since the order in which the validators are added would influence the outcome.

cel/validator.go
TristonianJones: SGTM.

cel/validator.go
TristonianJones: I'm okay with adding a TODO and coming back to this on a subsequent PR. I thought maybe I could make the check generic in the sense that if the specific overload expects a `list(dyn)` element type (I believe format is one of the few, possibly only), then homogeneous aggregate literal checks could be disabled since that was the developer's intent, but I'm not sure how string.format does it will be how other functions expect it to work.

cel/env.go
TristonianJones: Done.

cel/validator.go
l46kok: This is OK, we can bring in the change with the linked PR

l46kok (approved):

PR #766, opened 2023-07-06
Simplified partial state evaluation

TristonianJones:
Introduce the `PartialVars` method on `cel.Env` which allows users to
assume that any variable which has been declared in the environment,
but which has not been specified in the input `Activation` is unknown.

This fixes a usability issue with the `cel.PartialVars` method which expects
unknown patterns to be manually configured rather than inferred from
context.

### DIFF
--- a/cel/env.go
+++ b/cel/env.go
@@ -622,11 +622,11 @@ func (e *Env) maybeApplyFeature(feature int, option EnvOption) (*Env, error) {
 func (e *Env) computeUnknownVars(vars interpreter.Activation) []*interpreter.AttributePattern {
 	var unknownPatterns []*interpreter.AttributePattern
 	for _, v := range e.variables {
-		if _, found := vars.ResolveName(v.Name()); found {
+		varName := v.Name()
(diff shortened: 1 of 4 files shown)


cel/env.go
```diff
@@ -603,6 +617,20 @@ func (e *Env) maybeApplyFeature(feature int, option EnvOption) (*Env, error) {
 	return e, nil
 }
 
+// computeUnknownVars determines a set of missing variables based on the input activation and the
+// environment's configured declaration set.
+func (e *Env) computeUnknownVars(vars interpreter.Activation) []*interpreter.AttributePattern {
+	var unknownPatterns []*interpreter.AttributePattern
+	for _, v := range e.variables {
```
jcking, line 624: Perhaps just call v.Name() once and assign to a variable? Not sure how intelligent the Go compiler can get, considering its an interface.

TristonianJones: Done.

jcking (approved):

PR #770, opened 2023-07-08
Unify type adapter / provider representations

### LINKED ISSUE #568: Replace `exprpb.Type` references with `cel.Type`

TristonianJones:
The ref.TypeAdapter, ref.TypeProvider, and ref.TypeRegistry have been deprecated in favor of types.Adapter, types.Provider, and types.Registry respectively.

All existing uses of ref.TypeProvider / ref.TypeAdapter interfaces are wrapped into an interop layer under the covers which ensures that all internal objects refer to the new types and interfaces.

This shift marks the deprecation of the last of the exprpb.Type methods, thus simplifying implementation of custom type providers as implementers need only consider the *types.Type and not the ref.Type, exprpb.Type, or any other alternative.

Closes #568

### DIFF
--- a/cel/env.go
+++ b/cel/env.go
@@ -94,7 +94,7 @@ func (ast *Ast) Source() Source {
 
 // FormatType converts a type message into a string representation.
 //
-// Deprecated: prefer FormatCelType
+// Deprecated: prefer FormatCELType
 func FormatType(t *exprpb.Type) string {
 	return checker.FormatCheckedType(t)
 }
@@ -113,8 +113,8 @@ type Env struct {
 	variables       []*decls.VariableDecl
 	functions       map[string]*decls.FunctionDecl
 	macros          []parser.Macro
-	adapter         ref.TypeAdapter
-	provider        ref.TypeProvider
+	adapter         types.Adapter
+	provider        types.Provider
 	features        map[int]bool
 	appliedFeatures map[int]bool
 	libraries       map[string]bool
@@ -314,8 +314,8 @@ func (e *Env) Extend(opts ...EnvOption) (*Env, error) {
 	// Copy the adapter / provider if they appear to be mutable.
 	adapter := e.adapter
 	provider := e.provider
-	adapterReg, isAdapterReg := e.adapter.(ref.TypeRegistry)
-	providerReg, isProviderReg := e.provider.(ref.TypeRegistry)
+	adapterReg, isAdapterReg := e.adapter.(*types.Registry)
+	providerReg, isProviderReg := e.provider.(*types.Registry)
 	// In most cases the provider and adapter will be a ref.TypeRegistry;
 	// however, in the rare cases where they are not, they are assumed to
 	// be immutable. Since it is possible to set the TypeProvider separately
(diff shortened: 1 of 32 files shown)


cel/env.go
```diff
@@ -740,6 +757,90 @@ func getStdEnv() (*Env, error) {
...
+
+// FindStructType returns a types.Type instance for the given fully-qualified typeName if one exists.
+//
+// This method proxies to the underyling ref.TypeProvider's FindType method and converts protobuf type
+// into a native type representation. If the conversion fails, the type is listed as not found.
+func (p *interopCELTypeProvider) FindStructType(typeName string) (*types.Type, bool) {
+	if et, found := p.FindType(typeName); found {
+		t, err := types.ExprTypeToType(et)
+		if err != nil {
+			return nil, false
```
jcking, line 801: Swallowing the error feels...bad. Should the interface allow returning error? Maybe instead of `bool` we use `error` and a singleton `NotFoundErr` or something to communicate it? Or use 3 return values?

Not a requirement, just a thought.

cel/env.go
```diff
@@ -740,6 +757,90 @@ func getStdEnv() (*Env, error) {
...
+// FindStructFieldType returns a types.FieldType instance for the given fully-qualified typeName and field
+// name, if one exists.
+//
+// This method proxies to the underyling ref.TypeProvider's FindFieldType method and converts protobuf type
+// into a native type representation. If the conversion fails, the type is listed as not found.
+func (p *interopCELTypeProvider) FindStructFieldType(structType, fieldName string) (*types.FieldType, bool) {
+	if ft, found := p.FindFieldType(structType, fieldName); found {
+		t, err := types.ExprTypeToType(ft.Type)
+		if err != nil {
+			return nil, false
```
jcking, line 817: Same as above.

cel/env.go
```diff
@@ -740,6 +757,90 @@ func getStdEnv() (*Env, error) {
...
+//
+// This method proxies to the underlying types.Provider FindStructType method and converts the types.Type
+// value to a protobuf Type representation.
+//
+// Failure to convert the type will result in the type not being found.
+func (p *interopLegacyTypeProvider) FindType(typeName string) (*exprpb.Type, bool) {
+	if t, found := p.FindStructType(typeName); found {
+		et, err := types.TypeToExprType(t)
+		if err != nil {
+			return nil, false
```
jcking, line 843: Ditto.

cel/env.go
```diff
@@ -740,6 +757,90 @@ func getStdEnv() (*Env, error) {
...
+//
+// This call proxies to the types.Provider FindStructFieldType method and converts the types.FIeldType
+// value to a protobuf-based ref.FieldType representation if found.
+//
+// Failure to convert the FieldType will result in the field not being found.
+func (p *interopLegacyTypeProvider) FindFieldType(structType, fieldName string) (*ref.FieldType, bool) {
+	if cft, found := p.FindStructFieldType(structType, fieldName); found {
+		et, err := types.TypeToExprType(cft.Type)
+		if err != nil {
+			return nil, false
```
jcking, line 861: Same.

cel/env.go
TristonianJones: In this case, the not found result is usually surfaced as an error by something higher up, like the type-checker. Ideally, these conversion steps are needed in the long term, but they're definitely needed for backwards compatibility in the short-term.

I don't like swallowing the error, but the behavior is likely the same in either case: a misconfigured type doesn't exist, act accordingly. I felt this was better than a panic, but could see the argument for it.

jcking (dismissed):

jcking (approved):

PR #773, opened 2023-07-11
Open the ASTValidator API to external authors

TristonianJones:
Remove the private interface method from the `ASTValidator`

Also, introduce a `ValidatorConfig` which is computed prior to calling `Validate`.
The `ValidatorConfig` allows validators to interact with each other. Currently, to
interact with another validator's configuration, the user must also implement the
`ASTValidator` interface as well. These concerns may be separated in the future.

### DIFF
--- a/cel/env.go
+++ b/cel/env.go
@@ -204,12 +204,12 @@ func (e *Env) Check(ast *Ast) (*Ast, *Issues) {
 	if err != nil {
 		errs := common.NewErrors(ast.Source())
 		errs.ReportError(common.NoLocation, err.Error())
-		return nil, NewIssues(errs)
+		return nil, NewIssuesWithSourceInfo(errs, ast.SourceInfo())
 	}
 
 	res, errs := checker.Check(pe, ast.Source(), chk)
 	if len(errs.GetErrors()) > 0 {
-		return nil, NewIssues(errs)
+		return nil, NewIssuesWithSourceInfo(errs, ast.SourceInfo())
 	}
 	// Manually create the Ast to ensure that the Ast source information (which may be more
 	// detailed than the information provided by Check), is returned to the caller.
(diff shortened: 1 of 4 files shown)


cel/env.go
```diff
@@ -638,12 +645,21 @@ type Error = common.Error
...
 
 // NewIssues returns an Issues struct from a common.Errors object.
 func NewIssues(errs *common.Errors) *Issues {
+	return NewIssuesWithSourceInfo(errs, nil)
+}
+
+// NewIssuesWithSourceInfo returns an Issues struct from a common.Errors object with SourceInfo metatata
+// which can be used with the `ReportErrorAtID` method for additional error reports within the context
+// information that's inferred from an expression id.
+func NewIssuesWithSourceInfo(errs *common.Errors, info *exprpb.SourceInfo) *Issues {
```
jcking, line 659: Do we need to have a native type for SourceInfo? Not necessariy related to this specific PR.

cel/env.go
```diff
@@ -685,6 +701,32 @@ func (i *Issues) String() string {
...
 
+// ReportErrorAtID reports an error message with an optional set of formatting arguments.
+//
+// The source metadata for the expression at `id`, if present, is attached to the error report.
+// To ensure that source metadata is attached to error reports, use NewIssuesWithSourceInfo.
+func (i *Issues) ReportErrorAtID(id int64, message string, args ...any) {
+	i.errs.ReportErrorAtID(id, locationByID(id, i.info), message, args...)
+}
+
+func locationByID(id int64, sourceInfo *exprpb.SourceInfo) common.Location {
```
l46kok, line 717: This is from the type-checker right? Could we share this implementation?

cel/env.go
TristonianJones: I'll be creating one soon-ish, but hadn't created a native `SourceInfo` type. It seems likely that we'll add the end positions to the native type in addition to the start offsets.

cel/env.go
TristonianJones: I added a comment about refactoring SourceInfo and incorporating this code into the implementation, but I'm reluctant to expose a public interface from checker for it at this time. I'll try to figure out that issue in a later PR.

l46kok (approved):

PR #775, opened 2023-07-12
String format validator

### LINKED ISSUE #638: String formatting
**Feature request checklist**

*(This is a design proposal that is meant to be largely implemented by https://github.com/google/cel-go/pull/617 - as per offline discussion with @TristonianJones, the escaping functionality that was originally part of the formatting proposal will be implemented as a separate function, so a tracking issue seems like a good fit to cover both.*)

# Overview

String formatting is handled with `string.format`, which has the following signature:

```
string.format(formatStr, args)
```

Where `formatStr` is the formatting string, and `args` is a list containing arguments relevant to the formatting string. There is a second overloaded signature for locale identifiers (see [Internationalization](#Internationalization) for more):

```
string.format(formatStr, args, locale)
```

Implementation-wise, the formatting string is walked in a linear fashion to build up the output string, either copying from the formatting string or inserting members of args where appropriate. A potential optimization would be to do this once and cache/memoize the resulting built/formatted string, but benchmarking should probably be done to judge whether or not that would be valuable. Another optimization would be to build up an optimized, opcode/machine-readable version of any format calls on string constants (which would also fit in nicely with compile-time checking of formatting directives).

# Formatting syntax

The formatting string uses a syntax similar to the `printf`/`sprintf` implementations found in languages like C and Go. Formatting clauses that start with `%` get replaced with an item from `args` at the index implied by how many clauses there were prior.  For instance:

```
"%s, %s, %s".format(["A", “B”, “C”])
```

This would evaluate to `A, B, C`.

Conversion from non-string-types is handled by the equivalent of wrapping them with a call to CEL’s `ConvertToType`, except where datatype-specific formatting clauses are used (see [Formatting clauses](#Formatting-clauses)). The full list of supported CEL primitive types is:

* `string`
* `bool`
* `bytes`
* `double`
* `duration`
* `int`
* `list`
* `timestamp`

For arbitrary objects, their fields are printed as a list of comma-separated key-value pairs in ascending alphabetical order, all wrapped by a pair of curly brackets. This is similar to the behavior of Go’s `%#v` verb. For instance, given an object `someObject`, with a field `someStr` with a value of `“a string”` and a field `someInt` with a value of `42`, `string.format(“%s”, [someObject])` would yield:

```
“{someInt: 42, someStr: \”a string\”}”
```
As shown in the example above, string values printed as a part of this process are quoted/escaped as if they were passed to `string.format` after being escaped (See [Escaping strings](#Escaping strings)).

For escaping `%` (which is the character used to specify formatting directives), `%%` evaluates to a single `%`. For instance, `string.format(“%%”, [])` escapes to `“%”`. Note that embedding the equivalent Unicode/ASCII escape sequence (`0x0025`) is supported as well.

Some formatting clauses support a precision.  For instance, to format a double with 2 decimals of precision:

```
“%.2f”.format([0.269])
```

This would give:

```
0.27
```

The [Formatting clauses](#Formatting-clauses) section goes into more detail about formatting clauses.

Outside of the instances mentioned above for argument substitution, all other characters are used unmodified in the output string.

## Formatting clauses

As mentioned above, formatting clauses are specified as characters following the `%` character. For the initial release, `f` (for fixed-point), `e` (for scientific notation), `b` (for binary formatting), `h` (for hexadecimal formatting), and `o` (for octal formatting) will be supported, but others could be added in the future.

Naturally, an argument that doesn’t support the formatting directive results in an error. For instance, the following is not allowed:

```
“%.4f”.format([“not a number”])
```

For fixed-point formatting clauses, the character f must have a precision specified. In other words, the following is not allowed:

```
“%f”.format([0.12345])
```

For cases where the number of requested digits after the decimal is greater than the precision of the actual number given, the resulting string is padded with zeros. For example:

```
“%.3f”.format([0.3])
```

This gives:

```
“0.300”
```

For cases where the given number has higher precision than the number of requested digits, the resulting string uses a rounded version of the given number. For example:

```
“%.2f”.format([0.347])
```

This gives:

```
“0.35”
```

This mirrors the behavior found in Go and Python.

For scientific notation, there is the `e` formatting clause. `E` (in upper-case) is also valid, which changes the case of the `e` character used in the resulting string. In both cases, the character must be followed by the precision that should be used, so the following is not allowed:

```
string.format(“%e”, [0.3])
```

As mentioned earlier, the case of the formatting clause character can be changed to upper-case in order to change the case of the e in the resulting string. So for instance, given:

```
string.format(“%.6e”, [1052.032911275])
```

This gives:

```
“1.052033E+03”
```

Formatting integers as binary can be done with the `b` formatting clause. For instance, given:

```
string.format(“%b”, [5])
```

This outputs:

```
“101”
```

Likewise, `h` (and `H` for uppercase hexadecimal digits) is supported for formatting integers in base 16/hexadecimal. Again, given:

```
string.format(“%h”, [15])
```

This gives:

```
“F”
```

For formatting integers in base 8/octal there is the `o` formatting clause. As an example, if given:

```
string.format(“%o”, [11])
```

This will output:

```
“13” 
```

# Error handling

When the formatting string is a constant, the CEL compiler will check at compile/type-checking-time if the arguments passed to format match with the formatting directives. Any mismatches result in a compile-time error. Unused arguments also cause a compile-time error. For instance, the following will result in both errors:

```
“%d %d”.format([0, 1, 2])
```

In other cases (if the string is a variable or otherwise its own expression), errors that occur during formatting are indicated by adding additional output to the end of the formatted string.

# Internationalization

Since formatting things like numbers is locale-specific, `string.format` needs a way to deal with different locales. CEL allows function overloading, and so `string.format` can take a third argument that represents a locale identifier. If any locale-specific formatting is used, the locale identifier will modify the output as is appropriate. For example, given a formatting string that includes a fixed-point formatting clause and a French locale identifier:

```
string.format(“%.2f”, [3.14], “fr”)
```

This gives:

```
“3,14”
```

The full list of formatting clauses that are affected by the locale identifier is:

* `f` (locale affects whether `.` or `,` is used as the decimal point)
* `E` and `e` (locale affects whether `.` or `,` is used as the decimal point)

An unrecognized locale identifier is considered a compile-time error if the identifier is a string constant. If the identifier is not a constant (the user passed in a string variable instead), then a warning will be appended to the formatted string, and the default locale (`en_US`) will be used instead.

If the locale argument is not passed, or if an invalid non-constant locale identifier is passed, then the locale defaults to `en_US`. This is the case regardless of the locale that the CEL implementation is operating in or built with.

# String escaping

There will also be a separate function implemented as part of this proposal - `string.escape`, which escapes non-printable characters.

For example, given:

```
“this is a string \n with no newline”.escape()
```

This would result in:

```
“\”this is a string \\n with no newline\””
```

If printed, this would result in the user seeing a literal “\n” instead of a newline. This function is useful for cases where a string needs to be printed (as a part of an error, for instance) while safely ensuring surrounding formatting is preserved. The complete list of escape sequences that string.escape covers is:

* `\a` (bell)
* `\b` (backspace)
* `\f` (form feed)
* `\n` (newline)
* `\r` (carriage return)
* `\t` (horizontal tab)
* `\v` (vertical tab)

# Alternatives considered

## Automatically deriving locale

As an alternative to the manually-specified locale identifier that can be passed as the third argument to `string.format`, the locale could be derived from the OS/environment variables.

However, Golang does not provide a built-in means of detecting locale, and the way locale is stored varies from OS to OS. As far as I’m aware, CEL doesn’t limit support to one particular OS or another (cel-go should work everywhere Golang does), so it is not clear which OSes need implementations.

## Indexed argument substitution

With Python’s f-strings, it is possible to do something like:

```
“{2} {1} {0}”.format(“C”, “B”, “A”)
```

Which outputs:

```
“A B C”
```

However, this can be accomplished with implicit incrementing by reordering the arguments. For indexed arguments, there is also the problem of adding/removing an argument to the string; any addition not done at the end of the string (or at least after all other substitutions) will result in needing to re-number all other indices.

Another option is having both (Python does this), but for the sake of keeping CEL and its spec simple, I believe it is best to pick one.

TristonianJones:
Shift string format validation into the `ext.Strings()` library.

This change also removes all references to the string format extensions within
the cel and interpreter packages, and shifts validation from an `InterpretableDecorator`
to an `cel.ASTValidator` which is applied immediately after type-checking.

Note to reviewers: the majority of the changes are present in the `ext/formatting.go`
file, and contained within the `stringFormatValidator`. There are some subtle differences
such as the new validator checks by function name only, and not by overload id (this
will be addressed in a future PR), and the argument length check applies to the original
call signature (size 1) rather than the internal call signature which includes the call
target as the first argument (size 2). Most of the rest of the code is quite similar
to what existed beforehand.

Closes https://github.com/google/cel-go/issues/638


TristonianJones:
@DangerOnTheRanger would you mind taking a look to see if this matches your expectations?

ext/strings_test.go
```diff
@@ -1254,36 +1247,38 @@ func TestStringFormat(t *testing.T) {
...
 			cel.Abbrevs("google.expr.proto3.test"),
 			cel.Types(&proto3pb.TestAllTypes{}),
-			cel.ASTValidators(cel.ValidateHomogeneousAggregateLiterals()),
 			NativeTypes(
 				reflect.TypeOf(&TestNestedType{}),
 				reflect.ValueOf(&TestAllTypes{}),
 			),
 		}
+		if !skipCompileCheck {
+			opts = append(opts, cel.ASTValidators(cel.ValidateHomogeneousAggregateLiterals()))
```
DangerOnTheRanger: I'm a little confused as to why `ValidateHomogeneousAggregateLiterals` got moved inside this guard here - is there some kind of additional checking that will accidentally get caught if we check literals while not checking `string.format`?

ext/strings_test.go
```diff
@@ -1254,36 +1247,38 @@ func TestStringFormat(t *testing.T) {
...
 		}
-		return evalExpr(env, expr, dynArgs, skipCompileCheck, expectedRuntimeCost, expectedEstimatedCost, t)
+		return evalExpr(env, expr, dynArgs, expectedRuntimeCost, expectedEstimatedCost, t)
 	}
 	checkCase := func(output ref.Val, expectedOutput string, err error, expectedErr string, t *testing.T) {
 		if err != nil {
 			if expectedErr != "" {
-				if err.Error() != expectedErr {
+				if !strings.Contains(err.Error(), expectedErr) {
 					t.Fatalf("expected %q as error message, got %q", expectedErr, err.Error())
```
DangerOnTheRanger, line 1280: Super tiny nit: I think the `strings.Contains` change is good, though maybe it might be better for this line to now read something like `"expected error message to contain %q, got %q"` as a result.

ext/formatting.go
```diff
@@ -0,0 +1,878 @@
...
+	// TODO: expose the OverloadID() method on the NavigableCallExpr to refine this check.
+	if !call.IsMemberFunction() || call.FunctionName() != "format" {
+		return false
+	}
+	formatString := call.Target()
+	if formatString.Kind() != ast.LiteralKind && formatString.Type() != cel.StringType {
+		return false
+	}
+	args := call.Args()
+	if len(args) != 1 {
```
DangerOnTheRanger, line 477: Out of curiosity, did you think there was anything wrong with checking for the arg length of the original call signature? It makes sense to eventually move back to the `overloadID` check instead of the function name (to avoid accidentally looking at other member functions that are also named `format`) but this seems fine to me.

DangerOnTheRanger (commented):
Just had a couple small comments, nothing major or anything - LGTM otherwise.

ext/formatting.go
TristonianJones: In the AST, the `format` is a receiver-style call: `<string>.format(<list>)`, so the call will have a `target` which will be the format string, and a single argument. During program planning, the call structure is normalized and the target is moved to the first argument. Hence, the subtle change here. Make sense?

ext/strings_test.go
TristonianJones: I can leave it enabled. I can't recall why I moved it.

jpbetz (commented):
The way the options are configured (enabled by default) LGTM.

I'll defer to @DangerOnTheRanger on any specifics of the formatting code.

ext/formatting.go
DangerOnTheRanger: Right, I think I somehow misunderstood and thought you were wanting to change/replace the `len(args) != 1` check, sorry - this sounds good to me.

DangerOnTheRanger:
LGTM

jcking (approved):

PR #776, opened 2023-07-12
Update types.Unknown with attribute trails

### LINKED ISSUE #771: Include missing attributes on types.Unknown
**Feature request checklist**

**Change**
Currently, the `types.Unknown` value is a simple type alias to a list of integers; however,
the unknown response is significantly richer in C++, as it includes the list of missing
attributes in addition to the expression ids which were unknown.

Note: this will be a breaking API change for users of `types.Unknown`; however, it will
make the unknown responses significantly more usable than they are today.

**Alternatives considered**
It's possible to reconstruct the missing attributes and calls post-evaluation; however,
doing so accurately requires state tracking which is a significantly expensive overhead
and likely not worth the cost in most cases.

TristonianJones:
Update the types.Unknown definition.

## Breaking change

Use `*types.Unknown` rather than `types.Unknown` during
type casting as the underlying type has changed from a slice
to a structured object.

Specifically, shift `types.Unknown` from a simple list of ints
to a structured object which contains the set of unknown 
expression ids, as well as the attribute trails associated with
each. 

In the future this object may be further augmented with
information about missing function invocations.

Closes #771

### DIFF
--- a/common/types/unknown.go
+++ b/common/types/unknown.go
@@ -15,45 +15,173 @@
 package types
 
 import (
+	"fmt"
 	"reflect"
+	"strings"
+	"unicode"
 
 	"github.com/google/cel-go/common/types/ref"
 )
 
-// Unknown type implementation which collects expression ids which caused the
-// current value to become unknown.
-type Unknown []int64
+var (
+	unspecifiedAttribute = &AttributeTrail{qualifierPath: []any{}}
+)
+
+// NewAttributeTrail creates a new simple attribute from a variable name.
+func NewAttributeTrail(variable string) *AttributeTrail {
+	if variable == "" {
+		return unspecifiedAttribute
+	}
+	return &AttributeTrail{variable: variable}
+}
+
+// AttributeTrail specifies a variable with an optional qualifier path. An attribute value is expected to
+// correspond to an AbsoluteAttribute, meaning a field selection which starts with a top-level variable.
+//
+// The qualifer path elements adhere to the AttributeQualifier type constraint.
+type AttributeTrail struct {
+	variable      string
+	qualifierPath []any
+}
+
+// Equals returns whether two attribute values have the same variable name and qualifier paths.
+func (a *AttributeTrail) Equal(other *AttributeTrail) bool {
+	if a.Variable() != other.Variable() || len(a.QualifierPath()) != len(other.QualifierPath()) {
+		return false
+	}
+	for i, q := range a.QualifierPath() {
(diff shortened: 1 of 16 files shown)


TristonianJones:
FYI @josephschorr

common/types/unknown.go
```diff
@@ -15,45 +15,173 @@
...
+
+// QualifyAttribute qualifies an attribute using a valid AttributeQualifier type.
+func QualifyAttribute[T AttributeQualifier](attr *AttributeTrail, qualifier T) *AttributeTrail {
+	attr.qualifierPath = append(attr.qualifierPath, qualifier)
+	return attr
+}
+
+// Unknown type which collects expression ids which caused the current value to become unknown.
+type Unknown struct {
+	attributeTrails map[int64]*AttributeTrail
```
jnthntatum: You may want to consider making this a multimap or a list of pairs to support cases where the expr node can be tied to multiple attributes (comprehensions). I know you are deferring handling comprehensions, but it might spare another breaking change later.

common/types/unknown_test.go
```diff
@@ -22,43 +22,239 @@ import (
...
+			equal: true,
+		},
+		{
+			a:     QualifyAttribute[string](NewAttributeTrail("a"), "b"),
+			b:     NewAttributeTrail("a"),
+			equal: false,
+		},
+		{
+			a:     QualifyAttribute[string](NewAttributeTrail("a"), "b"),
+			b:     QualifyAttribute[int64](NewAttributeTrail("a"), 1),
```
jnthntatum, line 72: could you add a test case for a[1] == a[1u]? I'm not sure whether these should be treated as equal or not since maps treat them as equivalent with heterogeneous numeric equality

interpreter/interpreter_test.go
```diff
@@ -1450,7 +1450,7 @@ func TestInterpreter(t *testing.T) {
 				want = tc.out.(ref.Val)
 			}
 			got := prg.Eval(vars)
-			_, expectUnk := want.(types.Unknown)
+			_, expectUnk := want.(*types.Unknown)
```
jnthntatum, line 1506: could you add or update a test case that covers this?

jnthntatum (changes requested):

common/types/unknown.go
TristonianJones: Good point, will do.

TristonianJones:
@jnthntatum PTAL, thanks!

jnthntatum (approved):

PR #779, opened 2023-07-18
ContextProtoVars() to simplify proto-based inputs

TristonianJones:
Additional method to create an `interpreter.Activation` given a protobuf message as input.

This PR addresses feedback present in #684 and provides an evaluation equivalent for `DeclareContextProto`

### DIFF
--- a/common/types/pb/type.go
+++ b/common/types/pb/type.go
@@ -467,13 +467,13 @@ func unwrapDynamic(desc description, refMsg protoreflect.Message) (any, bool, er
 		unwrappedAny := &anypb.Any{}
 		err := Merge(unwrappedAny, msg)
 		if err != nil {
-			return nil, false, err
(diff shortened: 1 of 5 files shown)


common/types/pb/type.go
```diff
@@ -564,8 +564,10 @@ func zeroValueOf(msg proto.Message) proto.Message {
 }
 
 var (
+	jsonValueTypeURL = "types.googleapis.com/google.protobuf.Value"
```
l46kok, line 567: This is just to unmarshal an empty `any{}` into JSON's NullValue right?

l46kok (approved):

common/types/pb/type.go
TristonianJones: That's right. It was missing before and so an unset any field was actually erroring on access. It would be a good conformance test case.

PR #787, opened 2023-08-04
Fix cost estimates to propagate result sizes

jpbetz:
Found that we dropped some result sizes while computing estimated costs while digging into https://github.com/kubernetes/kubernetes/issues/119749.  This plumbs them in.

### DIFF
--- a/checker/cost.go
+++ b/checker/cost.go
@@ -533,14 +533,34 @@ func (c *coster) functionCost(function, overloadID string, target *AstNode, args
 
 	if est := c.estimator.EstimateCallCost(function, overloadID, target, args); est != nil {
 		callEst := *est
-		return CallEstimate{CostEstimate: callEst.Add(argCostSum())}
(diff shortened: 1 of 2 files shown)


checker/cost.go
```diff
@@ -533,14 +533,34 @@ func (c *coster) functionCost(function, overloadID string, target *AstNode, args
 
 	if est := c.estimator.EstimateCallCost(function, overloadID, target, args); est != nil {
 		callEst := *est
-		return CallEstimate{CostEstimate: callEst.Add(argCostSum())}
+		return CallEstimate{CostEstimate: callEst.Add(argCostSum()), ResultSize: est.ResultSize}
```
TristonianJones, line 536: Would helper methods to create size and cost based estimates help avoid missing sizes in the future?

jpbetz: Yeah, I think it would. How about I add it as follow up so we can backport this as a minimal fix?

TristonianJones (approved):

PR #788, opened 2023-08-05
Split Expr from NavigableExpr with interpreter support

TristonianJones:
Introduce a Go-native `Expr` representation

This PR splits the `Expr` interface from the `NavigableExpr` interface and introduces
support for the native value within the `interpreter` call paths.

The following changes will be occur in the following sequence once this PR is submitted:
* Update the type-checker to use the `Expr` value.
* Update the parser / unparser / pruner to use the `Expr` value.
* Introduce new Macro-related helpers to assist with backwards compatibility support
  for existing custom macros.

#789


checker/checker.go
```diff
@@ -64,13 +64,15 @@ func Check(parsedExpr *exprpb.ParsedExpr, source common.Source, env *Env) (*ast.
...
-	return &ast.CheckedAST{
-		Expr:         parsedExpr.GetExpr(),
-		SourceInfo:   parsedExpr.GetSourceInfo(),
-		TypeMap:      m,
-		ReferenceMap: c.references,
-	}, errs
+	}
+	info, err := ast.ProtoToSourceInfo(parsedExpr.GetSourceInfo())
+	if err != nil {
+
```
jcking: ?

jcking (dismissed):

checker/checker.go
TristonianJones: Missed this when I was trying to break up the larger PR into smaller ones. Thank you.

jcking (approved):

Issue #789, opened 2023-08-06
Migrate to a native AST and Expr representation


TristonianJones:
Migrate to a native AST and Expr representation.

The original cel-go implementation was heavily based on protobuf definitions
for types, declarations, and expressions. While this was expedient, it has also
been cumbersome. With PR #568, the types and declarations were replaced
by Go-native structs. This issue tracks the migration of the `Expr`, `SourceInfo`,
and `AST` objects. 

The motives for making this change are many:

* Performance
* Remove (most) hard dependencies on protobuf
* Allow for better interop between serialization formats and type systems

TristonianJones:
As part of the optimizer work, it's clear that the split between `cel.Ast` and `ast.AST` is confusing. The two structures exist separately to ease migration, but should be merged before this issue can be considered closed.

charithe:
Apologies if this is not the right place to ask this. Is the plan to unify the two AST data types into one in a future release?

I am a bit confused by how we are supposed to use the new native types right now. Most of the methods to work with parsed expressions take arguments of type `ast.Expr`. However, the return type of `Compile` and `Check` methods of `cel.Env` is `cel.Ast` which doesn't have a method to obtain the underlying `ast.Expr` object. I am currently forced to do something along the lines of the following:

```go
celAST, _ := env.Compile(exprString)
checkedExpr, _ := cel.AstToCheckedExpr(celAST)
astAST, _ := ast.ToAST(checkedExpr)
expr := astAST.Expr() // finally have an `ast.Expr` that I can use with `ast.PreOrderVisit` etc.
```

Is there a better way to do this with the current version of CEL (0.18.1) or should I wait for the AST types to be unified?

charithe:
@TristonianJones apologies for pinging you directly. I am stuck on the above issue of not being able to access the underlying `ast.Expr` of the `cel.Ast` without jumping through a lot of hoops. Would you be open to a PR that introduces two functions to obtain the `Expr` and `SourceInfo` of an `Ast` object like the following or adding them as methods to `cel.Ast`?

```diff
diff --git a/cel/io.go b/cel/io.go
index 3133fb9..50bb878 100644
--- a/cel/io.go
+++ b/cel/io.go
@@ -101,22 +101,36 @@ func AstToString(a *Ast) (string, error) {
 	return parser.Unparse(a.impl.Expr(), a.impl.SourceInfo())
 }
 
+// AstToExpr returns the underlying expr of this Ast
[...]

PR #793, opened 2023-08-09
Migrate the type-checker to a native AST representation

TristonianJones:
This change shifts the parse / check top-level APIs to use the native AST and Expr
representations. 

Updates to the parser internals, macro expansion signatures, cost estimation,
and unparsing APIs will be provided in subsequent PRs.

#789

### DIFF
--- a/common/ast/conversion.go
+++ b/common/ast/conversion.go
@@ -298,7 +298,7 @@ func EntryExprToProto(e EntryExpr) (*exprpb.Expr_CreateStruct_Entry, error) {
 
 func protoCall(id int64, call CallExpr) (*exprpb.Expr, error) {
 	var err error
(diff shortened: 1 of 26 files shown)


common/ast/conversion.go
```diff
@@ -293,7 +298,7 @@ func EntryExprToProto(e EntryExpr) (*exprpb.Expr_CreateStruct_Entry, error) {
 
 func protoCall(id int64, call CallExpr) (*exprpb.Expr, error) {
 	var err error
-	var target *exprpb.Expr
+	var target *exprpb.Expr = nil
```
jcking: Isn't this the default already in Go?

jcking (dismissed):

jcking (approved):

PR #796, opened 2023-08-11
Creating a Function that Reverses a String

bboogler:
## Commit Message
Creating a reverse function for CEL that takes a string, and returns a new string whose characters are in the reverse order of the original.

Function that can take "gums" and return "smug".
```

### DIFF
--- a/ext/strings.go
+++ b/ext/strings.go
@@ -267,6 +267,19 @@ const (
 //
 //	'TacoCat'.upperAscii()      // returns 'TACOCAT'
 //	'TacoCÆt Xii'.upperAscii()  // returns 'TACOCÆT XII'
+//
+// # Reverse
+//
+// Returns a new string whose characters are the same as the target string, only formatted in
+// reverse order.
+// This function relies on converting strings to rune arrays in order to reverse
+//
+//	<string>.reverse() -> <string>
+//
+// Examples:
+//
+//	'gums'.reverse() // returns 'smug'
+//	'John Smith'.reverse() // returns 'htimS nhoJ'
 func Strings(options ...StringsOption) cel.EnvOption {
 	s := &stringLib{
 		version:        math.MaxUint32,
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: ext/strings.go [hunk 1]
```diff
@@ -445,6 +458,12 @@ func (lib *stringLib) CompileOptions() []cel.EnvOption {
 					s := str.(types.String)
 					return stringOrError(upperASCII(string(s)))
 				}))),
+		cel.Function("reverse",
```
TristonianJones: Please introduce a new 'version' to optionally enable this feature. By default max version is used, but for teams that want to control function rollout having a new version of the library they can opt into is important.

ext/strings_test.go
```diff
@@ -105,6 +105,10 @@ var stringTests = []struct {
 	// Upper ASCII tests.
 	{expr: `'tacoCat'.upperAscii() == 'TACOCAT'`},
 	{expr: `'tacoCαt'.upperAscii() == 'TACOCαT'`},
+	// Reverse tests.
+	{expr: `'gums'.reverse() == 'smug'`},
+	{expr: `'palindromes'.reverse() == 'semordnilap'`},
+	{expr: `'John Smith'.reverse() == 'htimS nhoJ'`},
```
TristonianJones, line 111: Could you please add some unicode characters in the reverse? While it's clear in the implementation that you're swapping runes and not bytes, for someone else who wants to implement this function, such a test would be useful for conformance checks.

TristonianJones (changes requested):

TristonianJones:
@bboogler also, please update the `ext/README.md` with the documentation for your new function and a note about which version of the library where it was introduced.

ext/strings_test.go
bboogler: Unicode character tests have been added. It should be noted that for arguments such as "\u180etext\u200b\u200c\u200d\u2060\ufeff" the result is the reverse of the string the unicode characters represent, which in this example would be "​txet᠎". However, should "\u180etext\u200b\u200c\u200d\u2060\ufeff" be passed in reverse "ffefu\0602u\d002u\c002u\b002u\txete081u\", then an unknown escape error is encountered as it possesses escape characters. The only way to avoid this situation of passing in invalid unicode is to ensure that they are passed in as raw string literals. For example, passing in `ffefu\0602u\d002u\c002u\b002u\txete081u\` correctly returns "\u180etext\u200b\u200c\u200d\u2060\ufeff".

ext/strings.go
bboogler: New version introduced.

TristonianJones (approved):

TristonianJones:
@bboogler do you mind syncing with HEAD? That should fix the build error you've encountered.

bboogler:
@TristonianJones I don't mind, just synced. Blocked on  /gcbrun.

bboogler (commented):
The reverse function for taking a string argument and returning a new string whose characters are the original's in reverse order, is implemented and documented with this code.

PR #804, opened 2023-08-15
Static optimizers for constant folding

TristonianJones:
Support for constant folding and configurable optimizers.

Constant folding introduces several API changes to support the construction
ASTs which are intended to be type-check and unparser friendly. The constant
folding algorithm exercises many parts of this new framework and does
best effort folding through comprehension expressions as well.

Static optimizers are standalone instances and not configured as part of the standard
environment simply because their options, configurations, and sequencing may be
use-case specific and might not fit well within the parse -> compile -> eval flow.


cel/optimizer.go
```diff
@@ -0,0 +1,391 @@
...
+	return &Ast{
+		source: a.Source(),
+		impl:   optimized,
+	}, nil
+}
+
+// normalizeIDs ensures that the metadata present with an AST is reset in a manner such
+// that the ids within the expression correspond to the ids within macros. This function
+// ensures that
+func normalizeIDs(e *Env, optimized *ast.AST) {
```
l46kok, line 94: Perhaps this is a naming or a type consistency problem but in general, the switch between `cel.Ast` and `ast.AST` is a bit confusing. Do you think there's a better way to distinguish the AST impl. vs the AST used as part of the public interface?

cel/optimizer.go
```diff
@@ -0,0 +1,391 @@
...
+	// Return the optimized result.
+	return &Ast{
+		source: a.Source(),
+		impl:   optimized,
+	}, nil
+}
+
+// normalizeIDs ensures that the metadata present with an AST is reset in a manner such
+// that the ids within the expression correspond to the ids within macros. This function
+// ensures that
```
l46kok: Incomplete comment

cel/folding_test.go
```diff
@@ -0,0 +1,195 @@
...
+package cel_test
+
+import (
+	"testing"
+
+	"github.com/google/cel-go/cel"
+	"github.com/google/cel-go/test/proto3pb"
+)
+
+func TestConstantFoldingOptimizer(t *testing.T) {
```
l46kok, line 31: Can we add test cases around:

- Logical operators
- Operator ordering and its expected optimization behavior (e.g: `1 + 2 + x` vs `x + 1 + 2`)
- ID normalization

cel/folding.go
```diff
@@ -0,0 +1,450 @@
...
+	// Update the fold expression to be a literal.
+	expr.SetKindCase(ctx.NewLiteral(out))
+	return nil
+}
+
+// maybePruneBranches inspects the non-strict call expression to determine whether
+// a branch can be removed. Evaluation will naturally prune logical and / or calls,
+// but conditional will not be pruned cleanly, so this is one small area where the
+// constant folding step reimplements a portion of the evaluator.
+func maybePruneBranches(expr ast.NavigableExpr) bool {
```
l46kok: Can we prune `in` as well? e.g: `1 in [1, x + 2, 6]` -> true

cel/folding.go
```diff
@@ -0,0 +1,450 @@
...
+				return a
+			}
+		}
+		foldableExprs = ast.MatchDescendants(root, constantExprMatcher)
+	}
+	// Once all of the constants have been folded, try to run through the remaining comprehensions
+	// one last time. In this case, there's no guarantee they'll run, so we only update the
+	// target comprehension node with the literal value if the evaluation succeeds.
+	for _, compre := range ast.MatchDescendants(root, ast.KindMatcher(ast.ComprehensionKind)) {
+		tryFold(ctx, a, compre)
```
l46kok, line 94: Can we test this behavior? Perhaps with a deeply nested comprehension

cel/folding.go
```diff
@@ -0,0 +1,450 @@
...
+	}
+
+	// If the output is a list, map, or struct which contains optional entries, then prune it
+	// to make sure that the optionals, if resolved, do not surface in the output literal.
+	pruneOptionalElements(ctx, root)
+
+	// Ensure that all intermediate values in the folded expression can be represented as valid
+	// CEL literals within the AST structure.
+	ast.PostOrderVisit(root, ast.NewExprVisitor(func(e ast.Expr) {
```
l46kok, line 104: Looks like you can also use `MatchDescendants` here with the literal kind

common/ast/expr.go
```diff
@@ -606,6 +609,15 @@ func (e *baseListExpr) Elements() []Expr {
 	return e.elements
 }
 
+func (e *baseListExpr) IsOptional(index int32) bool {
+	for _, optIndex := range e.OptionalIndices() {
```
l46kok: Would it be feasible to create a map equivalent and access that instead?

cel/folding.go
```diff
@@ -0,0 +1,450 @@
...
+type constantFoldingOptimizer struct{}
+
+// Optimize queries the expression graph for scalar and aggregate literal expressions within call and
+// select statements and then evaluates them and replaces the call site with the literal result.
+//
+// Note: only values which can be represented as literals in CEL syntax are supported.
+func (*constantFoldingOptimizer) Optimize(ctx *OptimizerContext, a *ast.AST) *ast.AST {
+	root := ast.NavigateAST(a)
+
+	// Walk the list of foldable expression and continue to fold until there are no more folds left.
```
l46kok, line 68: Perhaps we could add a max iteration limit to avoid an infinite loop

l46kok (commented):
Submitting what I have so far -- the literal adaptation and optional handling is sufficiently complex. It could be a good idea to create abstractions for them to reuse them across other optimizer implementations.

cel/optimizer.go
TristonianJones: I'm not sure yet. I don't exactly like that I have `cel.Ast` and `ast.AST`, but I've tried to minimize the delta between them and stick with `cel.Ast` as the top-level object since I'll have to migrate it at some point. It's likely that I'll completely deprecate `cel.Ast` in favor of `cel.AST` which includes a `common.Source` reference and embeds an `ast.AST` instance. 

How does that sound to you? If that seems reasonable, I'll add an issue to track it.

cel/folding_test.go
TristonianJones: My mistake on the branches, I realize I had them in the inlining tests. Adding some specifically to the folding tests (along with the others requested).

cel/folding.go
TristonianJones: We can, and we can also prune statements like `x in []` -> false; however, I haven't done so yet. Would you like to see them added in this PR?

cel/optimizer.go
l46kok: That sounds good. Having a single consistent AST type to work with is definitely the most ideal state to be at.

cel/folding.go
l46kok: Up to you, a separate PR is totally fine

cel/folding.go
TristonianJones: I've added a test which does not fold the comprehension because at the lowest level the comprehension contains a variable. Otherwise, the `tryFold` is more easily exercised within the inliner tests which can inject `cel.bind()` into the testing context.

cel/folding.go
TristonianJones: I can, but the Match call generates new `ast.NavigableExpr` instances, so as a minor optimization, I used `PostOrderVisit` which will not allocate any new objects but just traverse the existing ones.

common/ast/expr.go
TristonianJones: Yes, sure. I left it as an exercise for the future if this proves to be a hotspot, but can do it now.

cel/folding.go
TristonianJones: Good point, I'll add a configuration option to the `NewConstantFoldingOptimizer` call.

cel/optimizer.go
TristonianJones: I added a note on #789 regarding the AST.

cel/folding.go
TristonianJones: Too late! I added support here. :)

cel/folding.go
```diff
@@ -0,0 +1,549 @@
...
+		return true
+	}
+	expr.SetKindCase(ctx.NewCall(function, newArgs...))
+	return true
+}
+
+// pruneOptionalElements works from the bottom up to resolve optional elements within
+// aggregate literals.
+//
+// Note, may aggregate literals will be resolved as arguments to functions or select
```
l46kok: nit: may aggregate?

cel/folding.go
```diff
@@ -0,0 +1,549 @@
...
+func pruneOptionalMapEntries(ctx *OptimizerContext, e ast.Expr) {
+	m := e.AsMap()
+	entries := m.Entries()
+	updatedEntries := []ast.EntryExpr{}
+	modified := false
+	for _, e := range entries {
+		entry := e.AsMapEntry()
+		key := entry.Key()
+		val := entry.Value()
+		if !entry.IsOptional() || val.Kind() != ast.LiteralKind {
```
l46kok, line 282: `LiteralKind` check doesn't seem to be needed since the entry is already being checked for being an optional

l46kok (dismissed):

cel/folding.go
TristonianJones: Optionals that are not literals can't be elided, so that's what this check is doing. If the entry is not optional, or the value-side of the optional hasn't been resolved to a literal, then preserve the entry as-is.

TristonianJones: I've added a comment to indicate what's happening here for future reference.

TristonianJones:
PTAL

l46kok (approved):

PR #806, opened 2023-08-15
Migrate cel.Ast to be a thin layer on ast.AST

TristonianJones:
Code cleanup related to #789

### DIFF
--- a/common/ast/ast.go
+++ b/common/ast/ast.go
@@ -99,7 +99,7 @@ func (a *AST) SetReference(id int64, r *ReferenceInfo) {
 
 // IsChecked returns whether the AST is type-checked.
 func (a *AST) IsChecked() bool {
-	return a != nil && (len(a.TypeMap()) > 0 || len(a.ReferenceMap()) > 0)
+	return a != nil && len(a.TypeMap()) > 0
 }
 
 // NewAST creates a base AST instance with an ast.Expr and ast.SourceInfo value.
--- a/common/ast/ast.go
+++ b/common/ast/ast.go
@@ -27,7 +27,6 @@ type AST struct {
 	sourceInfo *SourceInfo
 	typeMap    map[int64]*types.Type
 	refMap     map[int64]*ReferenceInfo
-	checked    bool
 }
 
 // Expr returns the root ast.Expr value in the AST.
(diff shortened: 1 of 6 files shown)


common/ast/ast.go
```diff
@@ -100,7 +99,7 @@ func (a *AST) SetReference(id int64, r *ReferenceInfo) {
 
 // IsChecked returns whether the AST is type-checked.
 func (a *AST) IsChecked() bool {
-	return a != nil && a.checked
+	return a != nil && (len(a.TypeMap()) > 0 || len(a.ReferenceMap()) > 0)
```
l46kok: Would checking the type map sufficient here? Can there be an AST that has references but no types?

TristonianJones: You're right, I think you just need the type map

TristonianJones:
PTAL

l46kok (approved):

tonyhb:
@TristonianJones would it be possible to grab the *ast.AST from *cel.Ast?  I'd like to walk the expressions to do some, well, things to it :)

TristonianJones:
@tonyhb are you able to use the new Optimizer interfaces? They'll allow you to access the internal AST. In the future, I'll replace cel.Ast, but it's a bit of a compatibility challenge at the moment and I haven't had a chance to dig in.

tonyhb:
I had just implemnted the constant folding optimizer and noticed that it's Optimize call gets an *ast.AST, and that the static optimizer copies *Ast.impl direclty into *ast.AST — so I think that should do it.  Thanks for the heads up.

I think you might be interested in why we need this:

We're compiling arbitrary expressions, and matching on arbitrary events (similar to pub/sub where this is used).  The main difference is that expressions can be dynamically created, leading to a factorial combination of expression checking.  Each event received must check _every_ expression for matches;  any new event or expression saved compounds expression checking.

The only way to simplify this is if we explicitly _dont_ evaluate each event.  It's a pain.

The plan is:  parse the AST and check each predicate.  For each predicate, create and merge the expression into a specific tree type for quick evaluation.

EG:  If the expression tests string equality, throw the string match into an adaptive radix tree, where the leafs point to its expression.  This way we can filter out all invalid expressions without having to test each one.  We'll have to do the same for LT/GT on ints, etc.

It's an early plan right now, and we've only literally just started research on it.

Anyway, thanks again for the help!

TristonianJones:
@tonyhb I think what you want is the disjunctive normal form of the expression to compute a trie / graph of the predicates and interested parties. This relates to our own thoughts related to expression set optimizers. Keep me posted! This sounds exciting

tonyhb:
Yep, literally that.  Then merging all of the applicable trees into one set so that we can optimize which expressions need evaluating given input data.  We just chatted about this last night for the first time, will let you know how investigation goes this weekend!

PR #812, opened 2023-08-18
Make `types.CheckedWellknowns` public

jarreds:
Allow external proto types to be exposed as well known types to override the default `protoObj` behaviors.

### DIFF
--- a/common/types/provider.go
+++ b/common/types/provider.go
@@ -351,7 +351,7 @@ func (p *Registry) registerAllTypes(fd *pb.FileDescription) error {
 	for _, typeName := range fd.GetTypeNames() {
 		// skip well-known type names since they're automatically sanitized
 		// during NewObjectType() calls.
-		if _, found := checkedWellKnowns[typeName]; found {
+		if _, found := CheckedWellKnowns[typeName]; found {
 			continue
 		}
 		err := p.RegisterType(NewObjectTypeValue(typeName))
--- a/common/types/types.go
+++ b/common/types/types.go
@@ -514,7 +514,7 @@ func NewOpaqueType(name string, params ...*Type) *Type {
 // type may also indicate additional traits through the use of the optional traits vararg argument.
 func NewObjectType(typeName string, traits ...int) *Type {
 	// Function sanitizes object types on the fly
-	if wkt, found := checkedWellKnowns[typeName]; found {
+	if wkt, found := CheckedWellKnowns[typeName]; found {
 		return wkt
 	}
 	traitMask := 0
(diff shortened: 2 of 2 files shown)


jarreds:
Note: `common/types/pb.CheckedWellKnowns` is public, so this functionality was possible in previous versions, but is no longer w/ the most recent releases.

TristonianJones:
@jarreds Are you able to use the `NewObjectType` call to accomplish what you would like? It seems like this would do the right thing without digging into the internals of how.

The reason this mapping was public earlier was because it was used in a couple of different places and CEL was primarily built on top of protos; however, as of today, most of the uses of the core protobufs have been removed.

jarreds:
@TristonianJones thank you, perhaps you can guide me, because I have not quite figured out another way to get it to work. We have a proto `structpb`'ish type that supports a bunch of other types beyond the default JSON shaped primitives. I have this working by doing something like:

```golang
	types.CheckedWellKnowns["acme.type.v1.Value"] = types.DynType
	types.CheckedWellKnowns["acme.type.v1.ListValue"] = types.NewListType(celtypes.DynType)
	types.CheckedWellKnowns["acme.type.v1.Struct"] = types.NewMapType(celtypes.StringType, celtypes.DynType)
```

Perhaps you can guide me on how to use `NewObjectType` to accomplish this?

Thanks again for considering this.

TristonianJones:
Hi @jarreds could you create your own method for proxying the creation of `NewObjectType` with the sort of checks you'd like to see introduced prior to invoking the CEL version? In general, it's not a good idea to mutate the state of what should be a constant variable.

jarreds:
Totally agree on not mutating a global var that in an ideal world would be constant. Let me play around with the option you mentioned.

jarreds:
FWIW I still have not figured out a way to get this working w/ `NewObjectType`. Like I mentioned, the following gets me to the behavior I'm looking for:

```golang
	types.CheckedWellKnowns["acme.type.v1.Value"] = types.DynType
	types.CheckedWellKnowns["acme.type.v1.ListValue"] = types.NewListType(celtypes.DynType)
	types.CheckedWellKnowns["acme.type.v1.Struct"] = types.NewMapType(celtypes.StringType, celtypes.DynType)
```

The main problem I'm encountering w/ `NewObjectType` is that there is not an invocation that I've found that allows me to construct the correct shaped types. The issues I see:

1. `types.NewListType` does not take a type name and neither does `types.NewMapType`, so there doesn't seem to be a way I've found to register them with the runtime for a specific type name. This is why the `CheckedWellKnowns` hack allows us to short-circuit and return the raw types above for the type names.
2. There doesn't seem to be any way to register `types.DynType` for a specific type name either.

I imagine `registry.RegisterType` would do the right thing if I could get the type name set on those type constructors mentioned in the list, but I haven't had a chance to modify CEL-go to try that out yet.

TristonianJones:
@jarreds Is part of the problem that the protobuf types are being registered through `cel.Types` (or similar)? If so, you could provide your own custom type registry which intercepts the call.

PR #813, opened 2023-08-18
Introduce pre-order / post-order visitor pattern

TristonianJones:
Introduce a visitor with pre and post-order traversal methods.

This change also ensures that `MatchDescendents()` is bottom-up traversal
and that `Children()` can be used to model top-down traversal.

There are a couple of missing LICENSE notices added as well.

### DIFF
--- a/common/ast/navigable.go
+++ b/common/ast/navigable.go
@@ -188,6 +188,9 @@ const (
 	postOrder
 )
 
+// TODO: consider exposing a way to configure a limit for the max visit depth.
+// It's possible that we could want to configure this on the NewExprVisitor()
+// and through MatchDescendents() / MaxID().
 func visit(expr Expr, visitor Visitor, order visitOrder, depth, maxDepth int) {
 	if maxDepth > 0 && depth == maxDepth {
 		return
--- a/common/ast/navigable.go
+++ b/common/ast/navigable.go
@@ -1,3 +1,17 @@
+// Copyright 2023 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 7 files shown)


common/ast/navigable_test.go
```diff
@@ -120,6 +135,44 @@ func TestNavigateAST(t *testing.T) {
 	}
 }
 
+func TestNavigableExprChildren(t *testing.T) {
```
l46kok: Could we also test for post order traversal?

common/ast/navigable.go
```diff
@@ -87,34 +101,146 @@ func AllMatcher() ExprMatcher {
...
+
+type visitOrder int
+
+const (
+	preOrder = iota + 1
+	postOrder
+)
+
+func visit(expr Expr, visitor Visitor, order visitOrder, depth, maxDepth int) {
+	if maxDepth > 0 && depth == maxDepth {
```
l46kok, line 195: Thoughts on aborting on abnormally deep AST? [Something similar to this](https://github.com/google/cel-java/commit/26545aa8935427be712bafd3f0abb378f367f2f6#diff-ffb0c1526486c5e8de07dbaba6750bf21a7c2cf22b8bc8ebfba30f40005e88dcR95)

common/ast/navigable_test.go
TristonianJones: The post order ends up getting tested indirectly via `MaxID`, but I can add an explicit test to show it within this one.

l46kok: The method is called in MaxID but I didn't come across any tests which validate the actual visiting order. Regardless, I think an explicit test is valuable.

common/ast/navigable.go
TristonianJones: I think I might want to set this as an option on the visitor. Java handles exceptions well, but go doesn't like panics, and I'm not sure I want to put an error return status on the visit. Let me think about this more and come back to it.

common/ast/navigable_test.go
TristonianJones: Agreed. I chose to show pre and post explicitly with an example pre order using `Children` as well.

common/ast/navigable.go
TristonianJones: I've added a TODO. We'll finalize the behavior before the next release tag.

l46kok (approved):

PR #817, opened 2023-08-19
Ensure stable ordering of overload candidates

### LINKED ISSUE #816: Possible regression in type handling
In CEL 1.16,

```cel
dyn([]).sum() == 0
```

This was always true

but in CEL 1.17, this sometimes returns false.  It appears that this may be related to what type `dyn([])` is assigned.  If I run `dyn([]).sum()` it most often returns an int, but sometimes returns a uint or duration.

See https://github.com/kubernetes/kubernetes/blob/8a8d63a732dea705fd38fab7d79e3730642bb2f7/staging/src/k8s.io/apiserver/pkg/cel/library/lists.go#L139 for the `sum()` implementation.

TristonianJones:
Ensure stable ordering of overload candidates

There was a map traversal happening during dynamic dispatch
which would cause random ordering of overload candidates, and
in cases where the overload resolution was ambiguous, the result
might randomly fail.

Closes #816

### DIFF
--- a/cel/cel_test.go
+++ b/cel/cel_test.go
@@ -2081,7 +2081,8 @@ func TestDynamicDispatch(t *testing.T) {
 		),
 	)
 	out, err := interpret(t, env, `
-		[1, 2].first() == 1
+		dyn([]).first() == 0
+		&& [1, 2].first() == 1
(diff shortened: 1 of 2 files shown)


jpbetz (approved):
LGTM

PR #819, opened 2023-08-21
rename google -> authzed in go.mod

ecordell:


ecordell:
Opened by mistake, but just for reference: https://github.com/authzed/spicedb/pull/1501

TristonianJones:
Thanks for the added context. We've been trying to move away from using the protobuf in top-level APIs. Most changes have been backward compatible, but #798 will be a breaking change. I can probably add a backwards compatible shim, at least for two major K8s releases, before I tag v0.18

PR #827, opened 2023-08-25
Inlining optimizer

TristonianJones:
Inliner with support for identifiers and simple select expressions.

Note: optional field selections are not supported for inlining matches
at this time.

### DIFF
--- a/cel/inlining.go
+++ b/cel/inlining.go
@@ -122,7 +122,9 @@ func (opt *inliningOptimizer) Optimize(ctx *OptimizerContext, a *ast.AST) *ast.A
 	return a
 }
 
-// inlineExpr
+// inlineExpr replaces the current expression with the inlined one, unless the location of the inlining
+// happens within a presence test, e.g. has(a.b.c) -> inline alpha for a.b.c in which case an attempt is
+// made to determine whether the inlined value can be presence or existence tested.
 func (opt *inliningOptimizer) inlineExpr(ctx *OptimizerContext, prev, inlined ast.Expr, inlinedType *Type) {
 	switch prev.Kind() {
 	case ast.SelectKind:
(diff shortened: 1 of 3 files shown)


cel/inlining.go
```diff
@@ -0,0 +1,212 @@
...
+			ctx.NewCall(operators.NotEquals,
+				ctx.NewMemberCall(overloads.Size, inlined),
+				ctx.NewLiteral(types.IntZero),
+			))
+		return
+	}
+	ctx.ReportErrorAtID(prev.ID(), "unable to inline expression type %v into presence test", inlinedType)
+}
+
+func isBindable(matches []ast.NavigableExpr, inlined ast.Expr, inlinedType *Type) bool {
```
l46kok, line 182: Can you help me understand the rationale behind binding the specified types here? Specifically, it's a bit unclear why size and indexing need to be binded. Are these to accommodate inlining of presence tests?

cel/inlining_test.go
```diff
@@ -0,0 +1,577 @@
...
+					t:     cel.BoolType,
+					expr:  `'hello'.contains('lo')`,
+				},
+				{
+					name: "b",
+					t:    cel.BoolType,
+				},
+			},
+			inlined: `cel.bind(alpha, "hello".contains("lo"), alpha && b && alpha)`,
+			folded:  `cel.bind(alpha, true, alpha && b && alpha)`,
```
l46kok, line 100: Shouldn't this one fold to just `b`? I guess the constant folding might have to be augmented to better handle binds (ironically, I think just choosing to not bind for this example would rightly produce `b`).

l46kok: Just noticed that we aren't folding simple literals correctly -- sent https://github.com/google/cel-go/pull/833

TristonianJones: This one should fold to `b`, but the constant folding isn't quite smart enough to reason about the `bind` logic to determine that the expression could be represented as `alpha && b && alpha`. Specifically, because `b` appears and it's not known, the bind() doesn't fold.

cel/inlining.go
TristonianJones: Ah, it has to do with whether the expression is in a presence test. So if `has(a.b.c) ? a.b.c : blah` is the expression and `a.b.c` is being inlined then the bind would need to know whether inlined type can be represented as something like a presence test, either directly, or via size check.

l46kok: I see, I think size makes sense especially for presence testing strings. What about indexer?

TristonianJones: Maps would qualify as an indexer, but it's not a guarantee that the index expression will be a string, so I should probably skip indexers to start and come back to it.

l46kok (dismissed):

l46kok (approved):

PR #848, opened 2023-10-24
Add basic examples to the appengine repl

jnthntatum:
Add basic examples to the appengine repl

- Add examples covering basic CEL features and extensions
- Add evaluate button to repl console

### DIFF
--- a/repl/appengine/web/src/app/reference_panel/reference-panel-component.html
+++ b/repl/appengine/web/src/app/reference_panel/reference-panel-component.html
@@ -16,8 +16,46 @@
 
 <h1>References</h1>
 
-<p>See <a href="https://github.com/google/cel-spec">github.com/google/cel-spec</a> for CEL language
-    overview. </p>
-<p>See
-    <a href="https://github.com/google/cel-go/tree/master/repl/main">github.com/google/cel-go/tree/master/repl/main</a>
-    for REPL syntax guide.</p>
+<mat-accordion>
+    <mat-expansion-panel>
+        <mat-expansion-panel-header>
+        <mat-panel-title>
+            CEL Specification
+        </mat-panel-title>
+        </mat-expansion-panel-header>
+        <p>See <a href="https://github.com/google/cel-spec">github.com/google/cel-spec</a> for CEL language
+            overview.</p>
+    </mat-expansion-panel>
+    <mat-expansion-panel>
+        <mat-expansion-panel-header>
+            <mat-panel-title>
+            REPL Syntax
+            </mat-panel-title>
+        </mat-expansion-panel-header>
+        <p>See <a href="https://github.com/google/cel-go/tree/master/repl/main">github.com/google/cel-go/tree/master/repl/main</a>
+            for REPL syntax guide.</p>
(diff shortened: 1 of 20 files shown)


TristonianJones (dismissed):

repl/appengine/web/src/app/reference_panel/reference-panel-component.html
```diff
@@ -16,8 +16,46 @@
...
 <h1>References</h1>
 
-<p>See <a href="https://github.com/google/cel-spec">github.com/google/cel-spec</a> for CEL language
-    overview. </p>
-<p>See
-    <a href="https://github.com/google/cel-go/tree/master/repl/main">github.com/google/cel-go/tree/master/repl/main</a>
-    for REPL syntax guide.</p>
\ No newline at end of file
+<mat-accordion>
+    <mat-expansion-panel>
```
l46kok: Expansion panels unfortunately won't push the height of its parent container when expanded. So you end up seeing a cut-off content with a scrollbar when the container isn't tall enough to display the expanded content.

You could either change the primary container to something that supports fullscreen (mat-sidenav-container) or just set the viewport heights in your container to be full (not standard angular practice, but it works):

```
.content {
    display: block;
    margin-top: 5vh;
    width: 100%;
    height: 95vh;
}
```

repl/appengine/web/src/app/reference_panel/reference-panel-component.html
```diff
@@ -16,8 +16,46 @@
...
+        <mat-expansion-panel-header>
+            <mat-panel-title>
+            REPL Syntax
+            </mat-panel-title>
+        </mat-expansion-panel-header>
+        <p>See <a href="https://github.com/google/cel-go/tree/master/repl/main">github.com/google/cel-go/tree/master/repl/main</a>
+            for REPL syntax guide.</p>
+    </mat-expansion-panel>
+    <mat-expansion-panel>
```
l46kok: Consider expanding the panels by default
```suggestion
    <mat-expansion-panel [expanded]="true">
```

repl/appengine/web/src/app/reference_panel/reference-panel-component.html
```diff
@@ -16,8 +16,46 @@
 
 <h1>References</h1>
 
-<p>See <a href="https://github.com/google/cel-spec">github.com/google/cel-spec</a> for CEL language
-    overview. </p>
-<p>See
-    <a href="https://github.com/google/cel-go/tree/master/repl/main">github.com/google/cel-go/tree/master/repl/main</a>
-    for REPL syntax guide.</p>
\ No newline at end of file
+<mat-accordion>
```
l46kok: Consider allowing independent expansions:

```suggestion
<mat-accordion multi>
```

repl/appengine/web/src/app/reference_panel/reference-panel-component.html
jnthntatum: Done: I had to restructure the top level app a bit but should behave more like a full screen page.

l46kok (approved):

PR #850, opened 2023-10-26
Sets cost estimation and tracking options

TristonianJones:
Introduce library-based options for cost estimation and cost tracking

*Breaking change* - This change introduced a minor shift to ensure that all instances
of the `interpreter.CostTracker` were instantiated using `interpreter.NewCostTracker()`.
This change revealed a bug where presence test, `has()`, costs were not being counted
when calling `ContextEval()` or `Eval()` with state tracking enabled. This bug has been
fixed, but revealed an inconsistency in expectations between the two method calls.

Note, these options only match against type-checked expressions. This is the 
same behavior as is used internally within estimators; however, it is probably
worth indicating the cost estimates are only accurate against type-checked
expressions at a future date.

### DIFF
--- a/interpreter/runtimecost.go
+++ b/interpreter/runtimecost.go
@@ -15,7 +15,6 @@
 package interpreter
 
 import (
-	"fmt"
 	"math"
 
 	"github.com/google/cel-go/common"
@@ -134,7 +133,7 @@ func PresenceTestHasCost(hasCost bool) CostTrackerOption {
 func NewCostTracker(estimator ActualCostEstimator, opts ...CostTrackerOption) (*CostTracker, error) {
 	tracker := &CostTracker{
 		Estimator:           estimator,
-		functionTrackers:    map[string]FunctionTracker{},
+		overloadTrackers:    map[string]FunctionTracker{},
 		presenceTestHasCost: true,
 	}
 	for _, opt := range opts {
@@ -146,14 +145,13 @@ func NewCostTracker(estimator ActualCostEstimator, opts ...CostTrackerOption) (*
 	return tracker, nil
 }
 
-// FunctionCostTracker binds a function overload to a runtime FunctionTracker implementation.
+// OverloadCostTracker binds an overload ID to a runtime FunctionTracker implementation.
 //
-// FunctionCostTracker instances augment or override ActualCostEstimator decisions, allowing for  versioned and/or
+// OverloadCostTracker instances augment or override ActualCostEstimator decisions, allowing for  versioned and/or
 // optional cost tracking changes.
-func FunctionCostTracker(function, overloadID string, fnTracker FunctionTracker) CostTrackerOption {
+func OverloadCostTracker(overloadID string, fnTracker FunctionTracker) CostTrackerOption {
 	return func(tracker *CostTracker) error {
-		functionKey := fmt.Sprintf("%s|%s", function, overloadID)
-		tracker.functionTrackers[functionKey] = fnTracker
+		tracker.overloadTrackers[overloadID] = fnTracker
 		return nil
 	}
 }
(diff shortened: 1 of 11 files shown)


cici37:
The cost might not set properly for other ext library as well. It might be out of scope for this PR but do we want to track it somewhere in an issue? (I am happy to create one if you are ok with it :) )

TristonianJones:
@cici37 Yes, please, if you wouldn't mind filing an issue to introduce costs for the other extension libraries that would be great. Most of these costs are currently tracked within the K8s cost estimation extensions and should be moved back into the CEL libraries.

interpreter/runtimecost.go
```diff
@@ -144,9 +146,25 @@ func NewCostTracker(estimator ActualCostEstimator, opts ...CostTrackerOption) (*
...
 
+// FunctionCostTracker binds a function overload to a runtime FunctionTracker implementation.
+//
+// FunctionCostTracker instances augment or override ActualCostEstimator decisions, allowing for  versioned and/or
+// optional cost tracking changes.
+func FunctionCostTracker(function, overloadID string, fnTracker FunctionTracker) CostTrackerOption {
+	return func(tracker *CostTracker) error {
+		functionKey := fmt.Sprintf("%s|%s", function, overloadID)
+		tracker.functionTrackers[functionKey] = fnTracker
+		return nil
```
cici37, line 155: return nil directly here? Does that mean there will never be error here?

interpreter/runtimecost.go
```diff
@@ -144,9 +146,25 @@ func NewCostTracker(estimator ActualCostEstimator, opts ...CostTrackerOption) (*
...
 }
 
+// FunctionCostTracker binds a function overload to a runtime FunctionTracker implementation.
+//
+// FunctionCostTracker instances augment or override ActualCostEstimator decisions, allowing for  versioned and/or
+// optional cost tracking changes.
+func FunctionCostTracker(function, overloadID string, fnTracker FunctionTracker) CostTrackerOption {
+	return func(tracker *CostTracker) error {
+		functionKey := fmt.Sprintf("%s|%s", function, overloadID)
+		tracker.functionTrackers[functionKey] = fnTracker
```
cici37: What if no such `functionKey` in functionTrackers?

cici37 (commented):
Could we merge this into 0.17 release and cut another 0.17.7 patch release for this fix? I would like to avoid bumping to 0.18 only because of the bug fix if possible. :)

interpreter/runtimecost.go
TristonianJones: That's correct, not from this particular `CostTrackerOption`

interpreter/runtimecost.go
TristonianJones: @cici37 you mean later on in the code? The function trackers override or extend the base estimator / tracker implementations, so the default behavior for the base tracker would be used.

cici37: It will panic/npe and break all existing tests while testing in k8s side..
```
if _, ok := tracker.functionTrackers[functionKey]; ok {
			tracker.functionTrackers[functionKey] = fnTracker
		}
```
solved it..

TristonianJones: @cici37 I don't follow what issue you're seeing. The `functionTrackers` is just a map and there's an assignment to a map key. This should always work unless the `functionTrackers` is not initialized because the `CostTracker` is initialized directly rather than with `NewCostTracker()`

cici37: The functionTrackers is not initialized at this point. And ContextEval func is used from k8s. Thanks for addressing this!

TristonianJones:
@cici37 PTAL

cici37:
/lgtm
/approve
Thanks for helping with this!

jnthntatum (approved):

PR #853, opened 2023-10-30
Fix expression inlining when working with macros

TristonianJones:
Fix expression inlining when working with macros

Inlining was only correctly accounting for situations when the expression to be
inlined did not contain any macros. Once macros were introduced, the identifier
references as well as the macro 'pointer' concept were not correctly being
propagated during inlining. 

To preserve the 'pointer' concept, macro expression sub-graphs need to be
deleted when updating macro references. In the prior version of this code, 
the sub-graphs were being preserved in the macro call metadata which resulted
in issues with round-tripping the expressions to a string, even though the logic
within the core AST was correct.

There are a few places in the code where expression IDs are being massaged in
order to account for lining up the id space between expressions and then for
propagating id changes down through macros. If there are ways to improve /
consolidate this logic in the future, then this is likely an area worth investing in.

### DIFF
--- a/cel/optimizer.go
+++ b/cel/optimizer.go
@@ -15,11 +15,8 @@
 package cel
 
 import (
-	"fmt"
-
 	"github.com/google/cel-go/common"
 	"github.com/google/cel-go/common/ast"
-	"github.com/google/cel-go/common/types"
 	"github.com/google/cel-go/common/types/ref"
 )
 
@@ -76,22 +73,29 @@ func (opt *StaticOptimizer) Optimize(env *Env, a *Ast) (*Ast, *Issues) {
 		info := optimized.SourceInfo()
 		expr := optimized.Expr()
 		normalizeIDs(freshIDGen.renumberStable, expr, info)
-
 		// Sanitize the macro call references once the optimized expression has been computed
 		// and the ids normalized between the expression and the macros.
-		sanitized := baseFac.CopyExpr(expr)
-		sanitizedExprMap := make(map[int64]ast.Expr)
-		ast.PostOrderVisit(sanitized, ast.NewExprVisitor(func(e ast.Expr) {
-			if _, found := info.GetMacroCall(e.ID()); found {
-				e.SetKindCase(nil)
+		exprRefMap := make(map[int64]struct{})
+		ast.PostOrderVisit(expr, ast.NewExprVisitor(func(e ast.Expr) {
+			if e.ID() == 0 {
+				return
 			}
-			sanitizedExprMap[e.ID()] = baseFac.CopyExpr(e)
+			exprRefMap[e.ID()] = struct{}{}
 		}))
 		// Update the macro call id references to ensure that macro pointers are
 		// updated consistently across macros.
-		for id, call := range info.MacroCalls() {
-			resetMacroCall(call, sanitizedExprMap)
-			info.SetMacroCall(id, call)
+		for _, call := range info.MacroCalls() {
+			ast.PostOrderVisit(call, ast.NewExprVisitor(func(e ast.Expr) {
+				if e.ID() == 0 {
+					return
+				}
+				exprRefMap[e.ID()] = struct{}{}
+			}))
+		}
+		for id := range info.MacroCalls() {
+			if _, found := exprRefMap[id]; !found {
+				info.ClearMacroCall(id)
+			}
 		}
 
 		// Recheck the updated expression for any possible type-agreement or validation errors.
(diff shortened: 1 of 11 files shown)


cel/optimizer.go
```diff
@@ -388,3 +343,66 @@ func (opt *optimizerExprFactory) NewStruct(typeName string, fields []ast.EntryEx
...
+	// Update the expression
+	prev.SetKindCase(updated)
+	if len(opt.sourceInfo.MacroCalls()) == 0 {
+		return
+	}
+
+	// prev is not macro, updated is not macro
+	// prev is macro, updated is not macro
+	// prev is macro, updated is macro
+	// prev is not macro, updated is macro
```
l46kok: Do you need this comment around? or perhaps document the exhaustive case on the function comment

cel/optimizer.go
```diff
@@ -70,12 +69,39 @@ func (opt *StaticOptimizer) Optimize(env *Env, a *Ast) (*Ast, *Issues) {
...
 		// Normalize expression id metadata including coordination with macro call metadata.
-		normalizeIDs(env, optimized)
+		freshIDGen := newIDGenerator(0)
+		info := optimized.SourceInfo()
+		expr := optimized.Expr()
+		normalizeIDs(freshIDGen.renumberStable, expr, info)
+		// Sanitize the macro call references once the optimized expression has been computed
+		// and the ids normalized between the expression and the macros.
+		exprRefMap := make(map[int64]struct{})
+		ast.PostOrderVisit(expr, ast.NewExprVisitor(func(e ast.Expr) {
```
l46kok: You could skip visiting of expr nodes if macro calls do not exist.

cel/optimizer.go
```diff
@@ -70,12 +69,39 @@ func (opt *StaticOptimizer) Optimize(env *Env, a *Ast) (*Ast, *Issues) {
...
+			ast.PostOrderVisit(call, ast.NewExprVisitor(func(e ast.Expr) {
+				if e.ID() == 0 {
+					return
+				}
+				exprRefMap[e.ID()] = struct{}{}
+			}))
+		}
+		for id := range info.MacroCalls() {
+			if _, found := exprRefMap[id]; !found {
+				info.ClearMacroCall(id)
```
l46kok: This is likely worth adding some coverage on, to ensure we're not left with an unclean macro metadata

cel/optimizer.go
```diff
@@ -388,3 +343,66 @@ func (opt *optimizerExprFactory) NewStruct(typeName string, fields []ast.EntryEx
...
+	_, prevIsMacro := opt.sourceInfo.GetMacroCall(prev.ID())
+
+	// Determine whether the updated expression was a macro.
+	updatedMacro, updatedIsMacro := opt.sourceInfo.GetMacroCall(updated.ID())
+
+	if updatedIsMacro {
+		// If the updated call was a macro, then updated id maps to prev id,
+		// and the updated macro moves into the prev id slot.
+		opt.sourceInfo.ClearMacroCall(updated.ID())
+		opt.sourceInfo.SetMacroCall(prev.ID(), updatedMacro)
```
l46kok: What would happen if the previous ID was also a macro? Wouldn't we need to move the prev macro around as well?

cel/optimizer.go
```diff
@@ -388,3 +343,66 @@ func (opt *optimizerExprFactory) NewStruct(typeName string, fields []ast.EntryEx
...
+				call.SetKindCase(nil)
+			} else {
+				call.SetKindCase(opt.fac.CopyExpr(macroExpr))
+			}
+			call.RenumberIDs(func(id int64) int64 {
+				if id == updated.ID() {
+					return prev.ID()
+				}
+				return id
+			})
```
l46kok: I actually don't fully understand why this step is needed. My take on this is that we're accounting for orphaned macro expr nodes after update, then normalizing the IDs afterwards. Some comment may help.

cel/optimizer.go
TristonianJones: Done.

cel/optimizer.go
TristonianJones: The branches are currently covered, are you looking for tests to make sure it's removed?

cel/optimizer.go
TristonianJones: Done in the top-level comment.

cel/optimizer.go
TristonianJones: If the previous expression was a macro, then all the other macros are only referencing it as a pointer (expression with unset body), and so only the previous macro body in the macro map needs to be updated

cel/optimizer.go
TristonianJones: I've added more comments

cel/optimizer.go
l46kok: Yep confirming the removal was the ask

TristonianJones: Done.

TristonianJones:
PTAL @l46kok

cel/optimizer.go
```diff
@@ -388,3 +350,85 @@ func (opt *optimizerExprFactory) NewStruct(typeName string, fields []ast.EntryEx
...
+				call.SetKindCase(opt.fac.CopyExpr(macroExpr))
+			}
+			// Since SetKindCase does not renumber the id, ensure the references to
+			// the old 'updated' id are mapped to the target id.
+			call.RenumberIDs(func(id int64) int64 {
+				if id == updated.ID() {
+					return target.ID()
+				}
+				return id
+			})
```
l46kok, line 467: Could we add test coverage for this?

TristonianJones: This ends up being difficult to test since the `UpdateExpr` call is only exposed on the transient `OptimizerContext` object. I happen to know this works since unparsing would fail if it didn't.

TristonianJones: @l46kok I've updated the APIs to make these kinds of checks about the internal behavior a bit easier to observe, but it's probably not quite as granular as you're looking for.

jnthntatum (approved):
LGTM

PR #855, opened 2023-11-02
Add option to include test types in the repl.

jnthntatum:
Adds a repl option to include packages of linked in message types.

This allows using the types without preparing a hermetic set of descriptors to load from file.

example:
`%load_descriptors --pkg 'spec-tests'`


repl/commands.go
```diff
@@ -47,6 +47,12 @@ optional language features.
 %option --container 'google.protobuf'
 %option --extension 'all'`
 
+	loadDescriptorsUsage = `LoadDescriptors loads a protobuf descriptor file (google.protobuf.FileDescriptorSet)
+from disk or from a predefined package. Supported packages are "cel-spec-test-types"
+(TestAllTypes) and "google-rpc" (AttributeContext).
+%load_descriptors 'path/to/descriptor_set.binarypb'
+%load_descriptors --pkg 'cel-spec-test-types'`
```
l46kok, line 54: Could we update the REPL's readme to include this new flag?

l46kok (dismissed):

repl/commands.go
jnthntatum: Done.

l46kok (approved):

PR #865, opened 2023-11-29
Set membership test rewriting optimizer

TristonianJones:
Translates expressions like `a in [1, 2, 3]` to `a in {1: true, 2: true, 3: true}`

The benefit of this change is that most of the set relationship operations over lists
are O(n) where n represents the number of elements in the list. The `cel.OptOptimize`
flag will perform a similar optimization at runtime; however, this optimizer performs
the optimization at compile time.

### DIFF
--- a/ext/sets_test.go
+++ b/ext/sets_test.go
@@ -382,6 +382,17 @@ func TestSetsMembershipRewriter(t *testing.T) {
 			},
 			out: types.True,
 		},
+		{
+			expr:      `a in {3: true}`,
+			optimized: `a in {3: true}`,
+			vars: []cel.EnvOption{
+				cel.Variable("a", cel.IntType),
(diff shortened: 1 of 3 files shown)


ext/sets_test.go
```diff
@@ -328,6 +330,119 @@ func TestSets(t *testing.T) {
 	}
 }
 
+func TestSetsMembershipRewriter(t *testing.T) {
```
l46kok, line 334: Up to you: maybe add some tests around macros.

l46kok (dismissed):

ext/sets_test.go
TristonianJones: Done. Added for macros and enums.

l46kok (approved):

PR #880, opened 2024-01-07
Fix type substitution within parameterized type names

TristonianJones:
Fix parameterized type substitutions for results from `type()` calls

### DIFF
--- a/checker/checker_test.go
+++ b/checker/checker_test.go
@@ -2233,6 +2233,35 @@ _&&_(_==_(list~type(list(dyn))^list,
 			| NotAMessage{}
 			| ...........^`,
 		},
+		{
+			in: `{}.map(c,[c,type(c)])`,
+			out: `__comprehension__(
+				// Variable
+				c,
+				// Target
+				{}~map(dyn, dyn),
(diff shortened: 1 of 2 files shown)


checker/checker_test.go
```diff
@@ -2233,6 +2233,35 @@ _&&_(_==_(list~type(list(dyn))^list,
 			| NotAMessage{}
 			| ...........^`,
 		},
+		{
+			in: `{}.map(c,[c,type(c)])`,
+			out: `__comprehension__(
+				// Variable
+				c,
+				// Target
+				{}~map(dyn, dyn),
```
l46kok, line 2242: Should the type call be preserved around dyn after substitution? (e.g: `{}~map(type(dyn), dyn)`)? I don't think it makes any difference in the runtime since it should get unwrapped, but this is what Java's type-checker currently produces.

TristonianJones: If the Java checker produces `map(type(dyn), dyn)` then that would be an error in Java since `c` (the key), is of any possible type.

l46kok (approved):

PR #901, opened 2024-02-27
Track sizes of comprehension results

### LINKED ISSUE #900: Estimated size information lost when `.map()` result is chained
Estimated size information lost with `.map()` when `.filter()` is called on the result (presumably applies to other chaining operations).

**To Reproduce**

- [ ]  parser
- [x]  checker
- [ ]  interpreter

```cel
[1,2,3,4,5].map(x, x).filter(x, x % 2 == 0)
```

Test setup:

`checker/cost_test.go`:

```go
		{
			name:   ".map.filter size",
			expr:   `[1,2,3,4,5].map(x, x).filter(x, x % 2 == 0)`,
			wanted: CostEstimate{Min: 97, Max: 97},
		},
```

**Expected behavior**

Expected a bounded cost estimate. got: [97, 18446744073709551615]

**Additional context**

Both these test cases pass when added:

```go
		{
			name:   ".filter size",
			expr:   `[1,2,3,4,5].filter(x, x % 2 == 0)`,
			wanted: CostEstimate{Min: 41, Max: 101},
		},
		{
			name:   ".map size",
			expr:   `[1,2,3,4,5].map(x, x)`,
			wanted: CostEstimate{Min: 86, Max: 86},
		},
```

jpbetz:
Setting this as draft because I'm not convinced it's entirely correct.  It's assuming that the result size of a comprehension matches the iteration. Does this miscount break `.filter()`, `.exists()` ...?

Fixes https://github.com/google/cel-go/issues/900

### DIFF
--- a/checker/cost_test.go
+++ b/checker/cost_test.go
@@ -503,6 +503,71 @@ func TestCost(t *testing.T) {
 			},
 			wanted: CostEstimate{Min: 3, Max: 3},
 		},
+		{
+			name:   ".filter list literal",
+			expr:   `[1,2,3,4,5].filter(x, x % 2 == 0)`,
+			wanted: CostEstimate{Min: 41, Max: 101},
+		},
+		{
+			name:   ".map list literal",
+			expr:   `[1,2,3,4,5].map(x, x)`,
+			wanted: CostEstimate{Min: 86, Max: 86},
+		},
+		{
+			name:   ".map.filter list literal",
+			expr:   `[1,2,3,4,5].map(x, x).filter(x, x % 2 == 0)`,
+			wanted: CostEstimate{Min: 117, Max: 177},
+		},
+		{
+			name:   ".map.exists list literal",
+			expr:   `[1,2,3,4,5].map(x, x).exists(x, x == 5) == true`,
+			wanted: CostEstimate{Min: 108, Max: 118},
+		},
+		{
+			name:   ".map.map list literal",
+			expr:   `[1,2,3,4,5].map(x, x).map(x, x)`,
+			wanted: CostEstimate{Min: 162, Max: 162},
+		},
+		{
+			name:   ".map list literal selection",
+			expr:   `[1,2,3,4,5].map(x, x)[4]`,
+			wanted: CostEstimate{Min: 87, Max: 87},
+		},
+		{
+			name:   "nested array selection",
+			expr:   `[[1,2],[1,2],[1,2],[1,2],[1,2]][4]`,
+			wanted: CostEstimate{Min: 61, Max: 61},
+		},
+		{
+			name:   "nested array selection",
+			expr:   `{'a': [1,2], 'b': [1,2], 'c': [1,2], 'd': [1,2], 'e': [1,2]}.b`,
+			wanted: CostEstimate{Min: 81, Max: 81},
+		},
+		{
+			// Estimated cost does not track the sizes of nested aggregate types
+			// (lists, maps, ...) and so assumes a worst case cost when an
+			// expression applies a comprehension to a nested aggregated type,
+			// even if the size information is available.
+			// TODO: This should be fixed.
+			name:   "comprehension on nested list",
+			expr:   `[1,2,3,4,5].map(x, [x, x]).all(y, y.all(y, y == 1))`,
+			wanted: CostEstimate{Min: 157, Max: 18446744073709551615},
+		},
+		{
+			// Make sure we're accounting for not just the iteration range size,
+			// but also the overall comprehension size. The chained map calls
+			// will treat the result of one map as the iteration range of the other,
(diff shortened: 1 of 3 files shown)


checker/cost_test.go
```diff
@@ -503,6 +503,31 @@ func TestCost(t *testing.T) {
...
+			wanted: CostEstimate{Min: 117, Max: 177},
+		},
+		{
+			name:   ".map.exists list literal",
+			expr:   `[1,2,3,4,5].map(x, x).exists(x, x % 2 == 0) == true`,
+			wanted: CostEstimate{Min: 108, Max: 123},
+		},
+		{
+			name:   ".map.map list literal",
+			expr:   `[1,2,3,4,5].map(x, x).map(x, x)`,
```
stevekuznetsov, line 528: Do we need to check the return type of the expression in `map()`? What about something like: 

```
[1,2,3,4,5].map(x, [x, x])
```

jpbetz: We have a gap today where we don't track composed (nested) sized objects well.  I think we need to do better, but it would be a larger than than this fix.

jpbetz:
@TristonianJones would you be willing to review?

Any problem test cases you can think of would be really helpful. 
You'd mentioned a few additional places where we might need to track size, I had trouble isolating them but if you see any let me know.

checker/cost_test.go
stevekuznetsov: The runtime calculations would be a backstop if we end up undershooting the cost here, right?

jpbetz: Runtime cost tallies actual operations as they execute, so it's a really good way to make sure we don't overuse CPU even if expected cost checking lets an expression in that's too expensive.

TristonianJones: It would be great to add the same tests to the runtime cost tests as well

jpbetz: Good idea. Added.

TristonianJones: @jpbetz even if we don't have confidence in the correctness of `[1,2,3,4,5].map(x, [x, x])`, we should add a test case to document the current behavior and make sure we understand the impact of the fix.

checker/cost_test.go
```diff
@@ -503,6 +503,46 @@ func TestCost(t *testing.T) {
...
+			wanted: CostEstimate{Min: 117, Max: 177},
+		},
+		{
+			name:   ".map.exists list literal",
+			expr:   `[1,2,3,4,5].map(x, x).exists(x, x == 5) == true`,
+			wanted: CostEstimate{Min: 108, Max: 118},
+		},
+		{
+			name:   ".map.map list literal",
+			expr:   `[1,2,3,4,5].map(x, x).map(x, x)`,
```
TristonianJones, line 528: Would you mind adding new cases for the following:

```go
// Make sure we're accounting for not just the iteration range size,
// but also the overall comprehension size. The chained map calls
// will treat the result of one map as the iteration range of the other,
// so they're planned in reverse; however, the `+` should verify that
// the comprehension result has a size.
[1,2,3,4,5].map(x, x).map(x, x) + [1]
```

Also, the case that @stevekuznetsov mentioned around a nested comprehension was interesting as well:

```go
[1,2,3].all(i, i in [1,2,3].map(j, j + j))
```

TristonianJones (changes requested):

checker/cost_test.go
jpbetz: Great. Added!

checker/cost_test.go
jpbetz: Added. As expected, there is a variation of this that fails to compute a bounded estimated cost. I've included a test showing it along with an explanation.

checker/cost_test.go
```diff
@@ -503,6 +503,71 @@ func TestCost(t *testing.T) {
...
+			name:   "nested array selection",
+			expr:   `{'a': [1,2], 'b': [1,2], 'c': [1,2], 'd': [1,2], 'e': [1,2]}.b`,
+			wanted: CostEstimate{Min: 81, Max: 81},
+		},
+		{
+			// Estimated cost does not track the sizes of nested aggregate types
+			// (lists, maps, ...) and so assumes a worst case cost when an
+			// expression applies a comprehension to a nested aggregated type,
+			// even if the size information is available.
+			// TODO: This should be fixed.
```
TristonianJones, line 551: This might be difficult to track since you have to know something more about the iteration variable size to understand the implications from `createList` or `createMap`. At least the common cases appear to have been resolved.

TristonianJones (approved):

checker/cost_test.go
jpbetz: Yeah, even without tackling the nesting (and more subtle tracking) in this PR, I think we're making forward progress.

PR #923, opened 2024-04-16
Math extensions update to include bitwise ops and floating point helpers

TristonianJones:
Support for floating point helpers and bitwise operators.


ext/math_test.go
```diff
@@ -112,6 +112,74 @@ func TestMath(t *testing.T) {
 				"numbers": []float64{-21.0, -10.5, 1.0},
 			},
 		},
+
+		// Tests for math bitwise operators
+		// Signed bitwise ops
+		{expr: "math.bitAnd(1, 2) == 0"},
```
l46kok, line 118: Up to you but probably worth adding some tests around mixed int/uint literals, for no matching overload and ones wrapped in `dyn` to demonstrate that it fails during eval

ext/math.go
```diff
@@ -86,28 +87,311 @@ import (
...
 //	math.least(dyn('string')) // runtime error
+//
+// # Math.BitOr
+//
+// Introduced at version: 1
+//
+// Var-args macro which accepts a variable quantity of [1, 32) integers and
+// performs a bitwise-OR operation on all of them. The macro is expanded into
+// a call to the internal-only overload of math.@bitOr(number),
+// math.@bitOr(number, number), or math.@bitOr(list)
```
l46kok: Is vararg macro expansion and the list overload coming in follow-up PRs?

ext/README.md
```diff
@@ -117,6 +118,249 @@ Examples:
...
+    math.bitAnd(-3, -5)   // returns -7
+
+### Math.BitXor
+
+Introduced at version: 1
+
+    math.bitXor(<int>, <int>) -> <int>
+    math.bitXor(<uint>, <uint>) -> <uint>
+
+Performs a bitwise-XOR  operation over two int or uint values.
```
l46kok:
```suggestion
Performs a bitwise-XOR operation over two int or uint values.
```

ext/math.go
```diff
@@ -86,28 +87,311 @@ import (
...
+//
+//	math.abs(<double>) -> <double>
+//	math.abs(<int>) -> <int>
+//	math.abs(<uint>) -> <uint>
+//
+// Examples:
+//
+//	math.abs(-1)  // returns 1
+//	math.abs(1)   // returns 1
+//	math.abs(-9223372036854775808) // overlflow error
```
l46kok:
```suggestion
//	math.abs(-9223372036854775808) // overflow error
```

ext/math_test.go
```diff
@@ -319,11 +387,32 @@ func TestMathRuntimeErrors(t *testing.T) {
...
+		{
+			expr: "math.abs(-9223372036854775808)",
+			err:  "overflow",
+		},
 	}
 
 	env := testMathEnv(t,
 		cel.Variable("a", cel.DynType),
 		cel.Variable("b", cel.IntType),
+		cel.Variable("ints", cel.ListType(cel.IntType)),
```
l46kok: Unused, but maybe you meant to include overloads for lists

ext/math.go
```diff
@@ -86,28 +87,311 @@ import (
...
+//
+// Perform a left shift of bits on the first parameter, by the amount of bits
+// specified in the second parameter. The first parameter is either a uint or
+// an int. The second parameter must be an int.
+//
+// When the second parameter is 64 or greater, 0 will be always be returned
+// since the number of bits shifted is greater than or equal to the total bit
+// length of the number being shifted.
+//
+//	math.bitShiftLeft(<int>, <int>) -> <int>
```
l46kok, line 163: Should we just define the second parameter to be uint? In most languages, negative shift is either undefined or an error.

ext/README.md
```diff
@@ -117,6 +118,249 @@ Examples:
...
+
+    math.abs(<double>) -> <double>
+    math.abs(<int>) -> <int>
+    math.abs(<uint>) -> <uint>
+
+Examples:
+
+    math.abs(-1)  // returns 1
+    math.abs(1)   // returns 1
+    math.abs(-9223372036854775808) // overlflow error
```
l46kok, line 298:
```suggestion
    math.abs(-9223372036854775808) // overflow error
```

ext/math.go
TristonianJones: Within ZetaSQL, the shift argument is an int. I kept it for parity there.

ext/math_test.go
TristonianJones: Thanks, it was from a prior test involving the vararg macros

ext/math.go
TristonianJones: I had a var-arg macro and then removed it. The comment was extant.

TristonianJones:
PTAL

ext/math.go
l46kok: We should document this behavior, this will have to be special cased in both Java/C++ to produce an error

l46kok (dismissed):

ext/math.go
TristonianJones: Oh, I see you mean documenting the error case for a negative shift?

l46kok: Yep, negative shifts isn't clearly established across most languages (It either results in an error, always returns 0, shift direction is flipped, with ambiguity introduced between logical/arithmetic shifts etc.).

TristonianJones: Done. I'll add conformance tests as well.

l46kok (approved):

PR #924, opened 2024-04-17
Simple Policy Compiler

TristonianJones:
Introduce a module for compiling a graph of expressions into a single expression.

The input format is YAML and allows for customization of the tag handling such that
CEL-based policy formats like those found in Kubernetes Admission Control could be
supported using a common foundation.

Currently, the compilation toolchain only supports first-match semantics for expression
evaluation; however, alternative evaluation semantics such as `or`, `and`, `last-match`,
and `aggregate` evaluation semantics may be supported in the future.


policy/parser.go
```diff
@@ -0,0 +1,615 @@
...
+
+func (p *parserImpl) parseYaml(src *Source) *Policy {
+	// Parse yaml representation from the source to an object model.
+	var docNode yaml.Node
+	err := sourceToYaml(src, &docNode)
+	if err != nil {
+		p.iss.ReportErrorAtID(0, err.Error())
+		return nil
+	}
+	return p.parsePolicy(p, docNode.Content[0])
```
l46kok, line 314: Consider producing an error to prohibit multiple YAML documents in a single file (via `---`)

policy/compiler.go
```diff
@@ -0,0 +1,199 @@
...
+	iss := cel.NewIssuesWithSourceInfo(errs, c.info)
+	rule, ruleIss := c.compileRule(p.Rule(), c.env, iss)
+	iss = iss.Append(ruleIss)
+	if iss.Err() != nil {
+		return nil, iss
+	}
+	ruleRoot, _ := env.Compile("true")
+	opt := cel.NewStaticOptimizer(&ruleComposer{rule: rule})
+	ruleExprAST, iss := opt.Optimize(env, ruleRoot)
+	return ruleExprAST, iss.Append(iss)
```
l46kok: append doesn't appear to be needed here

policy/compiler.go
```diff
@@ -0,0 +1,199 @@
...
+}
+
+func (c *compiler) compileRule(r *Rule, ruleEnv *cel.Env, iss *cel.Issues) (*compiledRule, *cel.Issues) {
+	var err error
+	compiledVars := make([]*compiledVariable, len(r.Variables()))
+	for i, v := range r.Variables() {
+		exprSrc := c.relSource(v.Expression())
+		varAST, exprIss := ruleEnv.CompileSource(exprSrc)
+		if exprIss.Err() == nil {
+			ruleEnv, err = ruleEnv.Extend(cel.Variable(fmt.Sprintf("variables.%s", v.Name().Value), varAST.OutputType()))
```
l46kok: Not needed now but for the future -- I imagine it might be helpful exposing an option to control the prefix for variables being exposed depending on context. For time being, you could consider pulling this out into a constant.

policy/parser.go
```diff
@@ -0,0 +1,615 @@
...
+		return v
+	}
+	for i := 0; i < len(node.Content); i += 2 {
+		key := node.Content[i]
+		keyID := ctx.CollectMetadata(key)
+		fieldName := key.Value
+		val := node.Content[i+1]
+		if val.Style == yaml.FoldedStyle || val.Style == yaml.LiteralStyle {
+			val.Line++
+			val.Column = key.Column + 2
```
l46kok: YAML doesn't strictly impose 2 spaces for indentations. Consider adding by 1 to allow any number of spaces to be matched. Otherwise, the compilation error will look odd if someone accidentally uses an indentation level of 1.

policy/compiler.go
```diff
@@ -0,0 +1,199 @@
...
+		}
+		iss = iss.Append(exprIss)
+	}
+	compiledMatches := []*compiledMatch{}
+	for _, m := range r.Matches() {
+		condSrc := c.relSource(m.Condition())
+		condAST, condIss := ruleEnv.CompileSource(condSrc)
+		iss = iss.Append(condIss)
+		if m.HasOutput() && m.HasRule() {
+			iss.ReportErrorAtID(m.Condition().ID, "either output or rule may be set but not both")
```
l46kok, line 98: Can we add a test for this?

policy/compiler.go
```diff
@@ -0,0 +1,199 @@
...
+
+type ruleComposer struct {
+	rule *compiledRule
+}
+
+// Optimize implements an AST optimizer for CEL which composes an expression graph into a single
+// expression value.
+func (opt *ruleComposer) Optimize(ctx *cel.OptimizerContext, a *ast.AST) *ast.AST {
+	ruleExpr, _ := optimizeRule(ctx, opt.rule)
+	ctx.UpdateExpr(a.Expr(), ruleExpr)
```
l46kok: Does this handle setting multiple macro calls (e.g: multiple binds)? I maybe wrong, but it seemed like this only handles updates of a single macro expression. Or does this just have to be moved to within the loop of where new bind macros are created? Either way, you may want to add unparser tests for compiled policies.

policy/parser.go
```diff
@@ -0,0 +1,615 @@
...
+
+// SetExpression sets the variable expression.
+func (v *Variable) SetExpression(e ValueString) {
+	v.expression = e
+}
+
+// Match declares a condition (defaults to true) as well as an output or a rule.
+// Either the output or the rule field may be set, but not both.
+type Match struct {
+	condition ValueString
```
l46kok, line 157: Any reason for condition being a non-pointer?

policy/parser.go
```diff
@@ -0,0 +1,615 @@
...
+func (defaultTagVisitor) RuleTag(ctx ParserContext, id int64, fieldName string, node *yaml.Node, r *Rule) {
+	ctx.ReportErrorAtID(id, "unsupported rule tag: %s", fieldName)
+}
+
+func (defaultTagVisitor) MatchTag(ctx ParserContext, id int64, fieldName string, node *yaml.Node, m *Match) {
+	ctx.ReportErrorAtID(id, "unsupported match tag: %s", fieldName)
+}
+
+func (defaultTagVisitor) VariableTag(ctx ParserContext, id int64, fieldName string, node *yaml.Node, v *Variable) {
+	ctx.ReportErrorAtID(id, "unsupported match tag: %s", fieldName)
```
l46kok:
```suggestion
	ctx.ReportErrorAtID(id, "unsupported variable tag: %s", fieldName)
```

policy/parser.go
```diff
@@ -0,0 +1,615 @@
...
+	return id
+}
+
+func (p *parserImpl) parsePolicy(ctx ParserContext, node *yaml.Node) *Policy {
+	ctx.CollectMetadata(node)
+	policy, id := ctx.NewPolicy(node)
+	if p.assertYamlType(id, node, yamlMap) == nil {
+		return policy
+	}
+	for i := 0; i < len(node.Content); i += 2 {
```
l46kok, line 433: I'm reasonably certain that the parser would enforce the yaml tags to be a key-value pair, but it would help to either add a check for this or at least a comment on why `+=2` is safe.

policy/compiler.go
TristonianJones: That's fair. I'll add a TODO and a constant

policy/compiler.go
TristonianJones: I think there is one in the last parser error test. Or were you thinking of a different test?

policy/parser.go
TristonianJones: Mostly just that it's required. The others are optional.

policy/compiler.go
TristonianJones: Oh, I'll add a comment. The optimize step here is weird because it starts off with `true` and replaces the expression completely with the policy expression fragments

policy/compiler.go
TristonianJones: Oh, I see. I handled this in the parser. I'm not sure best how to handle this in the compiler without bypassing the input parse step and going to the internal rep. I'll add a TODO

policy/compiler.go
TristonianJones: This was a bug. Fixed.

policy/parser.go
TristonianJones: I tried adding a test for this, but wasn't able to get the parsing to fail like I'd hoped. I'll have to revisit the test case, but the code check is there.

policy/compiler.go
l46kok: I'm primarily looking to see if the optimized policies can actually be unparsed. Relooking at the UpdateExpr code, I feel like it actually would not populate the binds correctly for the macro source, but maybe I'm wrong.

TristonianJones: I've added round-tripping on the generated expressions. There was a small bug where the last macro was not being added into the list due to how I was updating the root expression.

l46kok (approved):

PR #932, opened 2024-04-24
Improve token representation in AST

kortschak:
This change includes vendor directory addition of go-cmp. Unfortunately the go-cmp checks require an end-run around test package hermeticity, but I have made this as minimal as possible.

Please take a look.

# Pull Requests Guidelines

development branch.

## Commit Messages

```

```

## Reviews

* Perform a self-review.

## Merging

  if applicable.

### DIFF
--- a/common/ast/ast_test.go
+++ b/common/ast/ast_test.go
@@ -72,7 +72,11 @@ func TestASTCopy(t *testing.T) {
 		if err != nil {
 			t.Errorf("ast.ToAST() failed: %v", err)
 		}
-		if !reflect.DeepEqual(checked, checkedRoundtrip) {
+		same := reflect.DeepEqual(checked.Expr(), checkedRoundtrip.Expr()) &&
+			reflect.DeepEqual(checked.ReferenceMap(), checkedRoundtrip.ReferenceMap()) &&
+			reflect.DeepEqual(checked.TypeMap(), checkedRoundtrip.TypeMap()) &&
+			reflect.DeepEqual(checked.SourceInfo().MacroCalls(), checkedRoundtrip.SourceInfo().MacroCalls())
+		if !same {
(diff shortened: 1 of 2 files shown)


kortschak:
Fixed vendoring issue.

common/ast/ast_test.go
```diff
@@ -72,8 +73,30 @@ func TestASTCopy(t *testing.T) {
 		if err != nil {
 			t.Errorf("ast.ToAST() failed: %v", err)
 		}
-		if !reflect.DeepEqual(checked, checkedRoundtrip) {
-			t.Errorf("Roundtrip got %v, wanted %v", checkedRoundtrip, checked)
+		opts := []cmp.Option{
```
TristonianJones: One alternative would be to compare everything but position information with `reflect.DeepEqual`. The comparison set is pretty small (smaller than the proposed change):

```
checked.Expr(),
checked.ReferenceMap()
checked.TypeMap()
checked.SourceInfo().MacroCalls()
```

It's probably worth just stating that we don't care about the source info on round-tripping since it's not crucial to any part of the expression evaluation or tooling.

kortschak: Which would you prefer? The change is smaller, but the failure output is largely meaningless in the current code (a set of pointers) while in the change the actually differences are shown. If that's not important for you, then the smaller change would win, otherwise the options are needed for the diff rendering.

TristonianJones: I think it would be preferable to reduce the dependency set. I know the errors will not be as useful when there are issues; however, I also don't expect to be modifying these files much except for bug fixes, so any regression should standout right away.

TristonianJones: Thank you, Dan!

TristonianJones (approved):

PR #938, opened 2024-04-27
Introduce a flag for supporting 'null' as absent on optional field selection

### LINKED ISSUE #937: field selection with ? on null now produces an error
**Describe the bug**
I think this change happened in https://github.com/google/cel-go/pull/922. 

For example, given yaml like:

```
foo:
  bar: null
```

and CEL like:

```
foo.?bar.?baz.orValue("not there")
```

behavior changes from returning "not there" to throwing a no such key error.

**To Reproduce**

- [ ]  parser
- [ ]  checker
- [x]  interpreter

**Expected behavior**
Expected "not there" to be returned. 

**Additional context**

I tried reverting the change from https://github.com/google/cel-go/pull/922 and it makes things work the way I expect.

TristonianJones:
Due to a bug in the original implementation of optional values, a field selection
on a `null` value would produce `optional.none()` rather than an error. This
bug was fixed in #922; however, the bug has been relied upon as a feature
within upstream software.

To preserve the legacy behavior, configure the following option:

```
cel.NewEnv(cel.OptionalTypes(), cel.OptionalFieldSelectionNoneIfNull(true)))
```

Closes #937

### DIFF
--- a/cel/cel_test.go
+++ b/cel/cel_test.go
@@ -2659,6 +2659,71 @@ func TestOptionalValuesEval(t *testing.T) {
 	}
 }
 
+func TestOptionalValuesEvalNoneIfNull(t *testing.T) {
+	env := testEnv(t,
+		OptionalTypes(),
+		OptionalFieldSelectionNoneIfNull(true),
+	)
+	adapter := env.TypeAdapter()
+	tests := []struct {
+		expr string
+		in   map[string]any
+		out  any
+	}{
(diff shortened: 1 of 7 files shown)


mtaufen (commented):
LGTM

jcking:
LGTM assuming we also add a conformance test to cel-spec

jcking (approved):

PR #939, opened 2024-04-30
Update the presence testing fix to be opt-out

TristonianJones:
In conversation with @mtaufen regarding issue #937, it seems the safer fix
is to revert to the old behavior unless the error behavior has been opted-out
of by the CEL integrator.

This is in line with CEL team's policy to ensure that all non-error -> error fixes
be flag guarded.

### DIFF
--- a/cel/options.go
+++ b/cel/options.go
@@ -63,7 +63,7 @@ const (
 	featureVariadicLogicalASTs
 
 	// Enable optional field selection to treat null values as optional.none()
-	featureOptionalFieldSelectionNoneIfNull
+	featureEnableErrorOnBadPresenceTest
 )
 
 // EnvOption is a functional interface for configuring the environment.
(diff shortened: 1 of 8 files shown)


cel/options.go
```diff
@@ -63,7 +63,7 @@ const (
 	featureVariadicLogicalASTs
 
 	// Enable optional field selection to treat null values as optional.none()
```
l46kok: Comment should be updated

interpreter/attributes.go
```diff
@@ -129,11 +129,11 @@ type NamespacedAttribute interface {
 // AttrFactoryOption specifies a functional option for configuring an attribute factory.
 type AttrFactoryOption func(*attrFactory) *attrFactory
 
-// OptionalFieldSelectionNoneIfNull indicates that when a null value is encountered during
+// EnableErrorOnBadPresenceTest indicates that when a null value is encountered during
```
l46kok: Same here

l46kok (approved):

PR #941, opened 2024-05-13
Add support for struct tag name overrides in native types

patrickpichler:
In order to customize how to fields of native types can be accessed via CEL, struct tag support has been added. It is now possible to provide a struct field tag, that will act as the name for the corresponding field.

Here an example

```
type Person struct {
  Name string
  Age  int `cel:"age"`
}
```

and here is how to access the `Age` field from CEL:
```
person.age
```
----
# Pull Requests Guidelines

development branch.

## Commit Messages

```

```

## Reviews

* Perform a self-review.

## Merging

  if applicable.

### DIFF
--- a/ext/native.go
+++ b/ext/native.go
@@ -78,12 +78,28 @@ var (
 // same advice holds if you are using custom type adapters and type providers. The native type
 // provider composes over whichever type adapter and provider is configured in the cel.Env at
 // the time that it is invoked.
-func NativeTypes(refTypes ...any) cel.EnvOption {
+func NativeTypes(args ...any) cel.EnvOption {
 	return func(env *cel.Env) (*cel.Env, error) {
-		tp, err := newNativeTypeProvider(env.CELTypeAdapter(), env.CELTypeProvider(), refTypes...)
+		nativeTypes := make([]any, 0, len(args))
+		tpOptions := nativeTypeOptions{}
+
+		for _, v := range args {
+			switch v := v.(type) {
+			case NativeTypesOption:
+				err := v(&tpOptions)
+				if err != nil {
+					return nil, err
+				}
+			default:
+				nativeTypes = append(nativeTypes, v)
+			}
+		}
+
+		tp, err := newNativeTypeProvider(tpOptions, env.CELTypeAdapter(), env.CELTypeProvider(), nativeTypes...)
 		if err != nil {
 			return nil, err
 		}
+
 		env, err = cel.CustomTypeAdapter(tp)(env)
 		if err != nil {
 			return nil, err
@@ -92,12 +108,27 @@ func NativeTypes(refTypes ...any) cel.EnvOption {
 	}
 }
 
-func newNativeTypeProvider(adapter types.Adapter, provider types.Provider, refTypes ...any) (*nativeTypeProvider, error) {
+type NativeTypesOption func(*nativeTypeOptions) error
+
+type nativeTypeOptions struct {
+	// parseStructTags controls if CEL should support struct field renames, by parsing
+	// struct field tags.
+	parseStructTags bool
+}
+
+func ParseStructTags(enabled bool) NativeTypesOption {
+	return func(ntp *nativeTypeOptions) error {
+		ntp.parseStructTags = true
+		return nil
+	}
+}
+
+func newNativeTypeProvider(tpOptions nativeTypeOptions, adapter types.Adapter, provider types.Provider, refTypes ...any) (*nativeTypeProvider, error) {
 	nativeTypes := make(map[string]*nativeType, len(refTypes))
 	for _, refType := range refTypes {
 		switch rt := refType.(type) {
 		case reflect.Type:
-			result, err := newNativeTypes(rt)
+			result, err := newNativeTypes(tpOptions.parseStructTags, rt)
 			if err != nil {
 				return nil, err
 			}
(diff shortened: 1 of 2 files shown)


TristonianJones:
Thanks @patrickpichler! This seems like a very nice win.

ext/native_test.go
```diff
@@ -678,9 +698,15 @@ func mustParseTime(t *testing.T, timestamp string) time.Time {
 	return out
 }
 
+type TestStructWithMultipleSameNames struct {
+	Name        string
+	custom_name string `cel:"Name"`
```
TristonianJones, line 826: This is an interesting collision case. Can you document this behavior on the native.go library comments?

ext/native.go
```diff
@@ -155,6 +156,14 @@ func (tp *nativeTypeProvider) FindStructType(typeName string) (*types.Type, bool
 	return tp.baseProvider.FindStructType(typeName)
 }
 
+func toFieldName(f reflect.StructField) string {
+	if name, found := f.Tag.Lookup("cel"); found {
```
TristonianJones, line 216: Please document the tag being searched with an example in the top-level comments. Also make sure these comments are in sync with the README.md

ext/native.go
```diff
@@ -155,6 +156,14 @@ func (tp *nativeTypeProvider) FindStructType(typeName string) (*types.Type, bool
 	return tp.baseProvider.FindStructType(typeName)
 }
```
TristonianJones, line 210: The signature for `NativeTypes` takes an optional set of `any` values which are expected to be types; however, given this change enables tag handling in a very specific way and in a manner which could impact performance, could you consider introducing a functional option and filtering out the option values from the type values: 

```go
type NativeTypesOption func(*nativeTypeProvider) (*nativeTypeProvider, error)
```

You'd probably have to filter between options and types like so:

```go
nativeTypes := make([]any, 0, len(refTypes))
opts := []NativeTypesOption{}
for _, v := range refTypes {
  switch v := v.(type) {
    case NativeTypeOption:
      opts = append(opts, v)
    default:
      nativeTypes = append(nativeTypes, v)
  }
}
```

I would like to make sure that we're opting into the behavior change, and doing so in a way which won't impact performance.

ext/native.go
patrickpichler: Do you have a specific place in mind in the README? It doesn't appear as if native is mentioned at all 😅

ext/native.go
patrickpichler: The struct tag parsing is now opt-in. Each `nativeType` struct now needs to store an additional flag to know how to retrieve the name for a field.

quantumsheep:
Wow I just discovered this PR after creating #942! My PR fixes an issue not fixed in yours but completely related to the same use-case

ext/native.go
```diff
@@ -77,12 +78,45 @@ var (
...
-func NativeTypes(refTypes ...any) cel.EnvOption {
+//
+// There is also the possibility to rename the fields of native structs by setting the `cel` tag
+// for fields you want to override. In order to enable this feature, pass in the `EnableStructTag`
+// option. Here is an example to see it in action:
+//
+// ```go
+// package identity
+//
+//		type Account struct {
```
TristonianJones: nit: would you mind updating the formatting here? I think some tabs made it into the comments

ext/native.go
```diff
@@ -91,20 +125,35 @@ func NativeTypes(refTypes ...any) cel.EnvOption {
...
-func newNativeTypeProvider(adapter types.Adapter, provider types.Provider, refTypes ...any) (*nativeTypeProvider, error) {
+type NativeTypesOption func(*nativeTypeOptions) error
+
+type nativeTypeOptions struct {
+	// parseStructTags controls if CEL should support struct field renames, by parsing
+	// struct field tags.
+	parseStructTags bool
+}
+
+func ParseStructTags(enabled bool) NativeTypesOption {
```
TristonianJones, line 138: nit: please add comments to the new functional option.

ext/native.go
```diff
@@ -91,20 +125,35 @@ func NativeTypes(refTypes ...any) cel.EnvOption {
 	}
 }
 
-func newNativeTypeProvider(adapter types.Adapter, provider types.Provider, refTypes ...any) (*nativeTypeProvider, error) {
+type NativeTypesOption func(*nativeTypeOptions) error
```
TristonianJones, line 129: nit: please add comments to the new type declaration

ext/native.go
TristonianJones: Ah, then ignore the README.md comment. In the local file is fine. Thanks!

ext/native.go
```diff
@@ -505,23 +585,41 @@ func newNativeTypes(rawType reflect.Type) ([]*nativeType, error) {
...
+func newNativeType(parseStructTags bool, rawType reflect.Type) (*nativeType, error) {
 	refType := rawType
 	if refType.Kind() == reflect.Pointer {
 		refType = refType.Elem()
 	}
 	if !isValidObjectType(refType) {
 		return nil, fmt.Errorf("unsupported reflect.Type %v, must be reflect.Struct", rawType)
 	}
+	fieldNames := make(map[string]struct{})
+
```
TristonianJones, line 602: nit: you might want to only process this for-loop if `parseStructTags == true` since that would be the only situation which could cause field name collisions.

ext/native.go
```diff
@@ -569,9 +667,22 @@ func (t *nativeType) Value() any {
 	return t.typeName
 }
 
+// fieldByName returns the corresponding reflect.StructField for the give name either by matching
+// field tag or field name.
+func (t *nativeType) fieldByName(fieldName string) (reflect.StructField, bool) {
+	for i := 0; i < t.refType.NumField(); i++ {
```
TristonianJones, line 684: If we already have the `t.parseStructTags` info, then it's worth early returning with `t.refType.FieldByName(fieldName)`

patrickpichler:
@TristonianJones I've updated the code, would be nice if you could have another look  :)

TristonianJones (approved):

TristonianJones:
@patrickpichler thanks for the contribution!

PR #942, opened 2024-05-15
Allow custom field name resolving for native types

quantumsheep:
My use-case:
```go
type CustomNativeType struct {
	reflect.Type
}

func (t *CustomNativeType) FieldByName(name string) (reflect.StructField, bool) {
	name = strcase.ToPascal(name)
	return t.Type.FieldByName(name)
}

type Foo struct {
	MyField string
}

func main() {
	fooType := &CustomNativeType{Type: reflect.TypeOf(Foo{})}

	env, err := cel.NewEnv(
		ext.NativeTypes(fooType),
		cel.Variable("foo", cel.ObjectType("package.Foo")),
	)
	if err != nil {
		panic(err)
	}

	// Evaluate "foo.my_field"
}
```

### DIFF
--- a/ext/native.go
+++ b/ext/native.go
@@ -188,16 +188,17 @@ func (tp *nativeTypeProvider) FindStructFieldType(typeName, fieldName string) (*
 	if !ok {
 		return nil, false
 	}
+	refFieldName := refField.Name
 	return &types.FieldType{
 		Type: celType,
 		IsSet: func(obj any) bool {
 			refVal := reflect.Indirect(reflect.ValueOf(obj))
-			refField := refVal.FieldByName(fieldName)
+			refField := refVal.FieldByName(refFieldName)
 			return !refField.IsZero()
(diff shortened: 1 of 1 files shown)


ext/native.go
```diff
@@ -188,16 +188,17 @@ func (tp *nativeTypeProvider) FindStructFieldType(typeName, fieldName string) (*
 	if !ok {
 		return nil, false
 	}
+	refFieldName := refField.Name
```
TristonianJones, line 191: I'm not sure I follow how the input field name differs from the `refField.Name` value in this case. Could you add a test to illustrate the change in behavior?

TristonianJones:
@quantumsheep would you mind helping me understand how this change accomplishes your goal. I think the idea is to proxy to another object which implements an interface similar to what golang reflection does, but I'm not sure I understand how this PR accomplishes that.

PR #943, opened 2024-05-15
Support context propagation without breaking changes

goccy:
This PR is related to https://github.com/google/cel-go/pull/925 .

First of all, thank you for the review the other day.
As suggested, I tried to avoid changing the public interface and instead added new types to support it. Although there are many changes, as you can see from the test code, I added the functionality without altering the existing behavior 🎉 

Additionally, during the changes, it was unnecessary to pass context.Context to `AddQualifierContext` / `QualifyContext` / `QualifyIfPresentContext`, so if there are any issues, it's possible to revert these to use the original types. Please let me know your thoughts.

# Pull Requests Guidelines

development branch.

## Commit Messages

```

```

## Reviews

* Perform a self-review.

## Merging

  if applicable.


TristonianJones:
@goccy thanks for taking another look at this. I'm sorry I haven't been able to review the change yet. It's a large change and I haven't had a sufficiently large block of time to review it in the manner it deserves to be. I will try to get to the review in more detail next week.

TristonianJones:
@goccy I think I'm going to revive #368 as it will have fewer impacts and potential for breakages while still giving you most of what you want. The difference is that the implementation won't block on concurrent calls unless it needs to, instead iterating through calls in batches and re-evaluating. It's a definite overhead, but the highest cost should be the I/O rather than the repeated evaluation.

goccy:
@TristonianJones  Thank you for your reply. But Unfortunately, we are looking for the capability to propagate context.Context synchronously, not through asynchronous functions. Therefore, we believe ContextEval is the most suitable API for this purpose.

TristonianJones:
Hi @goccy, I'll use `ContextEval` as the entry point, but will make resolution of context related functions done via the iterative eval approach I mentioned earlier since it will work for both synchronous and asynchronous functions.

goccy:
@TristonianJones I understand ! It seems to be what I expected if it's just a difference in implementation. Thank you!

TristonianJones:
Fixed by https://github.com/cel-expr/cel-go/pull/1369

PR #945, opened 2024-05-22
Custom Tag Handler support for CEL Policy Compilation

TristonianJones:
The example content demonstrates how to map K8s Admission Policy
into the CEL Policy compilation.

### DIFF
--- a/common/ast/ast.go
+++ b/common/ast/ast.go
@@ -343,12 +343,11 @@ func (s *SourceInfo) GetLocationByOffset(offset int32) common.Location {
 	line := 1
 	col := int(offset)
 	for _, lineOffset := range s.LineOffsets() {
-		if lineOffset <= offset {
-			line++
-			col = int(offset - lineOffset)
-		} else {
+		if lineOffset > offset {
 			break
 		}
+		line++
+		col = int(offset - lineOffset)
 	}
 	return common.NewLocation(line, col)
 }
--- a/common/ast/ast.go
+++ b/common/ast/ast.go
@@ -321,17 +321,7 @@ func (s *SourceInfo) ClearOffsetRange(id int64) {
 // of the expression node at the id.
 func (s *SourceInfo) GetStartLocation(id int64) common.Location {
 	if o, found := s.GetOffsetRange(id); found {
-		line := 1
-		col := int(o.Start)
-		for _, lineOffset := range s.LineOffsets() {
-			if lineOffset < o.Start {
-				line++
-				col = int(o.Start - lineOffset)
-			} else {
-				break
-			}
-		}
-		return common.NewLocation(line, col)
+		return s.GetLocationByOffset(o.Start)
 	}
 	return common.NoLocation
 }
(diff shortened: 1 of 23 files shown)


policy/compiler.go
```diff
@@ -74,17 +74,21 @@ func (c *compiler) compileRule(r *Rule, ruleEnv *cel.Env, iss *cel.Issues) (*com
...
-			ruleEnv, err = ruleEnv.Extend(cel.Variable(fmt.Sprintf("%s.%s", variablePrefix, v.Name().Value), varAST.OutputType()))
-			if err != nil {
-				iss.ReportErrorAtID(v.Expression().ID, "invalid variable declaration")
-			}
-			compiledVars[i] = &compiledVariable{
-				name: v.name.Value,
-				expr: varAST,
-			}
+		varName := v.Name().Value
+		varType := cel.DynType
```
seirl: nit: I would add a comment here to explain why we need a DynType by default, it's not obvious.

policy/parser.go
```diff
@@ -32,13 +32,25 @@ const (
...
+
 // Policy declares a name, rule, and evaluation semantic for a given expression graph.
 type Policy struct {
 	name     ValueString
 	rule     *Rule
 	semantic semanticType
 	info     *ast.SourceInfo
 	source   *Source
+
+	metadata map[string]any
```
seirl, line 53: Are we only adding it to Policy? If we want to parse some rule-specific metadata, it will be a bit harder to store.

policy/compiler.go
TristonianJones: Done.

policy/parser.go
TristonianJones: For the moment, I'm just going to expose top-level metadata on Policy since it's available in all of the calls. Once we know more about how people are using this, I'm happy to evolve the API.

common/ast/ast.go
```diff
@@ -343,21 +333,26 @@ func (s *SourceInfo) GetStartLocation(id int64) common.Location {
...
+func (s *SourceInfo) GetLocationByOffset(offset int32) common.Location {
+	line := 1
+	col := int(offset)
+	for _, lineOffset := range s.LineOffsets() {
+		if lineOffset <= offset {
+			line++
+			col = int(offset - lineOffset)
+		} else {
+			break
+		}
```
l46kok: super nit: structuring this with a guard clause would make this read better imo. Up to you if you want to make the change:

```suggestion
        if lineOffset > offset {
            break
        }
        line++
        col = int(offset - lineOffset)
```

policy/parser.go
```diff
@@ -424,7 +498,7 @@ func (p *parserImpl) CollectMetadata(node *yaml.Node) int64 {
 	return id
 }
 
-func (p *parserImpl) parsePolicy(ctx ParserContext, node *yaml.Node) *Policy {
+func (p *parserImpl) ParsePolicy(ctx ParserContext, node *yaml.Node) *Policy {
```
l46kok, line 511: Add doc comments for exported names

TristonianJones:
PTAL. Thanks!

l46kok (approved):

PR #948, opened 2024-05-29
policy: parse explanations in match blocks

seirl:
This adds a secondary output to every block, an "explanation" of why a value was returned. This can be structured or non-structured text. For now the parser is simply storing the value, but the compiler cannot process it yet.


TristonianJones (approved):
I'm okay with this since adding compilation is probably not as straight-forward as just mutating the policy output. I think we should aim for structured outputs

PR #952, opened 2024-06-02
feat: load map data on first access

vivshankar:
The change here is to introduce a version of DynamicMap to load map contents based on a loader function
on first access only. This is to prevent the need to preload data, particularly from external sources, unless
the type is actually used.

Given `baseMap` is internal, rather than clone the entire file, the change here is to introduce the MapValueLoader
and an implementation is provided for DynamicMap and the corresponding reflectMapAccessor. This approach
can be extended to the other map types too.

### DIFF
--- a/common/types/map.go
+++ b/common/types/map.go
@@ -31,15 +31,34 @@ import (
 	structpb "google.golang.org/protobuf/types/known/structpb"
 )
 
+// MapValueLoader is a function type to load the map contents.
+type MapValueLoader func() (interface{}, int, error)
+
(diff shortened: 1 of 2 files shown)


TristonianJones:
@vivshankar It would be much more preferable to use a lazy variable `func() any` or `func() ref.Val` as a value provided to the `Activation` if you'd like to do some sort of lazy-loading at a top level. 

Alternatively, you could use a custom `types.Adapter` and override the `NativeToValue` call to produce map objects which are lazily initialized and proxy to the types and methods currently available on maps.

PR #955, opened 2024-06-05
Fixed partial variables extended bug and split PartialVarsEnvExtended test into two

jduraniglesias:
Changed the varsCopy variable to unconditionally copy e.variables. Split the partial variables extended test into TestPartialVarsEnv and TestPartialVarsExtendedEnv which tests, respectively, if internals are initialized and if sub environment is correctly copied. 

Fixes an edge case with env.PartialVars on an extended environment if the base was used for type checking already.

### DIFF
--- a/cel/env.go
+++ b/cel/env.go
@@ -309,17 +309,14 @@ func (e *Env) Extend(opts ...EnvOption) (*Env, error) {
 	copy(chkOptsCopy, e.chkOpts)
 
 	// Copy the declarations if needed.
-	varsCopy := []*decls.VariableDecl{}
 	if chk != nil {
 		// If the type-checker has already been instantiated, then the e.declarations have been
 		// validated within the chk instance.
 		chkOptsCopy = append(chkOptsCopy, checker.ValidatedDeclarations(chk))
-	} else {
-		// If the type-checker has not been instantiated, ensure the unvalidated declarations are
-		// provided to the extended Env instance.
-		varsCopy = make([]*decls.VariableDecl, len(e.variables))
-		copy(varsCopy, e.variables)
 	}
+	//
+	varsCopy := make([]*decls.VariableDecl, len(e.variables))
+	copy(varsCopy, e.variables)
 
 	// Copy macros and program options
 	macsCopy := make([]parser.Macro, len(e.macros))
--- a/cel/env.go
+++ b/cel/env.go
@@ -314,7 +314,6 @@ func (e *Env) Extend(opts ...EnvOption) (*Env, error) {
 		// validated within the chk instance.
 		chkOptsCopy = append(chkOptsCopy, checker.ValidatedDeclarations(chk))
 	}
-	//
 	varsCopy := make([]*decls.VariableDecl, len(e.variables))
 	copy(varsCopy, e.variables)
 
--- a/repl/evaluator.go
+++ b/repl/evaluator.go
@@ -585,6 +585,7 @@ func (e *Evaluator) AddOption(opt Optioner) error {
 	return nil
 }
 
+// EnablePartialEval enables the option to allow partial evaluations.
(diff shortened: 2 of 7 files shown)


cel/env.go
```diff
@@ -309,17 +309,14 @@ func (e *Env) Extend(opts ...EnvOption) (*Env, error) {
...
 		chkOptsCopy = append(chkOptsCopy, checker.ValidatedDeclarations(chk))
-	} else {
-		// If the type-checker has not been instantiated, ensure the unvalidated declarations are
-		// provided to the extended Env instance.
-		varsCopy = make([]*decls.VariableDecl, len(e.variables))
-		copy(varsCopy, e.variables)
 	}
+	//
+	varsCopy := make([]*decls.VariableDecl, len(e.variables))
+	copy(varsCopy, e.variables)
```
TristonianJones, line 318: Hi Javi,

I think the issue here might be that the validated declarations are expected to contain the variable declarations already, but in a more optimized form. It's possible that this change won't cause any harm, but it's worth running some env benchmarks

repl/evaluator.go
```diff
@@ -579,6 +585,17 @@ func (e *Evaluator) AddOption(opt Optioner) error {
 	return nil
 }
```
TristonianJones, line 587: nit: please add a doc comment to this method

test/bench/bench.go
```diff
@@ -40,6 +40,20 @@ type Case struct {
 	Out ref.Val
 }
 
+type EvalCase struct {
```
TristonianJones: Perhaps call this an `DynamicEnvCase` since this is related to tests related to changes in environment? Also, please add a top-level comment.

test/bench/bench.go
```diff
@@ -178,6 +192,53 @@ var (
 	}
 )
 
+var (
```
TristonianJones: nit: you can place the variable below within the same `var` block as above.

test/bench/bench.go
```diff
@@ -178,6 +192,53 @@ var (
...
+	ReferenceCheckerCases = []*EvalCase{
```
TristonianJones: nit: Consider renaming this to ReferenceDynamicEnvCases and add a comment to the exposed variable?

test/bench/bench.go
```diff
@@ -188,6 +249,13 @@ func RunReferenceCases(b *testing.B, env *cel.Env) {
 	}
 }
 
+func RunReferenceCheckerCases(b *testing.B) {
```
TristonianJones: Please add a top-level comment for the exposed method.

test/bench/bench.go
```diff
@@ -241,3 +309,17 @@ func RunCase(b *testing.B, env *cel.Env, bc *Case) {
 		})
 	}
 }
+
+func RunEvalCase(b *testing.B, bc *EvalCase) {
```
TristonianJones: nit: Please rename to `RunDynamicEnvCase` and add a top-level comment.

test/bench/bench.go
```diff
@@ -40,6 +40,21 @@ type Case struct {
 	Out ref.Val
 }
 
+// Case represents an expression compiled in a dynamic environment.
```
TristonianJones: nit: the format for Go comments should match the type or function name, `DynamicEnvCase represents ...`

test/bench/bench.go
```diff
@@ -188,6 +249,14 @@ func RunReferenceCases(b *testing.B, env *cel.Env) {
 	}
 }
 
+// RunReferenceCheckerCases evaluates the set of ReferenceDynamicEnvCases.
```
TristonianJones: nit: Can we rename this to match the test cases that are being run? i.e. `RunReferenceDynamicEnvCases`

TristonianJones (approved):

PR #956, opened 2024-06-05
Fix issue related to native byte arrays in type system

### LINKED ISSUE #953: Type checker and runtime types differ for native byte array in function overloads
**Describe the bug**
When using `ext.NativeTypes` with a byte array field, there CEL runtime  determines different types for at compile time
and at runtime.

**To Reproduce**

- [ ]  parser
- [x]  checker
- [x]  interpreter

```go
import (
	"encoding/hex"
        "fmt"
	"github.com/google/cel-go/common/types"
	"github.com/google/cel-go/common/types/ext"
)

type File struct {
	Hash [32]byte
}

func main() {
	reg := types.NewEmptyRegistry()
	env, err := cel.NewEnv(
		cel.CustomTypeAdapter(reg),
		cel.CustomTypeProvider(reg),
		ext.NativeTypes(
			reflect.TypeOf(&File{}),
		),
		cel.Variable("file", cel.ObjectType("main.File")),
		cel.Function("hex",
			cel.Overload("hex_bytes", []*cel.Type{types.BytesType}, types.StringType,
				cel.UnaryBinding(func(value ref.Val) ref.Val {
					bytes := value.(types.Bytes)
					return types.String(hex.EncodeToString(bytes))
				}))),
	)
	if err != nil {
		panic(err)
	}
	ast, iss := env.Compile(`hex(file.Hash)`)
	if iss.Err() != nil {
		panic(iss.Err())
	}
	prog, err := env.Program(ast)
	if err != nil {
		panic(err)
	}
	out, _, err = prog.Eval(map[string]any{
		"file": File{
			Hash: [32]byte{},
		},
	})
	if err != nil {
		panic(err)
	}
        fmt.Println(out)
}
```

**Actual behavior**
Program fails with
```
panic: no such overload: hex(list)
```

**Expected behavior**
Program should not panic and return the hex encoded bytes of the `file.Hash` field. 

**Additional context**
I've been trying to debug the issue and if i switch the overload function to accept a `types.ListType`, it fails at 
compile time. It only works if both overloads are defined. 

I think the issue is caused by the way the `nativeTypeProvider` implements `NativeToValue`. It only only ever returns a
`types.Bytes` for `[]bytes`, not for byte arrays ([here](https://github.com/google/cel-go/blob/master/ext/native.go#L328)), whereas the `convertToCELTypes` function specifies the `types.Bytes` type for byte arrays ([here](https://github.com/google/cel-go/blob/f7b83ae0424c68b21d1209d8dcf97f668dbf214a/ext/native.go#L387)).

patrickpichler:
Byte arrays are now always considered bytes by both the type checker and the runtime. This fixes a bug, when there was a native struct containing an byte array and a function that takes the byte arrays as input, the program would fail at runtime, as it would expect a dynamic list instead of bytes.

For an concrete example see the TestFunctionOverloadForNative test in the native tests.

Fixes #953

### DIFF
--- a/ext/native.go
+++ b/ext/native.go
@@ -325,7 +325,16 @@ func (tp *nativeTypeProvider) NativeToValue(val any) ref.Val {
 	// This isn't quite right if you're also supporting proto,
 	// but maybe an acceptable limitation.
 	switch refVal.Kind() {
-	case reflect.Array, reflect.Slice:
+	case reflect.Array:
+		refElem := refVal.Type().Elem()
+		if refElem == reflect.TypeOf(byte(0)) {
+			tmp := reflect.New(refVal.Type())
+			tmp.Elem().Set(refVal)
+
+			return types.Bytes(tmp.Elem().Bytes())
+		}
+		return types.NewDynamicList(tp, val)
+	case reflect.Slice:
 		switch val := val.(type) {
 		case []byte:
 			return tp.baseAdapter.NativeToValue(val)
--- a/common/overloads/overloads.go
+++ b/common/overloads/overloads.go
@@ -265,6 +265,7 @@ const (
 // Bytes conversion functions.
 const (
 	BytesToBytes  = "bytes_to_bytes"
+	IntsToBytes  = "ints_to_bytes"
 	StringToBytes = "string_to_bytes"
 )
 
--- a/ext/native_test.go
+++ b/ext/native_test.go
@@ -23,6 +23,7 @@ import (
 	"testing"
 	"time"
 
+	"encoding/hex"
 	"google.golang.org/protobuf/encoding/protojson"
 	"google.golang.org/protobuf/proto"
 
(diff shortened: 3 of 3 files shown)


ext/native.go
```diff
@@ -325,7 +325,16 @@ func (tp *nativeTypeProvider) NativeToValue(val any) ref.Val {
 	// This isn't quite right if you're also supporting proto,
 	// but maybe an acceptable limitation.
 	switch refVal.Kind() {
-	case reflect.Array, reflect.Slice:
+	case reflect.Array:
+		refElem := refVal.Type().Elem()
+		if refElem == reflect.TypeOf(byte(0)) {
+			tmp := reflect.New(refVal.Type())
```
patrickpichler, line 331: I am not sure about this part of the code. I think this will create a copy of the byte array, but I did not find any other way of retrieving a byte slice, as the passed in values are not addressable. 

If anyone knows a better way of doing this, I would love to hear it.

anjmao: Slice of bytes can be created using usafe without a copy. 
```go
var arr = [3]byte{1, 2, 3}
var arrSlice []byte = unsafe.Slice((*byte)(unsafe.Pointer(&arr[0])), len(arr))
```

patrickpichler: I sadly failed to do something like this with reflection. Apparently the reason for arrays being passed via interfaces are not addressable was a design decision to prevent subtle bugs, as arrays will still be passed by copy. 

In this case, this causes the not so nice side effect, that now we need to copy the array a second time (at least this is how i understand it, I might be wrong, but would be highly surprised).

patrickpichler:
@TristonianJones I see that a test was failing in CI, but I cannot reproduce it at all locally. Do you have any idea why that could happen? I also used go version 1.20 to run the tests.

PR #958, opened 2024-06-06
Fix for working with byte arrays

TristonianJones:
@patrickpichler I've moved the fix into a more central location. It's not perfectly optimal, but it is functional and I think that's more important at the moment.

### DIFF
--- a/common/types/bytes.go
+++ b/common/types/bytes.go
@@ -59,6 +59,9 @@ func (b Bytes) Compare(other ref.Val) ref.Val {
 func (b Bytes) ConvertToNative(typeDesc reflect.Type) (any, error) {
 	switch typeDesc.Kind() {
 	case reflect.Array:
+		if len(b) != typeDesc.Len() {
+			return nil, fmt.Errorf("[%d]byte not assignable to [%d]byte array", len(b), typeDesc.Len())
+		}
 		refArrPtr := reflect.New(reflect.ArrayOf(len(b), typeDesc.Elem()))
 		refArr := refArrPtr.Elem()
 		for i, byt := range b {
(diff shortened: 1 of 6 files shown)


common/types/bytes.go
```diff
@@ -58,7 +58,14 @@ func (b Bytes) Compare(other ref.Val) ref.Val {
 // ConvertToNative implements the ref.Val interface method.
 func (b Bytes) ConvertToNative(typeDesc reflect.Type) (any, error) {
 	switch typeDesc.Kind() {
-	case reflect.Array, reflect.Slice:
+	case reflect.Array:
```
patrickpichler, line 61: Why is the custom conversion needed here? 

I've tested this on the go playground and it should be possible to convert from a byte slice to an byte array, as long as the byte slice has more items than the requested array.

https://go.dev/play/p/fxZM4NPhWD8?v=goprev

TristonianJones: There seems to be some issue with converting between byte and uint8 in the Bazel builds. I'm not sure why, but this change works around the problem

patrickpichler: There should probably be a check to ensure that the `len` of the requested array is equals to the `len` of the returned array. Otherwise this would fail on type conversion.

e.g. 
```go
b := types.Bytes([]byte{1})
val, _ := b.ConvertToNative(reflect.TypeOf([10]byte{})
val.([10]byte) // <- this will panic
```

patrickpichler: Probably 
`refArrPtr := reflect.New(reflect.ArrayOf(len(b), typeDesc.Elem()))`
can be changed to 
`refArrPtr := reflect.New(typeDesc.Elem)`

but the sanity check for `len` should still be put in place imho

TristonianJones: Thanks for the catch!

jcking (approved):

PR #961, opened 2024-06-10
Introduce protobuf message testing to policies

TristonianJones:
Cover protobuf message types in policies

Also, addresses a subtle bug where function overload implementations
were being clobbered when a duplicate function overload signature
without an implementation was introduced in a later environment.

### DIFF
--- a/common/decls/decls.go
+++ b/common/decls/decls.go
@@ -162,7 +162,7 @@ func (f *FunctionDecl) AddOverload(overload *OverloadDecl) error {
 		if oID == overload.ID() {
 			if o.SignatureEquals(overload) && o.IsNonStrict() == overload.IsNonStrict() {
 				// Allow redefinition of an overload implementation so long as the signatures match.
-				if !o.hasBinding() || overload.hasBinding() {
+				if overload.hasBinding() {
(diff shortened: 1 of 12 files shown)


common/decls/decls.go
```diff
@@ -162,7 +162,9 @@ func (f *FunctionDecl) AddOverload(overload *OverloadDecl) error {
 		if oID == overload.ID() {
 			if o.SignatureEquals(overload) && o.IsNonStrict() == overload.IsNonStrict() {
 				// Allow redefinition of an overload implementation so long as the signatures match.
-				f.overloads[oID] = overload
+				if !o.hasBinding() || overload.hasBinding() {
```
l46kok: Seems like `overload.hasBinding()` is the only necessary condition since the intention is to prevent emptying out the existing binding.

l46kok (dismissed):

common/decls/decls.go
TristonianJones: Good point. Removing the branch

TristonianJones: PTAL

l46kok (approved):

PR #965, opened 2024-06-12
Policy test case fix for restricted destination, prevent panic when return is missing

l46kok:
restricted_destination produces a boolean output, but the test wasn't actually validating its result. Includes a minor fix to gracefully error when return type is missing in policy config.


TristonianJones (approved):
Thanks!

PR #978, opened 2024-07-08
Add function accessor to Env

jpbetz:
We're working on improving our backward compatibility linting to ensure that we never accidentally remove library functions once they are added. This would make it a lot easier for us to inspect the environments we actually build and use when linting.

### DIFF
--- a/cel/env_test.go
+++ b/cel/env_test.go
@@ -271,6 +271,22 @@ func TestLibraries(t *testing.T) {
 	}
 }
 
+func TestFunctions(t *testing.T) {
+	e, err := NewEnv(OptionalTypes())
+	if err != nil {
+		t.Fatalf("NewEnv() failed: %v", err)
+	}
+	for _, expected := range []string{"optional.of", "or"} {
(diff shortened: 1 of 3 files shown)


cel/env_test.go
```diff
@@ -271,6 +271,22 @@ func TestLibraries(t *testing.T) {
 	}
 }
 
+func TestFunctions(t *testing.T) {
+	e, err := NewEnv(OptionalTypes())
+	if err != nil {
+		t.Fatalf("NewEnv() failed: %v", err)
+	}
+	for _, expected := range []string{"optional.of", "or"} {
```
TristonianJones, line 279: You could use the list of the optional functions here, but I would probably also recommend adding a test which covers a function library that has a version attached to it, possibly updating the `ext/strings.go` test to validate the same feature and assert that a function does not exist when an older version is pinned?

jpbetz: This is a good idea. I've added a test that functions as a backward compatibility linter for `ext/strings.go`.

TristonianJones (approved):

Issue #352, opened 2020-05-11
Flatten or reduce a list of lists


btomasini:
**Feature request checklist**

- [x] There are no issues that match the desired change
- [x] The change is large enough it can't be addressed with a simple Pull Request
- [ ] If this is a bug, please file a [Bug Report](./ISSUE_TEMPLATE).

**Change**
I don't know if this is a feature request or a request for assistance. I have a cel expression which maps a list and in the predicate does a lookup against a map<string, list<string>>.  The result is a list of lists of stings, eg [['a'],[],['b', 'c']].  I would like to combine these into a single list of ['a','b','c'].  Here are some options I have thought of:

1. A flatmap macro.
2. A flatten function
3. A reducer in which I append the lists together

So far I am closest to implementing option 2 but am having some troubles with the InstanceOverloads not matching the input type.

Are there any built-in features of cel-go which would support this?

I am closest to implementing option 2 on my own but am having some troubles with the instance overloads not matching. If there is no built-in feature of cel-go for this, I will post a code sample here of my current work.

TristonianJones:
Hi @btomasini. A function makes the most sense (Option 2). Do you happen to have an example, test case, or gist that you could share?

btomasini:
@TristonianJones 

My original attempt failed as I treated it as an instance override rather than a global.  I plan to try again soon and will let you know what I come up with it.

TristonianJones:
@btomasini If you'd like some general support for CEL, you're also welcome to reach out to cel-go-discuss@googlegroups.com. It's usually a better forum for free-form discussion.

TristonianJones:
Hi @btomasini if you need any further help, I'd be happy to help in the discussion forums. For now, I'm closing the issue.

PR #980, opened 2024-07-10
Add list flatten()

fabriziosestito:
Adds list flatten, fixing #352

### DIFF
--- a/ext/lists.go
+++ b/ext/lists.go
@@ -38,15 +38,27 @@ import (
 //
 // # Flatten
 //
-// Flattens the given list of nested lists.
+// Flattens the given list to a single level.
 //
 //	<list>.flatten(<list>) -> <list>
 //
 // Examples:
 //
 // [1,[2,3],[4]].flatten() // return [1, 2, 3, 4]
-// [1,[2,3],[],[],[4]].flatten() // return [1, 2, 3, 4]
-// [1,[2,3],[[[4]]]].flatten() // return [1, 2, 3, 4]
+// [1,[2,[3,4]]].flatten() // return [1, 2, [3, 4]]
+// [1,2,[],[],[3,4]].flatten() // return [1, 2, 3, 4]
+//
+// # FlattenDeep
+//
+// Flattens the given list to the deepest level.
+//
+//	<list>.flattenDeep(<list>) -> <list>
+//
+// Examples:
+//
+// [1,[2,3],[4]].flattenDeep() // return [1, 2, 3, 4]
+// [1,[2,[3,4]]].flattenDeep() // return [1, 2, 3, 4]
+// [1,[2,[3,[4]]]].flattenDeep() // return [1, 2, 3, 4]
 func Lists() cel.EnvOption {
 	return cel.Lib(listsLib{})
 }
@@ -82,7 +94,17 @@ func (listsLib) CompileOptions() []cel.EnvOption {
 				[]*cel.Type{listType}, listType,
 				cel.UnaryBinding(func(arg ref.Val) ref.Val {
 					list := arg.(traits.Lister)
-					flatList := flatten(list)
+					flatList := flattenHelper(list, false)
+					return types.DefaultTypeAdapter.NativeToValue(flatList)
+				}),
+			),
+		),
+		cel.Function("flattenDeep",
+			cel.MemberOverload("list_flatten_deep",
+				[]*cel.Type{listType}, listType,
+				cel.UnaryBinding(func(arg ref.Val) ref.Val {
+					list := arg.(traits.Lister)
+					flatList := flattenHelper(list, true)
 					return types.DefaultTypeAdapter.NativeToValue(flatList)
 				}),
 			),
@@ -115,7 +137,7 @@ func slice(list traits.Lister, start, end types.Int) (ref.Val, error) {
 	return types.DefaultTypeAdapter.NativeToValue(newList), nil
 }
 
-func flatten(list traits.Lister) []ref.Val {
+func flattenHelper(list traits.Lister, deep bool) []ref.Val {
 	iter := list.Iterator()
 	var newList []ref.Val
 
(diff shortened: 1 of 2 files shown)


flavio (commented):
Looks pretty neat, thanks!

seirl (commented):
Is there a reason to make this recursive by default? The common case is to flatten a list of list. I feel like a recursive flatten would better be a separate `.flattenRec()` function.

fabriziosestito:
@seirl I took inspiration from Elixir's [List.flatten](https://hexdocs.pm/elixir/1.12/List.html#flatten/1).
Also, Ruby's [Array.flatten](https://docs.ruby-lang.org/en/3.3/Array.html#method-i-flatten) is recursive by default, but it accepts an optional `level` argument. We could consider this behaviour, as it is more flexible.

However, I don't have a strong opinion and I am ok with what you proposed.

seirl:
Interesting. FWIW:

- in Rust, `std::iter::Flatten` flattens a single level.
- in Python, `itertools.chain()` flattens a single level.

The reason I'm mentioning this is because `flatten` could also be useful to flatten a `list(optional(T))` into a `list(T)`, but the semantics are more complex if you have a recursive flatten.

fabriziosestito:
@seirl Alright, I will change this to single-level flattening.

fabriziosestito:
@seirl I've followed your suggestion, now `flatten` flattens to a single level, while `flattenDeep` flattens recursively. I've preferred the `deep` naming as I find it clearer for the user and it is also used by [lodash](https://lodash.com/docs/4.17.15#flattenDeep).

ext/lists.go
```diff
@@ -65,6 +89,26 @@ func (listsLib) CompileOptions() []cel.EnvOption {
 				}),
 			),
 		),
+		cel.Function("flatten",
+			cel.MemberOverload("list_flatten",
+				[]*cel.Type{listType}, listType,
```
TristonianJones: It seems somewhat unlikely that you'll get the right behavior at check-time for these signatures since the `list(T)` as input and `list(T)` as output could lead to situations where no reduction in type complexity occurs. You might need to specify `list(dyn)` to `list(dyn)` otherwise the subsequent expressions might choke at type-check time such as the example in #982

fabriziosestito: @TristonianJones thanks for looking at it. I've changed the signatures to `list(dyn) -> list(dyn)` in https://github.com/google/cel-go/pull/980/commits/f12810c53371bc3d977817d650daabcdc96d149b

seirl: Doesn't this prevent us from enforcing strong type-checking when we use `flatten()`?
My codebase requires that the output type of an expression shouldn't be `dyn` at check-time to enforce strong type-checking. Should I not do this?

ext/lists.go
```diff
@@ -65,6 +90,26 @@ func (listsLib) CompileOptions() []cel.EnvOption {
 				}),
 			),
 		),
+		cel.Function("flatten",
```
TristonianJones: @fabriziosestito Thanks for the feature updates! When adding new functions you'll need to introduce a concept of versions to the library. See `ext/strings.go` as an example of how this should look. The idea is that users should be able to opt-in to the new version or remain frozen.

ext/lists.go
```diff
@@ -65,6 +90,26 @@ func (listsLib) CompileOptions() []cel.EnvOption {
...
+			cel.MemberOverload("list_flatten",
+				[]*cel.Type{listDyn}, listDyn,
+				cel.UnaryBinding(func(arg ref.Val) ref.Val {
+					list := arg.(traits.Lister)
+					flatList := flattenHelper(list, false)
+					return types.DefaultTypeAdapter.NativeToValue(flatList)
+				}),
+			),
+		),
+		cel.Function("flattenDeep",
```
TristonianJones: Instead of `flattenDeep`, can we consider adding an overload of `flatten()` that takes an explicit argument indicating how many levels deep to flatten, e.g. `list_flatten_int`? This is safer than unbounded recursion. I'd be open to having `-1` as the depth signal unbounded recursion, but we can at least make it really clear what the compute costs could be for flattening.

ext/lists.go
```diff
@@ -65,6 +90,26 @@ func (listsLib) CompileOptions() []cel.EnvOption {
...
+			cel.MemberOverload("list_flatten_deep",
+				[]*cel.Type{listDyn}, listDyn,
+				cel.UnaryBinding(func(arg ref.Val) ref.Val {
+					list := arg.(traits.Lister)
+					flatList := flattenHelper(list, true)
+					return types.DefaultTypeAdapter.NativeToValue(flatList)
+				}),
+			),
+		),
```
TristonianJones: Could you add check-time and runtime cost estimations? See https://github.com/google/cel-go/pull/850 for an example of how this was done and tests on `ext/sets.go` and `ext/sets_test.go`? This will help with onboarding into Kubernetes. Otherwise the `flatten()` will look like it has an infinite cost (which should only be true when someone uses `flatten(-1)`.

TristonianJones (changes requested):
Thank you for the update. I've added a few comments on how to structure this so it's safer for production rollout and works with cost estimation tools!

ext/lists.go
TristonianJones: @seirl Unfortunately this is a limit of the type-checker in that it can't reason about reduction in type-parameter complexity, especially when it's arbitrary, e.g. for `flatten(list(T), int) -> T`, how would the type-checker be able to infer the proper type of `T` when `T` could be `list(list(list(V)))` but the runtime output could be `list(V)` or some other permutation of a list.

The only other approach you could take would be to introduce something like the example output structure into the type-signature, but that's going to be tricky, e.g. `flatten(list(T), list(V)) -> list(V)`

TristonianJones: To @seirl's point if we added overloads for `flatten(list(list(T))) -> list(T)`, then we can get some stronger type-checking at least for depth of 1.

TristonianJones (approved):
I'll update the signature of single-level flatten post-merge. Thanks for the contribution and the patience!

fabriziosestito:
@TristonianJones, thank you for merging. I'm currently on vacation with limited laptop access. I'll handle the cost estimation if the task is still open when I return. Regarding the function signature, shouldn't it be something like `flatten<T>(list(T | list(T))) -> list(T)`?

seirl:
@fabriziosestito the one-level .flatten() should only flatten List[List[T]], so [1, [2, 3]].flatten() should be an error.

If you use only strongly-typed CEL, all your lists need to be homogeneous, so [1, [2, 3]] doesn't make sense as a value.

TristonianJones:
@seirl I took care of the signature change in a different PR. The test case will pass checking because the mixed type list is `list(dyn)`

PR #981, opened 2024-07-12
Use bytes interface value for zero native type struct fields instead of dynamic list

patrickpichler:
If a struct has a byte array/slice field and it happens to be the zero value, a dynamic list object was returned. This caused issues with functions taking `Bytes` via arguments, as the type checker will accept such functions, but they might fail at runtime.

To work around this, the raw bytes array/slice is returned in case of an zero value.

### DIFF
--- a/ext/native.go
+++ b/ext/native.go
@@ -279,7 +279,7 @@ func (tp *nativeTypeProvider) FindStructFieldType(typeName, fieldName string) (*
 		GetFrom: func(obj any) (any, error) {
 			refVal := reflect.Indirect(reflect.ValueOf(obj))
 			refField := valueFieldByName(tp.options.parseStructTags, refVal, fieldName)
-			return getFieldValue(tp, refField), nil
+			return getFieldValue(refField), nil
 		},
 	}, true
 }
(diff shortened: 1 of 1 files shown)


ext/native.go
```diff
@@ -711,6 +711,9 @@ func getFieldValue(adapter types.Adapter, refField reflect.Value) any {
 	if refField.IsZero() {
 		switch refField.Kind() {
 		case reflect.Array, reflect.Slice:
+			if refField.Type().Elem() == reflect.TypeOf(byte(0)) {
+				return refField.Interface()
+			}
 			return types.NewDynamicList(adapter, []ref.Val{})
```
patrickpichler: @TristonianJones thinking about the logic here, what do you think about also returning the interface value of zero types (besides pointers probably). The way I understand it, shouldn't then further down the chain native type be converted via the type adapter?

TristonianJones: What you suggest seems like the right choice to me since something elsewhere in the stack should do the proper type conversion using the adapter. This is likely an oversight on my part ... worth trying for sure!

patrickpichler: The code has been adjusted to always use the value of `.Interface()`.

TristonianJones: It looks like this change works in some cases, but causes breakages in others. My guess is that nested type accesses need to be proxied through the proper `adapter` instance, so only scalar types should return `.Interface()` and the complex types should go through the previous logic related to lists, maps, and structs.

patrickpichler: Ah my bad! Looks like I didn't run the tests again after my change! The problem has been fixed.

ext/native.go
```diff
@@ -704,21 +704,16 @@ func (t *nativeType) hasField(fieldName string) (reflect.StructField, bool) {
...
 		switch refField.Kind() {
-		case reflect.Array, reflect.Slice:
-			return types.NewDynamicList(adapter, []ref.Val{})
-		case reflect.Map:
-			return types.NewDynamicMap(adapter, map[ref.Val]ref.Val{})
 		case reflect.Struct:
 			if refField.Type() == timestampType {
 				return types.Timestamp{Time: time.Unix(0, 0)}
 			}
-			return reflect.New(refField.Type()).Elem().Interface()
```
TristonianJones, line 721: It's probably worth noting how this differs from the `nativeToValue` behavior in `native.go`. The `time.Unix(0, 0)` value can be returned on its own in the timestamp block since the proto semantics are that uninitialized timestamp accesses are treated as unix time 0. 

``` go
// Proto semantics will return a time value initialized at Unix time 0 if the value is unset,
// but golang will return the zero value for `time.Time` which is not the same.
if refField.Type() == timestampType {
	return time.Unix(0, 0)
}
```

TristonianJones (dismissed):
One minor nit, but otherwise LGTM

ext/native.go
patrickpichler: Updated to now simply return `time.Unix(0,0)` 👍

TristonianJones (approved):

Issue #985, opened 2024-07-25
[Policy engine] Support conditional nested rules


seirl:
In the policy engine, it would be useful to support conditional nested rules, because that would allow us to define variables that only make sense if a specific condition is met:

```
rule:
  match:
    - condition: a.hasValue()
      rule:
        variables:
          - name: a_derived
            expression: "a.value().derived"
        match:
          - condition: a_derived > 42
            output: "false"
          - condition: a_derived < 42
            output: "false"
          - output: true
    - condition: "b.hasValue()"
      output: "b.value()"
    - output: "false"
```

However this is currently not supported and will fail with the error message `subrule early terminates policy`. I think this check is here to prevent things like:

```
rule:
  match:
    - rule:
        match:
          - condition: a > 42
            output: "false"
          - output: true
    - output: "false"
```

where the last "output" block can never be reached.

To support this, the check should be relaxed, but we still want to catch the error above. I think a good refined check would be:

**"If the nested rule is guarded by a trivially true condition, it should not early terminate the policy".**

This would allow the first example (because there is a condition `a.hasValue()` guarding the nested rule), but not the second (because there is no condition for the rule and for the last output block).

seirl:
Come to think of it, I don't really see the use-case for unconditional rules (at least ones that are not the last one in the list). For me the reason they are useful is mostly to define variables that only make sense when the condition is met, otherwise you can just flatten the nested rule in the upper rule.

PR #989, opened 2024-07-30
Variable and nested rule limits

### LINKED ISSUE #972: [Policy Compiler] Specify iteration limit in RuleComposer
For safety and consistency across other optimizer implementations, [an iteration limit should be specified](https://github.com/google/cel-go/blob/24a38efc7346e0a9f1b3c6c4dcadc650101b6259/policy/compiler.go#L162), or be configurable. This ensures the optimization work would eventually terminate in case pathologically large ASTs are supplied or there is a bug with the optimizer implementation.

TristonianJones:
Provide a configurable limit for variable and nested rules

Ensure that policies are no so expressive that they cause problems
for stacks with limited recursion depths. The limit defaults to 100.
This change also makes a critical fix to support conditionally nested
rules. Previously the condition was being omitted during composition.

Closes  #972

### DIFF
--- a/policy/composer.go
+++ b/policy/composer.go
@@ -63,7 +63,7 @@ func (opt *ruleComposerImpl) optimizeRule(ctx *cel.OptimizerContext, r *Compiled
 	vars := r.Variables()
 	optionalResult := true
 
-	// Build the rule sub-graph.
+	// Build the rule subgraph.
 	for i := len(matches) - 1; i >= 0; i-- {
 		m := matches[i]
 		cond := ctx.CopyASTAndMetadata(m.Condition().NativeRep())
@@ -109,7 +109,7 @@ func (opt *ruleComposerImpl) optimizeRule(ctx *cel.OptimizerContext, r *Compiled
 		}
 	}
 
-	// Bind variables in reverse order to declaration on top of rule-subgrap
+	// Bind variables in reverse order to declaration on top of rule-subgraph.
 	for i := len(vars) - 1; i >= 0; i-- {
 		v := vars[i]
 		varAST := ctx.CopyASTAndMetadata(v.Expr().NativeRep())
--- a/policy/composer.go
+++ b/policy/composer.go
@@ -24,7 +24,10 @@ import (
 // NewRuleComposer creates a rule composer which stitches together rules within a policy into
 // a single CEL expression.
 func NewRuleComposer(env *cel.Env, p *Policy) *RuleComposer {
-	return &RuleComposer{env: env, p: p}
+	return &RuleComposer{
+		env: env,
+		p:   p,
+	}
 }
 
 // RuleComposer optimizes a set of expressions into a single expression.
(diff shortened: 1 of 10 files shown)


policy/composer.go
```diff
@@ -90,10 +97,19 @@ func optimizeRule(ctx *cel.OptimizerContext, r *CompiledRule) (ast.Expr, bool) {
...
+				operators.Conditional,
+				cond,
+				nestedRule,
+				matchExpr,
+			)
+		}
 	}
 
-	vars := r.Variables()
+	// Bind variables in reverse order to declaration on top of rule-subgrap
```
l46kok:
```suggestion
	// Bind variables in reverse order to declaration on top of rule-subgraph
```

policy/config.go
```diff
@@ -87,7 +88,12 @@ func (ec *ExtensionConfig) AsEnvOption(baseEnv *cel.Env) (cel.EnvOption, error)
 	if !found {
 		return nil, fmt.Errorf("unrecognized extension: %s", ec.Name)
 	}
-	return fac(uint32(ec.Version)), nil
+	// If the version is less than or equal to zero set the version to the max value.
```
l46kok:
```suggestion
	// If the version is less than zero set the version to the max value.
```

Unless if you meant to include 0, but I think we treat 0 as when the library is first introduced

policy/testdata/limits/config.yaml
```diff
@@ -0,0 +1,22 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+name: "limits"
+extensions:
+  - name: "strings"
+    version: -1
```
l46kok: Just a thought: having a named constant would make this clearer (ex: latest). We can defer, I don't know how trivial it is to make the config deserializer support this.

policy/composer.go
```diff
@@ -90,10 +97,19 @@ func optimizeRule(ctx *cel.OptimizerContext, r *CompiledRule) (ast.Expr, bool) {
 			ctx.ReportErrorAtID(nestedRule.ID(), "subrule early terminates policy")
```
l46kok, line 97: We can effectively tip-toe around the subrule termination guard here because we're only checking for the optional properties of the immediate ancestor:

```
name: "subrule optional"
rule:
  match:
    - rule:
        match:
          - rule:
              match:
                - condition: >
                    x > 100
                  output: "'hi'"
          - output: "'this is effectively the default'"
    - output: "'we can never reach this'"
```

I think this problem likely existed before this fix, but probably good to take in consideration with https://github.com/google/cel-go/issues/985 if we end up supporting conditional nested rules.

policy/config.go
TristonianJones: You're right, it needs to be `less than zero`, but I updated `version` to be a string so we can use `latest` as the version

policy/testdata/limits/config.yaml
TristonianJones: Fixed. Good suggestion.

policy/composer.go
TristonianJones: Definitely, the issue is mostly related to a lack of a condition expression when entering a rule which contains an `output` expression. I'll update this code path to reflect an `unreachable output` error in a subsequent PR for #985. It should be possible to detect through just a static pass over the outputs to see if there's more than one without any condition attached (directly, or in ancestor)

l46kok (approved):

PR #991, opened 2024-08-01
Improved support for nested rules

### LINKED ISSUE #985: [Policy engine] Support conditional nested rules

TristonianJones:
Introduces a proper check to ensure all branches within a subrule are reachable,
thus enabling support for conditionally nested rules.

Closes #985

### DIFF
--- a/policy/compiler.go
+++ b/policy/compiler.go
@@ -68,6 +68,7 @@ func (r *CompiledRule) OutputType() *cel.Type {
 // HasOptionalOutput returns whether the rule returns a concrete or optional value.
 // The rule may return an optional value if all match expressions under the rule are conditional.
 func (r *CompiledRule) HasOptionalOutput() bool {
+	optionalOutput := false
 	for _, m := range r.Matches() {
 		if m.NestedRule() != nil && m.NestedRule().HasOptionalOutput() {
 			return true
@@ -75,8 +76,9 @@ func (r *CompiledRule) HasOptionalOutput() bool {
 		if m.ConditionIsLiteral(types.True) {
 			return false
 		}
+		optionalOutput = true
 	}
-	return true
+	return optionalOutput
 }
 
 // CompiledVariable represents the variable name, expression, and associated type-check declaration.
@@ -116,6 +118,7 @@ type CompiledMatch struct {
 	nestedRule *CompiledRule
 }
 
+// SourceID returns the source identifier associated with the compiled match.
 func (m *CompiledMatch) SourceID() int64 {
 	return m.exprID
 }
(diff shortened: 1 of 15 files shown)


policy/compiler.go
```diff
@@ -84,17 +100,28 @@ func (v *CompiledVariable) Declaration() *decls.VariableDecl {
...
+
 // Condition returns the compiled predicate expression which must evaluate to true before the output
 // or subrule is entered.
 func (m *CompiledMatch) Condition() *cel.Ast {
 	return m.cond
 }
 
+// ConditionIsLiteral indicates whether the condition for the match is a literal with a given value.
+func (m *CompiledMatch) ConditionIsLiteral(val ref.Val) bool {
+	c := m.cond.NativeRep().Expr()
```
l46kok, line 134: This may panic due to `cond` being `nil` if we fail to compile the condition for a match: https://github.com/google/cel-go/pull/991/files#diff-329a773e11a74db809329cb56d571b04b394de161cef34e6ac66f9bf2fd21c45R256

TristonianJones: On a compiled match, the condition is always non-nil since it's either trivially `true` or a custom expression. On a parsed match, the value can be `nil`, but there are better guarantees here.

l46kok: If the custom expression cannot be compiled, the condition AST will not be set, and we'll still enter this literal check when doing the reachable check through `HasOptionalOutput()` function, at which point it should panic because `cond` is `nil`.

I think this will panic on an input like:

```
rule:
  match:  
    - condition: bad_condition
      output: "'foo'"
```

TristonianJones: Ah, good point. I realized the `cel.Ast` wasn't as `nil`-safe as it could be, so the check should now work just fine even if `cond` is `nil` given the update to the nil-safety for `cel.Ast`

TristonianJones:
PTAL

l46kok (approved):

PR #992, opened 2024-08-02
Check for output type agreement during the compile phase

### LINKED ISSUE #973: [Policy Compiler] Surface type-mismatches error in a more readable fashion
Policy compiler will fail to type-check the composed policies (expectedly) if their output types are different. Example:

```
name: foo_policy
rule:
    match:
       - condition: x
         output: "error"
       - output: true
```

```
Compile() failed: ERROR: testdata/nested_rule/policy.yaml:-1:0: found no matching overload for '_?_:_' applied to '(bool, map(string, bool), string)'
```

We should try to raise this error in a more readable fashion, and ideally with the source positions wrt/ yaml. 

(xref b/350064634 for Java impl)

TristonianJones:
Validate that `CompiledRule` and `CompiledMatch` output expression types agree.

Closes #973

### DIFF
--- a/policy/compiler.go
+++ b/policy/compiler.go
@@ -115,11 +115,15 @@ func (m *CompiledMatch) NestedRule() *CompiledRule {
 	return m.nestedRule
 }
 
+// OutputType returns the cel.Type associated with output expression.
 func (m *CompiledMatch) OutputType() *cel.Type {
 	if m.output != nil {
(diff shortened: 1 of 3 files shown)


l46kok (dismissed):

TristonianJones:
@l46kok PTAL. I missed a comment and updated `nil` safety

l46kok (approved):

PR #993, opened 2024-08-02
Support for typename import aliases

TristonianJones:
Allow users to specify type name imports in policies.

Imported types are aliased by the simple type name:

```yaml
# Create an import alias to `TypeName`
imports:
  - name: my.custom.really.ridiculously.long.package.TypeName

rule:
  match:
    - output: "TypeName{}"
```
 
As a future refinement, the imports will support a custom `alias` to allow
for alias disambiguation between types which come from different packages
but have the same simple name.

### DIFF
--- a/policy/parser.go
+++ b/policy/parser.go
@@ -65,6 +65,7 @@ func (p *Policy) SourceInfo() *ast.SourceInfo {
 	return p.info
 }
 
+// Imports returns the list of imports associated with the policy.
 func (p *Policy) Imports() []*Import {
 	return p.imports
 }
@@ -94,8 +95,9 @@ func (p *Policy) MetadataKeys() []string {
 	return keys
 }
 
-func (p *Policy) AddImport(name ValueString) {
-	p.imports = append(p.imports, &Import{name: name})
+// AddImport adds an import to the policy.
+func (p *Policy) AddImport(i *Import) {
+	p.imports = append(p.imports, i)
 }
 
 // SetName configures the policy name.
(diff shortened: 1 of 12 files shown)


TristonianJones:
FYI @seirl, I've added proper support for this feature.

policy/parser.go
```diff
@@ -88,6 +94,10 @@ func (p *Policy) MetadataKeys() []string {
 	return keys
 }
 
+func (p *Policy) AddImport(name ValueString) {
```
l46kok: Add doc comments on all exported names

policy/testdata/pb/policy.yaml
```diff
@@ -13,8 +13,20 @@
 # limitations under the License.
 
 name: "pb"
+
+imports:
+   - name: google.expr.proto3.test.TestAllTypes.NestedEnum
+   - name: >
+       "google.expr.proto3.test.ImportedGlobalEnum"
```
l46kok: I'm actually kind of surprised the import for this works. My impression was that we would have to "prune" the additional indentation and the quotes that comes along with the import due to the multiline syntax. Is there something in the abbreviation mechanism that makes this possible?

l46kok: Thinking about this further -- we should remove the container setting here to properly test this, so that it's not accidentally resolving these imports through the explicitly set container: https://github.com/google/cel-go/blob/master/policy/testdata/pb/config.yaml#L16.

TristonianJones: I adjusted the container to remove `proto3` which will cause namespace resolution to fail and forces me to add another alias. The reason I've kept the container at all is to simplify test case generation.

TristonianJones:
PTAL

l46kok (approved):

PR #995, opened 2024-08-03
Foldable Maps and Lists

TristonianJones:
Support for Foldable map and list types prior to runtime support for two iter-var values in comprehensions v2

### DIFF
--- a/common/types/map.go
+++ b/common/types/map.go
@@ -396,7 +396,7 @@ func (a *jsonStructAccessor) Iterator() traits.Iterator {
 // Fold calls the FoldEntry method for each (key, value) pair in the map.
 func (a *jsonStructAccessor) Fold(f traits.Folder) {
 	for k, v := range a.st {
-		if !f.FoldEntry(MapType, k, v) {
+		if !f.FoldEntry(k, v) {
 			break
 		}
 	}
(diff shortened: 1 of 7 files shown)


common/types/traits/iterator.go
```diff
@@ -34,3 +34,16 @@ type Iterator interface {
...
+type Foldable interface {
+	// Fold invokes the Folder.FoldEntry for all entries in the type
+	Fold(Folder)
+}
+
+// Folder performs a fold on a given entry and indicates whether to continue folding.
+type Folder interface {
+	// FoldEntry indicates the calling type and the (key, value) pair associated with the entry.
+	// If the output is true, continue folding. Otherwise, terminate the fold.
+	FoldEntry(t ref.Type, key, val any) bool
```
jcking: What is the purpose of passing the type to the `FoldEntry`? is this something that is going to be used later?

TristonianJones: It has to do with whether the fold is happening within one or two argument comprehensions to I know which value to assign where; however, I could remove this and just maintain support for the single and dual arg comprehensions separately in later CLs (might need to for performance reasons anyway)

TristonianJones: Removed. PTAL

common/types/map.go
```diff
@@ -94,6 +94,24 @@ func NewProtoMap(adapter Adapter, value *pb.Map) traits.Mapper {
 	}
 }
 
+// NewMutableMap constructs a mutable map from an adapter and a set of map values.
+func NewMutableMap(adapter Adapter, mutableValues map[ref.Val]ref.Val) traits.MutableMapper {
```
jcking, line 98: What happens if a mutable map is part of `mutableValues`? is that possible?

TristonianJones: That shouldn't be possible since the nested comprehensions will ensure the outputs are immutable. Unless, of course, someone breaks a lot of assumptions with their own extension.

jcking (approved):

PR #1001, opened 2024-08-12
Implement native conformance test runner

jcking:


conformance/BUILD.bazel
```diff
@@ -24,47 +31,60 @@ ALL_TESTS = [
...
+
+go_test(
+    name = "go_default_test",
+    size = "small",
+    srcs = [
+        "conformance_test.go",
+    ],
+    tags = [
+        "manual",
+        "notap",
```
TristonianJones, line 60: Should we put `notap` here?

conformance/conformance_test.go
```diff
@@ -0,0 +1,265 @@
...
+	tests     testsFlag
+	skipTests skipTestsFlag
+
+	envWithMacros *cel.Env
+	envNoMacros   *cel.Env
+)
+
+func init() {
+	flag.BoolVar(&dashboard, "dashboard", false, "Dashboard.")
+	flag.Var(&tests, "tests", "Paths to run, separate by a comma.")
```
TristonianJones, line 75: Could you document an example path?

conformance/conformance_test.bzl
```diff
@@ -0,0 +1,58 @@
...
+    native.sh_test(
+        name = name,
+        size = "small",
+        srcs = ["//conformance:conformance_test.sh"],
+        args = ["$(location //conformance:go_default_test)"] + _conformance_test_args(data, skip_tests, dashboard),
+        data = ["//conformance:go_default_test"] + data,
+        tags = [
+            "guitar",
+            "manual",
+            "notap",
```
TristonianJones, line 56: We should probably nix references to guitar and notap here.

conformance/BUILD.bazel
jcking: Yes, this test target is not meant to be run directly. It is invoked via the conformance_test macros.

conformance/conformance_test.bzl
jcking: Its easier to leave them, as adding them back in downstream will be tedious.

conformance/conformance_test.go
jcking: This tool is never meant to be invoked directly, only through the macros. Bazel passes through the paths in a way that `github.com/bazelbuild/rules_go/go/runfiles` understands. Which is a little programmatic.

TristonianJones (approved):

PR #1002, opened 2024-08-14
Add syntax for escaped field selectors.

jnthntatum:
Introduces syntax for escaped fields that are not normally syntactically legal field selections with backticks.

This slightly relaxes the parse requirements for fields to cover more types of field like keys e.g. 

```
tags.`skewer-case-name` 
proto.`com.example.ext.field_name`
paths.`/api/v1`
```

This PR is limited to fields, but we could consider generalizing to cover namespaced names and simple identifier for variables in a later update (e.g. `in.example.Foo` or `package`)


TristonianJones (commented):
I like this as a first approach at relaxing the syntax to support escapes need for fields named after reserved words and for supporting proto extension fields natively. 

The type-checker should continue to validate field names which is great, and there's no weirdness with invalid function names, just escaping on the namespaces. It seems to help satisfy a common pain point. It might not solve top-level .in domains within namespace names, but it's a great start.

test/proto2pb/test_all_types.proto
```diff
@@ -112,7 +112,7 @@ message NestedTestAllTypes {
 // with fully qualified names.
 message ExampleType {
   optional string name = 1;
-
+  optional int64 in = 2;
```
TristonianJones, line 115: Can you make sure to update the canonical test protos with the same field name in a PR on cel-spec?

ext/protos_test.go
```diff
@@ -42,12 +42,21 @@ func TestProtos(t *testing.T) {
 		{expr: `proto.getExt(ExampleType{}, google.expr.proto2.test.nested_example) == ExampleType{}`},
 		{expr: `proto.getExt(ExampleType{}, google.expr.proto2.test.ExtendedExampleType.extended_examples) == []`},
 		{expr: `proto.getExt(ExampleType{}, google.expr.proto2.test.ExtendedExampleType.enum_ext) == GlobalEnum.GOO`},
+		{expr: "ExampleType{`in`: 64}.`in` == 64"},
+		// TODO(jnthntatum): legal parse, but can't assign extension fields without updates to type checker and runtime
```
TristonianJones: This is interesting. Are you thinking of doing this in a follow-up PR?

TristonianJones (dismissed):
LGTM. My questions which can be addressed post-submission.

test/proto2pb/test_all_types.proto
jnthntatum: Will do.

ext/protos_test.go
jnthntatum: Done. I added a reference, can you re-approve?

TristonianJones (approved):

PR #1004, opened 2024-08-16
Update to flatten list to have a stronger type-check signature

TristonianJones:


ext/lists.go
```diff
@@ -117,24 +119,35 @@ func (lib listsLib) CompileOptions() []cel.EnvOption {
 		opts = append(opts,
 			cel.Function("flatten",
 				cel.MemberOverload("list_flatten",
```
l46kok, line 121: should the name of the overload be `list_flatten_list` to include the arg?

ext/lists_test.go
```diff
@@ -36,8 +36,8 @@ func TestLists(t *testing.T) {
 		{expr: `[1,2,3,4].slice(0, 10)`, err: "cannot slice(0, 10), list is length 4"},
 		{expr: `[1,2,3,4].slice(-5, 10)`, err: "cannot slice(-5, 10), negative indexes not supported"},
 		{expr: `[1,2,3,4].slice(-5, -3)`, err: "cannot slice(-5, -3), negative indexes not supported"},
-		{expr: `[].flatten() == []`},
-		{expr: `[1,2,3,4].flatten() == [1,2,3,4]`},
+		{expr: `dyn([]).flatten() == []`},
+		{expr: `dyn([1,2,3,4]).flatten() == [1,2,3,4]`},
```
l46kok, line 40: From the usability standpoint, wouldn't we want flatten to behave like an identity function if it doesn't contain any nested list? requiring an extra wrapping via dyn seems odd

ext/lists.go
l46kok: Please disregard, this is correct as is.

l46kok (approved):

ext/lists_test.go
l46kok: We can come back to this. We'll need a better way to reason about type reduction, or implement union types. For now, `DisableTypeGuards` should be ok.

PR #1006, opened 2024-08-22
Support for context proto declarations

TristonianJones:
Introduce support for using a single context declaration

Context declarations take a protobuf message as input and treat all
top-level fields as variables. Only one context declaration is supported
at this time.


TristonianJones:
@jcking friendly ping

jcking (approved):

PR #1009, opened 2024-08-27
Allow configurable tag name overrides in native types

matthchr:
Allowing tags other than "cel" enables more flexibility, particularly when the types in question aren't owned by those configuring the CEL environment.

The main use-case for this is enabling an experience in CEL that matches the shape of an object in JSON (or YAML). This includes:
 - Property casing: 'Property' in Go would likely be 'property' in JSON/YAML.
 - Embedded structures: metav1.ObjectType `json:"metadata,omitempty"` is common on Kuberentes resources, allowing users to request CEL use the JSON tags enables access via the ".metadata" property even though there is no such property on the actual Native type.

## Reviews

* Perform a self-review.

## Merging

  if applicable.

### DIFF
--- a/ext/native.go
+++ b/ext/native.go
@@ -129,15 +129,31 @@ func NativeTypes(args ...any) cel.EnvOption {
 type NativeTypesOption func(*nativeTypeOptions) error
 
 type nativeTypeOptions struct {
-	// parseStructTags controls if CEL should support struct field renames, by parsing
-	// struct field tags.
-	parseStructTags bool
+	// structTagToParse controls if CEL should support struct field renames, by parsing
+	// struct field tags. This must be set to the tag to parse, such as "cel" or "json".
+	structTagToParse string
 }
 
 // ParseStructTags configures if native types field names should be overridable by CEL struct tags.
+// This is equivalent to ParseStructTag("cel")
 func ParseStructTags(enabled bool) NativeTypesOption {
 	return func(ntp *nativeTypeOptions) error {
-		ntp.parseStructTags = true
+		if enabled {
+			ntp.structTagToParse = "cel"
+		} else {
+			ntp.structTagToParse = ""
+		}
+		return nil
+	}
+}
+
+// ParseStructTag configures the struct tag to parse. The 0th item in the tag is used as the name of the CEL field.
+// For example:
+// If the tag to parse is "cel" and the struct field has tag cel:"foo", the CEL struct field will be "foo".
+// If the tag to parse is "json" and the struct field has tag json:"foo,omitempty", the CEL struct field will be "foo".
+func ParseStructTag(tag string) NativeTypesOption {
+	return func(ntp *nativeTypeOptions) error {
+		ntp.structTagToParse = tag
 		return nil
 	}
 }
@@ -147,7 +163,7 @@ func newNativeTypeProvider(tpOptions nativeTypeOptions, adapter types.Adapter, p
 	for _, refType := range refTypes {
 		switch rt := refType.(type) {
 		case reflect.Type:
-			result, err := newNativeTypes(tpOptions.parseStructTags, rt)
+			result, err := newNativeTypes(tpOptions.structTagToParse, rt)
 			if err != nil {
 				return nil, err
 			}
@@ -155,7 +171,7 @@ func newNativeTypeProvider(tpOptions nativeTypeOptions, adapter types.Adapter, p
 				nativeTypes[result[idx].TypeName()] = result[idx]
 			}
 		case reflect.Value:
-			result, err := newNativeTypes(tpOptions.parseStructTags, rt.Type())
+			result, err := newNativeTypes(tpOptions.structTagToParse, rt.Type())
 			if err != nil {
 				return nil, err
 			}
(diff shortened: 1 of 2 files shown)


ext/native.go
```diff
@@ -273,12 +275,12 @@ func (tp *nativeTypeProvider) FindStructFieldType(typeName, fieldName string) (*
 		Type: celType,
 		IsSet: func(obj any) bool {
 			refVal := reflect.Indirect(reflect.ValueOf(obj))
-			refField := valueFieldByName(tp.options.parseStructTags, refVal, fieldName)
```
matthchr, line 276: `valueFieldByName` is (I think) not needed, since we've already found the structTag to field mapping above via:
`refField, isDefined := t.hasField(fieldName)` we can just use `refField.Name` to do a direct lookup without needing to re-parse tags.

TristonianJones:
Hi @matthchr, thanks for the contribution. Would you mind signing the CLA? I'll try to give this a review by the end of the week. 

/gcbrun

matthchr:
I've signed the CLA. Let me know if you have any questions as well - I'm happy to talk through our use-case. I had originally done a different approach, something like this:
```
type NativeFieldAdapter = func(rt reflect.Type, name string) (reflect.StructField, bool)

func defaultNativeFieldAdapter(rt reflect.Type, name string) (reflect.StructField, bool) {
	return rt.FieldByName(name)
}
```

that users could configure, before I realized that actually all I really needed was to be able to configure the tag to parse for the feature that already existed, because what I wanted to do was use JSON tags rather than a new CEL tag.

ext/native.go
TristonianJones: That makes sense. Thanks.

ext/native.go
```diff
@@ -208,13 +221,18 @@ func (tp *nativeTypeProvider) FindStructType(typeName string) (*types.Type, bool
...
 		return f.Name
 	}
 
-	if name, found := f.Tag.Lookup("cel"); found {
-		return name
+	tag, found := f.Tag.Lookup(structTagToParse)
+	if found {
+		splits := strings.Split(tag, ",")
+		if len(splits) > 0 {
+			name := splits[0]
```
TristonianJones, line 242: @matthchr Are we sure the name will always be in the first ordinal position on the tag split?

ext/native_test.go
```diff
@@ -750,7 +816,7 @@ func TestNativeTypesWithOptional(t *testing.T) {
 }
 
 func TestNativeTypeConvertToType(t *testing.T) {
-	nt, err := newNativeType(true, reflect.TypeOf(&TestAllTypes{}))
+	nt, err := newNativeType("cel", reflect.TypeOf(&TestAllTypes{}))
```
TristonianJones: Would you mind making these parameterized tests for the known tag types: `cel`, `json`?

ext/native.go
matthchr: Good question. I've added the following comment:

```
			// We make the assumption that the leftmost entry in the tag is the name.
			// This seems to be true for most tags that have the concept of a name/key, such as:
			// https://pkg.go.dev/encoding/xml#Marshal
			// https://pkg.go.dev/encoding/json#Marshal
			// https://pkg.go.dev/go.mongodb.org/mongo-driver/bson#hdr-Structs
			// https://pkg.go.dev/gopkg.in/yaml.v2#Marshal
```

As far as I can tell, the convention seems to be that the 0th entry is the name/key for tags that _have_ a name/key. Some don't have such a concept, such as https://pkg.go.dev/encoding/asn1, and using this option with `asn1` wouldn't make any sense because of that.

I quickly went through the list here: https://go.dev/wiki/Well-known-struct-tags - the one that jumped out at me as not following this rule was protobuf, which seems to do:
`protobuf:"bytes,1,opt,name=something,proto3"` (at least for proto3?), where the name could presumably be anywhere as long as it's annotated with the `name=` key within the tag.

I suppose if we really wanted to cater to cases like this, we could add to the `ParseStructTag(tag string)` option to look something like:
`func ParseStructTagDetailed(tag string, parser func(tag string) string) NativeTypesOption` where the user can define a tag-parsing function that could work for any arbitrary tag.

That feels like maybe an addition left to a later time if it's actually needed though, as it could be done in a nonbreaking way and the vast majority of tags (json, yaml, etc) follow the 0th-item rule already.

ext/native_test.go
matthchr: Done

ext/native.go
TristonianJones: Proto is handled as a special-case already, so I'm not too worried about it. However, let's document that the name is expected to be the 0th entry in the tag as part of the CEL option.

matthchr: > However, let's document that the name is expected to be the 0th entry in the tag as part of the CEL option.

Done. `ParseStructTag` now says:
```
// ParseStructTag configures the struct tag to parse. The 0th item in the tag is used as the name of the CEL field.
// For example: 
// If the tag to parse is "cel" and the struct field has tag cel:"foo", the CEL struct field will be "foo".
// If the tag to parse is "json" and the struct field has tag json:"foo,omitempty", the CEL struct field will be "foo".
```

TristonianJones (approved):

matthchr:
Thanks for merging!

PR #1021, opened 2024-09-12
ext: add list.sort()

cezar-guimaraes:
I skipped creating an issue because this seemed like a small enough change to me, but please let me know if an issue is preferable.

This PR adds a `<list(T)>.sort() -> <list(T)>` member. T must implement `(traits.Comparer)`.

I believe that the check if the element type implements `traits.Comparer` only happens at runtime (how can I write tests for this?). An alternative implementation would create separate member overloads for `list(T), T in {set of comparable types}` similar to kube [apiserver list extensions](https://github.com/kubernetes/apiextensions-apiserver/blob/v0.25.16/pkg/apiserver/schema/cel/library/lists.go#L135)

Let me know if I should update the PR to create separate overloads.

### DIFF
--- a/ext/README.md
+++ b/ext/README.md
@@ -405,6 +405,19 @@ Examples:
     [1,2,3,4].slice(1, 3) // return [2, 3]
     [1,2,3,4].slice(2, 4) // return [3 ,4]
 
+### Sort
+
+Sorts a list with comparable elements. If the element type is not comparable
+
+    <list(T)>.sort() -> <list(T)>
+
+Examples:
+
+    [3, 2, 1].sort() // return [1, 2, 3]
+    ["b", "c", "a"].sort() // return ["a", "b", "c"]
+    [1, "b"].sort() // error
+    [[1, 2, 3]].sort() // error
+
 ## Sets
 
 Sets provides set relationship tests.
--- a/ext/README.md
+++ b/ext/README.md
@@ -408,8 +408,10 @@ Examples:
 ### Sort
 
 Sorts a list with comparable elements. If the element type is not comparable
+or the element types are not the same, the function will produce an error.
 
     <list(T)>.sort() -> <list(T)>
+    T in {int, uint, double, bool, duration, timestamp, string, bytes}
 
 Examples:
 
--- a/ext/README.md
+++ b/ext/README.md
@@ -407,6 +407,8 @@ Examples:
 
 ### Sort
 
+**Introduced in version 2**
+
 Sorts a list with comparable elements. If the element type is not comparable
 or the element types are not the same, the function will produce an error.
 
--- a/ext/lists.go
+++ b/ext/lists.go
@@ -189,11 +189,15 @@ func (lib listsLib) CompileOptions() []cel.EnvOption {
 	}
 	if lib.version >= 2 {
 		sortDecl := cel.Function("sort",
-			templatedOverloads(comparableTypes, func(t *cel.Type) cel.FunctionOpt {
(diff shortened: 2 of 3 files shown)


ext/lists.go
```diff
@@ -159,6 +174,27 @@ func (lib listsLib) CompileOptions() []cel.EnvOption {
 			),
 		)
 	}
+	if lib.version >= 2 {
+		opts = append(opts,
+			cel.Function("sort",
+				cel.MemberOverload("list_sort",
+					[]*cel.Type{listType}, listType,
```
TristonianJones: A slight refinement on this approach would be to create a singleton binding for all comparable types: `bool`, `bytes`, `int`, `uint`, `string`, `timestamp`, `duration`. I realize it's a bit annoying but it will really help with ensuring the function is only used with comparable types.

cezar-guimaraes: I agree! Added in 802efa851da23e7c4f05678e5029fc1e235b393c

cezar-guimaraes: I actually created MemberOverloads for `list(int), list(uint), ...` because I couldn't find out how to constrain the element type must be comparable via a `cel.SingletonBinding`, only that the operand `list` must implement traits.Lister. Let me know if there is a better way

ext/lists.go
```diff
@@ -159,6 +187,29 @@ func (lib listsLib) CompileOptions() []cel.EnvOption {
 			),
 		)
 	}
+	if lib.version >= 2 {
+		sortDecl := cel.Function("sort",
+			templatedOverloads(comparableTypes, func(t *cel.Type) cel.FunctionOpt {
+				return cel.MemberOverload(
+					fmt.Sprintf("list_%s_sort", t.TypeName()),
+					[]*cel.Type{cel.ListType(t)}, cel.ListType(t),
+					cel.UnaryBinding(func(arg ref.Val) ref.Val {
```
TristonianJones: I think it's possible to introduce this function as a `cel.SingletonUnaryBinding` since the implementation is not type-specific. This allows people to add new signatures for comparable types without having to generate new functions to take advantage of `sort()`

TristonianJones:
@cezar-guimaraes I just had one more comment, but otherwise this is looking good.

ext/lists.go
cezar-guimaraes: Got it. Can you check again? I added all list traits since adding only the necessary ones (`indexer` and `sizer`) would also match Map

ext/lists.go
TristonianJones: I see your point, I added helper type traits for list and maps in #1026 to at least better clarify intent. The check you're doing now is quite reasonable.

ext/README.md
```diff
@@ -405,6 +405,21 @@ Examples:
     [1,2,3,4].slice(1, 3) // return [2, 3]
     [1,2,3,4].slice(2, 4) // return [3 ,4]
 
+### Sort
+
+Sorts a list with comparable elements. If the element type is not comparable
```
TristonianJones, line 412: I really appreciate that you added version checks for lists. Can you make sure to indicate which version this feature was introduced in the `.md` and `.go` files?

cezar-guimaraes: Done

cezar-guimaraes:
Rebased the branch to use the new helper type introduced in #1026
1ad695c

TristonianJones (approved):

TristonianJones:
@cezar-guimaraes thanks for your contribution and your patience!

PR #1037, opened 2024-10-04
Add functions to the lists extension.

seirl:
These list functions are used in an internal Google project, but are useful enough that they can be upstreamed to the Go implementation, and later to other runtimes.

The new functions are:

- `range()` to return a list of integers from 0 to n-1.
- `reverse()` to return the reverse of a list.
- `distinct()` to return the unique elements in a list.
- `sortByAssociatedKeys()` to sort a list by a list of keys, used to implement the `sortBy()` macro.

The new macros are:

- `sortBy()`, used like: `<listExpr>.sortBy(<elementBinding>, <keyExpression>)` to sort a list by a "key expression" applied to each element, similar to `l.sort(key=lambda e: keyExpression)` in Python.


ext/README.md
```diff
@@ -393,6 +393,68 @@ Example:
...
+Returns the distinct elements of a list. If the element type is not comparable
+or the element types are not the same, the function will produce an error.
+
+	<list(T)>.distinct() -> <list(T)>
+	T in {int, uint, double, bool, duration, timestamp, string, bytes}
+
+Examples:
+
+	[1, 2, 2, 3, 3, 3].distinct() // return [1, 2, 3]
+	["b", "c", "a"].sort() // return ["a", "b", "c"]
```
TristonianJones: I think maybe you meant this example to be different?

ext/README.md
```diff
@@ -393,6 +393,68 @@ Example:
...
+	[1, "b"].distinct() // error
+	[[1, 2, 3]].distinct() // error
```
TristonianJones: Are these errors type-check errors or runtime errors? If they are runtime errors, then couldn't the values be determined via CEL equality? In general, distinct should work across all types. The one wrinkle is really numeric equality which is heterogeneous. For example, users should probably expect the list to be preserved, so this might be the one exception to the `distinct` in terms of equality rule: `[0u, 0.0, 0].distinct()`

ext/lists.go
```diff
@@ -39,6 +41,48 @@ var comparableTypes = []*cel.Type{
...
+//
+// Returns a list of integers from 0 to n-1.
+//
+//	range(<int>) -> <list(int)>
+//
+// Examples:
+//
+//	range(5) -> [0, 1, 2, 3, 4]
+//
+// ## Reverse
```
TristonianJones: nit: prefer single hash mark

ext/README.md
```diff
@@ -393,6 +393,68 @@ Example:
...
+
+### Flatten
+
+**Introduced in version 2**
```
TristonianJones: I think `flatten` was introduced into version 1

ext/README.md
```diff
@@ -393,6 +393,68 @@ Example:
...
+    [1,[2,[3,[4]]]].flatten(2) // return [1, 2, 3, [4]]
+    [1,[2,[3,[4]]]].flatten(-1) // error
+
+### Range
+
+**Introduced in version 2**
+
+Returns a list of integers from 0 to n-1.
+
+	range(<int>) -> <list(int)>
```
TristonianJones: Consider using a namespace `lists.range` to better connote where this feature comes from.

ext/lists.go
```diff
@@ -83,7 +127,35 @@ var comparableTypes = []*cel.Type{
...
+// Examples:
+//
+//	[
+//	  Player { name: "foo", score: 0 },
+//	  Player { name: "bar", score: -10 },
+//	  Player { name: "baz", score: 1000 },
+//	].sortBy(e, e.score).map(e, e.name)
+//	== ["bar", "foo", "baz"]
+//
+// # SortByAssociatedKeys
```
TristonianJones: Could you help me understand the use case for this? This seems like an implementation detail of the `sortBy` macro.

ext/lists.go
```diff
@@ -276,29 +450,122 @@ func flatten(list traits.Lister, depth int64) ([]ref.Val, error) {
 }
 
 func sortList(list traits.Lister) (ref.Val, error) {
+	return sortListByAssociatedKeys(list, list)
+}
+
+func sortListByAssociatedKeys(list traits.Lister, keys traits.Lister) (ref.Val, error) {
```
TristonianJones: nit: `list, keys traits.Lister`

ext/lists.go
```diff
@@ -276,29 +450,122 @@ func flatten(list traits.Lister, depth int64) ([]ref.Val, error) {
...
+	}
+
+	sort.Slice(sortedIndices, func(i, j int) bool {
+		iKey := keys.Get(sortedIndices[i])
+		jKey := keys.Get(sortedIndices[j])
+		return iKey.(traits.Comparer).Compare(jKey) == types.IntNegOne
+	})
+
 	sorted := make([]ref.Val, 0, listLength)
+	for i := types.IntZero; i < listLength; i++ {
```
TristonianJones: Consider making this a range:

```go
for _, sortedIdx := range sortedIndices {
  sorted = append(sorted, list.Get(sortedIdx))
}
```

ext/lists.go
```diff
@@ -276,29 +450,122 @@ func flatten(list traits.Lister, depth int64) ([]ref.Val, error) {
...
+//
+// into something equivalent to:
+//
+//	 cel.bind(
+//		  __sortBy_input__,
+//		  myListExpr,
+//		  __sortBy_input__.sortByAssociatedKeys(__sortBy_input__.map(e, -math.abs(e))
+//		)
+func sortByMacro(meh cel.MacroExprFactory, target ast.Expr, args []ast.Expr) (ast.Expr, *cel.Error) {
+	varIdent := meh.NewIdent("__sortBy_input__")
```
TristonianJones: Consider using `@` as a prefix to indicate the variable is not intended to be accessed.

ext/lists.go
```diff
@@ -276,29 +450,122 @@ func flatten(list traits.Lister, depth int64) ([]ref.Val, error) {
...
+		mapCompr,
+	)
+
+	bindExpr := meh.NewComprehension(
+		meh.NewList(),
+		"#unused",
+		varName,
+		target,
+		meh.NewLiteral(types.False),
+		varIdent,
```
TristonianJones, line 520: Prefer making a copy of `varIdent`, using `ast.CopyExpr`

ext/lists.go
```diff
@@ -276,29 +450,122 @@ func flatten(list traits.Lister, depth int64) ([]ref.Val, error) {
...
+		targetKind != ast.ComprehensionKind && targetKind != ast.CallKind {
+		return nil, meh.NewError(target.ID(), fmt.Sprintf("sortBy can only be applied to a list, identifier, comprehension, call or select expression"))
+	}
+
+	mapCompr, err := parser.MakeMap(meh, varIdent, args)
+	if err != nil {
+		return nil, err
+	}
+	callExpr := meh.NewMemberCall("sortByAssociatedKeys",
+		varIdent,
```
TristonianJones: Prefer making a copy of `varIdent`

ext/lists.go
```diff
@@ -276,29 +450,122 @@ func flatten(list traits.Lister, depth int64) ([]ref.Val, error) {
...
+			seen[val] = true
+		}
 	}
 
-	sort.Slice(sorted, func(i, j int) bool {
-		return sorted[i].(traits.Comparer).Compare(sorted[j]) == types.IntNegOne
+	// Sort for determinism across implementations.
+	sort.Slice(uniqueList, func(i, j int) bool {
+		return uniqueList[i].(traits.Comparer).Compare(uniqueList[j]) == types.IntNegOne
 	})
```
TristonianJones: I would prefer if we left the distinct check as unsorted by default, but rather treated as an O(n^2) insertion based on equality. In SQL, a `SELECT DISTINCT` does not have ordering implications, but rather preserves the record order as it was presented.

TristonianJones (changes requested):

ext/lists.go
seirl: I think we can have it as an implementation detail for now (prefixed with @ to show that the function is internal), and if there is a need for it later on we can always expose it.

seirl:
@TristonianJones Addressed all the comments, PTAL :slightly_smiling_face:

ext/README.md
TristonianJones: Could you help me understand the reason for not using something like CEL equality for `distinct` here in order to support more complex data types within `distinct`?

ext/README.md
```diff
@@ -422,6 +482,22 @@ Examples:
     [1, "b"].sort() // error
     [[1, 2, 3]].sort() // error
 
+### SortBy
+
+**Introduced in version 2**
+
+Sorts a list by a key value, i.e., the order is determined by the result of
+an expression applied to each element of the list.
```
TristonianJones, line 489: Nit: could you add the function signature of `sortBy` here for reference?

ext/lists.go
```diff
@@ -277,31 +425,124 @@ func flatten(list traits.Lister, depth int64) ([]ref.Val, error) {
...
+	return bindExpr, nil
+}
+
+func distinctList(list traits.Lister) (ref.Val, error) {
+	listLength := list.Size().(types.Int)
+	if listLength == 0 {
+		return list, nil
+	}
+	uniqueList := make([]ref.Val, 0, listLength)
+	seen := make(map[ref.Val]bool, listLength)
```
TristonianJones: It won't be very fast, but I think it's more accurate to write the following since it will account for the numeric case mentioned and support any generic data type that CEL supports:

```go
for i := types.IntZero; i < listLength; i++ {
    val := list.Get(i)
    seen := false
    for j := types.IntZero; j < types.Int(len(uniqueList)); j++ {
      if i = j {
         continue
      } 
      other := uniqueList[i]
      if val.Equal(other) == types.True {
        seen = true
        break
      }
    }
    if !seen {
      uniqueList = append(uniqueList, val)  
    }
}
```

Technically CEL should error if you try to do something like this `{1u: 'value', 1:'other value'}` and the two-variable comprehensions absolutely will error under those circumstances. We should commit to strict `Equal` checks and live with the potential edge case as we do for map construction. It's not as fast, and maybe you'll want to optimize the distinct check if you can infer that the environment supports homogenous aggregate literals and that the element type can be expected to be consistent and primitive (note, bytes won't work with map keys, so it's a little stricter than this).

seirl: Done and added a test for the `[1, 1.0]` case.

seirl:
PTAL. Also can you confirm that you're okay with the naming for the internal sortBy function (`.@sortByAssociatedKeys`)?

ext/README.md
TristonianJones: Can you update the docs here to match the implementation? Also perhaps add tests related to distinct with a list of lists?

ext/lists_test.go
```diff
@@ -36,15 +43,30 @@ func TestLists(t *testing.T) {
...
+		{expr: `[-3, 1, -5, -2, 4].sortBy(e, -(e * e)) == [-5, 4, -3, -2, 1]`},
+		{expr: `[-3, 1, -5, -2, 4].map(e, e * 2).sortBy(e, -(e * e)) == [-10, 8, -6, -4, 2]`},
+		{expr: `lists.range(3).sortBy(e, -e) == [2, 1, 0]`},
+		{expr: `["a", "c", "b", "first"].sortBy(e, e == "first" ? "" : e) == ["first", "a", "b", "c"]`},
+		{expr: `[].distinct() == []`},
+		{expr: `[1].distinct() == [1]`},
+		{expr: `[-2, 5, -2, 1, 1, 5, -2, 1].distinct() == [-2, 5, 1]`},
+		{expr: `['c', 'a', 'a', 'b', 'a', 'b', 'c', 'c'].distinct() == ['c', 'a', 'b']`},
+		{expr: `[1, 2.0, "c", 3, "c", 1].distinct() == [1, 2.0, "c", 3]`},
+		{expr: `[1, 1.0, 2].distinct() == [1, 2]`},
```
TristonianJones, line 71: Can you add some more tests for `distinct` with complex types, like lists of lists and protobufs?

TristonianJones:
Looking pretty good, and the naming of the internal sorting function is good for me. Thanks!

ext/lists.go
```diff
@@ -218,6 +277,77 @@ func (lib listsLib) CompileOptions() []cel.EnvOption {
...
+						return types.WrapErr(err)
+					}
+					return result
+				}),
+			),
+		))
+		opts = append(opts, cel.Function("reverse",
+			cel.MemberOverload("list_reverse",
+				[]*cel.Type{listType}, listType,
+				cel.FunctionBinding(func(args ...ref.Val) ref.Val {
```
TristonianJones, line 328: nit, this can be a `UnaryBinding`

ext/lists.go
```diff
@@ -218,6 +277,77 @@ func (lib listsLib) CompileOptions() []cel.EnvOption {
...
+						return types.WrapErr(err)
+					}
+					return result
+				}),
+			),
+		))
+		opts = append(opts, cel.Function("distinct",
+			cel.MemberOverload("list_distinct",
+				[]*cel.Type{listType}, listType,
+				cel.FunctionBinding(func(args ...ref.Val) ref.Val {
```
TristonianJones: nit, this can be a `UnaryBinding`

TristonianJones:
@seirl looking good, but we'll need to sync #1045 to fix the BUILD error

seirl:
Done!

TristonianJones (approved):

PR #1039, opened 2024-10-14
Update tag-based parsing to use lambda for additional customization

matthchr:
This allows callers to override a bit more about struct field parsing. In particular, it's useful when parsing tags to perform escaping of tags that aren't supported CEL field names.
This allows implementation of something similar to what Kuberentes does with CEL field names:
https://kubernetes.io/docs/reference/using-api/cel/#escaping.

# Pull Requests Guidelines

development branch.

## Commit Messages

```

```

## Reviews

* Perform a self-review.

## Merging

  if applicable.

### DIFF
--- a/ext/native.go
+++ b/ext/native.go
@@ -128,10 +128,36 @@ func NativeTypes(args ...any) cel.EnvOption {
 // NativeTypesOption is a functional interface for configuring handling of native types.
 type NativeTypesOption func(*nativeTypeOptions) error
 
+// NativeTypesFieldNameHandler is a handler for mapping a reflect.StructField to a CEL field name.
+// This can be used to override the default Go struct field to CEL field name mapping.
+type NativeTypesFieldNameHandler = func(field reflect.StructField) string
+
+func fieldNameByTag(structTagToParse string) func(field reflect.StructField) string {
+	return func(field reflect.StructField) string {
+		tag, found := field.Tag.Lookup(structTagToParse)
+		if found {
+			splits := strings.Split(tag, ",")
+			if len(splits) > 0 {
+				// We make the assumption that the leftmost entry in the tag is the name.
+				// This seems to be true for most tags that have the concept of a name/key, such as:
+				// https://pkg.go.dev/encoding/xml#Marshal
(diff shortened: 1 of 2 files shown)


ext/native.go
```diff
@@ -128,20 +128,46 @@ func NativeTypes(args ...any) cel.EnvOption {
 // NativeTypesOption is a functional interface for configuring handling of native types.
 type NativeTypesOption func(*nativeTypeOptions) error
 
+// NativeTypesFieldNameHandler is a handler for mapping a reflect.StructField to a CEL field name.
+// This can be used to override the default Go struct field to CEL field name mapping.
+type NativeTypesFieldNameHandler = func(field reflect.StructField) string
+
+func fieldNameByTag(structTagToParse string) func(field reflect.StructField) string {
```
matthchr, line 135: Should this be exported? If it was, we could possibly just drop the `ParseStructTag` option below as it was added recently and there hasn't been a tagged cel-go version since its addition.

Callers who wanted pure tags parsing could then do `ext.ParseStructField(ext.FieldNameByTag)` to achieve the same thing, or they could customize the field parsing further by doing `ext.ParseStructField(myFunc)`.

I didn't do this for now mostly because it means deleting the added `ParseStructTags`. I think there should be no users of this yet though as no new tag published yet...

If the maintainers prefer me to do this just say so and I can update the PR to do so.

matthchr:
Realized this might be somewhat related to https://github.com/google/cel-go/pull/1002

matthchr:
My main use-case for this is to control escaping of tags that aren't valid CEL names, _but_ there are probably other viable use-cases as well like conditionally adding field name suffixes/prefixes or similar. Felt like rather than try to encode everything in the native types handler just exposing a function that enabled the caller to configure it how they want made sense.

ext/native.go
TristonianJones: Specifying a string tag name is likely going to be most common. For more esoteric parsing, I don't mind supporting a separate option that accepts a functional handler as you've set it up currently.

TristonianJones (approved):

PR #1048, opened 2024-10-21
Runtime support for cel.@block

TristonianJones:
Introduces runtimes support for `cel.@block`

CEL block is used by static optimizers to group subexpressions together
such that the following is true:

* Each subexpression value may be referenced by its ordinal through the 
  variable: `@index<ordinal>`
* Subexpressions must be organized such that the expression at `@index<N>` 
   may only depend on variables represented in `@index<N-1>`
* Index expressions may not occur within a comprehension loop since each
   comprehension represents a scope which declares its own variables.

Depends on https://github.com/google/cel-spec/pull/402 before submission


ext/bindings.go
```diff
@@ -94,3 +157,74 @@ func celBind(mef cel.MacroExprFactory, target ast.Expr, args []ast.Expr) (ast.Ex
...
+	bs.slotActivationPool.Put(sa)
+}
+
+type slotActivation struct {
+	interpreter.Activation
+	slotExprs []interpreter.Interpretable
+	slotVals  []ref.Val
+}
+
+func (sa *slotActivation) ResolveName(name string) (any, bool) {
```
jcking: As discussed offline, we should add a recursive check here to ensure the sub expression we are evaluating is not the slot being resolved, to prevent bad expressions from infinite looping and killing the stack.

TristonianJones: Done, PTAL

ext/bindings.go
```diff
@@ -94,3 +189,148 @@ func celBind(mef cel.MacroExprFactory, target ast.Expr, args []ast.Expr) (ast.Ex
...
+type slotVal struct {
+	value   *ref.Val
+	visited bool
+}
+
+type dynamicSlotActivation struct {
+	interpreter.Activation
+	slotExprs []interpreter.Interpretable
+	slotCount int
+	slotVals  []*slotVal
```
jcking, line 248: I personally would just avoid the pointer here.

TristonianJones: The pointer ends up being pretty convenient since it means that I can update the `slotVal` state without having to set the slot index multiple times.

TristonianJones: Any other concerns? I've written it both ways and the visit logic is easier to follow with the pointers. Object pooling generally solves any overhead concerns. See #1056

TristonianJones: I wound up running tests against the pooled allocations versus stack-based allocs and overall the pointers with pooled memory were a 5% win, so I'm sticking with that for now and just merged the policy composition support. Happy to come back to this as a follow up if desired.

PR #1052, opened 2024-10-25
Fix out of range error for non-negative string indexing offsets

### LINKED ISSUE #1051: IndexOf in the ext pkg returns an error when it runs on an empty string
**Describe the bug**

`"".indexOf("foo")` returns `index out of range: 0` instead of `-1` when it runs against an empty string

**To Reproduce**

- [ ]  parser
- [ ]  checker
- [x]  interpreter

```cel
"".indexOf("foo")
```

Test setup:
```go
env, err := cel.NewEnv(ext.Strings())
if err != nil {
	t.Fatalf("Failed to create CEL env: %v", err)
}

celAST, compileIssues := env.Compile(`"".indexOf("foo")`)
assert.NotNil(t, celAST)
assert.Equal(t, "", compileIssues.String())

checkedExpr, err := cel.AstToCheckedExpr(celAST)
assert.Nil(t, err)
assert.NotNil(t, checkedExpr)

program, err := env.Program(celAST)
assert.NoError(t, err)
assert.NotNil(t, program)

_, _, err = program.Eval(map[string]interface{}{})
assert.NoError(t, err)
```

**Expected behavior**
Returns -1 as it does when it does not find the substring

**Additional context**
We use optionals, and we often do `myValue.orValue("")` to run conditionals on strings, but here doing `myValue.orValue("").indexOf` throws an error unexpectedly

TristonianJones:
Closes #1051

### DIFF
--- a/ext/strings.go
+++ b/ext/strings.go
@@ -561,8 +561,12 @@ func indexOfOffset(str, substr string, offset int64) (int64, error) {
 	off := int(offset)
 	runes := []rune(str)
 	subrunes := []rune(substr)
-	if off < 0 || off >= len(runes) {
+	if off < 0 {
 		return -1, fmt.Errorf("index out of range: %d", off)
+	}
+	// If the offset exceeds the length, return -1 rather than error.
+	if off >= len(runes) {
+		return -1, nil
 	}
 	for i := off; i < len(runes)-(len(subrunes)-1); i++ {
 		found := true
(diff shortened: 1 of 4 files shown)


ext/strings.go
jnthntatum, line 1: Can you also update the documentation (e.g. line 122)

jnthntatum (dismissed):

ext/strings.go
TristonianJones: Done. PTAL

jnthntatum (approved):

PR #1067, opened 2024-11-13
First and last

bigkevmcd:
# Pull Requests Guidelines

development branch.

## Commit Messages

```
Add support for getting the first and last items from a list.

This adds support for doing something like `list.first()` and `list.last()` which return the first and last elements of a list respectively.

`.first()` is syntactic sugar, but `.last()` makes it simpler to get the last element from a list without having to compute the length of the list twice.

This can shorten a long expression instead of `nested.elements[nested.elements.size()-1]` you can rewrite this as `nested.elements.last().value()`.

Use of the optional functionality requires the registration of the `cel.OptionalTypes()` env option.
```

## Reviews

* Perform a self-review.

## Merging

  if applicable.


seirl:
Alternative suggestion, maybe supporting the negative indexing semantics from python (-1 == last, -2 == first before last...) would be another option. But it wouldn't have the optional return value when the list is empty, which can be nice too.

TristonianJones:
@bigkevmcd you raise an interesting point regarding optional. I think my concern with having either `null_type` or `optional_type(T)` as a return type is that at first glance users won't know which it is. You can specify a `cel.Lib` as an `cel.EnvOption`, so it's possible to do any of the following things:

1. Move `first` and `last` into another version of `cel.OptionalTypes()`
1. Auto-enable optionals by providing `cel.OptionalTypes()` as an element in the `CompileOptions()` response
1. Test whether optional types are enabled before adding the method signature

I think option 1 is the cleanest since the functions would be unambiguous in their signature and encourage the use of optional values. Option 2 runs into unexpected complexity being enabled, and option 3 runs into problems where `cel.EnvOption` order impacts which features are enabled (this is already a challenge with `cel.CustomTypeProvider`, and `cel.ClearMacros`)

seirl:
> Alternative suggestion, maybe supporting the negative indexing semantics from python (-1 == last, -2 == first before last...) would be another option. But it wouldn't have the optional return value when the list is empty, which can be nice too.

I remembered `_[?_]`. So with negative indexing you could have:

- `list.first()` -> `list[?0]`
- `list.last()` -> `list[?-1]`

TristonianJones:
@seirl while the use of a negative index would certainly be very clean, I'm worried that users who have implemented custom CEL lists will not be able to support the negative index without a rewrite of the expression to `list.size() > 0 : optional.of(list[list.size()-1]) : optional.none()`

seirl:
OK, I didn't realize this would be an issue, I expected that the negative index translation would happen before the custom list implementation.

I've also seen people use `l.reverse()[?0]` but it makes a copy of the list.

cel/library.go
```diff
@@ -260,6 +260,27 @@ func (stdLibrary) ProgramOptions() []ProgramOption {
...
+// [1, 2, 3].first().value() == 1
+
+// # Last
+//
+// Introduced in version: 2
+//
+// Returns an optional with the last value from the right hand list, or
+// optional.None.
+//
+// [1, 2, 3].last().value() == 3
```
TristonianJones, line 280: Would you mind adding examples of empty lists for `first()` and `last()`? Also, don't forget to update the `README.md` since often folks look there for high-level docs within the extensions folder. Thanks!

cel/library.go
```diff
@@ -260,6 +260,27 @@ func (stdLibrary) ProgramOptions() []ProgramOption {
...
+//
+// This is syntactic sugar for msg.elements[msg.elements.size()-1].
+
```
TristonianJones, line 283: nit: extra blank line

cel/library.go
bigkevmcd: @TristonianJones Add this to `ext/README.md` ?

TristonianJones: Yes, thank you for clarifying, the `ext/README.md` is the one I meant.

TristonianJones (approved):

TristonianJones:
@bigkevmcd, thanks for the contribution!

PR #1075, opened 2024-11-21
Introduce versioning options to all extensions

TristonianJones:
Whether extensions support versioning or not, all extensions now support a version option
in order to remain insulated from versioned features added at a later date.


l46kok (approved):

jpbetz:
LGTM. Thanks for doing this!

PR #1083, opened 2024-12-04
Fix two-variable comprehension pruning

### LINKED ISSUE #1080: Two variable comprehensions potentially getting lost during pruning
**Describe the bug**

I am not quite sure whether this is a bug or not. I was reading the source to track down a different problem and came across this, which seems odd to me.

https://github.com/google/cel-go/blob/5d18e93d204801cbe02158a0ddb3a3abce29e4ed/interpreter/prune.go#L418-L433

Shouldn't it check `compre.IterVar2` and return a `p.NewComprehensionTwoVar(...)` if it's not empty?

Similarly, shouldn't there be a check for a two variable comprehension in the following bit as well?

https://github.com/google/cel-go/blob/5d18e93d204801cbe02158a0ddb3a3abce29e4ed/interpreter/prune.go#L284-L287

TristonianJones:
Added support for two-variable comprehension prune / unparse

Closes #1080

### DIFF
--- a/ext/comprehensions_test.go
+++ b/ext/comprehensions_test.go
@@ -359,6 +359,62 @@ func TestTwoVarComprehensionsVersion(t *testing.T) {
 	}
 }
 
+func TestTwoVarComprehensionsUnparse(t *testing.T) {
+	tests := []struct {
+		name     string
+		expr     string
+		unparsed string
+	}{
+		{
+			name:     "transform map entry",
+			expr:     `[0, 0u].transformMapEntry(i, v, {v: i})`,
+			unparsed: `[0, 0u].transformMapEntry(i, v, {v: i})`,
+		},
+		{
+			name:     "transform map",
+			expr:     `{'a': 'world', 'b': 'hello'}.transformMap(i, v, i == 'a' ? v.upperAscii() : v)`,
+			unparsed: `{"a": "world", "b": "hello"}.transformMap(i, v, (i == "a") ? v.upperAscii() : v)`,
+		},
+		{
+			name:     "transform list",
+			expr:     `[1.0, 2.0, 2.0].transformList(i, v, i / 2.0 == 1.0)`,
+			unparsed: `[1.0, 2.0, 2.0].transformList(i, v, i / 2.0 == 1.0)`,
+		},
+		{
+			name:     "existsOne",
+			expr:     `{'a': 'b', 'c': 'd'}.existsOne(k, v, k == 'b' || v == 'b')`,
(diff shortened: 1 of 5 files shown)


ext/comprehensions_test.go
```diff
@@ -345,6 +345,62 @@ func TestTwoVarComprehensionsRuntimeErrors(t *testing.T) {
 	}
 }
 
+func TestTwoVarComprehensionsUnparse(t *testing.T) {
```
jnthntatum, line 364: could you update this test to include a PruneAST example that would have failed?

TristonianJones: I think all of these tests would have failed since the second variable would have been omitted. Happy to add more cases if you have some in mind

TristonianJones: Oh, of course, I should actually add tests for pruning. Sorry about that

TristonianJones: PTAL, the change ended up being much more invasive than I had originally anticipated.

interpreter/prune.go
```diff
@@ -281,13 +281,35 @@ func (p *astPruner) prune(node ast.Expr) (ast.Expr, bool) {
...
 		if node.Kind() == ast.ComprehensionKind {
 			compre := node.AsComprehension()
 			visit(macro, clearIterVarVisitor(compre.IterVar(), p.state))
+			if compre.HasIterVar2() {
+				visit(macro, clearIterVarVisitor(compre.IterVar2(), p.state))
+			}
+			loopCond := compre.LoopCondition()
+			iterRange := compre.IterRange()
+			pruneMacroCall =
+				(loopCond.Kind() == ast.LiteralKind && loopCond.AsLiteral() == types.False) &&
```
jnthntatum: nit/clarity -- could you add a comment or add a predicate function for this test (checking for trivial comprehension)? does this also apply to .optMap?

jnthntatum (changes requested):

interpreter/prune.go
TristonianJones: Perhaps I should just restrict the pruning to `cel.bind` calls only since `optMap` although not wrong would look very strange. Thoughts?

jnthntatum: Just testing for cel.bind is easier to follow imo, but up to you.

jnthntatum (dismissed):

interpreter/prune.go
TristonianJones: Done, PTAL.

jnthntatum (approved):

PR #1089, opened 2024-12-06
Add support for typed conformance tests.

jnthntatum:
- Update test runner to support typed tests
- add conversion from native type to canonical proto rep
- bump cel-spec version to 0.19.1

### DIFF
--- a/common/types/types.go
+++ b/common/types/types.go
@@ -768,6 +768,19 @@ func ProtoAsType(t *celpb.Type) (*Type, error) {
 	}
 }
 
+// TypeToProto converts from a CEL-native type representation to canonical CEL celpb.Type protobuf type.
+func TypeToProto(t *Type) (*celpb.Type, error) {
+	exprType, err := TypeToExprType(t)
+	if err != nil {
+		return nil, err
+	}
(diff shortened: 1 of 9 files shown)


common/types/types.go
```diff
@@ -768,6 +768,19 @@ func ProtoAsType(t *celpb.Type) (*Type, error) {
 	}
 }
 
+// TypeToProto converts from a CEL-native type representation to canonical CEL celpb.Type protobuf type.
```
TristonianJones, line 771: Would you mind moving this logic to the `common/decls/decls.go` file? There's some precedent for interop between the alpha and canonical protobuf types in that file.

jnthntatum: This is the inverse of the preceding function in this file -- are you sure you'd prefer it in decls/decls.go? There aren't any dependency changes so not sure I follow your comment

TristonianJones: My apologies, I had recently refactored some of the methods for type conversion in the `decls.go` and forgotten about the ones in `types.go` as well. This is fine.

TristonianJones (approved):

PR #1092, opened 2024-12-09
Policy nested rule fix

### LINKED ISSUE #1087: Policy Compiler composition fails for doubly-nested non-optional match rules.
**Describe the bug**
When the policy compiler has a doubly-nested match rule in which the output is not optional, the composer thinks that the innermost nested rule has an optional type.

**Minimal reproducible example 1**

```yaml
rule:
  match:
    - condition: "1 > 0"
      rule:
        match:
          - rule:
              match:
                - output: "1002"

    - output: "1003"
```

Output:

```
ERROR: <input>:1:1: found no matching overload for '_?_:_' applied to '(bool, optional_type(int), int)'
 | true
 | ^
```

Expected: This should compile properly and return something equivalent to:

```
1 > 0 ? 1002 : 1003
```

**Minimal reproducible example 2**

```yaml
rule:
  match:
    - condition: "1 > 0"
      rule:
        match:
          - condition: "42 > 43"
            output: "1001"

          - rule:
              match:
                - output: "1002"

    - output: "1003"
```

Output:

```
ERROR: <input>:1:1: found no matching overload for '_?_:_' applied to '(bool, optional_type(int), int)'
 | true
 | ^
```

Expected: This should compile properly and return something equivalent to:

```
1 > 0 ? (42 > 43 ? 1001 : 1002) : 1003
```

TristonianJones:
More robust handling for optional / non-optional policy outputs.

Closes #1087

### DIFF
--- a/policy/composer.go
+++ b/policy/composer.go
@@ -122,10 +122,10 @@ func (opt *ruleComposerImpl) optimizeRule(ctx *cel.OptimizerContext, r *Compiled
 		if nestedHasOptional {
 			step := newOptionalOutputStep(ctx, cond, nestedRule)
 			output = step.combine(output)
-		} else {
-			step := newNonOptionalOutputStep(ctx, cond, nestedRule)
-			output = step.combine(output)
+			continue
 		}
+		step := newNonOptionalOutputStep(ctx, cond, nestedRule)
+		output = step.combine(output)
 	}
 
 	matchExpr := output.expr()
@@ -151,6 +151,8 @@ func (opt *ruleComposerImpl) rewriteVariableName(ctx *cel.OptimizerContext) ast.
 	})
 }
 
+// registerVariable creates an entry for a variable name within the cel.@block used to enumerate
+// variables within composed policy expression.
 func (opt *ruleComposerImpl) registerVariable(ctx *cel.OptimizerContext, v *CompiledVariable) {
 	varName := fmt.Sprintf("variables.%s", v.Name())
 	indexVar := fmt.Sprintf("@index%d", opt.nextVarIndex)
@@ -166,22 +168,32 @@ func (opt *ruleComposerImpl) registerVariable(ctx *cel.OptimizerContext, v *Comp
 	opt.nextVarIndex++
 }
 
+// sortedVariables returns the variables ordered by their declaration index.
 func (opt *ruleComposerImpl) sortedVariables() []varIndex {
 	return opt.varIndices
 }
 
+// outputStep interface represents a policy output expression.
 type outputStep interface {
+	// isOptional indicates whether the output step has an optional result.
+	//
+	// Individual conditional attributes are not optional; however, rules and subrules can have optional output.
 	isOptional() bool
 
+	// condition returns the condition associated with the output.
 	condition() ast.Expr
 
+	// isConditional returns true if the condition expression is not trivially true.
 	isConditional() bool
 
+	// expr returns the output expression for the step.
 	expr() ast.Expr
 
+	// combine assembles two output expressions into a single output step.
 	combine(other outputStep) outputStep
 }
 
+// baseOutputStep encapsulates the common features of an outputStep implementation.
 type baseOutputStep struct {
 	ctx  *cel.OptimizerContext
 	cond ast.Expr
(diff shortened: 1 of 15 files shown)


policy/composer.go
```diff
@@ -192,6 +168,187 @@ func (opt *ruleComposerImpl) registerVariable(ctx *cel.OptimizerContext, v *Comp
...
+//
+// non-optional.combine(non-optional) // non-optional
+// (non-optional && conditional).combine(optional) // optional
+// (non-optional && unconditional).combine(optional) // non-optional
+//
+// The last combination case is unusual, but effectively it means that the non-optional value prunes away
+// the potential optional output.
+func (s nonOptionalOutputStep) combine(step outputStep) outputStep {
+	if step == nil {
+		// The input `step`` may be nil if this is the first outputStep
```
l46kok: nit: extra backtick in step

policy/composer.go
```diff
@@ -88,42 +88,29 @@ func (opt *ruleComposerImpl) Optimize(ctx *cel.OptimizerContext, a *ast.AST) *as
...
 	for i := matchCount - 1; i >= 0; i-- {
 		m := matches[i]
 		cond := ctx.CopyASTAndMetadata(m.Condition().NativeRep())
-		// If the condition is trivially true, not of the matches in the rule causes the result
-		// to become optional, and the rule is not the last match, then this will introduce
-		// unreachable outputs or rules.
-		triviallyTrue := m.ConditionIsLiteral(types.True)
 
 		// If the output is non-nil, then determine whether the output should be wrapped
 		// into an optional value, a conditional, or both.
```
seirl: This part of the comment looks out-of-date.

policy/composer.go
```diff
@@ -192,6 +168,187 @@ func (opt *ruleComposerImpl) registerVariable(ctx *cel.OptimizerContext, v *Comp
 	opt.nextVarIndex++
 }
 
+// sortedVariables returns the variables ordered by their declaration index.
 func (opt *ruleComposerImpl) sortedVariables() []varIndex {
 	return opt.varIndices
 }
+
+// outputStep interface represents a policy output expression.
+type outputStep interface {
```
seirl: nit: why "step" if it represents an expression? (maybe expanding the comment is enough to explain)

seirl: Just reading the description, I don't understand the difference between this and CompiledRule. It seems to me that CompiledRule could also define all the methods in this interface. Could we make the difference explicit? What processing happens to turn a CompiledRule into this?

seirl (dismissed):
Thank you for the quick fix!

policy/composer.go
TristonianJones: The output step could either be a `CompiledMatch` or a `CompiledRule`, and it represents some intermediate state of expression composition. I've updated the comment to indicate the purpose of this interface. It's likely something I'll elaborate on in further PRs to introduce new composition semantics.

policy/testdata/nested_rule5/policy.yaml
```diff
@@ -0,0 +1,30 @@
...
+name: nested_rule5
+rule:
+  match:
+    - condition: x > 0
+      rule:
+        match:
+          - rule:
+              match:
+                - condition: "x > 2"
+                  output: "true"
```
l46kok, line 24: Could we add a failure test case for showing incompatible output types for these nested rules involving optionals? For example, changing this output to `optional.of(true)` should make composition fail.

TristonianJones: Do you want to show the compilation error, composition error, or runtime error? I think if I put `optional.of(true)` that's a compile error since the output types don't match between rules. If I use `dyn(optional.of(true))` I think that works, but it just means the output type is potentially an `optional.of(optional.of(true))`

l46kok: Compilation error post composition due to output type disagreement was what I was thinking. Since we're using optional types to indicate optional steps, it's likely valuable to include a case where the output itself having an optional value doesn't clobber with the optional steps from nested rules.

seirl: The check that output types agree for all the subrules happens before composition, IIUC.

l46kok: Oh thanks for pointing that out. The type agreement check occurs during composition for Java, so I assumed it must've been the same here. We should be fine

l46kok (dismissed):

l46kok (approved):

PR #1096, opened 2024-12-16
Address non-const format string lint findings

jnthntatum:


cel/env.go
```diff
@@ -217,7 +217,7 @@ func (e *Env) Check(ast *Ast) (*Ast, *Issues) {
 	chk, err := e.initChecker()
 	if err != nil {
 		errs := common.NewErrors(ast.Source())
-		errs.ReportError(common.NoLocation, err.Error())
```
TristonianJones, line 220: Do you think it might be better to introduce a new method without the format arguments?

jnthntatum: Done where the calls weren't one-offs: PTAL. A recent change makes this a hard error when developing with `go test` so I was just trying to fix that for now. If you want me to refactor more of the existing formatter calls I could add an `f` suffix to the format string versions to make it a bit clearer.

TristonianJones (approved):

PR #1117, opened 2025-01-30
Add LateFunctionBinding declaration and fix constant folding

zeitgeist87:
Some functions are intended to have side effects and therefore should not be folded even if all parameters are constants. Currently the constant folding optimizer throws an error if such a function does not provide an implementation. This change allows functions with side effects to return Unknown.

### DIFF
--- a/cel/folding.go
+++ b/cel/folding.go
@@ -68,13 +68,18 @@ func (opt *constantFoldingOptimizer) Optimize(ctx *OptimizerContext, a *ast.AST)
 	// Walk the list of foldable expression and continue to fold until there are no more folds left.
 	// All of the fold candidates returned by the constantExprMatcher should succeed unless there's
 	// a logic bug with the selection of expressions.
-	foldableExprs := ast.MatchDescendants(root, constantExprMatcher)
+	constantExprMatcherCapture := func(e ast.NavigableExpr) bool { return constantExprMatcher(ctx, a, e) }
+	foldableExprs := ast.MatchDescendants(root, constantExprMatcherCapture)
 	foldCount := 0
 	for len(foldableExprs) != 0 && foldCount < opt.maxFoldIterations {
 		for _, fold := range foldableExprs {
 			// If the expression could be folded because it's a non-strict call, and the
 			// branches are pruned, continue to the next fold.
 			if fold.Kind() == ast.CallKind && maybePruneBranches(ctx, fold) {
+				continue
+			}
+			// Late-bound function calls cannot be folded.
+			if fold.Kind() == ast.CallKind && isLateBoundFunctionCall(ctx, a, fold) {
 				continue
 			}
 			// Otherwise, assume all context is needed to evaluate the expression.
@@ -85,7 +90,7 @@ func (opt *constantFoldingOptimizer) Optimize(ctx *OptimizerContext, a *ast.AST)
 			}
 		}
 		foldCount++
-		foldableExprs = ast.MatchDescendants(root, constantExprMatcher)
+		foldableExprs = ast.MatchDescendants(root, constantExprMatcherCapture)
 	}
 	// Once all of the constants have been folded, try to run through the remaining comprehensions
 	// one last time. In this case, there's no guarantee they'll run, so we only update the
@@ -137,6 +142,15 @@ func tryFold(ctx *OptimizerContext, a *ast.AST, expr ast.Expr) error {
 	// Update the fold expression to be a literal.
 	ctx.UpdateExpr(expr, ctx.NewLiteral(out))
 	return nil
+}
+
+func isLateBoundFunctionCall(ctx *OptimizerContext, a *ast.AST, expr ast.Expr) bool {
+	call := expr.AsCall()
+	function := ctx.Functions()[call.FunctionName()]
+	if function == nil {
+		return false
+	}
+	return function.HasLateBinding()
 }
 
 // maybePruneBranches inspects the non-strict call expression to determine whether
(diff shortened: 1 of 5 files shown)


cel/folding_test.go
```diff
@@ -313,6 +316,83 @@ func TestConstantFoldingOptimizer(t *testing.T) {
...
+			folded: `3`,
+		},
+		{
+			expr:   `withSideEffect(3)`,
+			folded: `withSideEffect(3)`,
+		},
+		{
+			expr:  `noImpl(3)`,
+			error: `constant-folding evaluation failed: no such overload: noImpl`,
+		},
```
TristonianJones, line 343: To keep things simple, the constant folding optimizer should validate that all functions / overloads referenced in the ast.Expr have bindings before attempting to fold.

cel/folding_test.go
```diff
@@ -313,6 +316,83 @@ func TestConstantFoldingOptimizer(t *testing.T) {
...
+			Overload("noSideEffect_int_int",
+				[]*Type{IntType},
+				IntType, FunctionBinding(func(args ...ref.Val) ref.Val {
+					return args[0]
+				}))),
+		Function("withSideEffect",
+			Overload("withSideEffect_int_int",
+				[]*Type{IntType},
+				IntType, FunctionBinding(func(args ...ref.Val) ref.Val {
+					return &types.Unknown{}
```
TristonianJones: In the side-effect case, we can propose marking a function as volatile or late-bound also skip trying to fold these functions. 

```go
cel.Function("withSideEffect", 
    cel.Overload("withSideEffect_int_int", []*cel.Type{cel.IntType, cel.IntType}, cel.IntType),
    cel.LateFunctionBinding())
```

This marker will let users specify that the function definition will be provided later on the `cel.Activation`. 

```go
// validates that the function has a late binding flag, and packages up context needed to resolve
// the function within the Activation by the interpreter.
lateFn, err := env.LateBoundFunction("withSideEffect", func(rhs, lhs ref.Val) ref.Val {
   // do something interesting
})
input, err := cel.NewActivation(data)
input.AddFunctionBinding(lateFn)
```

I'd like to do this to support late-bound functions inside of cel-go, so perhaps we can work toward the late-bound function support and solve two issues at once?

zeitgeist87:
@TristonianJones I added the `cel.LateFunctionBinding()` and adapted the constant folding optimizer to use it. However, this is just the declaration and it does not add support for adding late function bindings to the Activation.

cel/folding_test.go
zeitgeist87: I tried implementing this:

```
func checkForMissingBindings(ctx *OptimizerContext, expr ast.Expr) error {
	var result error = nil
	visitor := ast.NewExprVisitor(func(e ast.Expr) {
		if e.Kind() == ast.CallKind {
			function := ctx.Functions()[e.AsCall().FunctionName()]
			if function != nil {
				bindings, err := function.Bindings()
				if err != nil {
					result = err
				} else if len(bindings) == 0 {
					result = fmt.Errorf("function %s has no bindings and cannot be folded", function.Name())
				}
			} else {
				result = fmt.Errorf("function %s was not found and cannot be folded", function.Name())
			}
		}
	})
	ast.PostOrderVisit(expr, visitor)
	return result
}
```

However, this fails for built-in functions like `_?._` and it does not take the overloads from the type checker into account, which would also be a bit tricky for singletons. If we go with the `cel.LateFunctionBinding()` approach, then the only benefit of this would be a slightly more readable error message. I don't think it is worth the added complexity. WDYT?

cel/folding.go
```diff
@@ -139,6 +145,32 @@ func tryFold(ctx *OptimizerContext, a *ast.AST, expr ast.Expr) error {
...
+	if function == nil {
+		return overloads
+	}
+	for _, o := range function.OverloadDecls() {
+		for _, id := range overloadIds {
+			if id == o.ID() {
+				overloads = append(overloads, o)
+			}
+		}
+	}
```
TristonianJones: Could you add a function to select an overload by name from the `decl.FunctionDecl`? I think that will simplify some of this code and be generally useful.

cel/folding.go
```diff
@@ -139,6 +145,32 @@ func tryFold(ctx *OptimizerContext, a *ast.AST, expr ast.Expr) error {
...
+	return overloads
+}
+
+func isLateBoundFunctionCall(ctx *OptimizerContext, a *ast.AST, expr ast.Expr) bool {
+	call := expr.AsCall()
+	overloadDecls := getFunctionOverloadDecls(ctx.Functions()[call.FunctionName()], a.GetOverloadIDs(expr.ID()))
```
TristonianJones: Consider iterating over the functions and overloads inline since you just need to check for existence rather than allocate the overload subset.

TristonianJones: Alternatively, you could expose `HasLateBinding()` at the `decls.FunctionDecl` level since this will be useful in generating an `Activation` with slots carved out for associating the function implementation with the activation instance.

cel/folding_test.go
TristonianJones: I think the `HasLateBinding()` being related to a specific configuration of a specific function / overload probably removes the need for a `checkForMissing` binding step. We should just error if we can't evaluate the constant fold and that's okay.

cel/folding.go
zeitgeist87: I instead went with the approach you suggested below of adding `HasLateBinding()` to `decls.FunctionDecl`.

cel/folding.go
zeitgeist87: I added `HasLateBinding()` to `decls.FunctionDecl`. Since we enforce that functions cannot mix late and non-late bindings, we also don't need to iterate over the overloads here. This greatly simplifies the code.

common/decls/decls.go
```diff
@@ -232,6 +232,9 @@ func (f *FunctionDecl) AddOverload(overload *OverloadDecl) error {
 			}
 			return fmt.Errorf("overload redefinition in function. %s: %s has multiple definitions", f.Name(), oID)
 		}
+		if overload.HasLateBinding() != o.HasLateBinding() {
```
TristonianJones, line 235: Did you want to check `f.HasLateBinding()` here?

TristonianJones: Nevermind, this is cheaper how you have it. Thanks!

TristonianJones (approved):

PR #1119, opened 2025-01-31
Minor update on cost order

TristonianJones:
Added test to validate cost order is preserved, and updated cost order to fix current behavior on K8s


cici37:
I have tested the fix in K8s side. Thank you soooo much for the fix!! :)

/lgtm

jcking (approved):

Issue #1050, opened 2024-10-24
Support `ProgramOption`s to limit and disable comprehension macros


sergiitk:
**Feature request checklist**

- [x] There are no issues that match the desired change
- [ ] The change is large enough it can't be addressed with a simple Pull Request
- [x] If this is a bug, please file a [Bug Report](./ISSUE_TEMPLATE).

**Change**
To make CEL environment setup consistent across CEL implementations, I propose to add a runtime option equivalent to CEL-cpp `InterpreterOptions.enable_comprehension` to CEL-Go and CEL-Java (corresponding ticket https://github.com/google/cel-java/issues/484).

In Java, runtime CEL comprehensions can be limited using `comprehensionMaxIterations(int value)` runtime option. Setting `value` to 0  disables the comprehension macros. If the limit is reached, program evaluation throws `CelEvaluationException` with `CelErrorCode.ITERATION_BUDGET_EXCEEDED`.

There is no option equivalent to `comprehensionMaxIterations()` in CEL-Go . Similar to google/cel-java#484, the new Go option equivalent to cpp's `enable_comprehension` could use `comprehensionMaxIterations(0)` under-the-hood.

**Example**

```go
// Actual option names up to the implementer.
prg, err := env.Program(ast, EvalOptions(OptDisableComprehension))

// Equivalent:
prg, err := env.Program(ast, ComprehensionMaxIterations(0))
```

**Related**
Internal ref: go/grpc-cel-integration

sergiitk:
I was able to achieve the desired behavior using the new Config feature:

```go
celConfig := env.NewConfig("grpc-unified-matcher").
	AddVariables(
		// variables
	).
	AddValidators(
		env.NewValidator("cel.validator.comprehension_nesting_limit").SetConfig(map[string]any{"limit": 0}),
	)

matcherEnv, err := cel.NewCustomEnv(cel.FromConfig(celConfig))
```

In my specific setup where I get AST from an external source, I needed to additionally run the check phase before building the program to surface the validation errors (similar to https://github.com/google/cel-go/issues/1078#issuecomment-2744181932):

```go
func NewCheckedCelProgram(ast *cel.Ast) (cel.Program, error) {
	ast, checkIssues := celRuntimeEnv.Check(ast)
	if checkIssues != nil {
		return nil, fmt.Errorf("error checking AST: %v", checkIssues)
	}
	prg, err := celRuntimeEnv.Program(ast)
	if err != nil {
[...]

PR #1125, opened 2025-02-07
Canonical environment description and stdlib subsetting

### LINKED ISSUE #290: Make subsetting the standard environment simple

### LINKED ISSUE #844: Option to Remove Built-in CEL Functions
**Feature request checklist**

**Change**
Currently, there is no option to remove built-in CEL functions when using `cel.NewEnv`. Also, there is no option to include standard built-in CEL functions when using `cel.NewCustomEnv`. The proposed change is to add an option to `cel.NewEnv` for excluding built-in functions or an option to `cel.NewCustomEnv` for including standard CEL functions.

### LINKED ISSUE #899: Can we allow users to disable specific functions in Checker?
We only want to use a subset of CEL.

### LINKED ISSUE #1078: Support `ProgramOption`s to disable string conversion and list/string concatenation
**Feature request checklist**

**Change**
To make CEL environment setup consistent across CEL implementations, I propose to add equivalent runtime options that control string conversion and list/string concatenation equivalent to those found in CEL-cpp:

```cpp
// Enable string() overloads.
bool enable_string_conversion = true;

// Enable string concatenation overload.
bool enable_string_concat = true;

// Enable list concatenation overload.
bool enable_list_concat = true;
```

**Example**
```go
// Actual option names up to the implementer.
opts := cel.EvalOptions(cel.OptDisableDisableStringConcat |
	cel.OptDisableDisableListConcat | cel.OptDisableDisableListConcat)
prg, err := env.Program(ast, opts)
```

**Related**
* Corresponding cel-java request: google/cel-java#502
* Internal ref: go/grpc-cel-integration

TristonianJones:
Support for subsetting the CEL standard environment via a canonical environment
description which is YAML-serializable.

This change also removes a behavior which would potentially break forward compatible
changes, as now type declarations are not included into the variable declaration set.
In the past a duplicate declaration for `int` would accept the precedence of the variable
with the same name (erroring if there were type differences) rather than permitting
the user-provided variable to mask the type identifier.

Closes #1078
Closes #899
Closes #844
Closes #290

### DIFF
--- a/common/env/env.go
+++ b/common/env/env.go
@@ -1,0 +1,348 @@
+// Copyright 2025 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Package env provides a representation of a CEL environment.
+package env
+
+import (
+	"fmt"
+	"math"
+	"strconv"
+
+	"github.com/google/cel-go/common/decls"
+	"github.com/google/cel-go/common/types"
+)
+
+// NewConfig creates an instance of a YAML serializable CEL environment configuration.
+func NewConfig() *Config {
+	return &Config{
+		Imports:    []*Import{},
+		Extensions: []*Extension{},
+		Variables:  []*Variable{},
+		Functions:  []*Function{},
+	}
+}
+
+// Config represents a serializable form of the CEL environment configuration.
+//
+// Note: custom validations, feature flags, and performance tuning parameters are
+// not (yet) considered part of the core CEL environment configuration and should
+// be managed separately until a common convention for such configuration settings
+// can be developed.
+type Config struct {
+	Name            string           `yaml:"name"`
+	Description     string           `yaml:"description,omitempty"`
+	Container       string           `yaml:"container,omitempty"`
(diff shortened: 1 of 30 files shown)


common/env/env.go
```diff
@@ -0,0 +1,393 @@
...
+	case "list":
+		if len(td.Params) == 1 {
+			et, err := td.Params[0].AsCELType(tp)
+			if err != nil {
+				return nil, err
+			}
+			return types.NewListType(et), nil
+		}
+		return nil, fmt.Errorf("list type has unexpected param count: %d", len(td.Params))
+	case "optional", "optional_type":
```
l46kok: `optional_type` is the actual type name for optionals right? Should we just remove the case for `optional`, or is this to maintain backwards compatibility?

common/env/env.go
```diff
@@ -0,0 +1,393 @@
...
+	}
+}
+
+// Config represents a serializable form of the CEL environment configuration.
+//
+// Note: custom validations, feature flags, and performance tuning parameters are
+// not (yet) considered part of the core CEL environment configuration and should
+// be managed separately until a common convention for such configuration settings
+// can be developed.
+type Config struct {
```
l46kok, line 44: Is being able to subset type identifiers desirable?

l46kok (dismissed):

common/env/env.go
TristonianJones: It is, but I haven't figured out how to introduce this concept into the type registries / type providers yet.

common/env/env.go
TristonianJones: I'll just remove the back-compat, I don't think it's necessary

l46kok (approved):

sergiitk:
This resolved #1050 as well.

PR #1128, opened 2025-02-11
Support for cel.Env conversion to YAML-serializable config

TristonianJones:
Support for serializing `cel.Env` to the YAML-serializable `env.Config` object.

This update introduces mutable setters on the `env.Config` object, a new method
for marking a library disabled, and a host of YAML serialization tests.

The next piece of this development will be to formalize the extension handlers
as interfaces in the `cel` package to support the option for configuring a new
CEL environment from the serializable config value and configuring new env options
as appropriate when portions of the config refer to extensions not visible from
the `cel` package.

### DIFF
--- a/cel/env.go
+++ b/cel/env.go
@@ -165,14 +165,14 @@ type Env struct {
 // the serialize method. Since optimizers are a separate construct from the environment
 // and the standard expression components (parse, check, evalute), they are also not
 // supported by the serialize method.
-func (e *Env) Config() *env.Config {
-	conf := env.NewConfig()
+func (e *Env) Config(name string) *env.Config {
+	conf := env.NewConfig(name)
 	// Container settings
 	if e.Container != containers.DefaultContainer {
-		conf.Container = e.Container.Name()
-		for _, typeName := range e.Container.AliasSet() {
-			conf.Imports = append(conf.Imports, env.NewImport(typeName))
-		}
+		conf.SetContainer(e.Container.Name())
+	}
+	for _, typeName := range e.Container.AliasSet() {
+		conf.AddImports(env.NewImport(typeName))
 	}
 
 	libOverloads := map[string][]string{}
@@ -202,25 +202,24 @@ func (e *Env) Config() *env.Config {
 			libName = alias
 		}
 		if libName == "stdlib" && canSubset {
-			conf.StdLib = subsetLib.LibrarySubset()
+			conf.SetStdLib(subsetLib.LibrarySubset())
 			continue
 		}
 		version := uint32(math.MaxUint32)
 		if versionLib, isVersioned := lib.(LibraryVersioner); isVersioned {
 			version = versionLib.LibraryVersion()
 		}
-		conf.Extensions = append(conf.Extensions, env.NewExtension(libName, version))
+		conf.AddExtensions(env.NewExtension(libName, version))
 	}
 
 	// If this is a custom environment without the standard env, mark the stdlib as disabled.
 	if conf.StdLib == nil && !e.HasLibrary("cel.lib.std") {
-		conf.StdLib = env.NewLibrarySubset()
-		conf.StdLib.Disabled = true
+		conf.SetStdLib(env.NewLibrarySubset().SetDisabled(true))
 	}
 
 	// Serialize the variables
 	if e.contextProto != nil {
-		conf.ContextVariable = env.NewContextVariable(string(e.contextProto.FullName()))
+		conf.SetContextVariable(env.NewContextVariable(string(e.contextProto.FullName())))
 		skipVariables := map[string]bool{}
 		fields := e.contextProto.Fields()
 		for i := 0; i < fields.Len(); i++ {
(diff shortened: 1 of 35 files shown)


TristonianJones:
@aakash070 FYI

cel/env.go
```diff
@@ -151,6 +156,115 @@ type Env struct {
...
+	if e.contextProto != nil {
+		conf.SetContextVariable(env.NewContextVariable(string(e.contextProto.FullName())))
+		skipVariables := map[string]bool{}
+		fields := e.contextProto.Fields()
+		for i := 0; i < fields.Len(); i++ {
+			field := fields.Get(i)
+			variable, err := fieldToVariable(field)
+			if err != nil {
+				continue
+			}
```
l46kok: Would it be reasonable to report the failed conversion here?

cel/env_test.go
```diff
@@ -302,6 +304,126 @@ func TestFunctions(t *testing.T) {
...
+		{
+			name: "context proto - with extra variable",
+			opts: []EnvOption{
+				DeclareContextProto((&proto3pb.TestAllTypes{}).ProtoReflect().Descriptor()),
+				Variable("extra", StringType),
+			},
+			wantConfig: env.NewConfig("context proto - with extra variable").
+				SetContextVariable(env.NewContextVariable("google.expr.proto3.test.TestAllTypes")).
+				AddVariables(env.NewVariable("extra", env.NewTypeDesc("string"))),
+		},
```
l46kok: This can be removed, it's duplicated from above

cel/env.go
```diff
@@ -151,6 +156,115 @@ type Env struct {
...
+	for libName, lib := range e.libraries {
+		// Track the options which have been configured by a library and
+		// then diff the library version against the configured function
+		// to detect incremental overloads or rewrites.
+		libEnv, _ := NewCustomEnv()
+		libEnv, _ = Lib(lib)(libEnv)
+		for fnName, fnDecl := range libEnv.Functions() {
+			if len(fnDecl.OverloadDecls()) == 0 {
+				continue
+			}
```
l46kok, line 190: It feels like this should error (not sure if we have a use case for 0 overloads)

cel/env.go
```diff
@@ -151,6 +156,115 @@ type Env struct {
...
+			variable, err := fieldToVariable(field)
+			if err != nil {
+				continue
+			}
+			skipVariables[variable.Name()] = true
+		}
+		for _, v := range vars {
+			if _, found := skipVariables[v.Name()]; !found {
+				conf.AddVariableDecls(v)
+			}
```
l46kok, line 249: Could you help me understand why we're not adding the top level variables from the context proto? This is part of what `DeclareContextProto` does. I feel like I'd be surprised if they are suddenly gone from the serialized config.

cel/env.go
TristonianJones: Yes, let me update the signature of this call. I don't think it can ever actually fail, but better to have an error on config conversion than a crash.

cel/env.go
TristonianJones: We're supplying the top-level variables by indicating there's a `context_variable` in the config. Including the variables would be redundant. Also, possibly confusing if you try to roundtrip from a config with the `context_variable` set and suddenly you get the `context_variable` back out with lots of variables defined.

l46kok: I guess they'd just be added back to the env when deserialized, just unsure if we want to retain them somehow for auditing reasons

l46kok: Yeah I'm convinced we should skip them. Thanks for the elaboration

cel/env.go
TristonianJones: The zero overloads case can happen if the function is present but effectively disabled by subsetting, so it seems reasonable to omit from the config.

common/env/testdata/subset_env.yaml
```diff
@@ -0,0 +1,39 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+name: "subset-env"
+stdlib:
+  exclude_macros:
+    - map
+    - filter
+  exclude_functions:
```
l46kok, line 20: In Java, we treated configuring both include/exclude in the same path to be an error: https://github.com/google/cel-java/blob/main/checker/src/test/java/dev/cel/checker/CelStandardDeclarationsTest.java#L58

It's likely useful to do the same here to avoid surprises

l46kok (approved):

common/env/testdata/subset_env.yaml
TristonianJones: I don't have a great builder pattern I can follow with how the YAML serialization works. What I can do is add validation that only one of these options is set when trying to configure an environment from a config object, or when you attempt to subset and provide both options, something in the `StdLibSubset()` option that I can invoke on the `LibrarySubset` object to validate it's correctness. What do you think?

l46kok: SGTM, I think it'd be good to inform the user somehow pertaining to misconfiguration. I'll defer to you on where that's done.

PR #1129, opened 2025-02-13
Option to configure CEL via env.Config object

TristonianJones:
EnvOption to configure CEL via Config object

Introduces validation methods and checks on all env exposed classes,
specifically `env.Config.Validate()` is called before processing any config
object within the `cel.FromConfig()` option.

This change also introduces limitations on how an environment can be 
constructed when being used with standard library subsetting ... any
singleton library which is being subset must be subset at the time it is
configured. Any subsequent attempt to configure the singleton will either
be ignored or result in an error. The `FromConfig` option validates collisions
within the pre-existing config, but there is no subsequent validation
afterwards.

### DIFF
--- a/cel/env_test.go
+++ b/cel/env_test.go
@@ -15,9 +15,11 @@
 package cel
 
 import (
+	"errors"
 	"fmt"
 	"math"
 	"reflect"
+	"strings"
 	"sync"
 	"testing"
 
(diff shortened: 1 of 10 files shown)


cel/env_test.go
```diff
@@ -515,6 +517,15 @@ func BenchmarkEnvExtendEagerDecls(b *testing.B) {
 	}
 }
 
+func mustContextProto(t *testing.T, pb proto.Message) Activation {
```
l46kok, line 823: nit: unused

l46kok (dismissed):

cel/env_test.go
TristonianJones: Looks like all of my FromConfig tests were pruned during a rebase. I've restored them and this function is used now. Thanks!

l46kok (approved):

Issue #1131, opened 2025-02-18
[Config] Support `config` and `subset` fields for Extensions


TristonianJones:
As part of the serializable configuration effort, all `Extension`
libraries should support `config` as a map field to indicate when 
configuration data other than versions should be applied.

Likewise, `Extension` libraries should also support the `subset`
field akin to the `stdlib` field in order to make sure extensions
are trivial to subset when the implementor decides this is a worthwhile
feature to support.

PR #1132, opened 2025-02-18
Support for feature flags and validators in env.Config

TristonianJones:
Support for feature flags and validators in `env.Config` as both the
output from `cel.Env` and as an input to the `cel.FromConfig` option.

Note: only `ConfigurableASTValidators` are included as serializable
since some validators are included within other libraries whereas
configurable validators are intended to be standalone.

Note: this implies that libraries should support configuration similar
to validators. Filed https://github.com/google/cel-go/issues/1131 to track

### DIFF
--- a/cel/options.go
+++ b/cel/options.go
@@ -73,6 +73,26 @@ const (
 	featureIdentEscapeSyntax
 )
 
+var featureIDsToNames = map[int]string{
+	featureEnableMacroCallTracking:     "cel.feature.macro_call_tracking",
+	featureCrossTypeNumericComparisons: "cel.feature.cross_type_numeric_comparisons",
+	featureIdentEscapeSyntax:           "cel.feature.backtick_identifier_syntax",
+}
+
+func featureNameByID(id int) (string, bool) {
+	name, found := featureIDsToNames[id]
+	return name, found
+}
+
+func featureIDByName(name string) (int, bool) {
+	for id, n := range featureIDsToNames {
+		if n == name {
+			return id, true
+		}
+	}
+	return 0, false
+}
+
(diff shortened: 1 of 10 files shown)


cel/options.go
```diff
@@ -530,6 +550,8 @@ func configToEnvOptions(config *env.Config, provider types.Provider, optFactorie
 		}
 		envOpts = append(envOpts, VariableDecls(vars...))
 	}
+
+	// Cnofigure functions
```
l46kok: nit: typo

cel/options.go
```diff
@@ -541,22 +563,58 @@ func configToEnvOptions(config *env.Config, provider types.Provider, optFactorie
...
-			if opt, useOption := optFac(e); useOption {
-				envOpts = append(envOpts, opt)
-				extHandled = true
-				break
-			}
+
+	// Configure features
+	for _, feat := range config.Features {
+		if id, found := featureIDByName(feat.Name); found {
+			envOpts = append(envOpts, features(id, feat.Enabled))
```
l46kok, line 574: For any misconfigured features (e.g: typo), do you want to surface this back to the uesr?

l46kok (dismissed):

cel/options.go
TristonianJones: I know, I thought about that, but I think I need to skip the feature since it could not be supported on a specific CEL stack. I can come back to this, but I couldn't think of a good way to mark a features as "intended for support in config files".

cel/options.go
TristonianJones: Fixed. PTAL

l46kok (approved):

PR #1133, opened 2025-02-19
Update strings.format to adhere to the specification

jcking:


common/types/string.go
```diff
@@ -186,6 +186,10 @@ func (s String) Value() any {
 	return string(s)
 }
 
+func (s String) String() string {
+	return strconv.Quote(string(s))
```
TristonianJones: I feel like we should leave this one alone and maybe use %q in places where we print the string like in lists and maps. Seems less likely to hit a change-detector test issue that way too.

jcking: Unfortunately we cannot use `%q` because that would quote everything, including ints. We pass ref.Value to `%s`.

jcking: Done.

ext/formatting_v2.go
```diff
@@ -0,0 +1,772 @@
...
+	return string(fmter.buf), nil
+}
+
+type stringFormatterV2 struct{}
+
+func (c *stringFormatterV2) String(arg ref.Val) (string, error) {
+	return formatStringV2(arg)
+}
+
+func (c *stringFormatterV2) Decimal(arg ref.Val) (string, error) {
```
TristonianJones, line 204: Could you propagate the comments from the v1 file into v2?

jcking: Which ones? I thought I copy pasted pretty much everything?

TristonianJones: Oh, just the capitalized methods. They're all exposed, so `golint` will complain if they're not commented.

jcking: I thought it only did that for public types? Because formatting.go does have them commented either.

jcking: Done.

TristonianJones (approved):

PR #1136, opened 2025-02-24
Support for splitting nested branching operators within policies

TristonianJones:
Introduce support for splitting nested match blocks according to a configurable height limit.

When expressions are more than 25 levels deep, logical AND, logical OR, and conditional operators
will automatically be split into local variables within a `cel.@block`. This depth is configurable
in the event that a different limit is desired.

### DIFF
--- a/common/ast/ast.go
+++ b/common/ast/ast.go
@@ -465,6 +465,9 @@ func (v *maxIDVisitor) VisitEntryExpr(e EntryExpr) {
 
 type heightVisitor map[int64]int
 
+// VisitExpr computes the height of a given node as the max height of its children plus one.
+//
+// Identifiers and literals are treated as having a height of zero.
 func (hv heightVisitor) VisitExpr(e Expr) {
 	// default includes IdentKind, LiteralKind
 	hv[e.ID()] = 0
@@ -496,6 +499,7 @@ func (hv heightVisitor) VisitExpr(e Expr) {
 	}
 }
 
+// VisitEntryExpr computes the max height of a map or struct entry and associates the height with the entry id.
 func (hv heightVisitor) VisitEntryExpr(e EntryExpr) {
 	hv[e.ID()] = 0
 	switch e.Kind() {
--- a/common/ast/ast.go
+++ b/common/ast/ast.go
@@ -160,6 +160,13 @@ func MaxID(a *AST) int64 {
 	return visitor.maxID + 1
 }
 
+// Heights computes the heights of all AST expressions and returns a map from expression id to height.
+func Heights(a *AST) map[int64]int {
+	visitor := make(heightVisitor)
+	PostOrderVisit(a.Expr(), visitor)
+	return visitor
+}
+
 // NewSourceInfo creates a simple SourceInfo object from an input common.Source value.
 func NewSourceInfo(src common.Source) *SourceInfo {
 	var lineOffsets []int32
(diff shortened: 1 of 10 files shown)


common/ast/ast.go
```diff
@@ -455,3 +462,70 @@ func (v *maxIDVisitor) VisitEntryExpr(e EntryExpr) {
 		v.maxID = e.ID()
 	}
 }
+
+type heightVisitor map[int64]int
+
+func (hv heightVisitor) VisitExpr(e Expr) {
```
l46kok, line 471: Please document all exported functions

policy/composer.go
```diff
@@ -156,6 +195,62 @@ func (opt *ruleComposerImpl) rewriteVariableName(ctx *cel.OptimizerContext) ast.
...
+	// Split the expr into local variables based on expression depth within ternaries
+	ruleAST := ctx.NewAST(ruleExpr)
+	ruleNav := ast.NavigateAST(ruleAST)
+	// Branches are ordered from leaf to root via the ast.MatchDescendants call.
+	heights := ast.Heights(ruleAST)
+	branchMap := map[int64]int{}
+	branches := []ast.NavigableExpr{}
+	ast.MatchDescendants(ruleNav, func(e ast.NavigableExpr) bool {
+		// // Compute and record the height of the expression
+		// calculator.computeHeight(e)
```
l46kok: nit: can remove

policy/composer.go
```diff
@@ -156,6 +195,62 @@ func (opt *ruleComposerImpl) rewriteVariableName(ctx *cel.OptimizerContext) ast.
...
+
+		// If the expression is a comprehension, then all branches captured previously that relate to
+		// the comprehension body should be removed from the list of candidate branches for unnesting.
+		if e.Kind() == ast.ComprehensionKind {
+			// This only removes branches from the map, but not from the list of branches.
+			removeIneligibleConditions(e, branchMap)
+			return false
+		}
+		// Otherwise, if the expression is not a call, don't include it.
```
l46kok, line 216: Ternaries is our primary concern, so this can be a future enhancement -- but do we eventually want to split up nested lists/maps and field selections?

policy/composer.go
```diff
@@ -362,3 +474,42 @@ func isOptionalNone(e ast.Expr) bool {
...
+}
+
+func comprehensionSubExprIDs(e ast.NavigableExpr) []int64 {
+	compre := e.AsComprehension()
+	// Almost the same as e.Children(), but skips the iteration range
+	return enumerateExprIDs(
+		compre.AccuInit().(ast.NavigableExpr),
+		compre.LoopCondition().(ast.NavigableExpr),
+		compre.LoopStep().(ast.NavigableExpr),
+		compre.Result().(ast.NavigableExpr),
```
l46kok, line 490: Do you think we can a test case with nested macros with a deeply nested expr inside? Anecdotally, this is where I ran into the most trouble with CSE, [so I ended up creating some baseline cases for it](https://github.com/google/cel-java/blob/a9c96754f6852ad9818a9cbe58c71a4d165be897/optimizer/src/test/resources/subexpression_unparsed.baseline#L321-L367). For example, I'm not 100% confident that we should be including the loop step / result here.

policy/composer.go
TristonianJones: It's possible, but the worst offenders are definitely the ternaries and logical AND/OR

policy/composer.go
TristonianJones: Since the loop step can access per-element iteration state, it should be considered ineligible. After more thought, I can include the `result` expression even though it references internal state since it doesn't reference per-element iteration state.

policy/composer.go
TristonianJones: I've expanded the nesting check to all calls, though I believe I need to sort unnest candidates by height, and I haven't done this yet.

policy/composer.go
TristonianJones: Since we're not mangling expression identifiers within comprehensions, I'm going to continue with the more conservative approach and omit all comprehension sub-expressions.

l46kok (approved):

PR #1138, opened 2025-02-26
Separate unnest optimization from composer to capture type info

TristonianJones:
Introduce a separate optimizer for unnesting policy expressions.

When attempting to compose the variable set and then optimize the
remaining expression in the same optimizer, some type information was
lost regarding the core expression. Separating the optimizers fixes the
type-loss issue.

### DIFF
--- a/policy/composer.go
+++ b/policy/composer.go
@@ -72,13 +72,21 @@ type RuleComposer struct {
 // Compose stitches together a set of expressions within a CompiledRule into a single CEL ast.
 func (c *RuleComposer) Compose(r *CompiledRule) (*cel.Ast, *cel.Issues) {
 	ruleRoot, _ := c.env.Compile("true")
-	opt := cel.NewStaticOptimizer(
-		&ruleComposerImpl{
-			rule:             r,
-			varIndices:       []varIndex{},
-			exprUnnestHeight: c.exprUnnestHeight,
-		})
-	return opt.Optimize(c.env, ruleRoot)
+	composer := &ruleComposerImpl{
+		rule:       r,
+		varIndices: []varIndex{},
+	}
+	opt := cel.NewStaticOptimizer(composer)
+	ast, iss := opt.Optimize(c.env, ruleRoot)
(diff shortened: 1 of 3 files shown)


policy/composer.go
```diff
@@ -197,15 +198,94 @@ func (opt *ruleComposerImpl) rewriteVariableName(ctx *cel.OptimizerContext) ast.
...
+}
+
+type ruleUnnesterImpl struct {
+	nextVarIndex     int
+	varIndices       []varIndex
+	exprUnnestHeight int
+}
+
+func (opt *ruleUnnesterImpl) Optimize(ctx *cel.OptimizerContext, a *ast.AST) *ast.AST {
+	// If the input AST does not start with a cel.@block, return
```
l46kok: Did you mean to return early here? It would be preferable to

policy/composer.go
```diff
@@ -197,15 +198,94 @@ func (opt *ruleComposerImpl) rewriteVariableName(ctx *cel.OptimizerContext) ast.
...
+		varExprs = make([]ast.Expr, vars.Size())
+		copy(varExprs, vars.Elements())
+		for i := 0; i < vars.Size(); i++ {
+			v := varExprs[i]
+			indexVar := fmt.Sprintf("@index%d", i)
+			t := a.GetType(v.ID())
+			vi := varIndex{
+				index:    i,
+				indexVar: indexVar,
+				localVar: indexVar,
```
l46kok: It looks like we can safely remove this?

l46kok (dismissed):

policy/composer.go
TristonianJones: Since the optimizer always re-checks the expression I need to actually rebuild the `cel.@block`, and the expression unnesting could also generate a block if one doesn't yet exist, so I don't think I can early return unless I missed something.

policy/composer.go
TristonianJones: Yep, will remove.

l46kok (approved):

PR #1143, opened 2025-03-12
Add a new compiler tool which can be used to compile CEL expressions and policies using serialized environment

aakash070:
This change adds a new compiler tool which can be used to generate an ast for CEL expressions and policies provided in the form of:
1. CEL policy file
2. CEL expression string file
3. CEL checked expression
4. Raw CEL string

The compiler has the ability to extend a serialized environment file to create the environment which is used for compilation. The file can be in either Textproto or YAML format.


tools/compiler/testdata/custom_policy.celpolicy
```diff
@@ -0,0 +1,25 @@
...
+rule:
+  match:
+    - condition: |
+        variable1 == 1 || variable2 == "known"
+      output: "true"
+variable_types:
+  - variable_name: "variable1"
+    variable_type: "int"
+  - variable_name: "variable2"
+    variable_type: "string"
```
TristonianJones, line 20: Could you ensure that `variable_types` appears before the `rule` declaration?

tools/compiler/testdata/config.textproto
```diff
@@ -0,0 +1,138 @@
+# proto-file: cel.dev/expr/conformance/env_config.proto
```
TristonianJones: At a file level, we should really just validate the round-tripping logic yaml -> proto -> yaml and validate that the roundtrip is correct.  If desired, you could convert yaml -> textproto in memory and run tests against the converted text proto. Let's not check in the textproto file itself though.

tools/compiler/compiler_test.go
```diff
@@ -0,0 +1,224 @@
+package compiler
```
TristonianJones, line 15: nit: missing copyright

tools/compiler/compiler.go
```diff
@@ -0,0 +1,515 @@
...
+	}
+	c.env = env
+	return c.env, nil
+}
+
+func (c *compiler) CreatePolicyParser() (*policy.Parser, error) {
+	return policy.NewParser(c.policyParserOptions...)
+}
+
+func (c *compiler) PolicyCompilerOptions() []policy.CompilerOption {
```
TristonianJones, line 139: nit: please comment all exposed functions

tools/compiler/testdata/config.textproto
aakash070: Since we're not providing a round tripping logic yaml->proto->yaml as a capability, i am not sure if adding the same for tests would be beneficial as the transformation logic could get biased for the test scenario.
I have replaced the textproto file with an in-memory proto derived object instead, it should serve the same purpose without having to check in the file.

TristonianJones:
@aakash070 looks like the build deps for policy need to be opened up for compiler to depend on them.

tools/compiler/compiler.go
```diff
@@ -0,0 +1,515 @@
...
+
+// CreateAst creates a CEL AST from a file using the provided compiler:
+// - All policy metadata options as executed using the policy metadata map to extend the
+// environment.
+// - All policy compiler options are passed on to compile the parsed policy.
+//
+// The file must be in one of the following formats:
+// - .cel: CEL string expression
+// - .celpolicy: CEL policy
+func (f *FileExpression) CreateAst(compiler Compiler) (*cel.Ast, map[string]any, error) {
```
TristonianJones: nit: casing on `Ast` should be `AST` (sorry for the bad precedent from CEL early on with `cel.Ast`)

TristonianJones (approved):

PR #1149, opened 2025-04-02
Add test runner library

aakash070:
This change adds a test runner library under tools/celtest which provides native language support for setting up and executing tests. 
1. The library makes use of the compiler tool for setting up the CEL runtime environment.
2. It also provides callback options for augmenting the test runner with test suite path, test suite parser, file descriptor set path.

The change also includes a test harness in YAML which aligns with cel.expr.conformance.test.TestSuite.
The following changes will be raised separately:
1. The bazel rule to trigger the test runner with parameterized options
2. Sample usages for triggering test runner library for policies, expressions and compiled expressions.


test/suite.go
```diff
@@ -0,0 +1,60 @@
...
+	Tests []*Case `yaml:"tests"`
+}
+
+// Case is a test case to validate a CEL policy or expression. The test case
+// encompasses evaluation of the compiled expression using the provided input
+// bindings and asserting the result against the expected result.
+type Case struct {
+	Name        string `yaml:"name"`
+	Description string `yaml:"description"`
+	Input       Input  `yaml:"input"`
```
TristonianJones: Let's shift this to be `Input map[string]InputValue` and add a second field called `InputContext` which refers to a struct containing the `ContextExpr string`:

```go
type Case struct {
        Name        string `yaml:"name"`
	Description string `yaml:"description"`
	Input       map[string]*InputValue  `yaml:"input"`
	InputContext *InputContext `yaml:"input_context"`
        Output *Output `yaml:"output"`
}
```

```go
type InputContext struct {
   ContextExpr string `yaml:"context_expr"`
}
```

aakash070: Do we need a separate struct to contain the `ContextExpr` string or is it to maintain parity with the textproto schema? We could also move this up alongside the input map:
```
type Case struct {
        Name        string `yaml:"name"`
	Description string `yaml:"description"`
	Input       map[string]*InputValue  `yaml:"input"`
	InputContext string `yaml:"input_context"`
        Output *Output `yaml:"output"`
}
```

TristonianJones: @aakash070 I kept the same fields for parity between the two different file formats. Context inputs are the less common case. Another alternative is to keep the `InputContext` and mark it as inline which would hoist the `context_expr` field up without making users type `input_context` first:

```go
type Case struct {
        Name        string `yaml:"name"`
	Description string `yaml:"description"`
	Input       map[string]*InputValue  `yaml:"input,omitempty"`
	InputContext *InputContext `yaml:",inline,omitemtpy"`
        Output *Output `yaml:"output"`
}

type InputContext struct {
   ContextExpr string `yaml:"context_expr"`
}
```

```yml
# Populates a `Case` object with an `InputContext` struct where
# `ContextExpr` is set to the expression
test_case:
  name: ...
  context_expr: "qualified.ContextMsg{...}"
  output: ...
```

tools/celtest/test_runner.go
```diff
@@ -0,0 +1,698 @@
...
+
+// TestRunnerOption is a functional interface that can be used to configure a Test Runner in the
+// following ways:
+// - set the Compiler
+// - add Input Expressions
+// - set the test suite file path
+// - set the test suite parser based on the file format: YAML or Textproto
+type TestRunnerOption func(*TestRunner) (*TestRunner, error)
+
+// TestSuiteParserTextproto is a functional interface that provides a cel.expr.conformance.test.TestSuite
```
TristonianJones: nit: the options should describe what they do and can omit the `"is a functional interface that"` since the preference is that Go docs are phrased `Noun <verb> ...` meaning you'd want to skip right to `TestSuiteParserTextproto provides a `

tools/celtest/test_runner.go
```diff
@@ -0,0 +1,698 @@
...
+// message. The message can be serialized in a Textproto file or derived programmatically.
+type TestSuiteParserTextproto func(*testing.T, any) (*conformancepb.TestSuite, error)
```
TristonianJones: Can you help me understand the `derived programmatically` comment? Do you mean you could provide a `conformancepb.TestSuite` as input?

tools/celtest/test_runner.go
```diff
@@ -0,0 +1,698 @@
...
+// The TestRunner provides the following methods:
+// - Programs: Creates a list of CEL programs from the input expressions.
+// - Tests: Creates a list of tests from the test suite file.
+// - ExecuteTest: Executes a single
+type TestRunner struct {
+	compiler.Compiler
+	Expressions              []compiler.InputExpression
+	TestSuiteFilePath        string
+	FileDescriptorSetPath    string
+	TestSuiteParserTextproto TestSuiteParserTextproto
```
TristonianJones: Can these parsers be hidden as an implementation detail? Should the TestRunner just be able to parse test suites instead?

tools/celtest/test_runner.go
```diff
@@ -0,0 +1,698 @@
...
+}
+
+// NewTestRunner creates a Test Runner with the provided options.
+// The options can be used to:
+// - configure the Compiler used for parsing and compiling the input expressions
+// - configure the Test Runner used for parsing and executing the tests
+func NewTestRunner(opts ...any) (*TestRunner, error) {
+	testRunnerOpts := make([]TestRunnerOption, 0, len(opts)+1)
+	testCompilerOpts := make([]any, 0, len(opts))
+	if fileDescriptorSetPath != "" {
```
TristonianJones: I would recommend supporting option creation from flags be a separate step fro the `NewTestRunner` call since you could get some undesirable behavior when the flags and options collide. 

As a general comment it would be better to accept `TestRunnerOption` typed options only, and allow for the configuration of a `compiler` as a `TestRunnerOption` as well.

```go
func TestRunnerCompiler(compiler *Compiler) TestRunnerOption {
    return func(tr *TestRunner) (*TestRunner, error) {
      tr.Compiler = compiler
      return tr, nil
    }
    
  // Possible convenience option
 func TestRunnerCompilerFromConfig(conf *env.Config) TestRunnerOption {
     return func(tr *TestRunner) (*TestRunner, error) {
        env, err := cel.NewEnv(cel.FromConfig(conf))
        if err != nil {
           return nil, err
        }
        c, err := compiler.NewCompiler(compiler.Env(env))
        if err != nil {
            return nil, err
        }
        return TestRunnerCompiler(compiler)(tr)
     }
 }
```

This would allow you to compose the compiler and test runner creation better with fewer configuration errors:

```go
c, err := compiler.NewCompiler(compilerOpts)) 
runnerOpts = append(runnerOpts, TestRunnerCompiler(c))
return NewTestRunner(runnerOpts...)
```

tools/celtest/test_runner.go
```diff
@@ -0,0 +1,698 @@
...
+	fds := &descpb.FileDescriptorSet{}
+	if err := proto.Unmarshal(bytes, fds); err != nil {
+		return nil, fmt.Errorf("failed to unmarshal file descriptor set file %q: %v", fileDescriptorSetPath, err)
+	}
+	return fds, nil
+}
+
+// DefaultTestSuiteParserTextproto is a default parser for a test suite file serialized in a
+// Textproto file. The test suite is a cel.expr.conformance.test.TestSuite message.
+func DefaultTestSuiteParserTextproto(t *testing.T, path any) (*conformancepb.TestSuite, error) {
```
TristonianJones: If this is the default, do we want to let people access it?

tools/celtest/test_runner.go
```diff
@@ -0,0 +1,698 @@
...
+	TestSuiteParserTextproto TestSuiteParserTextproto
+	TestSuiteParserYAML      TestSuiteParserYAML
+}
+
+// Test represents a single test case to be executed. It encompasses the following:
+// - Name: The name of the test case.
+// - Input: The input to be used for evaluating the CEL expression.
+// - ResultMatcher: A function that takes in the result of evaluating the CEL expression and
+// returns a TestResult.
+type Test struct {
```
TristonianJones, line 226: If these are all internal fields, consider making them private and offering a constructor function to create the test which accepts these inputs?

```
func NewTest(name string, input cel.Activation, output ref.Val)

func NewTestWithMatcher(name string, input cel.Activation, matcher func(ref.Val, error) TestResult)
```

tools/celtest/test_runner.go
```diff
@@ -0,0 +1,698 @@
...
+	Name          string
+	Input         interpreter.Activation
+	ResultMatcher func(ref.Val, error) TestResult
+}
+
+// TestResult represents the result of a test case execution. It contains the validation result
+// along with the expected result and any errors encountered during the execution.
+type TestResult struct {
+	Success bool
```
TristonianJones, line 247: Does success mean the execution completed successfully or that the result matcher condition was satisfied? Consider adding a few comments on the fields to explain how they behave

TristonianJones (changes requested):
Please add test cases for the `test_runner.go` and `compiler.go` as part of this PR. Thank you!

tools/celtest/test_runner.go
aakash070: Thanks for pointing, yes, these can be made private methods now. 
I was thinking about a specific case when I made these as public methods, where the tests would be triggered by providing Compiler and Test Runner options programmatically rather than via the build rule. In such a case, the test suite parser would also need to be configured via a Test Runner option. But, I updated `TestSuiteParser` to handle such cases where the test suite file contains messages/objects in standard format and missed changing these back to private methods.

tools/celtest/test_runner.go
aakash070: We may need to allow the parsers to remain configurable by the users. This can help handle cases where the test suite is:
1. Either derived from a Textproto/YAML file whose contents are not in the standard test suite format, i.e. neither a cel.spec.expr.conformance.test.TestSuite message nor a test.TestSuite object.
2. Or created dynamically by the users at runtime. For example: there are cases where the users want to generate the test suite by leveraging the data stored in their db rather than maintaining test suite files.

tools/celtest/test_runner.go
aakash070: This refers to cases where `conformancepb.TestSuite` is generated dynamically rather than being stored in file format. This would permit the users to perform operations like db lookups or invoke rpcs if needed to create the test suite. 
Will update the documentation here.

tools/celtest/test_runner.go
aakash070: It validates whether the result matcher condition was satisfied. Will add comments.

tools/celtest/test_runner.go
aakash070: Updated the documentation

tools/celtest/test_runner.go
aakash070: We may still need to support cases where the tests are triggered via build rules but custom functions need to be configured in the CEL env. For such a case, we may need to accept cel.EnvOption as type of option and modify the env accordingly.

tools/celtest/test_runner.go
```diff
@@ -0,0 +1,708 @@
...
+	ParseYAML(any) (*test.Suite, error)
+}
+
+type tsParser struct {
+	TestSuiteParser
+}
+
+// ParseTextproto parses a test suite file in Textproto format.
+func (p *tsParser) ParseTextproto(path any) (*conformancepb.TestSuite, error) {
+	filePath := path.(string)
```
TristonianJones: If the path only works as a string, should it be switched from `any` to `string` in the interface and implementations?

tools/celtest/test_runner.go
```diff
@@ -0,0 +1,708 @@
...
+	if tr.Compiler == nil {
+		return nil, fmt.Errorf("compiler is not set")
+	}
+	e, err := tr.CreateEnv()
+	if err != nil {
+		return nil, err
+	}
+	var programs []cel.Program
+	for _, expr := range tr.Expressions {
+		ast, _, err := expr.CreateAST(tr.Compiler)
```
TristonianJones, line 337: Can we add a TODO to propagate the metadata from CreateAST back with the Program instances? Possibly as another struct which combines this information?

tools/celtest/test_runner.go
```diff
@@ -0,0 +1,708 @@
...
+func (tr *TestRunner) eval(expr string) (ref.Val, error) {
+	e, err := tr.CreateEnv()
+	if err != nil {
+		return nil, err
+	}
+	e, err = e.Extend(cel.OptionalTypes())
+	if err != nil {
+		return nil, fmt.Errorf("e.Extend() failed: %v", err)
+	}
+	ast, iss := e.Compile(expr)
```
TristonianJones, line 565: You can probably wrap the `CreateEnv` and extension with optional types into a `sync.DoOnce` to avoid building it each time. Or you can just build it at TestRunner init which would probably be easiest.

tools/celtest/test_runner.go
aakash070: I had kept this as `any` to allow for custom implementations of the interface to pass non-string input args to the methods. But that requirement can be handled later when needed. Will make this change.

tools/celtest/test_runner.go
aakash070: The underlying `Compiler` providing the `env` already maintains it as a  singleton object. Will modify the same to make use of `sync.Once` to set up the env.
Will also add the optional extension in the compiler initialization.

aakash070:
@TristonianJones I have added tests for `test_runner.go`. Tests existed for `compiler.go`.
I have moved one of the tests for custom policy metadata options from compiler to test_runner. Please check.

tools/celtest/test_runner.go
```diff
@@ -0,0 +1,707 @@
...
+		t.Run(test.name, func(t *testing.T) {
+			err := tr.ExecuteTest(t, programs, test)
+			if err != nil {
+				t.Fatalf("error executing test: %v", err)
+			}
+		})
+	}
+}
+
+func testRunnerOptions(testRunnerOpts []TestRunnerOption, testCompilerOpts ...any) []TestRunnerOption {
```
TristonianJones, line 100: Could you actually make this a functional option to configure common options from flags? Can be done in a follow-up

```go
func TestOptionsFromFlags() TestRunnerOption {
   return func(tr *TestRunner) (*TestRunner, error) {
      opts := []TestRunnerOption{
          DefaultTestSuiteParser(testSuitePath),
          AddFileDescriptorSet(...),
          ...
       }
       var err error
       for _, o := range opts {
         tr, err = o(tr)
         if err != nil {
           return nil, err
         }
       }
       return tr
   }
}

TristonianJones (approved):

tools/celtest/test_runner.go
TristonianJones: Resolving for now, in order to merge, but please update in the next PR

PR #1160, opened 2025-04-10
Prompt generation for AI-assisted authoring based on a CEL environment

TristonianJones:
Prompt generation for AI-assisted authoring based on a CEL environment

Currently a template is used to generate a prompt including variables, functions, and macros
available to a CEL author. This prompt is intended to serve as an example of how to use CEL
documentation methods to assist with prompt stuffing / generation for various workflows.

### DIFF
--- a/cel/BUILD.bazel
+++ b/cel/BUILD.bazel
@@ -18,8 +18,10 @@ go_library(
         "optimizer.go",
         "options.go",
         "program.go",
+        "prompt.go",
         "validator.go",
     ],
+    embedsrcs = ["//cel/templates"],
     importpath = "github.com/google/cel-go/cel",
     visibility = ["//visibility:public"],
     deps = [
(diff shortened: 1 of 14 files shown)


jcking (approved):

amcrn:
This feature is extremely helpful. After integrating against it, I've noticed a few enhancements that might be useful:

* Support customizing `hiddenFunctions` in `prompt.go`
* Support multi-line `OverloadExamples` (ex: if an extension does a pretty-print of JSON, the example JSON output is rendered unoptimally)
* If `OverloadExamples` is provided, the function signature is no longer printed
  * Example: if `strings.quote` has an example provided, then `strings.quote(string) -> string` is no longer printed.
  * Printing the signature can provide useful information, in addition to the examples.
* The `doc` on a `VariableDecls` isn't printed as `authoring.tmpl` only has `{{define "variable"}}{{.Name}} is a {{.Type}}`

All of these can be worked around by implementing a customized `Prompt`, which I figure is the intent, but I figure I'd pass it along if these observations are useful for upstream enhancements/changes. Thanks!

TristonianJones:
@amcrn Thanks for the feedback. I'm happy to accept and review PRs along these lines if you'd like to contribute them.

PR #1170, opened 2025-04-23
fix test runner test cases

aakash070:
This change removes additional quotes around raw expressions passed on to the test runner and fixes tests.


TristonianJones (commented):
Are there unit tests that are running for these cases? How were they not caught before?

aakash070:
There was no check to verify that tests are created by the test runner, and, since the provided file path did not have an extension, neither YAML nor Textproto test parser was able to create the tests. In such a case the passed expression could not be validated as no tests were triggered and hence the test case would not fail. Have added empty checks now.

TristonianJones (approved):

PR #1172, opened 2025-04-27
feature: Late Bind Function Calls at Evaluation Time

hyp0th3rmi4:
# Pull Requests Guidelines

This PR provides a minimal but fully functional prototype implementation for Late-binding functions at program evaluation time. It supports the discussion originated in: https://github.com/google/cel-go/discussions/1081, to demonstrate that it can be done.

The implementation comes fully tested in the new code that has been added, which has been kept at minimum, by trying to limit changes in the code base.

The model uses `functions.Overload` as the primary abstraction to supply run-time implementations of functions to replace the same overloads bound to the call nodes during tree parsing and checking.

# Files Changed:

- `interpreter/activation.go`: added a new interface LateBindActivation with ability to resolve overloads given an overload identifier.
- `intepreter/decorators.go`: added the implementation of an InterpretableDecorator that manipulate call evalXXX nodes to defer the function bind at evaluation time.
- `intepreter/interpreter.go`: added PlannerOption to inject latebind decorator.
- `interpreter/late_binding.go`: implementation of the IntepretableCall nodes that perform late binding.
- `cel/program.go`: integration of the creation fo the late binding planner option into the interpretable construction.
- `cel/options.go`: definition of the late binding options to be supplied to `Program.Eval(....)`.

## Limitations:

This implementation is based on the `functions.Overload` and the `interpreter.Dispatcher` abstractions. As a result, runtime signature checks are very limited, but the additional burden is kept to minimum. This is because differently from `decls.FunctionDecl` and `decls.OverloadDecl`, `functions.Overload` does no longer maintain type information to perform any check.

If an implementations based on the abstractions in the decls namespace is preferred, I can uplift the PR.


hyp0th3rmi4:
Some considerations on existing limitations:

## Function Filters

At present time we allow for all functions to be overloaded, maybe it would be a good idea to control the behaviour for operators that are implemented as function. The current code already avoids to manipulate `evalEq` and `evalNe` but perhaps we should be able to allow for filtering out arithmetic operators.

## Runtime Check for Signatures
This capability is implemented in a rather limited number, which relies on matching the specific function overload configured in the evalXXX node with the function overload supplied at runtime (unary -> unary, binary -> binary, function -> function). This is good as a first implementation but very low level. We can address this issue by moving from `functions.Overload` to `decls.FunctionDecl` and `decls.OverloadDecl`. If we do this, we may need to:

- expose the overload configured in `decls.OverloadDecl` via a `Binding()` method to provide access from outside the package.
- configure the lateBindEval interpretable with the selected `decls.OverloadDecl`
- uplift the lateBindActivation to rely on `decls.OverloadDecl` and no longer a dispatcher implementation.

Alternativaly, in the `cel` package we can provide a method that produces a LateBindActivation which is already checked against the defined functions in the environment implementation. With this approach, we do an ahead of time compatibility check versuse a real run-tme check.

TristonianJones:
Fixed by #1483

PR #1178, opened 2025-05-12
Create an util method to convert rpc status to eval status

ChinmayMadeshi:
We are removing dependency of eval proto on rpc status via ErrorSet and so for usages which are already dependent on rpc status, we have created an util for converting rpc status to eval status.


WORKSPACE
```diff
@@ -1,5 +1,6 @@
 workspace(name = "cel_go")
 
+load("@bazel//tools/build_defs/repo:local.bzl", "local_repository")
```
aakash070: maybe we can avoid pushing this?

cel/io_test.go
```diff
@@ -27,8 +27,10 @@ import (
 	"github.com/google/cel-go/common/operators"
 	"github.com/google/cel-go/common/types"
 
+    anypb "google.golang.org/protobuf/types/known/anypb"
```
aakash070: nit: please format the file with `go fmt`

WORKSPACE
ChinmayMadeshi: Yes, removed.

cel/io_test.go
ChinmayMadeshi: Done.

cel/io.go
```diff
@@ -114,6 +115,21 @@ func RefValueToValue(res ref.Val) (*exprpb.Value, error) {
 	return ValueAsAlphaProto(res)
 }
 
+// RPCStatusToEvalErrorStatus converts an RPC status to a celpb status. Status value is intended to
+// be wire and field compatible with `google.rpc.Status`.
+func RPCStatusToEvalErrorStatus(s *rpcpb.Status) (*celpb.Status, error) {
```
TristonianJones: Could you validate that we could just move the error and unknown conversion into `ValueAsProto` if we just handled error and unknown? Likewise we could offer a similar abstraction for ProtoAsValue

ChinmayMadeshi: ErrorSet and UnknownSet are a part of ExprValue. Value doesn't have them. And so the iam team is dependent on ExprValue. Should we change the function signature here to convert them ? or add a wrapper method on this ?

TristonianJones: I would consider introducing an `ExprValue` converter since it's part of the CEL package of values.

conformance/BUILD.bazel
```diff
@@ -48,6 +48,7 @@ _TESTS_TO_SKIP = [
     "fields/qualified_identifier_resolution/map_value_repeat_key_heterogeneous",
     "macros/map/map_extract_keys",
     "timestamps/duration_converters/get_milliseconds",
+    "optionals/optionals/map_optional_select_has",
```
TristonianJones, line 51: Why is this here? There's an identical test in `cel_test.go` for `has({'foo': optional.none()}.foo.value)` which passes, so this is pretty confusing.

ChinmayMadeshi: Without adding this test here, after pulling in the latest version of cel-spec the conformance tests are failing. https://github.com/google/cel-go/pull/1178/checks?check_run_id=42100401376 , hence i added this here upon discussion with Sokwhan and plan to create a github issue for this.

aakash070: Conformance tests have `featureEnableErrorOnBadPresenceTest` enabled for the env causing the failure.

cel/io.go
ChinmayMadeshi: Done

conformance/BUILD.bazel
TristonianJones: In the future if you can bump the cel-spec in a separate PR that will help with review.

cel/io.go
```diff
@@ -126,6 +128,55 @@ func ValueAsAlphaProto(res ref.Val) (*exprpb.Value, error) {
...
+	switch res := res.(type) {
+	case *types.Unknown:
+		return &celpb.ExprValue{
+			Kind: &celpb.ExprValue_Unknown{
+				Unknown: &celpb.UnknownSet{
+					Exprs: res.IDs(),
+				},
+			}}, nil
+	case *types.Err:
+		s := status.Convert(res.Unwrap()).Proto()
```
TristonianJones: Please don't take a direct dependency on grpc status here, as this is intended for gRPC APIs to propagate status up to proto APIs. Instead, just convert the error to a CEL status error proto.

cel/io.go
```diff
@@ -126,6 +128,55 @@ func ValueAsAlphaProto(res ref.Val) (*exprpb.Value, error) {
 	return alpha, err
 }
 
+// RPCStatusToEvalErrorStatus converts an RPC status to a celpb status. Status value is intended to
+// be wire and field compatible with `google.rpc.Status`.
+func RPCStatusToEvalErrorStatus(s *rpcpb.Status) (*celpb.Status, error) {
```
TristonianJones: Please remove this method. We don't need two levels of translation, just the one internal one.

cel/io.go
```diff
@@ -126,6 +126,37 @@ func ValueAsAlphaProto(res ref.Val) (*exprpb.Value, error) {
 	return alpha, err
 }
 
+// RefValToExprValue converts between ref.Val and cel.expr.ExprValue.
+// The result ExprValue is the serialized proto form.
+func RefValToExprValue(res ref.Val) (*celpb.ExprValue, error) {
```
TristonianJones: Would you mind updating this method to return v1alpha1.ExprValue as the output? The convention is as follows:

```
RefValToExprValue // v1alpha1
ValueAsAlphaExprValueProto // v1alpha1
ValueAsExprValueProto // canonical
```

The alpha proto implementations proxy into the canoncial value and convert the type to v1alpha1before returning.

MODULE.bazel
```diff
@@ -83,6 +83,7 @@ use_repo(
     "in_gopkg_yaml_v3",
     "org_golang_google_genproto_googleapis_api",
     "org_golang_google_genproto_googleapis_rpc",
+    "org_golang_google_grpc",
```
aakash070: can this be removed now?

aakash070 (approved):

TristonianJones (approved):

PR #1180, opened 2025-05-18
Add support for global constants to constant folder

zeitgeist87:
Global constants like enums are represented in the AST as identifiers. These identifiers can be looked up in the AST reference map. If the entry in the reference map has a value associated with it, we can treat it as a constant and attempt to turn it into a literal.

This reduces the size of the AST and unlocks more optimization opportunities.

### DIFF
--- a/cel/folding_test.go
+++ b/cel/folding_test.go
@@ -24,6 +24,7 @@ import (
 	"google.golang.org/protobuf/proto"
 
 	"github.com/google/cel-go/common/ast"
+	"github.com/google/cel-go/common/types"
 	"github.com/google/cel-go/common/types/ref"
 
 	proto3pb "github.com/google/cel-go/test/proto3pb"
@@ -32,8 +33,9 @@ import (
 
 func TestConstantFoldingOptimizer(t *testing.T) {
 	tests := []struct {
-		expr   string
-		folded string
+		expr        string
+		folded      string
+		knownValues map[string]any
 	}{
 		{
 			expr:   `[1, 1 + 2, 1 + (2 + 3)]`,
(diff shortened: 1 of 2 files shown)


TristonianJones (dismissed):

cel/folding_test.go
```diff
@@ -279,15 +280,59 @@ func TestConstantFoldingOptimizer(t *testing.T) {
...
+			folded: `[1, 2, 3]`,
+		},
 	}
 	e, err := NewEnv(
 		OptionalTypes(),
 		EnableMacroCallTracking(),
 		Types(&proto3pb.TestAllTypes{}),
-		Variable("x", DynType))
+		Variable("x", DynType),
+		Constant("c", IntType, types.Int(proto3pb.ImportedGlobalEnum_IMPORT_BAZ)),
```
TristonianJones, line 335: Would you mind adding a non-proto constant as well? Enums will be in the reference map, but it's worth seeing what this does with something that doesn't usually go into the reference map like a string value

cel/folding_test.go
```diff
@@ -279,15 +280,59 @@ func TestConstantFoldingOptimizer(t *testing.T) {
...
 		t.Fatalf("NewEnv() failed: %v", err)
 	}
+	e, err = e.Extend(
+		Constant("l", ListType(StringType), types.NewStringList(e.adapter, []string{"foo", "bar", "baz"})),
+	)
+	if err != nil {
+		t.Fatalf("Extend() failed: %v", err)
+	}
+	e, err = e.Extend(
+		Constant("o", ObjectType("google.expr.proto3.test.TestAllTypes"),
```
TristonianJones: Does this cause an error if you try to convert the AST to a proto when you don't run constant-folding? I had been toying with restricting the input types to `Constant` so I'm wondering if `Globals` is a better reference point considering the default `Activation` would inherit a runtime value like the one you're specifying here.

zeitgeist87: Yes, it returns an error. I get `failed: unsupported constant kind: google.expr.proto3.test.TestAllTypes`, but surprisingly also `failed: unsupported constant kind: list(dyn)` for the list example.

zeitgeist87: Only a handful of types is actually supported: https://github.com/google/cel-go/blob/f6d3c92171c2c8732a8d0a4b24d6729df4261520/common/ast/conversion.go#L591

TristonianJones: Yes, that's what I expected. Could you try this with `Globals` as an alternative which will automatically extend the input activation with a set of predefined constants? That might work better.

zeitgeist87: Yes, I can make it work using `Globals`. Is that what you had in mind: https://github.com/google/cel-go/commit/f28f6bbce581a781b05fd1b3c81f12105cf6aae3

However, we probably need a better way to turn a `ProgramOption` into an `EnvOption`. At least I couldn't find any. I think the only other way would be to implement the `Library` interface and use `Lib()`

zeitgeist87: I updated this PR with support for `Globals`.

zeitgeist87: However, this makes the constant folder pretty aggressive, so I added configuration options that are disabled by default.

cel/options.go
```diff
@@ -399,6 +399,16 @@ func registerFiles(reg customTypeRegistry, files *protoregistry.Files) error {
 // ProgramOption is a functional interface for configuring evaluation bindings and behaviors.
 type ProgramOption func(p *prog) (*prog, error)
 
+// Adds one or more ProgramOptions to the environment.
+//
+// The options are applied before any other options when creating a program for evaluation.
+func ProgOpts(options ...ProgramOption) EnvOption {
```
TristonianJones: Maybe call this `DefaultProgramOptions`? Alternatively, you could pass an `Activation` into `FoldVariablesWithKnownValues` and remove the `ProgOpts` option? 

You could also have a single option which is called `FoldKnownValues` which optionally takes an `Activation` to simplify the setup and combine the behaviors which might be hard to differentiate for the uninitiated.

zeitgeist87: Done.

TristonianJones (approved):

PR #1183, opened 2025-05-30
add support for handling unknown expression ids in test result

aakash070:
This change adds support for Unknown Set test result. A new test runner option has been added which enables the `OptPartialEval` eval option.
Note: This change adds partial evaluation support for cases where top level variables are missing in the input activations.


tools/celtest/test_runner.go
```diff
@@ -627,7 +652,30 @@ func (tr *TestRunner) createResultMatcherFromPB(t *testing.T, testCase *conforma
 			return failureResult
 		}, nil
 	case *conformancepb.TestOutput_Unknown:
-		//  TODO: to implement
+		// Validate that all expected unknown expression ids are returned by the evaluation result.
+		return func(out ref.Val, err error) TestResult {
+			expectedUnknownIDs := testOutput.Unknown.GetExprs()
+			if err == nil && types.IsUnknown(out) {
+				actualUnknownIDs := out.Value().(*types.Unknown).IDs()
+				if len(expectedUnknownIDs) != len(actualUnknownIDs) {
```
jnthntatum: could you use cmp here instead? https://pkg.go.dev/github.com/google/go-cmp/cmp#example-Diff-Testing

tools/celtest/test_runner.go
```diff
@@ -417,6 +419,17 @@ func fileDescriptorSet(path string) (*descpb.FileDescriptorSet, error) {
 	return fds, nil
 }
 
+// PartialEvalProgramOption returns a TestRunnerOption which enables partial evaluation for the CEL
+// program.
+//
+// Note: Partial evaluation where only root level variables are missing is supported.
```
jnthntatum: nit/clarity: you could just say you are using Env.PartialVars(activation) to infer unknown root vars based on the activation.

tools/celtest/testdata/raw_expr_tests.yaml
```diff
@@ -32,3 +32,19 @@ section:
             value: 42
         output:
           value: false
+  - name: "unknown"
```
jnthntatum, line 45: for completeness, could you add a case where multiple expr ids are returned as unknown?

jnthntatum (changes requested):
LGTM, some minor suggestions

jnthntatum (approved):

PR #1186, opened 2025-06-05
Add documentation for YAML quirks in celpolicy

jnthntatum:
Also add top-level description field.

The celpolicy parser has special handling for CEL expressions embedded in
the YAML document in order to preserve location information. This is used
to help the CEL compiler correctly highlight error location.

As a consequence, the CEL expression will include characters that are
normally not considered meaningful in YAML, and the literal and folded style
for multiline strings are treated identically.

### DIFF
--- a/policy/parser.go
+++ b/policy/parser.go
@@ -45,12 +45,13 @@ func NewPolicy(src *Source, info *ast.SourceInfo) *Policy {
 
 // Policy declares a name, rule, and evaluation semantic for a given expression graph.
 type Policy struct {
-	name     ValueString
-	imports  []*Import
-	rule     *Rule
-	semantic semanticType
-	info     *ast.SourceInfo
-	source   *Source
+	name        ValueString
+	description ValueString
+	imports     []*Import
+	rule        *Rule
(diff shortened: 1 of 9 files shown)


policy/parser.go
```diff
@@ -615,15 +653,19 @@ func (p *parserImpl) ParsePolicy(ctx ParserContext, node *yaml.Node) *Policy {
...
 		switch fieldName {
 		case "imports":
 			p.parseImports(ctx, policy, val)
 		case "name":
 			policy.SetName(ctx.NewString(val))
+		case "description":
+			policy.SetDescription(p.newStrictString(val))
+			// Since the description field was not supported initially, some
+			// clients rely on the ability to intercept it.
+			p.visitor.PolicyTag(ctx, keyID, fieldName, val, policy)
```
l46kok, line 668: I feel like maybe the visiting the tag should always be done (kind of like an overridable `postVisit`), but carving out a special case for description seems odd. We don't have to tackle it here -- right now, I think this best preserves backwards compatibility

l46kok (approved):

PR #1189, opened 2025-06-09
Add test runner option to configure a custom test suite parser

aakash070:
This change adds the following:
1. A TestRunnerOption which allows configuring a custom test suite parser in the test runner.
2. A TestRunnerOption which allows configuring a custom compiler in the test runner

### DIFF
--- a/tools/celtest/test_runner_test.go
+++ b/tools/celtest/test_runner_test.go
@@ -230,17 +230,19 @@ type tsparser struct {
 	TestSuiteParser
 }
 
+// ParseTextproto implements the ParseTextproto method of the TestSuiteParser interface.
 func (p *tsparser) ParseTextproto(_ string) (*conformancepb.TestSuite, error) {
 	return nil, nil
 }
 
+// ParseYAML implements the ParseYAML method of the TestSuiteParser interface.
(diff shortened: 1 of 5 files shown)


TristonianJones (dismissed):

tools/celtest/test_runner_test.go
```diff
@@ -210,3 +213,47 @@ func TestTriggerTests(t *testing.T) {
...
+		}
+		TriggerTests(t, testRunnerOpts...)
+	})
+}
+
+type tsparser struct {
+	TestSuiteParser
+}
+
+func (p *tsparser) ParseTextproto(path string) (*conformancepb.TestSuite, error) {
```
TristonianJones: If this is just internal to the test file, you don't need to expose it. The function can be lower-case. Here and below

aakash070: These are `TestSuiteParser` interface method implementations.

TristonianJones (dismissed):

tools/celtest/test_runner_test.go
TristonianJones: Would you mind documenting that these methods are implementing the interface?

TristonianJones (approved):

PR #1190, opened 2025-06-24
Add support for nested element type

MisLink:
Recursively find all Struct types, covering the containment of pointer types in the container.

### DIFF
--- a/ext/native.go
+++ b/ext/native.go
@@ -609,7 +609,8 @@ func newNativeTypes(fieldNameHandler NativeTypesFieldNameHandler, rawType reflec
 	var iterateStructMembers func(reflect.Type)
 	iterateStructMembers = func(t reflect.Type) {
 		if k := t.Kind(); k == reflect.Pointer || k == reflect.Slice || k == reflect.Array || k == reflect.Map {
-			t = t.Elem()
+			iterateStructMembers(t.Elem())
+			return
 		}
 		if t.Kind() != reflect.Struct {
 			return
--- a/ext/native_test.go
+++ b/ext/native_test.go
@@ -941,6 +941,67 @@ func TestNativeStructEmbedded(t *testing.T) {
 	}
 }
 
+type TestNestedStruct struct {
+	ListVal []*TestNestedType
+}
+
+func TestNativeNestedStruct(t *testing.T) {
+	var nativeTests = []struct {
+		expr string
+		in   any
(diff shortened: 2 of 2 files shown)


TristonianJones:
@MisLink I was able to successfully run an equivalent expression using the existing object structure without the nested type traversal. Perhaps you need to test a different case to exercise how the PR impacts users?

MisLink:
@TristonianJones Could you please share the test cases you used? I tried with the one I added myself, but it throws an error unless I apply this modification.

```
=== RUN   TestNativeNestedStruct
=== RUN   TestNativeNestedStruct/[0]
    /Users/mi/projects/cel-go/ext/native_test.go:985: env.Check(test.ListVal.exists(x, x.custom_name == "name")) failed: ERROR: <input>:1:25: unexpected failed resolution of 'ext.TestNestedType'
         | test.ListVal.exists(x, x.custom_name == "name")
         | ........................^
--- FAIL: TestNativeNestedStruct (0.00s)
    --- FAIL: TestNativeNestedStruct/[0] (0.00s)
```

TristonianJones:
@MisLink, my apologies, that wasn't a very good reply on my part. I added the following case to the `TestNativeTypes` function in `ext/native_test.go`:

```go
{
	expr: `t.ListVal.exists(x, x.custom_name == "name")`,
	in: map[string]any{
		"t": TestAllTypes{ListVal: []*TestNestedType{{NestedCustomName: "name"}}},
	},
	out:     true,
	envOpts: []any{cel.Variable("t", cel.ObjectType("ext.TestAllTypes")), ParseStructTags(true)},
},
```

MisLink:
@TristonianJones The `TestAllTypes` struct includes a field of type `*TestNestedType` at the beginning. The previous logic was able to handle this case correctly, which ensured that the `TestNestedType` type was included during the parsing phase. I didn’t trace the execution phase in detail, but in practice, both this type and its nested types could be handled properly at runtime.

However, if a struct contains only deeply nested types (as is the case in the new struct I added for testing), the parser is unable to detect those types, which leads to runtime errors during execution. This is exactly why I added that specific struct for testing purposes (apologies—I should have mentioned that earlier!)

TristonianJones:
@MisLink ah, I see, thanks for the explanation. I didn't catch that the first time.

TristonianJones (approved):

PR #1199, opened 2025-07-20
Init function bindings on environment init

beldmian:
While performing some performance testing for one of my projects, I saw that in case of creating many (a lot of) CEL programs in the same environment, creation of function bindings for them can take a significant percentage of CPU time, because they are being regenerated on every program creation, which, as I can see from code, feels really excessive. This PR suggest a way to perform calculation of bindings once for environment configuration and then reuse this bindings in program creation.

Probably, there also should be provided a way to change this behavior through options, let me know if that feels more acceptable with them.

# Pull Requests Guidelines

development branch.

## Commit Messages

```

```

## Reviews

* Perform a self-review.

## Merging

  if applicable.

### DIFF
--- a/cel/env.go
+++ b/cel/env.go
@@ -130,19 +130,21 @@ func FormatCELType(t *Type) string {
 // Env encapsulates the context necessary to perform parsing, type checking, or generation of
 // evaluable programs for different expressions.
 type Env struct {
-	Container        *containers.Container
-	variables        []*decls.VariableDecl
-	functions        map[string]*decls.FunctionDecl
+	Container       *containers.Container
+	variables       []*decls.VariableDecl
+	functions       map[string]*decls.FunctionDecl
+	macros          []Macro
+	contextProto    protoreflect.MessageDescriptor
+	adapter         types.Adapter
+	provider        types.Provider
+	features        map[int]bool
+	appliedFeatures map[int]bool
+	libraries       map[string]SingletonLibrary
+	validators      []ASTValidator
+	costOptions     []checker.CostOption
+
+	funcBindOnce     sync.Once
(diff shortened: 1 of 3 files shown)


TristonianJones:
@beldmian ideally the programs would be cached rather than recreated. Is that perhaps an option?

beldmian:
@TristonianJones Programs are not serializable (at least in general way), which is needed to make them cached in my use case, the nearest to program thing serializable, as I can see, is checked expression of its AST.

TristonianJones:
@beldmian I see, are you having to page in / out a lot of ASTs in the same environment?

beldmian:
@TristonianJones Yes, to be precise, I am precompiling a lot of cel expressions in one environment and then they are being shipped to another one (there they are being executed not really many times, but the execution should be fast and at the same time the consumption of resources should be small enough).

cel/env.go
```diff
@@ -793,6 +796,16 @@ func (e *Env) configure(opts []EnvOption) (*Env, error) {
 		}
 	}
 
+	// Precalculate function bindings
+	e.functionBindings = []*functions.Overload{}
```
TristonianJones: Could you perhaps move the binding resolution into the `newProgram()` call as `sync.Once`? In some cases the bindings won't need to be resolved at all, so it would be great to defer this to where the program is actually built.

beldmian: Yep, did it, also reverted changes in the tests since now errors behavior is unchanged

TristonianJones: Thank you!

TristonianJones (approved):

TristonianJones:
@beldmian thanks for the contribution!

PR #1200, opened 2025-07-22
Support cost estimation and tracking for the regex library

maskri17:


ext/regex.go
```diff
@@ -151,14 +154,27 @@ func (r *regexLib) CompileOptions() []cel.EnvOption {
...
+		cel.CostTrackerOptions(
+			interpreter.OverloadCostTracker("regex_extract_string_string", trackExtractCost()),
+			interpreter.OverloadCostTracker("regex_extractAll_string_string", trackExtractAllCost()),
+			interpreter.OverloadCostTracker("regex_replace_string_string_string", trackReplaceCost()),
+			interpreter.OverloadCostTracker("regex_replace_string_string_string_int", trackReplaceCost()),
+		),
+	}
 }
 
 func compileRegex(regexStr string) (*regexp.Regexp, error) {
```
TristonianJones: I'm not sure this is adding much since you're having to check the error state immediately after calling. Consider just doing this inline?

ext/README.md
```diff
@@ -945,3 +945,70 @@ Examples:
...
+`optional` string. If no match is found, it returns an optional none value. An error will
+be thrown for invalid regex or for multiple capture groups.
+
+    regex.extract(target: string, pattern: string) -> optional<string>
+
+Examples:
+
+    regex.extract('hello world', 'hello(.*)') == optional.of(' world')
+    regex.extract('item-A, item-B', 'item-(\\w+)') == optional.of('A')
+    regex.extract('HELLO', 'hello') == optional.empty()
```
TristonianJones: nit: use `optional.none()` instead

ext/regex.go
```diff
@@ -330,3 +342,80 @@ func extractAll(target, regexStr ref.Val) ref.Val {
 	}
 	return types.NewStringList(types.DefaultTypeAdapter, result)
 }
+
+func estimateExtractCost() checker.FunctionEstimator {
+	return func(c checker.CostEstimator, target *checker.AstNode, args []checker.AstNode) *checker.CallEstimate {
+		if len(args) == 2 {
```
TristonianJones, line 340: Please add comments about what the logic below is doing, e.g.

```go
// compute cost as a function of the size of the target string multiplied by a traversal cost, times the size
// of the regex multiplied by a regex traversal cost. Ensure the resulting size is minimally zero and maximally
// the size of the target string.
...
```

Currently the result size is always the same as the target size, which I think is just the upper bound, but not the actual lower bound.

ext/regex.go
```diff
@@ -330,3 +342,80 @@ func extractAll(target, regexStr ref.Val) ref.Val {
...
+
+func estimateExtractAllCost() checker.FunctionEstimator {
+	return func(c checker.CostEstimator, target *checker.AstNode, args []checker.AstNode) *checker.CallEstimate {
+		if len(args) == 2 {
+
+			targetSize := estimateSize(c, args[0])
+			targetCost := targetSize.Add(checker.FixedSizeEstimate(1)).MultiplyByCostFactor(common.StringTraversalCostFactor)
+			regexCost := estimateSize(c, args[1]).Add(checker.FixedSizeEstimate(1)).MultiplyByCostFactor(common.RegexStringLengthCostFactor)
+
+			resultSize := checker.SizeEstimate{Min: 0, Max: targetSize.Max}
```
TristonianJones: Nice, glad to see the extract all accounts for 0 as the min size and every char in the string as the max size. :)

ext/regex.go
```diff
@@ -330,3 +342,80 @@ func extractAll(target, regexStr ref.Val) ref.Val {
...
+			return &checker.CallEstimate{CostEstimate: checker.CostEstimate(targetCost).Multiply(regexCost), ResultSize: &targetSize}
+		}
+		return nil
+	}
+}
+
+func estimateExtractAllCost() checker.FunctionEstimator {
+	return func(c checker.CostEstimator, target *checker.AstNode, args []checker.AstNode) *checker.CallEstimate {
+		if len(args) == 2 {
+
```
TristonianJones: nit: extra blank line.

ext/regex.go
```diff
@@ -330,3 +342,80 @@ func extractAll(target, regexStr ref.Val) ref.Val {
...
+			targetSize := estimateSize(c, args[0])
```
TristonianJones, line 362: Please add a comment about how the cost is computed and the expected size.

ext/README.md
```diff
@@ -945,3 +945,70 @@ Examples:
 
     {'greeting': 'aloha', 'farewell': 'aloha'}
       .transformMapEntry(keyVar, valueVar, {valueVar: keyVar}) // error, duplicate key
+
+## Regex
+
+Regex introduces functions for regular expressions in CEL.
+
+Note: all functions use the 'regex' namespace. If you are currently using a variable named 'regex', the functions will likely work as intended, however there is some chance for collision. This library depends on the CEL optional type. Please ensure that the cel.OptionalTypes() is enabled when using regex extensions.
```
TristonianJones: Please try to wrap to 80-ish characters

ext/README.md
```diff
@@ -945,3 +945,70 @@ Examples:
...
+
+### Replace
+
+The `regex.replace` function replaces all non-overlapping substring of a regex
+pattern in the target string with a replacement string. Optionally, you can
+limit the number of replacements by providing a count argument. When the count
+is a negative number, the function acts as replace all. Only numeric (\N)
```
TristonianJones, line 963: Do we want to support negative numbers? In other places we usually treat this as an error, and this wasn't really something we addressed in the spec.

ext/regex.go
```diff
@@ -330,3 +342,80 @@ func extractAll(target, regexStr ref.Val) ref.Val {
...
+		}
+		return nil
+	}
+}
+
+func estimateReplaceCost() checker.FunctionEstimator {
+	return func(c checker.CostEstimator, target *checker.AstNode, args []checker.AstNode) *checker.CallEstimate {
+		l := len(args)
+		if l == 3 || l == 4 {
+			targetSize := estimateSize(c, args[0])
```
TristonianJones, line 386: I think to be most accurate we need to look at the size of the replacement as well:

```go
targetSize := estimateSIze(c, args[0])
replacementSize := estimateSize(c, args[2])

allReplacedSize := targetSize.Max * replacementSize.Max
noneReplacedSize := targetSize.Max
resultSize := &checker.SizeEstimate{Min: noneReplacedSize, Max: allReplacedSize}
if noneReplacedSize > allReplacedSize {
  resultSize =  &checker.SizeEstimate{Min: allReplacedSize, Max: noneReplacedSize}
}
```

ext/regex.go
```diff
@@ -330,3 +342,80 @@ func extractAll(target, regexStr ref.Val) ref.Val {
...
+			resultSize := checker.SizeEstimate{Min: 0, Max: targetSize.Max}
+			targetCost := targetSize.Add(checker.FixedSizeEstimate(1)).MultiplyByCostFactor(common.StringTraversalCostFactor)
+
+			regexSize := estimateSize(c, args[1]).Add(checker.FixedSizeEstimate(1)).MultiplyByCostFactor(common.RegexStringLengthCostFactor)
```
TristonianJones: Please rename regexSize to `regexCost`

ext/regex.go
```diff
@@ -330,3 +342,80 @@ func extractAll(target, regexStr ref.Val) ref.Val {
...
+			replacementCost := estimateSize(c, args[2]).Add(checker.FixedSizeEstimate(1)).MultiplyByCostFactor(common.StringTraversalCostFactor)
+			cost := targetCost.Multiply(regexSize).Add(checker.CostEstimate(targetSize).Multiply(replacementCost))
```
TristonianJones: nit: prefer using `targetSize.MultiplyByCost(replacementCost)`

ext/regex.go
```diff
@@ -330,3 +342,80 @@ func extractAll(target, regexStr ref.Val) ref.Val {
...
+			return &checker.CallEstimate{CostEstimate: cost, ResultSize: &resultSize}
+		}
+		return nil
+	}
+}
+
+func trackExtractCost() interpreter.FunctionTracker {
+	return func(args []ref.Val, _ ref.Val) *uint64 {
+		targetSize := float64(actualSize(args[0])) * float64(common.StringTraversalCostFactor)
+		regexSize := float64(actualSize(args[1])) * float64(common.StringTraversalCostFactor)
```
TristonianJones: Don't forget to use `RegexTraversalCostFactor` here

ext/regex.go
```diff
@@ -330,3 +342,80 @@ func extractAll(target, regexStr ref.Val) ref.Val {
...
+		cost := callCost + uint64(float64(targetSize*regexSize))
+		return &cost
+	}
+}
+
+func trackExtractAllCost() interpreter.FunctionTracker {
+	return func(args []ref.Val, _ ref.Val) *uint64 {
+		targetSize := float64(actualSize(args[0])) * float64(common.StringTraversalCostFactor)
+		regexSize := float64(actualSize(args[1])) * float64(common.StringTraversalCostFactor)
```
TristonianJones: Similarly, don't forget about regex traversal here and below.

ext/regex.go
```diff
@@ -330,3 +342,80 @@ func extractAll(target, regexStr ref.Val) ref.Val {
...
+		cost := callCost + uint64(float64(targetSize*regexSize))
+		return &cost
+	}
+}
+
+func trackExtractAllCost() interpreter.FunctionTracker {
+	return func(args []ref.Val, _ ref.Val) *uint64 {
```
TristonianJones: Don't forget to account for the size of the list that's created by looking at the result argument (currently listed as `_`, but should be renamed to `result`)

ext/regex.go
```diff
@@ -330,3 +342,80 @@ func extractAll(target, regexStr ref.Val) ref.Val {
...
+		targetSize := float64(actualSize(args[0])) * float64(common.StringTraversalCostFactor)
+		regexSize := float64(actualSize(args[1])) * float64(common.StringTraversalCostFactor)
+		cost := callCost + uint64(float64(targetSize*regexSize)) + common.ListCreateBaseCost
+		return &cost
+	}
+}
+
+func trackReplaceCost() interpreter.FunctionTracker {
+	return func(args []ref.Val, _ ref.Val) *uint64 {
```
TristonianJones: You can look at the result to more accurately estimate the amount of replacements performed, so that's something to consider in your equation.

ext/regex.go
maskri17: Can you please explain further not sure what inlining means? I was just following this https://github.com/google/cel-go/blob/83e2dcee2a2c640a731d7e69f86475a4c7a1e8e7/ext/sets.go#L131-L139

ext/regex.go
maskri17: Done this got me thinking that maybe we add the result string build size for regex.Extract as well. So I am thinking of adding it to both estimator and tracker func for it in the next changes

ext/README.md
maskri17: Tried to emulate the string.split like this: https://github.com/google/cel-go/blob/83e2dcee2a2c640a731d7e69f86475a4c7a1e8e7/ext/strings.go#L215-L217

ext/regex.go
maskri17: Added comments for all estimators

ext/regex.go
jnthntatum: I think @TristonianJones meant that the helper function `compileRegex` is just a wrapper around regexp.Compile so it's not that helpful here. just use regexp.Compile and remove the helper.

ext/README.md
```diff
@@ -945,3 +945,73 @@ Examples:
 
     {'greeting': 'aloha', 'farewell': 'aloha'}
       .transformMapEntry(keyVar, valueVar, {valueVar: keyVar}) // error, duplicate key
+
+## Regex
+
+Regex introduces functions for regular expressions in CEL.
+
+Note: all functions use the 'regex' namespace. If you are currently using a
```
jnthntatum: try to stick to 80ish column lines. don't worry about if you need to break that rule for examples etc, but there's no need for an extra long line here.

ext/regex_test.go
```diff
@@ -272,3 +269,112 @@ func testRegexEnv(t *testing.T, opts ...cel.EnvOption) *cel.Env {
...
+		expr          string
+		vars          []cel.EnvOption
+		in            map[string]any
+		hints         map[string]uint64
+		estimatedCost checker.CostEstimate
+		actualCost    uint64
+	}{
+		{
+			expr:          `regex.extract('hello world', 'hello (.*)') == optional.of('world')`,
+			estimatedCost: checker.CostEstimate{Min: 19, Max: 20},
```
jnthntatum: partly for my benefit, could you add a comment working out how the values for the estimate and observed costs were derived on 2-3 interesting test cases?

In particular, since everything is constant here, I'd expect the actual cost to line up pretty well with the cost estimate.

ext/regex.go
```diff
@@ -330,3 +342,104 @@ func extractAll(target, regexStr ref.Val) ref.Val {
...
+		return nil
+	}
+}
+
+func estimateExtractAllCost() checker.FunctionEstimator {
+	return func(c checker.CostEstimator, target *checker.AstNode, args []checker.AstNode) *checker.CallEstimate {
+		if len(args) == 2 {
+			targetSize := estimateSize(c, args[0])
+			// The target cost is the size of the target string, scaled by a traversal factor.
+			targetCost := targetSize.Add(checker.FixedSizeEstimate(1)).MultiplyByCostFactor(common.StringTraversalCostFactor)
```
jnthntatum, line 365: could you document somewhere why +1 (is it a correction for zero size estimate?)

ext/regex.go
```diff
@@ -330,3 +342,104 @@ func extractAll(target, regexStr ref.Val) ref.Val {
...
+			regexCost := estimateSize(c, args[1]).Add(checker.FixedSizeEstimate(1)).MultiplyByCostFactor(common.RegexStringLengthCostFactor)
+			// Estimate the build cost of the new string using a pessimistic upper bound.
+			replacementCost := targetSize.Multiply(replacementSize.Add(checker.FixedSizeEstimate(1)))
+
+			// Estimate the potential size range of the output string. The final size could be smaller
+			// (if the replacement is shorter than the match) or larger than the original.
+			allReplacedSize := targetSize.Max * replacementSize.Max
+			noneReplacedSize := targetSize.Max
+			resultSize := &checker.SizeEstimate{Min: noneReplacedSize, Max: allReplacedSize}
+			if noneReplacedSize > allReplacedSize {
```
jnthntatum: this can only be true when replacement size is '0', right? Imo it would be clearer write it that way.

maybe worth thinking about the extremes:

`regex.replace('foo bar', r'(foo bar)', r'\1\1\1\1\1...' )`

`regex.replace('foo bar', r'(foo bar)', '')`

ext/regex.go
```diff
@@ -330,3 +342,104 @@ func extractAll(target, regexStr ref.Val) ref.Val {
...
+		l := len(args)
+		if l == 3 || l == 4 {
+			targetSize := estimateSize(c, args[0])
+			replacementSize := estimateSize(c, args[2])
+			// The target cost is the size of the target string, scaled by a traversal factor.
+			targetCost := targetSize.Add(checker.FixedSizeEstimate(1)).MultiplyByCostFactor(common.StringTraversalCostFactor)
+			// The regex cost is the size of the regex pattern, scaled by a complexity factor.
+			regexCost := estimateSize(c, args[1]).Add(checker.FixedSizeEstimate(1)).MultiplyByCostFactor(common.RegexStringLengthCostFactor)
+			// Estimate the build cost of the new string using a pessimistic upper bound.
+			replacementCost := targetSize.Multiply(replacementSize.Add(checker.FixedSizeEstimate(1)))
```
jnthntatum: this is a SizeEstimate, right? consider renaming or moving to cost units here.

ext/regex.go
maskri17: I have added the cases and updated the conditions take a look

ext/regex.go
maskri17: thats exactly right and done

ext/regex.go
```diff
@@ -330,3 +334,107 @@ func extractAll(target, regexStr ref.Val) ref.Val {
...
+			}
+		}
+		return nil
+	}
+}
+
+func trackReplaceCost() interpreter.FunctionTracker {
+	return func(args []ref.Val, result ref.Val) *uint64 {
+		targetSize := float64(actualSize(args[0])) * float64(common.StringTraversalCostFactor)
+		regexSize := float64(actualSize(args[1])) * float64(common.RegexStringLengthCostFactor)
```
jnthntatum: the units are a bit confusing here, consider renaming targetCost and regexCost

size * (cost / size) -> cost

I think the tracking estimate should be close to the estimator for 'matches' in [runtimecost.go](https://source.corp.google.com/piper///depot/google3/third_party/cel/go/interpreter/runtimecost.go)

plus the cost of creating the result string

```
let targetCost = ceil( float(1 + target_size) * StringTraversalFactor)
let regexCost = ceil(float(pattern_size) * RegexLengthFactor)
let resultCost = ceil(float(result_size) * StringTraversalFactor)

let estimate = targetCost*regexCost + resultCost + callCost
```

ext/regex.go
```diff
@@ -330,3 +334,107 @@ func extractAll(target, regexStr ref.Val) ref.Val {
...
+		cost := callCost + uint64(float64(targetSize*regexSize)) + uint64(actualSize(result))
+		return &cost
+	}
+}
+
+func trackExtractAllCost() interpreter.FunctionTracker {
+	return func(args []ref.Val, result ref.Val) *uint64 {
+		targetSize := float64(actualSize(args[0])) * float64(common.StringTraversalCostFactor)
+		regexSize := float64(actualSize(args[1])) * float64(common.RegexStringLengthCostFactor)
+		cost := callCost + uint64(float64(targetSize*regexSize)) + uint64(actualSize(result))
```
jnthntatum: I think you also need to factor in the fixed cost for new list allocation (see ext.trackAllocatingListCall in ext/lists.go)

ext/regex.go
```diff
@@ -330,3 +334,107 @@ func extractAll(target, regexStr ref.Val) ref.Val {
...
+		return &cost
+	}
+}
+
+func trackExtractCost() interpreter.FunctionTracker {
+	return func(args []ref.Val, result ref.Val) *uint64 {
+		targetSize := float64(actualSize(args[0])) * float64(common.StringTraversalCostFactor)
+		regexSize := float64(actualSize(args[1])) * float64(common.RegexStringLengthCostFactor)
+		cost := callCost + uint64(float64(targetSize*regexSize)) + uint64(actualSize(result))
```
jnthntatum: I think the result size component needs to be scaled by the string traversal factor see strings.quote estimator in [runtimecost.go](https://source.corp.google.com/piper///depot/google3/third_party/cel/go/interpreter/runtimecost.go)

ext/regex.go
```diff
@@ -330,3 +334,107 @@ func extractAll(target, regexStr ref.Val) ref.Val {
...
+			// Fixed size estimate of +1 is added for safety from zero size args.
+			// The target cost is the size of the target string, scaled by a traversal factor.
+			targetCost := targetSize.Add(checker.FixedSizeEstimate(1)).MultiplyByCostFactor(common.StringTraversalCostFactor)
+			// The regex cost is the size of the regex pattern, scaled by a complexity factor.
+			regexCost := estimateSize(c, args[1]).Add(checker.FixedSizeEstimate(1)).MultiplyByCostFactor(common.RegexStringLengthCostFactor)
+
+			// The result is a list of strings. Worst Case: its contents are the size of the entire target.
+			resultSize := &checker.SizeEstimate{Min: 0, Max: targetSize.Max}
+			// The cost to allocate the result list is its base cost plus the size of its contents.
+			listAllocCost := resultSize.Add(checker.FixedSizeEstimate(common.ListCreateBaseCost)).MultiplyByCostFactor(1)
```
jnthntatum: nit: the unit handling is weird here.

prefer something like:

`listAllocCost := resultSize.MulitplyByCostFactor(1).Add(checker.FixedCostEstimate(common.ListCreateBaseCost))`

ext/regex.go
maskri17: no need to use traversal factor for result as we have the access to the actual size. Also, estimators didnt use traversal factor for result string either

ext/regex_test.go
```diff
@@ -272,3 +269,142 @@ func testRegexEnv(t *testing.T, opts ...cel.EnvOption) *cel.Env {
...
+		//   search cost, and the allocation cost for the 2-char result string '22'.
+		//   It aligns perfectly with the minimum estimate.
+		{
+			expr:          "regex.extract('4122345432', '22').orValue('777') == '22'",
+			estimatedCost: checker.CostEstimate{Min: 5, Max: 15},
+			actualCost:    5,
+		},
+		{
+			expr:          "regex.extract('4122345432', '22').or(optional.of('777')) == optional.of('22')",
+			estimatedCost: checker.CostEstimate{Min: 7, Max: 1844674407370955280},
```
jnthntatum: integer underflow? What's going on with this test case?

if this is expected, can you document why it happens?

ext/regex.go
```diff
@@ -330,3 +334,119 @@ func extractAll(target, regexStr ref.Val) ref.Val {
...
+
+func estimateExtractAllCost() checker.FunctionEstimator {
+	return func(c checker.CostEstimator, target *checker.AstNode, args []checker.AstNode) *checker.CallEstimate {
+		if len(args) == 2 {
+			targetSize := estimateSize(c, args[0])
+			// Fixed size estimate of +1 is added for safety from zero size args.
+			// The target cost is the size of the target string, scaled by a traversal factor.
+			targetCost := targetSize.Add(checker.FixedSizeEstimate(1)).MultiplyByCostFactor(common.StringTraversalCostFactor)
+			// The regex cost is the size of the regex pattern, scaled by a complexity factor.
+			regexCost := estimateSize(c, args[1]).Add(checker.FixedSizeEstimate(1)).MultiplyByCostFactor(common.RegexStringLengthCostFactor)
```
jnthntatum, line 367: For `matches` the regex cost factor is a rate relative to the size of the search string.

https://github.com/google/cel-go/blob/b24530c38d546ccf80e347b8fa17f864437ffb58/common/cost.go#L37C2-L39C36

So for stdlib matches we estimate (and track) this as:

`targetCost.Multiply(regexCost)` (multiply *not* add)

https://github.com/google/cel-go/blob/b24530c38d546ccf80e347b8fa17f864437ffb58/checker/cost.go#L792C4-L801C84

then specific to replace / extract / extract all, you'd add the cost for creating the output.

I'd imagine it should be similar here.

ext/regex_test.go
maskri17: That huge number in the estimatedCost.Max is not an integer underflow but an expected worst-case calculation.

The .or() function introduces runtime uncertainty, and since the cost estimator can't know which branch the code will take, it must prepare for the most expensive possible outcome. To guarantee safety, the model defaults to the largest possible value (math.MaxUint64) to account for this uncertainty similar to this: https://github.com/google/cel-go/blob/master/ext/lists_test.go#L253. I documented it.

ext/regex.go
maskri17: agreed

jnthntatum (approved):

PR #1201, opened 2025-07-23
Adding helper functions to map node IDs to the IDs of their parents and children.

kashish-a:
feat: Add AST parent and children mapping helpers

Introduces new `Children` and `Parents` helper functions for AST nodes.
These are **additional utilities** that do not alter existing AST
processing. They're vital for our CEL coverage solution, addressing
missing nodes in `EvalDetails` due to internal optimizations. By enabling
upward and downward propagation, these functions ensure complete and
accurate coverage data.

### DIFF
--- a/common/ast/ast.go
+++ b/common/ast/ast.go
@@ -533,3 +533,118 @@ func (hv heightVisitor) maxEntryHeight(entries ...EntryExpr) int {
 	}
 	return max
 }
+
+
+// Children returns a map from expression id to a list of the ids of its direct children.
+func Children(a *AST) map[int64][]int64 {
+	visitor := make(childrenVisitor)
+	PostOrderVisit(a.Expr(), visitor)
+	return visitor
+}
+
+type childrenVisitor map[int64][]int64
+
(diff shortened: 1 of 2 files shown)


aakash070 (approved):

common/ast/ast.go
```diff
@@ -533,3 +533,118 @@ func (hv heightVisitor) maxEntryHeight(entries ...EntryExpr) int {
 	}
 	return max
 }
+
+
+// Children returns a map from expression id to a list of the ids of its direct children.
+func Children(a *AST) map[int64][]int64 {
```
jnthntatum, line 539: normalized parent/child traversal is provided by the navigable AST types.

`
navExpr := ast.NavigateExpr(cAST, expr)
`

Could you describe what you need here that can't be accomplished with NavigableExpr? If there is a tradeoff to using id maps over the Navigable* types, can you document that here?

common/ast/navigable.go

aakash070:
closing this request, will be raising a separate pull request with the helper functions.

PR #1204, opened 2025-07-24
Adding annotation optimizer

kashish-a:
This PR adds the basic infrastructure for adding annotations to CEL expressions. The optimizer works by wrapping each node in a `cel.@annotation` call. The call receives the original expression and a list containing annotation info for that node.

Note that this PR is based on the PR containing type-checking support for annotations.


aakash070 (approved):

aakash070 (approved):

cel/options.go
```diff
@@ -275,6 +275,14 @@ func Abbrevs(qualifiedNames ...string) EnvOption {
 	}
 }
 
+// EnableAnnotations adds a function that is used internally to annotate expressions.
+func EnableAnnotations() EnvOption {
+	return Function("cel.@annotation",
+		decls.Overload("cel_annotation",
+			[]*types.Type{types.DynType, types.NewListType(types.NewMapType(types.StringType, types.DynType))}, types.DynType),
```
jnthntatum: Some thoughts on typing:

1) might be a bit safer to use a type parameter here instead of dyn?

depends a bit on when you expect this to actually make it into the type checker -- dyn makes checking more permissive in a way that you may not want. `"foo".startsWith(1)` -> `"foo".startsWith(@(1, {'anno': 'value'}))`

`cel.@annotation(T, map(string, dyn)) -> T`

2) Can you document on why list(map)? I don't know what the expected behavior would be if there are duplicates of the same kind of annotation on a given node. I would think it generally would indicate something is misconfigured (e.g. name collision).

For the annotation representation itself, map is probably okay to keep checking simple.

cel/optimizer.go
```diff
@@ -443,6 +454,44 @@ func (opt *optimizerExprFactory) NewStructField(field string, value ast.Expr, is
 	return opt.fac.NewStructField(opt.nextID(), field, value, isOptional)
 }
 
+// NewComprehension creates a comprehension expression with a set of optional indices.
```
jnthntatum: Can you update the description here? (where is optional indices coming from?) If you do want to give a specific example, filter's loop step is something like

loopStep: (i > 0)? __result__ + [i] : __result__

cel/optimizer.go
```diff
@@ -443,6 +454,44 @@ func (opt *optimizerExprFactory) NewStructField(field string, value ast.Expr, is
...
+// - accuVar: __result__
+// - accuInit: []
+// - loopCondition: true
+// - loopStep: __result__.append(i)
+// - result: __result__
+func (opt *optimizerExprFactory) NewComprehension(iterRange ast.Expr, iterVar string, accuVar string, accuInit ast.Expr, loopCondition ast.Expr, loopStep ast.Expr, result ast.Expr) ast.Expr {
+	return opt.fac.NewComprehension(opt.nextID(), iterRange, iterVar, accuVar, accuInit, loopCondition, loopStep, result)
+}
+
+// NewComprehensionTwoVar creates a comprehension expression with a set of optional indices.
```
jnthntatum: same here: looks like description needs to be updated.

cel/annotation_test.go
```diff
@@ -0,0 +1,365 @@
+// Copyright 2019 Google LLC
```
jnthntatum: nit: update copyright year for new file.

cel/annotation_test.go
```diff
@@ -0,0 +1,365 @@
...
+			expectError: "error generating annotation: factory failed",
+		},
+		{
+			name: "multiple annotations",
+			expr: "1",
+			factories: []AnnotationFactory{
+				&testAnnotationFactory{name: "ann1", value: testValue},
+				&testAnnotationFactory{name: "ann2", value: testValue},
+			},
+			want: `cel.@annotation(1, [{"name": "ann1", "value": "test", "is_expr": false}, {"name": "ann2", "value": "test", "is_expr": false}])`,
```
jnthntatum: just confirming: you are still planning to have a post process step that strips these from the main evaluable AST and collects them all in a macro_calls style map?

cel/annotation_test.go
```diff
@@ -0,0 +1,365 @@
...
+					value:      testValue,
+					applicable: func(e ast.Expr) bool { return e.Kind() == ast.CallKind },
+				},
+				&testAnnotationFactory{
+					name:       "literal_ann",
+					value:      testValue,
+					applicable: func(e ast.Expr) bool { return e.Kind() == ast.LiteralKind },
+				},
+			},
+			want: `cel.@annotation(size(cel.@annotation("hello", [{"name": "literal_ann", "value": "test", "is_expr": false}])), [{"name": "call_ann", "value": "test", "is_expr": false}])`,
```
jnthntatum: up to you:

I'm assuming that this is just passing around internal data in a way that won't break the type-checker, so not too important, but you could get a more condensed format with something like:

`{"<annot_name>": {"value": foo} }`  (assume is_expr is false if expr key not defined)

cel/annotation.go
```diff
@@ -0,0 +1,243 @@
...
+func createAnnotationNode(ctx *OptimizerContext, annotations []ast.Expr) ast.Expr {
+	return ctx.NewList(annotations, nil)
+}
+
+// createAnnotationExpr creates an AST node for a single Annotation struct (a map literal).
+func createAnnotationExpr(ctx *OptimizerContext, ann *Annotation) ast.Expr {
+	return ctx.NewMap(
+		[]ast.EntryExpr{
+			ctx.NewMapEntry(ctx.NewLiteral(types.String("name")), ctx.NewLiteral(types.String(ann.Name)), false),
+			ctx.NewMapEntry(ctx.NewLiteral(types.String("value")), ctx.NewLiteral(types.DefaultTypeAdapter.NativeToValue(ann.Value)), false),
```
jnthntatum, line 255: No action needed for now:

With this approach, you're limited to what literals are expressible in the normal type checking environment, right? I think that's probably ok for now, but may be worth revisiting depending on usage.

cases like:
- create Metadata proto that is not accessible in the main expression
- call an internal function (e.g. a lazy function that records a metric)

cel/annotation.go
```diff
@@ -0,0 +1,243 @@
+// Copyright 2019 Google LLC
```
jnthntatum: nit: copyright year

cel/options.go
```diff
@@ -275,6 +275,14 @@ func Abbrevs(qualifiedNames ...string) EnvOption {
 	}
 }
 
+// EnableAnnotations adds a function that is used internally to annotate expressions.
+func EnableAnnotations() EnvOption {
```
l46kok: Could we move this to `annotations.go`? Imo, the use-case is specific to annotations such that it probably belongs there.

cel/annotation.go
```diff
@@ -0,0 +1,243 @@
...
+
+// NewAnnotation creates a new annotation with the given name, value, and whether the value is an
+// evaluable CEL expression or a non-evaluable constant.
+//
+// Note: Evaluable CEL expressions are not supported yet.
+func NewAnnotation(name string, value any, isExpr bool) *Annotation {
+	return &Annotation{
+		Name:   name,
+		Value:  value,
+		IsExpr: isExpr,
```
l46kok: It's still unclear to me on whether we really need to annotate evaluable CEL expressions separately (and if we do, whether we should keep this as a field or a different struct). 

But given that we aren't leveraging this field just yet and to keep the PR focused, could we for now split this into a separate change?

cel/annotation_test.go
```diff
@@ -0,0 +1,365 @@
...
+				checkPart := func(name string, expr ast.Expr, want string) {
+					t.Helper()
+					partAst := ast.NewAST(expr, sourceInfo)
+					got, err := AstToString(&Ast{impl: partAst})
+					if err != nil {
+						t.Fatalf("AstToString(%s) failed: %v", name, err)
+					}
+					if got != want {
+						t.Errorf("%s got\n%s\nwant\n%s", name, got, want)
+					}
```
l46kok: nit: consider extracting this into a small helper function within the test and reuse that across

cel/annotation_test.go
```diff
@@ -0,0 +1,365 @@
...
+				}
+				checkPart("iterRange", comp.IterRange(), `cel.@annotation([cel.@annotation(1, [{"name": "ann", "value": "test", "is_expr": false}])], [{"name": "ann", "value": "test", "is_expr": false}])`)
+				checkPart("accuInit", comp.AccuInit(), `cel.@annotation([], [{"name": "ann", "value": "test", "is_expr": false}])`)
+				checkPart("loopCondition", comp.LoopCondition(), `cel.@annotation(true, [{"name": "ann", "value": "test", "is_expr": false}])`)
+				checkPart("loopStep", comp.LoopStep(), `cel.@annotation(cel.@annotation(@result, [{"name": "ann", "value": "test", "is_expr": false}]) + cel.@annotation([cel.@annotation(cel.@annotation(x, [{"name": "ann", "value": "test", "is_expr": false}]) * cel.@annotation(2, [{"name": "ann", "value": "test", "is_expr": false}]), [{"name": "ann", "value": "test", "is_expr": false}])], [{"name": "ann", "value": "test", "is_expr": false}]), [{"name": "ann", "value": "test", "is_expr": false}])`)
```
l46kok: Consider breaking the expression apart into more readable chunks, by adding newlines / tabs as needed. You could normalize the test string afterwards for comparison ([example](https://github.com/google/cel-go/blob/93e910b85eebd87df104429ddc57ed4a541688df/policy/compiler_test.go#L513-L518)).

cel/annotation_test.go
```diff
@@ -0,0 +1,365 @@
...
+				checkPart("result", comp.Result(), `cel.@annotation(@result, [{"name": "ann", "value": "test", "is_expr": false}])`)
+			},
+		},
+		{
+			name: "Annotating map comprehension with 2 variables",
+			expr: "{'a': 'b'}.exists(k, k == v)",
```
l46kok: Could you add another test case for nested macros? (something like `[1,2,3].map(x, x + 1).exists(y, y > 0)`). Generally, the optimizer APIs require you to call `updateExpr` somewhere to normalize the IDs against the macro_call map. I'm unsure if the unparsing will work as intended in its current form.

cel/annotation_test.go
kashish-a: Yes, that's correct. We plan to add a scraper that will generate a mapping from node IDs to their corresponding annotations.

cel/annotation_test.go
kashish-a: Annotations are stored as a list attached to each expression. This allows annotations from different optimized subtrees to be collected and preserved when creating new nodes during optimization. Using a list ensures that annotations from various parts of the AST are retained without name conflicts.

cel/annotation.go
kashish-a: `annotate evaluable CEL expressions separately` To clarify,  IsExpr depicts if the expression in the Value field of the Annotation struct is a constant or an evaluable expression. This is solely for extension to when we consider dynamic value in annotations.

cel/annotation.go
kashish-a: Yes, that is correct. Currently the checker supports `types.BoolType, types.BytesType, types.DoubleType, types.IntType, types.NullType, types.StringType, types.UintType` these types of literals.

aakash070:
Closing this pull request as these changes will be modified and raised in a different pr.

PR #1206, opened 2025-08-01
Adding cel.@annotation in env functions via an option

kashish-a:
Iintroduced a new environment option to include the cel.@annotation function, which users are now expected to add when creating their environment to enable annotations.

### DIFF
--- a/cel/options.go
+++ b/cel/options.go
@@ -275,6 +275,17 @@ func Abbrevs(qualifiedNames ...string) EnvOption {
 	}
 }
 
+// EnableAnnotations adds a function that is used internally to annotate expressions.
+func EnableAnnotations() EnvOption {
+	return func(e *Env) (*Env, error) {
+		e.functions["cel.@annotation"], _ = decls.NewFunction("cel.@annotation",
+			decls.Overload("cel.@annotation",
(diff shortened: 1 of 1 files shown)


cel/options.go
```diff
@@ -275,6 +275,17 @@ func Abbrevs(qualifiedNames ...string) EnvOption {
 	}
 }
 
+// EnableAnnotations adds a function that is used internally to annotate expressions.
+func EnableAnnotations() EnvOption {
+	return func(e *Env) (*Env, error) {
```
aakash070: shall we use cel.Function to create the EnvOption?

cel/options.go
```diff
@@ -275,6 +275,17 @@ func Abbrevs(qualifiedNames ...string) EnvOption {
...
+		e.functions["cel.@annotation"], _ = decls.NewFunction("cel.@annotation",
+			decls.Overload("cel.@annotation",
+				[]*types.Type{types.AnyType, types.NewListType(types.NewMapType(types.StringType, types.AnyType))}, types.DynType),
```
aakash070: Why are we passing the first argument as AnyType, should it not be DynType? Same for the map value type.

kashish-a:
closing this PR as the changes have been incorporated in https://github.com/google/cel-go/pull/1204

PR #1207, opened 2025-08-12
Support unwrapping unknown implementations of `proto.Message`

srikrsna:
`proto.Message` can have implementations outside of protobuf module, all such cases should be treated as a dynamic message.

### DIFF
--- a/common/types/pb/type.go
+++ b/common/types/pb/type.go
@@ -472,7 +472,7 @@ func unwrap(desc description, msg proto.Message) (any, bool, error) {
 		}
 		return v.GetValue(), true, nil
 	}
-	return msg, false, nil
+	return unwrapDynamic(desc, msg.ProtoReflect())
 }
 
 // unwrapDynamic unwraps a reflected protobuf Message value.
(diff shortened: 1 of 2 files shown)


common/types/pb/type.go
```diff
@@ -472,7 +472,7 @@ func unwrap(desc description, msg proto.Message) (any, bool, error) {
 		}
 		return v.GetValue(), true, nil
 	}
-	return msg, false, nil
+	return unwrapDynamic(desc, msg.ProtoReflect())
```
TristonianJones, line 475: Could you be a bit more explicit about which type definition this is? There are explicit checks for dynamic message earlier in the switch.

TristonianJones: It's probably worth a test case to validate the expected behavior. The change should be mostly harmless, but it's worth exercising the case where it demonstrates the desired effect before and after.

srikrsna: Because `proto.Message` is an interface, it can have implementations other than the ones listed here. A case that I encountered is using https://pkg.go.dev/buf.build/go/hyperpb#Message.

TristonianJones: Would you mind adding a test case so we don't regress? Thanks!

TristonianJones (approved):

PR #1208, opened 2025-08-14
Add logic to unwrap activiations and return evalState for the underlying activations

aakash070:
This change adds an inline `activationWrapper` interface to the `asEvalState` function. This can be implemented by composite objects like  `dynamicSlotActivation` and `constantSlotActivation` to return the contained `Activations`. 
This will fix state observation for ternary conditional expressions which are added as part of the composition optimizer during policy compilation when state tracking is enabled.

### DIFF
--- a/interpreter/interpreter.go
+++ b/interpreter/interpreter.go
@@ -137,6 +137,11 @@ func (esa evalStateActivation) asEvalState() EvalState {
 	return esa.state
 }
 
+// activationWrapper identifies an object that can be unwrapped to access the underlying activation.
+type activationWrapper interface {
+	Unwrap() Activation
+}
+
 // asEvalState walks the Activation hierarchy and returns the first EvalState found, if present.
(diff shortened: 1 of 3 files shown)


interpreter/interpreter.go
```diff
@@ -142,6 +142,18 @@ func asEvalState(vars Activation) (EvalState, bool) {
 	if conv, ok := vars.(evalStateConverter); ok {
 		return conv.asEvalState(), true
 	}
+	// Check if the current activation wraps another activation. This is used to support
+	// wrappers such as the @block() activation which may be composed of a dynamicSlotActivation or a
+	// constantSlotActivation. In this case, the underlying activation is the portion which interacts
+	// with the EvalState.
+	type activationWrapper interface {
```
TristonianJones: Would you mind moving this interface to another place within the file? I'd prefer not to declare it within the function block

TristonianJones (approved):

PR #1209, opened 2025-08-20
CEL test coverage calculation

aakash070:
This change adds coverage calculation logic for the CEL test runner:
1. The `cel_go_test` bazel macro accepts a new boolean flag `enable_coverage`. When set to `True` the coverage statistics for the test suite are logged.
2. Coverage metrics include:
* AST node coverage - Determines the percentage of AST nodes covered
* AST branch coverage -  Determines the percentage of logical branches covered
For each of the above metrics, an optional set of interesting nodes and branches are logged which can help determine which nodes or logical branches are missed by the tests.

Sample coverage logs for `//policy:restricted_destinations_policy`:
restricted_destinations_policy

### DIFF
--- a/test/cel_go_test.bzl
+++ b/test/cel_go_test.bzl
@@ -31,6 +31,7 @@ def cel_go_test(
         test_data_path = "",
         file_descriptor_set = "",
         filegroup = "",
+        enable_coverage = False,
         deps = [],
         data = [],
         **kwargs):
@@ -59,6 +60,7 @@ def cel_go_test(
           this must be in binary format with either a .binarypb or .pb or.fds extension. If you need
           to support a textformat file_descriptor_set, embed it in the environment file. (default None)
       filegroup: str label of a filegroup containing the test suite, config, and cel_expr.
+      enable_coverage: boolean indicating if coverage should be enabled for the test.
       deps: list of dependencies for the go_test rule
       data: list of data dependencies for the go_test rule
       **kwargs: additional arguments to pass to the go_test rule
@@ -104,6 +106,7 @@ def cel_go_test(
         "--test_suite_path=%s" % test_suite,
         "--config_path=%s" % config,
         "--base_config_path=%s" % base_config,
+        "--enable_coverage=%s" % enable_coverage,
     ]
 
     if cel_expr_format == ".cel" or cel_expr_format == ".celpolicy" or cel_expr_format == ".yaml":
--- a/tools/celtest/test_coverage_reporter.go
+++ b/tools/celtest/test_coverage_reporter.go
@@ -1,0 +1,143 @@
+// Copyright 2025 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Package celtest provides functions for testing CEL policies and expressions.
+package celtest
+
+import (
+	"fmt"
+	"strings"
+	"testing"
+
+	"github.com/google/cel-go/cel"
+	"github.com/google/cel-go/common/ast"
+	"github.com/google/cel-go/common/types"
+)
+
+// reportCoverage reports the coverage information for the provided programs.
+//   - For the node coverage, the coverage is reported as a percentage of the number of nodes which
+//     were evaluated during the test execution and hence are present in the program Coverage report.
+//   - For the branch coverage, every node which has a boolean return type is considered as a branch.
+//     The number of branches which were evaluated during the test execution and hence are present in
+//     the program Coverage report are reported as the branch coverage percentage.
+func reportCoverage(t *testing.T, programs []Program) {
+	t.Helper()
+	for _, p := range programs {
+		exprString, err := cel.AstToString(p.Ast)
+		if err != nil {
+			t.Logf("Error converting AST to string for a program: %v", err)
+			continue
+		}
+		rootNavigableExpr := ast.NavigateAST(p.Ast.NativeRep())
+		// Initialize coverage metrics
+		cr := &coverageReport{
+			nodes:                  0,
+			coveredNodes:           0,
+			branches:               0,
+			coveredBooleanOutcomes: 0,
+			unencounteredNodes:     []string{},
+			unencounteredBranches:  []string{},
+		}
+		traverseAndCalculateCoverage(t, rootNavigableExpr, p, true, "", cr)
+		printCoverageReport(t, exprString, cr)
+	}
+}
+
+type coverageReport struct {
+	nodes                  int64
+	coveredNodes           int64
+	branches               int64
+	coveredBooleanOutcomes int64
+	unencounteredNodes     []string
+	unencounteredBranches  []string
+}
+
+func traverseAndCalculateCoverage(t *testing.T, expr ast.NavigableExpr, p Program, logUnencountered bool,
+	preceedingTabs string, cr *coverageReport) {
+	t.Helper()
+	if expr == nil || len(p.CoverageStats) == 0 {
+		return
+	}
+	nodeID := expr.ID()
+	cr.nodes++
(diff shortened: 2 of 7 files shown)


tools/celtest/test_runner.go
```diff
@@ -451,15 +467,34 @@ func PartialEvalProgramOption() TestRunnerOption {
...
 	PolicyMetadata map[string]any
+	Ast            *cel.Ast
+	CoverageStats  map[int64]set
 }
 
+// set is a generic set implementation using a map where the keys are the elements of the set
+// and the values are empty structs. This approach is memory-efficient because an empty struct
+// occupies zero bytes. This will be used to store the values observed for each expression during
+// coverage calculation.
+type set map[any]struct{}
```
jnthntatum: could this just be map[ref.Val]struct{}?

tools/celtest/test_runner.go
```diff
@@ -890,19 +926,337 @@ func (tr *TestRunner) createResultMatcher(t *testing.T, testOutput *test.Output)
...
+// - If a node is visited, all its ancestors are visited.
+// - If a node is visited, and it has only one child, then the child is visited.
+func propagateVisitedness(root *cel.Ast, coverageStats map[int64]set, state interpreter.EvalState) {
+	if root == nil || coverageStats == nil {
+		return
+	}
+	// Propagate visitedness upwards for ancestors.
+	ast.PostOrderVisit(root.NativeRep().Expr(), parentVisitor{coverageStats: coverageStats, evalState: state})
+	// Propagate visitedness downwards for single-child nodes.
+	ast.PostOrderVisit(root.NativeRep().Expr(), childrenVisitor{coverageStats: coverageStats, evalState: state})
```
jnthntatum, line 986: IIUC, the children rule only propagates one level down?

Maybe that's okay, but what is the issue you're solving? Is the eval state not tracking specific kinds of leaf subexpressions?

tools/celtest/test_runner.go
```diff
@@ -890,19 +926,337 @@ func (tr *TestRunner) createResultMatcher(t *testing.T, testOutput *test.Output)
...
+// VisitExpr populates the coverageStats for the current expression based on the coverageStats of its direct children.
+func (pv parentVisitor) VisitExpr(e ast.Expr) {
+	switch e.Kind() {
+	case ast.SelectKind:
+		if pv.coverageStats[e.ID()] != nil {
+			return
+		}
+		if selVal, ok := pv.coverageStats[e.AsSelect().Operand().ID()]; ok {
+			pv.coverageStats[e.ID()] = make(set)
+			pv.coverageStats[e.ID()][selVal] = struct{}{}
```
jnthntatum: Could you describe when this happens? 

It seems odd to associate the observed operand value to the select expression.

tools/celtest/test_runner.go
```diff
@@ -890,19 +926,337 @@ func (tr *TestRunner) createResultMatcher(t *testing.T, testOutput *test.Output)
...
+				preceedingTabs = preceedingTabs + "\t\t"
+			}
+		}
+	}
+	for _, child := range expr.Children() {
+		traverseAndCalculateCoverage(t, child.(ast.NavigableExpr), p, logUnencounteredNodes, logUnencounteredBranches, preceedingTabs, cr)
+	}
+}
+
+func printCoverageReport(t *testing.T, exprString string, cr *coverageReport) {
```
jnthntatum: Could you attach an example report to the PR description?

jnthntatum (commented):
LGTM.

Still not sure on the best way to present the details on missing coverage: it may make sense to just present the high level stats by default and add a flag or other mechanism to opt into dumping the missing coverage snippets, but I'll defer to other reviewers.

tools/celtest/test_runner.go
aakash070: In some cases it was observed that for `select` kind nodes the trace details did not include the `operand` node, hence, adding this step to cover all such cases where if an observed node has just a single child then the child node is also marked as visited.

tools/celtest/test_runner.go
aakash070: This was not observed but added this as a fallback mechanism to mark the `select` kind node visited if the operand was visited.
> It seems odd to associate the observed operand value to the select expression. 

Yeah we can simple mark this is plain visited, will make the change.

jnthntatum (dismissed):

tools/celtest/test_coverage_reporter.go
```diff
@@ -0,0 +1,143 @@
...
+		}
+	}
+	for _, child := range expr.Children() {
+		traverseAndCalculateCoverage(t, child.(ast.NavigableExpr), p, logUnencountered, preceedingTabs, cr)
+	}
+}
+
+func printCoverageReport(t *testing.T, exprString string, cr *coverageReport) {
+	t.Helper()
+	t.Logf("--- Start Coverage Report ---\nExpression: %s", exprString)
```
l46kok, line 130: fyi: cel-go's unparser has the capability to format expressions for readability, wrapping on designated column limit / operators of your choice via `UnparserOption`. This does require directly using the unparser package here. Just mentioning since this could help with eyeballing between the reported missing subexpr coverage against the original expression, especially for large composed expressions.

tools/celtest/test_coverage_reporter.go
```diff
@@ -0,0 +1,143 @@
...
+	if cr.nodes == 0 {
+		t.Logf("No coverage stats found")
+		return
+	}
+	// Log Node Coverage results
+	nodeCoverage := float64(cr.coveredNodes) / float64(cr.nodes) * 100.0
+	t.Logf("AST Node Coverage: %.2f%% (%d out of %d nodes covered)", nodeCoverage, cr.coveredNodes, cr.nodes)
+	if len(cr.unencounteredNodes) > 0 {
+		t.Logf("Interesting Unencountered Nodes:\n%s", strings.Join(cr.unencounteredNodes, "\n"))
```
l46kok, line 139: nit: Consider removing filler words such as "Interesting"

tools/celtest/test_coverage_reporter.go
```diff
@@ -0,0 +1,143 @@
...
+		logUnencountered = false
+	}
+	// Check for Branch Coverage if the node is a boolean type
+	if interestingBoolNode {
+		cr.branches += 2
+		exprText, _ := cel.ExprToString(expr, p.Ast.NativeRep().SourceInfo())
+		if info, found := p.CoverageStats[nodeID]; !found {
+			if logUnencountered {
+				cr.unencounteredBranches = append(cr.unencounteredBranches,
+					"\n"+preceedingTabs+fmt.Sprintf("Expression ID %d ('%s'): Never evaluated (neither true nor false)", nodeID, exprText))
```
l46kok: nit: double negatives (or beginning with negatives in general) tends to be confusing. Maybe just say:

```
Expression foo: No coverage.
Expression foo: lacks 'true' coverage. (or only evaluated to "false")
```

test/cel_go_test.bzl
```diff
@@ -31,6 +31,7 @@ def cel_go_test(
         test_data_path = "",
         file_descriptor_set = "",
         filegroup = "",
+        enable_coverage = False,
```
l46kok, line 34: Will you eventually be enabling this by default or should this always be opt-in?

tools/celtest/test_runner.go
l46kok: Consider amending the comment to include this observation.

tools/celtest/test_coverage_reporter.go
```diff
@@ -0,0 +1,143 @@
...
+
+func traverseAndCalculateCoverage(t *testing.T, expr ast.NavigableExpr, p Program, logUnencountered bool,
+	preceedingTabs string, cr *coverageReport) {
+	t.Helper()
+	if expr == nil || len(p.CoverageStats) == 0 {
+		return
+	}
+	nodeID := expr.ID()
+	cr.nodes++
+	interestingBoolNode := expr.Type() == types.BoolType && expr.AsLiteral() == nil && expr.AsCall().FunctionName() != "cel.@block"
```
l46kok, line 85: For future readers -- consider commenting the special casing for cel.@block. My guess is that this is immaterial to the user looking at coverage information, since this call is often a by-product of tool-composed expressions.

tools/celtest/test_coverage_reporter.go
aakash070: This was added intentionally since we'll be filtering out some of the nodes/branches from being printed in the logs.

test/cel_go_test.bzl
aakash070: this will remain an opt-in feature.

jnthntatum (approved):

l46kok (approved):

PR #1219, opened 2025-09-25
feat: let pb value to native go type equals Message.filed

yywing:
Let pb value to native go type equals Message.filed.

env: variable `a` is `proto.Message`. `a.list` is `[]string`

```golang
func ReadList(v ref.Value) ref.Value {
      switch v.Value.(type) {
      case []string:
              return v
      default:
             painc("a")
      }
}
```
`readList(a.list)` will panic

### DIFF
--- a/common/types/map_test.go
+++ b/common/types/map_test.go
@@ -892,7 +892,7 @@ func TestProtoMapConvertToNative(t *testing.T) {
 	if mapVal4.Equal(mapVal) != True || mapVal.Equal(mapVal4) != True {
 		t.Errorf("mapVal4.Equal(mapVal) returned false, wanted true")
 	}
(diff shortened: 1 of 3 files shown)


TristonianJones:
@yywing unfortunately, this sort of approach comes with a high cost overhead since it requires eager conversion to a go-based list or map which could result in the conversion of lots of objects which aren't used. If you'd like to read a list, the safest approach is as follows:

```go
func ReadList(v ref.Val) []string {
  l, err := v.ConvertToNative(reflect.TypeFor[[]string]())
  if err != nil {
    // Although, I wouldn't recommend panicking, just here for illustrative purposes.
    panic(fmt.Printf("value could not be converted to []string: %v", v))
  }
  return l
}
```

PR #1220, opened 2025-09-25
Adding ability to use JSON Names for pb fields

### LINKED ISSUE #1214: Support JSON Name in pb field names
**Feature request checklist**

**Change**

Allow for the protobuf field name mapping to use the [JSONName](https://pkg.go.dev/google.golang.org/protobuf/reflect/protoreflect#FieldDescriptor) instead of the field names.

**Example**

`person.firstName` vs. `person.first_name`

**Alternatives considered**

I have temporarily forked this repo to support this, it was a pretty simple change, however I am not sure of a good way to make this configurable either through a global var or a build tag for a PR.

https://github.com/google/cel-go/compare/master...Blizzard:cel-go:blizzard

Ideally I don't want to maintain this for for such a small thing. The reason why this is needed in our use case is that we allow for CEL queries to filter data in a graphql API. With the graphql schema being based on the JSON of the data, it is way less confusing to our users than having to do the mental dance of the above example.

renevo:
This adds a simple variable to map the fields to the JSONName() instead of Name() on the protobuf field descriptors. 

This allows for using cel expressions against the same JSON representations in some scenarios. 

The default behavior is identical. Test has been added to validate field names with setting turned on and all tests pass.

Closes #1214

### DIFF
--- a/common/types/pb/type.go
+++ b/common/types/pb/type.go
@@ -30,6 +30,22 @@ import (
 	wrapperspb "google.golang.org/protobuf/types/known/wrapperspb"
 )
 
+var (
+	// UseJSONFieldNames enables protobuf JSON field names for field access instead of the short name of the declaration.
+	//
+	// This should be set before any types are loaded into the Db.
+	//
+	// For example, given a proto field declaration:
+	//
+	//   message Sample {
+	//     int32 sample_field = 1;
+	//   }
+	//
+	// The default field name for access is "sample_field". If UseJSONFieldNames is set to true,
+	// the field name for access becomes "sampleField".
+	UseJSONFieldNames = false
+)
+
 // description is a private interface used to make it convenient to perform type unwrapping at
 // the TypeDescription or FieldDescription level.
 type description interface {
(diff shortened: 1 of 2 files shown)


renevo:
Is there more information/work I can do on this PR to move one way or the other?

common/types/pb/type.go
```diff
@@ -47,7 +63,11 @@ func newTypeDescription(typeName string, desc protoreflect.MessageDescriptor, ex
 	fields := desc.Fields()
 	for i := 0; i < fields.Len(); i++ {
 		f := fields.Get(i)
-		fieldMap[string(f.Name())] = newFieldDescription(f)
+		if UseJSONFieldNames {
```
TristonianJones, line 66: If you move this flag up as a function argument, you should be able to pass the same argument into the `pb.Db` instance which generates the internal type descriptions.

TristonianJones: Specifically, you'll want to update the `common/types/pb.go` definition for the ``Db` struct:

```go
type fieldNameKind int

const (
   fieldNameKind proto = iota
   json
   jsonOrProto
)

type Db struct {
	revFileDescriptorMap map[string]*FileDescription
	// files contains the deduped set of FileDescriptions whose types are contained in the pb.Db.
	files []*FileDescription
	// extensions contains the mapping between a given type name, extension name and its FieldDescription
	extensions map[string]map[string]*FieldDescription
	
	// fieldNameConvention supports type resolution using either the protobuf name or its JSON name
    fieldNameConvention fieldNameKind
}
```

TristonianJones: You would then need to support a functional option to the `NewDb` call:

```go
// NewDb creates a new `pb.Db` with an empty type name to file description map.
func NewDb(DbOption ...opts) *Db {
	pbdb := &Db{ ... }
	for _, opt := range opts {
	   pbdb = opt(pbdb)
	}
```

renevo: let me look into that, thanks for the response.

renevo:
This was implemented in #1286 closing this out

PR #1227, opened 2025-11-10
Expose Match source ID to callers.

catalinadurbala:
Making the exprID of Matches accessible to callers.

Will be used as keys for additional per-Match metadata, with the end
goal of compiling an AST where the output is comprised of this metadata
instead.

### DIFF
--- a/policy/parser.go
+++ b/policy/parser.go
@@ -301,6 +301,11 @@ type Match struct {
 	rule        *Rule
 }
 
+// SourceID returns the source identifier associated with the match.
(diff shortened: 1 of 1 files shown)


TristonianJones (dismissed):

catalinadurbala:
Sorry, I accidentally added another commit that should've gone to a different PR. I've now removed it from this PR but it looks like it invalidated the check anyway :(

TristonianJones (approved):

PR #1228, opened 2025-11-14
Bazel configuration to better enable dependent builds

### LINKED ISSUE #1184: Bzlmod Dependency Issues
**Describe the bug**

I'm having a heck of time with `cel.dev/expr`, `cel-go`, and a new bzlmod version of our repo.

First off, the error I'm seeing:

```
ERROR: /private/var/tmp/_bazel_jward/63162deb35ba6b2f678241a64a1d59bb/external/gazelle++go_deps+com_github_google_cel_go/common/types/BUILD.bazel:8:11: no such package '@@[unknown repo 'dev_cel_expr' requested from @@gazelle++go_deps+com_github_google_cel_go]//': The repository '@@[unknown repo 'dev_cel_expr' requested from @@gazelle++go_deps+com_github_google_cel_go]' could not be resolved: No repository visible as '@dev_cel_expr' from repository '@@gazelle++go_deps+com_github_google_cel_go' and referenced by '@@gazelle++go_deps+com_github_google_cel_go//common/types:go_default_library'
```

It seems that Bazel itself has some dependency on `cel.dev/expr` via a `bazel_dep` on `cel-spec`. So what that means, if you an `import "cel.dev/expr"` in your code, Gazelle will actually place the dep in your BUILD.bazel as `"@cel-spec//:go_default_library"`. Getting all the version to align to even get bzlmod happy is pretty tricky, but in that case I still can't get it building.

So the list of what I have:

1. Using bzlmod in project with Bazel version 8.2.1
2. Have a dep on `cel.dev/expr` in one of my own packages
3. Also have a dep on `cel-go` in one of my own packages
4. `rules_go 0.54.1`
5. `gazelle 0.43.0`

The incantation I'm at that has caused the above error:

`MODULE.bazel`
```
bazel_dep(
    name = "cel-spec",
    version = "0.23.0",
)
```

`go.mod`
```
require (
	cel.dev/expr v0.23.0
	github.com/google/cel-go v0.24.1
)
```

Honestly, this isn't the incantation I even want, but it's the only one I found that could make bzlmod happy, and it still doesn't work. Ideally I could just pull in latest versions.

Here's some examples of the run-around:

<details>

<summary>Run `go get github.com/google/cel-go@v0.25.0`</summary>

```
$ bazel run @rules_go//go -- get github.com/google/cel-go@v0.25.0
Computing main repo mapping: 
Loading: 
Loading: 0 packages loaded
bazel: Entering directory `/private/var/tmp/_bazel_jward/63162deb35ba6b2f678241a64a1d59bb/execroot/_main/'
Analyzing: target @@rules_go+//go:go (0 packages loaded, 0 targets configured)
Analyzing: target @@rules_go+//go:go (0 packages loaded, 0 targets configured)

Analyzing: target @@rules_go+//go:go (19 packages loaded, 10305 targets configured)

INFO: Analyzed target @@rules_go+//go:go (92 packages loaded, 17359 targets configured).
INFO: Found 1 target...
bazel: Leaving directory `/private/var/tmp/_bazel_jward/63162deb35ba6b2f678241a64a1d59bb/execroot/_main/'
Target @@rules_go+//go/tools/go_bin_runner:go_bin_runner up-to-date:
  .bazel/bin/external/rules_go+/go/tools/go_bin_runner/bin/go
INFO: Elapsed time: 1.404s, Critical Path: 0.01s
INFO: 1 process: 11 action cache hit, 1 internal.
INFO: Build completed successfully, 1 total action
INFO: Running command line: .bazel/bin/external/rules_go+/go/tools/go_bin_runner/bin/go <args omitted>
go: upgraded cel.dev/expr v0.23.0 => v0.23.1
go: upgraded github.com/google/cel-go v0.24.1 => v0.25.0
rules_go: Running 'bazel mod tidy' since go.mod changed...
 no actions running
DEBUG: /private/var/tmp/_bazel_jward/63162deb35ba6b2f678241a64a1d59bb/external/rules_jvm_external+/private/lib/coordinates.bzl:76:18: Assuming com.google.code.gson:gson:jar:2.9.0 should be interpreted as com.google.code.gson:gson:2.9.0@jar
DEBUG: /private/var/tmp/_bazel_jward/63162deb35ba6b2f678241a64a1d59bb/external/rules_jvm_external+/private/lib/coordinates.bzl:76:18: Assuming info.picocli:picocli:jar:4.6.3 should be interpreted as info.picocli:picocli:4.6.3@jar
DEBUG: /private/var/tmp/_bazel_jward/63162deb35ba6b2f678241a64a1d59bb/external/gazelle+/internal/bzlmod/go_deps.bzl:607:40: 
Mismatch between versions requested for Go module cel.dev/expr:

  bazel_dep version (MODULE.bazel): 0.23.0 (as "cel-spec")
  Go module version (go.mod):       0.23.1
 
Either ensure that you have

  bazel_dep(module_name = "cel-spec", version = "0.23.1")

in your MODULE.bazel file or downgrade the Go module version via

  bazel run @rules_go//go -- get cel.dev/expr@v0.23.0
```
</details>

<details>

<summary>Remove `bazel_dep` on `cel-spec`</summary>

```
bazel_dep(
    name = "cel-spec",
    version = "0.23.0",
)
```

```
$ bazel mod tidy
 no actions running
DEBUG: /private/var/tmp/_bazel_jward/63162deb35ba6b2f678241a64a1d59bb/external/rules_jvm_external+/private/extensions/maven.bzl:351:18: The maven repository 'maven' has contributions from multiple bzlmod modules, and will be resolved together: ["core", "grpc-java"]
DEBUG: /private/var/tmp/_bazel_jward/63162deb35ba6b2f678241a64a1d59bb/external/rules_jvm_external+/private/lib/coordinates.bzl:76:18: Assuming com.google.code.gson:gson:jar:2.9.0 should be interpreted as com.google.code.gson:gson:2.9.0@jar
DEBUG: /private/var/tmp/_bazel_jward/63162deb35ba6b2f678241a64a1d59bb/external/rules_jvm_external+/private/lib/coordinates.bzl:76:18: Assuming info.picocli:picocli:jar:4.6.3 should be interpreted as info.picocli:picocli:4.6.3@jar
DEBUG: /private/var/tmp/_bazel_jward/63162deb35ba6b2f678241a64a1d59bb/external/gazelle+/internal/bzlmod/go_deps.bzl:607:40: 
Mismatch between versions requested for Go module cel.dev/expr:

  bazel_dep version (MODULE.bazel): 0.15.0 (as "cel-spec")
  Go module version (go.mod):       0.23.1
 
Either ensure that you have

  bazel_dep(module_name = "cel-spec", version = "0.23.1")

in your MODULE.bazel file or downgrade the Go module version via

  bazel run @rules_go//go -- get cel.dev/expr@v0.15.0
```
</details>

TristonianJones:
Disable cel.dev/expr gazelle build generation

Closes #1184


TristonianJones:
FYI @fmeum @ouillie

ouillie:
Here's an outline for how to verify if a solution will work for #1184:

Start by cloning @jarreds reproduction: https://github.com/jarreds/cel-go-repro.

<details><summary>Click for potentially superfluous details that might be helpful for understanding exactly what's happening.</summary>

As the README in that repo says, the problem surfaces when you introduce the bzlmod dependency on `grpc` then try `bazel build cel-test`. Looking at the output of `bazel mod graph`, the only chain of module dependencies that leads to `cel-spec` is `grpc` -> `xds` -> `cel-spec`. Looking at the [only version of `xds`](https://github.com/bazelbuild/bazel-central-registry/blob/main/modules/xds/0.0.0-20240423-555b57e/MODULE.bazel) in BCR, you'll notice it's using a pretty old version of `cel-spec` (0.15.0). You can override this by adding `bazel_dep(name = "cel-spec")` to the repro's MODULE file (either with a version, or with a `local_path_override`). FTR, I just tested with the latest `cel-spec` version (0.25.1) and observed the same failure as with 0.15.0, but I did the rest of my testing with 0.15.0 anyway.
</details>

Then, clone `cel-spec` locally, check out tag `v0.15.0`, apply [the patches](https://github.com/bazelbuild/bazel-central-registry/tree/main/modules/cel-spec/0.15.0/patches) listed for that version in BCR, and use a `local_path_override` in the reproduction repo to bring `cel-spec` in from this local clone.

Here's what I know: if you edit `go.mod` in the local `cel-spec` clone, and change the module name at the top to anything other than `cel.dev/expr`, the build in the reproduction repo should start working. For instance, it works when you change it to `module cel.dev/exp` (missing the final 'r'). I *cannot* confidently tell you *why* that works, but I laid out my suspicion in https://github.com/google/cel-go/issues/1184#issuecomment-3459134336. Basically, I think it has to do with Gazelle getting confused by conflicting definitions for a Go module called `cel.dev/expr`.

It's possible that entirely ridding the `cel-go` repo of the Gazelle dependency on `cel.dev/expr` would also fix this. I left a comment to that effect.

go.mod
```diff
@@ -1,15 +1,13 @@
 module github.com/google/cel-go
 
-go 1.22.0
-
-toolchain go1.23.0
+go 1.23.0
 
 require (
-	cel.dev/expr v0.24.0
+	cel.dev/expr v0.25.1
```
ouillie, line 6: ~I believe the fact that this is still here is a problem. It's still relying on the Go module dependency, and not exclusively on the bzlmod dependency, for this Go module.~

edit: this comment was based on a misunderstanding of how Gazelle works.

fmeum: This is correct as is, Gazelle recognizes that the Bazel module provides the Go module and only uses the former.

ouillie: Well, I tested this out locally using:

```
replace github.com/google/cel-go => <path to 'bzl-deps' branch of cel-go>
```

in the `go.mod` file of the reproduction repo, and the problem persisted, but I'm all out of ideas as to how this can be fixed (other than renaming the Go module as I described in https://github.com/google/cel-go/pull/1228#issuecomment-3540438196).

fmeum: A local `replace` would override the bazel_dep usage, I think, so that is not a realistic test setup.

ouillie: Any suggestions as to how this could be tested locally? Or any insights as to why renaming the module in `cel-spec`, or using [my workaround in the root repository](https://github.com/jarreds/cel-go-repro/compare/main...ouillie:cel-go-repro:main), does fix the issue?

fmeum: Your override forces the Go module to win over the Bazel module, which explains the behavior. Renaming the module does as well since the Go module is then seen as distinct from the Bazel module.

It's true that testing this is pretty difficult. `--override_repository` comes to mind as a way to "just" override a repo in Bazel without any changes observable to module extensions.

ouillie: I'm a little confused, because there is no `cel-go` Bazel module. There is a `cel-spec` Bazel module which corresponds to the Go import path `cel.dev/expr`, whereas `cel-go` refers to the Go import path `github.com/google/cel-go`.

Are you saying that by replacing `cel-go`, it's also overriding all the transitive dependencies of `cel-go` (including `cel-spec`) so that Gazelle no longer tries to use a `bazel_dep` for any of those where it normally would have?

TristonianJones: I'm happy to also make a cel-go BCR module if that helps the situation as well

ouillie: I don't think it will, but should be easy to test that out locally since it already has a MODULE.bazel file.

I'm starting to think this is a bug in Gazelle. It looks like Gazelle should actually already see the transitive `cel-spec` bazel_dep via `grpc` -> `xds` in the Bazel module graph, so adding it to `cel-go` as well shouldn't actually change anything (especially since, in my case, `cel-go` is also a transitive dependency pulled in via another go_dep, so its `MODULE.bazel` file just doesn't come into play for me in the first place).

Gazelle definitely has logic to try to handle Go modules supplied as bazel_deps, but it's possible this only really works for the root module, and not the other go_dep modules generated by Gazelle. Idk. Debugging this is hard.

Then again, if Gazelle were fixed (assuming it's not currently WAI), then everyone who uses gRPC would by default use that ancient version of `cel-spec` (from `xds`) to satisfy `cel-go`'s dependency on `cel-spec`, unless they manually add a bazel_dep on `cel-spec` to the root module. Or, if you make cel-go into a BCR module too, they could just add that as a bazel_dep. Either way, it's a bit of a leaky abstraction for people like me who don't actually depend on either one directly.

This all gets back to my previous assertion that the cleanest solution would have been to make the cel-spec bazel module completely language agnostic, which would've sidestepped this whole issue, but I'm assuming that's a non-starter.

TristonianJones: I've disabled build file generation with gazelle, mirroring your workaround @ouillie. PTAL

ouillie: Unfortunately, I don't think that's going to help either. `cel-go` is probably always consumed as a go_dep (because it's not published to BCR), so I believe Gazelle would not actually bother to even look at its MODULE.bazel file. It would just be ignored.

If it were distributed as a Bazel module, it looks like Gazelle would actually [fail if it encountered that `gazelle_override`](https://github.com/bazel-contrib/bazel-gazelle/blob/v0.47.0/internal/bzlmod/go_deps.bzl#L89) in a bazel_dep. So, that's clearly not something Gazelle wants libraries to do anyway (which makes sense; a lot of module-level customizations are only permitted in the root module).

So, that workaround does help, but only if everybody who encounters this problem (which only occurs when `cel-spec` is included as both bazel_dep and a go_dep via different dependency chains) adds it to the MODULE.bazel file of their *root* module.

Sorry I can't be more helpful. I'm sure you have a limited debugging budget like the rest of us. I'd be willing to sink a bit more time into this if I had a clearer understanding of whether or not Gazelle actually *intends* to support this corner case in the first place. There's clearly logic to try to substitute Go modules provided by bazel_deps vs go_deps, but it looks to me like it only works for the root module, and maybe doesn't work for transitive go_dep dependencies as in this case. But I'm not even totally sure of that.

l46kok (approved):

PR #1230, opened 2025-11-14
Remove strcase dependency

TristonianJones:
Preserve the prior strcase behavior and remove dependency on strcase


common/types/map.go
```diff
@@ -1036,3 +1036,32 @@ func InsertMapKeyValue(m traits.Mapper, k, v ref.Val) ref.Val {
 	}
 	return NewErr("insert failed: key %v already exists", k)
 }
+
+func upperCamelCase(s string) string {
```
jcking, line 1040: I think Unicode rules exist where doing this rune by rune may not be exactly correct. But I might be mis-remembering. Do we support non-ASCII?

TristonianJones: The `strcase.UpperCamelCase` function just covers rune-by-rune, so this is only here to preserve the existing behavior.

TristonianJones:
@jcking PTAL

jcking (approved):

PR #1232, opened 2025-11-17
Fix the formatting directives to support what's documented in cel-spec

TristonianJones:
The cel-spec formatting docs expect that int, uint are supported with '%e' and '%f' directives as floating point. The code attempts to do this, but failed to convert to floating point resulting in a very unusual text output.


hudlow:
IMO, we should have failing conformance tests before this merges.

jcking (approved):

Issue #1237, opened 2025-11-19
Upstream kubernetes CIDR.contains functionality


tdesrosi:
**Feature request checklist**

- [X] There are no issues that match the desired change
- [ ] The change is large enough it can't be addressed with a simple Pull Request
- [ ] If this is a bug, please file a [Bug Report](./ISSUE_TEMPLATE).

**Change**

The basic premise is to match upstream functionality in [kubernetes](https://github.com/kubernetes/kubernetes/blob/8d450ef773127374148abad4daaf28dac6cb2625/staging/src/k8s.io/apiserver/pkg/cel/library/cidr.go#L218) for CIDR/IP range built-in functions.

Kubernetes has implemented a robust set of CEL functions for handling IP addresses and CIDR ranges (e.g., isIP, cidr, ip, and containment checks). These are currently locked inside k8s.io/apiserver, but they are generally useful for any policy engine dealing with network logic (firewalls, access lists, etc.).

The implementation would introduce opaque types for IP and CIDR to ensure correctness, rather than treating them as raw strings.

**Example**

Match [kubernetes functionality](https://kubernetes.io/docs/reference/using-api/cel/): 

```
cidr('192.168.0.0/24').containsIP(ip('192.168.0.1'))
```

**Alternatives considered**

Using matches() for IP validation is error-prone and difficult to maintain (especially for IPv6). We also are using startsWith() in various places, but this makes CEL policies which check IP inclusion in a CIDR range long and repetitive.

Loading a WASM module for basic network logic is overkill and introduces performance overhead for simple checks.

**Process**

I (@tdesrosi) will being this work.

gudvinr:
What's stopping you from just importing k8s.io/apiserver? Or copying same files into your project directly. CEL isn't only for networking applications so doesn't seem like a great idea to bloat standard library with stuff for specific domain.

And cel-go isn't the only implementation of CEL. When you share types with protobuf, you don't need per-language library implementation. But if one implementation has this specific library and other doesn't, you can't share expressions anymore.

TristonianJones:
Hi @gudvinr, I've actually suggested copying the types and much of the implementation from K8s directly. Taking a dependency on k8s.io/server would lead to a circular dependency, and there's likely a portability aspect where we'll need to introduce something like a CIDR / IP interface to ensure future interop with K8s during a future CEL bump -- extensions are opt-in, but it would be great to have a single extension library for networking.

[...]

PR #1238, opened 2025-11-20
Upstreaming network.IP/CIDR to CEL-go from kubernetes

tdesrosi:
# Upstream CIDR and IP-related functions from kubernetes into cel-go

This is part of a broader effort to bring network functions from the kubernetes
project into CEL specifications upstream. This is related directly to
[issues/1237](https://github.com/google/cel-go/issues/1237).

These are currently locked inside k8s.io/apiserver, but they are generally
useful for any policy engine dealing with network logic (firewalls, access lists, etc.).


TristonianJones:
FYI @cici37 @jpbetz

ext/network.go
```diff
@@ -0,0 +1,540 @@
...
+	}
+	ip := net.ParseIP(string(s))
+	if ip == nil {
+		return types.NewErr("invalid ip address: %s", s)
+	}
+	return types.Bool(cidr.IPNet.Contains(ip))
+}
+
+func netCIDRContainsCIDR(lhs, rhs ref.Val) ref.Val {
+	parent, ok := lhs.(CIDR)
```
TristonianJones: These sorts of type guards can be removed since there are generic type guards generated during the function declaration

It should be safe to just write `parent := lhs.(CIDR)`

ext/network.go
```diff
@@ -0,0 +1,540 @@
...
+	return IP{IP: cidr.ExtraIP, Str: cidr.ExtraIP.String()}
+}
+
+// --- Opaque Type Wrappers ---
+
+// IP is an exported CEL value that wraps net.IP.
+// It implements ref.Val.
+type IP struct {
+	net.IP
+	Str string
```
TristonianJones: Is the `Str` field intended to be exposed?

ext/network.go
```diff
@@ -0,0 +1,540 @@
...
+
+func (i IP) Value() any {
+	return i.IP
+}
+
+// CIDR is an exported CEL value that wraps *net.IPNet.
+// It implements ref.Val.
+type CIDR struct {
+	*net.IPNet
+	ExtraIP net.IP
```
TristonianJones: What does `ExtraIP` do, and should it be a value when the embedded type is a pointer?

ext/network.go
```diff
@@ -0,0 +1,540 @@
...
+	Str     string
+}
+
+func (c CIDR) ConvertToNative(typeDesc reflect.Type) (any, error) {
+	if typeDesc == reflect.TypeOf(&net.IPNet{}) {
```
TristonianJones: Here and elsewhere, prefer `typeDesc == reflect.TypeFor[*net.IPNet]()` instead of instantiating the type. The rest of CEL should upgrade to do the same, but this new method came about more recently than the majority of CEL code.

ext/network.go
```diff
@@ -0,0 +1,540 @@
...
+	return networkIPType
+}
+
+func (i IP) Value() any {
+	return i.IP
+}
+
+// CIDR is an exported CEL value that wraps *net.IPNet.
+// It implements ref.Val.
+type CIDR struct {
```
TristonianJones, line 530: Consider copying the K8s definition directly: staging/src/k8s.io/apiserver/pkg/cel/cidr.go

Additionally you could create an interface for denoting CIDR types to support backwards compatibility with K8s, but we can follow up with that later.

ext/network.go
```diff
@@ -0,0 +1,540 @@
...
+		return types.MaybeNoSuchOverloadErr(val)
+	}
+	return IP{IP: cidr.ExtraIP, Str: cidr.ExtraIP.String()}
+}
+
+// --- Opaque Type Wrappers ---
+
+// IP is an exported CEL value that wraps net.IP.
+// It implements ref.Val.
+type IP struct {
```
TristonianJones, line 483: Please copy this type definition: staging/src/k8s.io/apiserver/pkg/cel/ip.go

ext/network.go
tdesrosi: Acknowledged, will change.

ext/network.go
tdesrosi: This won't be necessary using net/netip rather than net.

ext/network.go
tdesrosi: net.ParseCIDR("192.168.1.5/24") returned both an IP object and IPNet object, which masked address bits to zero automatically. netip.Prefix remembers the host bits, so it shouldn't be a problem any more.

ext/network.go
```diff
@@ -0,0 +1,481 @@
...
+	containsIPFunc       = "containsIP"
+	containsCIDRFunc     = "containsCIDR"
+	maskedFunc           = "masked"
+	prefixLengthFunc     = "prefixLength"
+	ipFromCIDRFunc       = "ip"
+)
+
+var (
+	// Definitions for the Opaque Types
+	networkIPType   = types.NewTypeValue("network.IP", traits.ReceiverType)
```
TristonianJones: You can omit the `traits.ReceiverType` here and below. It implies the type implements the `Receive` method which it doesn't.

TristonianJones: Prefer creating an `types.OpaqueType` since these are types which do not have any fields accessible.

ext/network.go
```diff
@@ -0,0 +1,481 @@
...
+
+func (*networkLib) CompileOptions() []cel.EnvOption {
+	return []cel.EnvOption{
+		// 1. Register Types
+		cel.Types(
+			networkIPType,
+			networkCIDRType,
+		),
+		// 2. Register Adapter (Bundled here so it applies automatically)
+		cel.CustomTypeAdapter(&networkAdapter{
```
TristonianJones: Prefer instead to create an inline `cel.EnvOption` like so:

```go
func (e *cel.Env) (*cel.Env, error) {
   adapter := &networkAdapter{Adapter: e.CELTypeAdapter()}
   return cel.CustomTypeAdapter(adapter)(e)
}
```

ext/network.go
TristonianJones: The type definition here doesn't look the same as the one from K8s. Could you please use the one from K8s?

ext/network.go
TristonianJones: Likewise to the IP type comment, could you please use the K8s definition verbatim?

ext/network.go
```diff
@@ -0,0 +1,481 @@
...
+			Adapter: types.DefaultTypeAdapter,
+		}),
+		// 3. Register Functions
+		cel.Function(isIPFunc,
```
TristonianJones: I would be very happy if we used the K8s content verbatim here (modulo some simple renames) as it'll simplify cost-tracking checks which K8s is doing presently and which someone will also need to upstream before making the improvements available to K8s to use.

https://github.com/kubernetes/kubernetes/blob/5bcb7599736327cd8c6d23e398002354a6e40f68/staging/src/k8s.io/apiserver/pkg/cel/library/ip.go

The primary cleanup would be to shift from an `IP` library to a `Network` library which aggregates the K8s `library/ip.go` and `library/cidr.go` since we'd very likely always use them together.

ext/network.go
```diff
@@ -0,0 +1,488 @@
...
+	containsIPFunc       = "containsIP"
+	containsCIDRFunc     = "containsCIDR"
+	maskedFunc           = "masked"
+	prefixLengthFunc     = "prefixLength"
+	ipFromCIDRFunc       = "ip"
+)
+
+var (
+	// Definitions for the Opaque Types
+	networkIPType   = types.NewOpaqueType("network.IP")
```
TristonianJones: Please make these variables exported with the same names that K8s uses:

```
var (
	IPType = cel.OpaqueType("net.IP")
)
```

ext/network.go
```diff
@@ -0,0 +1,488 @@
...
+	return []cel.EnvOption{
+		// 1. Register Types
+		cel.Types(
+			networkIPType,
+			networkCIDRType,
+		),
+
+		// 2. Register Functions
+		cel.Function(isIPFunc,
+			cel.Overload("isIP_string", []*cel.Type{cel.StringType}, cel.BoolType,
```
TristonianJones: For the greatest degree of compatibility at the AST level, you'll need to make sure that the function and overload names from K8s are copied here exactly, e.g. swap `ip_isLoopback` with `ip_is_loopback` ... there are a few others where the casing doesn't match.

ext/network.go
```diff
@@ -0,0 +1,488 @@
...
+	// Delegate to the wrapped adapter (e.g., Protobuf adapter)
+	return a.Adapter.NativeToValue(value)
+}
+
+// --- Implementation Logic ---
+
+func parseIPAddr(raw string) (netip.Addr, error) {
+	addr, err := netip.ParseAddr(raw)
+	if err != nil {
+		return netip.Addr{}, fmt.Errorf("IP Address %q parse error: %v", raw, err)
```
TristonianJones: To avoid tripping up change detector tests, it's useful to copy the error messages directly from K8s as well:

```
		return netip.Addr{}, fmt.Errorf("IP Address %q parse error during conversion from string: %v", raw, err)
```

ext/network.go
```diff
@@ -0,0 +1,488 @@
...
+	_, err := netip.ParsePrefix(string(s))
+	return types.Bool(err == nil)
+}
+
+func netIPString(val ref.Val) ref.Val {
+	s := val.(types.String)
+	str := string(s)
+	addr, err := parseIPAddr(str)
+	if err != nil {
+		return types.NewErr("%v", err)
```
TristonianJones: Prefer to use `types.WrapErr(err)` here

ext/network.go
```diff
@@ -0,0 +1,488 @@
...
+
+func netIsIP(val ref.Val) ref.Val {
+	s := val.(types.String)
+	_, err := parseIPAddr(string(s))
+	return types.Bool(err == nil)
+}
+
+func netIsCIDR(val ref.Val) ref.Val {
+	s := val.(types.String)
+	_, err := netip.ParsePrefix(string(s))
```
TristonianJones: Can you make sure the function implementations are being done consistently with K8s? There are some specific constraints that K8s enforces which I'm not seeing here. Might be way to copy the function implementation and make the main edit the removal of the unnecessary function guards.

https://github.com/kubernetes/kubernetes/blob/5bcb7599736327cd8c6d23e398002354a6e40f68/staging/src/k8s.io/apiserver/pkg/cel/library/cidr.go#L281

tdesrosi:
A note about these changes: To maintain strict AST-level compatibility with k8s, I'm going with a split registration pattern (declaring signatures in CompileOptions and bindings in ProgramOptions), rather than bundling them in CompileOptions via cel.Function.

When using the modern cel-go helper cel.Function() to register both the global overload (ip(string)) and the member overload (cidr.ip()) simultaneously, the internal dispatcher validation incorrectly flags the self-referencing overload ID "ip" as a collision (overload already exists).

To resolve this collision with the modern helper, we would be forced to rename the overload ID to something distinct like "ip_string". While functionally equivalent, this breaks strict parity with the Kubernetes AST reference data. If requested, we can go this route (and maintain parity minus the "ip" --> "ip_string" difference in overload IDs).

TristonianJones:
> A note about these changes: To maintain strict AST-level compatibility with k8s, I'm going with a split registration pattern (declaring signatures in CompileOptions and bindings in ProgramOptions), rather than bundling them in CompileOptions via cel.Function.

Actually, there really isn't any difference here between how the bindings are configured. K8s stages it a little differently, but it's materially identical between the two approaches.
 
> When using the modern cel-go helper cel.Function() to register both the global overload (ip(string)) and the member overload (cidr.ip()) simultaneously, the internal dispatcher validation incorrectly flags the self-referencing overload ID "ip" as a collision (overload already exists).

It looks like the function `ip` overloads aren't properly configured  as the overload_id values don't match those in K8s. Take another look as the two overloads in K8s are `cidr_ip` and `string_to_ip`.
 
It's worth looking at the other overloads as well.

tdesrosi:
Thanks Tristan, here's one more pass rechecking overloads and member overloads.

ext/network.go
```diff
@@ -0,0 +1,495 @@
...
+//
+// Examples:
+//
+//	cidr('10.0.0.0/8').containsIP(ip('10.0.0.1')) == true
+//	cidr('10.0.0.0/8').containsIP('10.0.0.1') == true
+//	cidr('10.0.0.0/8').containsCIDR('10.1.0.0/16') == true
+//	cidr('192.168.1.5/24').ip() == ip('192.168.1.5')
+//	cidr('192.168.1.5/24').masked() == cidr('192.168.1.0/24')
+//	cidr('192.168.1.0/24').prefixLength() == 24
+func Network() cel.EnvOption {
```
TristonianJones: Please add a version option here, even though it's not functional yet, as it will make forward compatible much clearer from a maintenance standpoint within K8s.

ext/network.go
```diff
@@ -0,0 +1,495 @@
...
+		cel.Function(isLinkLocalMcastFunc,
+			cel.MemberOverload("ip_is_link_local_multicast", []*cel.Type{IPType}, cel.BoolType,
+				cel.UnaryBinding(netIPIsLinkLocalMulticast)),
+		),
+		cel.Function(isLinkLocalUcastFunc,
+			cel.MemberOverload("ip_is_link_local_unicast", []*cel.Type{IPType}, cel.BoolType,
+				cel.UnaryBinding(netIPIsLinkLocalUnicast)),
+		),
+		cel.Function(containsIPFunc,
+			cel.MemberOverload("cidr_contains_ip", []*cel.Type{CIDRType, IPType}, cel.BoolType,
```
TristonianJones: The naming here is what CEL recommends as the convention, but K8s uses `cidr_contains_ip_ip`

ext/network.go
```diff
@@ -0,0 +1,495 @@
...
+
+func (*networkLib) CompileOptions() []cel.EnvOption {
+	return []cel.EnvOption{
+		// 1. Register Types
+		cel.Types(
+			IPType,
+			CIDRType,
+		),
+
+		// 2. Register Functions
```
TristonianJones, line 205: Looks like we're missing `cidr_to_string` and `ip_to_string` as part of the `string` conversion function.

ext/network.go
```diff
@@ -0,0 +1,526 @@
...
+//
+// Examples:
+//
+//	cidr('10.0.0.0/8').containsIP(ip('10.0.0.1')) == true
+//	cidr('10.0.0.0/8').containsIP('10.0.0.1') == true
+//	cidr('10.0.0.0/8').containsCIDR('10.1.0.0/16') == true
+//	cidr('192.168.1.5/24').ip() == ip('192.168.1.5')
+//	cidr('192.168.1.5/24').masked() == cidr('192.168.1.0/24')
+//	cidr('192.168.1.0/24').prefixLength() == 24
+func Network(version uint32) cel.EnvOption {
```
TristonianJones: Instead of passing version directly, prefer a functional `NetworkOption` that indicates the `NetworkVersion(int)`, similar to how the `BindingsOption` is configured:

https://github.com/google/cel-go/blob/e9f15ea3c1841839cb0c1136e7dad95cde3e896b/ext/bindings.go#L69-L78


https://github.com/google/cel-go/blob/e9f15ea3c1841839cb0c1136e7dad95cde3e896b/ext/bindings.go#L54-L56

ext/network.go
```diff
@@ -0,0 +1,526 @@
...
+var (
+	// Definitions for the Opaque Types
+	IPType   = types.NewOpaqueType("net.IP")
+	CIDRType = types.NewOpaqueType("net.CIDR")
+)
+
+type networkLib struct {
+	version uint32
+}
+
```
TristonianJones, line 192: Please add comments to all exported (UpperCase) functions

ext/network.go
```diff
@@ -0,0 +1,526 @@
...
+func (*networkLib) CompileOptions() []cel.EnvOption {
+	return []cel.EnvOption{
+		// 1. Register Types
+		cel.Types(
+			IPType,
+			CIDRType,
+		),
+
+		// 2. Register Functions
+		cel.Function(isIPFunc,
```
TristonianJones: minor nit, but could you order these functions by name alphabetically? The K8s files do this and it makes the review diff a bit easier to reason about. Perhaps group by CIDR first, then IP?

ext/network.go
```diff
@@ -0,0 +1,526 @@
...
+
+func (i IP) Value() any {
+	return i.Addr
+}
+
+type CIDR struct {
+	netip.Prefix
+}
+
+func (c CIDR) ConvertToNative(typeDesc reflect.Type) (any, error) {
```
TristonianJones, line 535: Please add unit tests for conversion functions.

ext/network.go
```diff
@@ -0,0 +1,526 @@
...
+	case types.TypeType:
+		return CIDRType
+	}
+	return types.NewErr("type conversion error from '%s' to '%s'", CIDRType, typeValue)
+}
+
+func (c CIDR) Equal(other ref.Val) ref.Val {
+	o, ok := other.(CIDR)
+	if !ok {
+		return types.ValOrErr(other, "no such overload")
```
TristonianJones: CEL types no longer error when the type is not the same. The current semantic is to treat the runtime equality check as 'false'

ext/network.go
```diff
@@ -0,0 +1,526 @@
...
+	case types.TypeType:
+		return IPType
+	}
+	return types.NewErr("type conversion error from '%s' to '%s'", IPType, typeValue)
+}
+
+func (i IP) Equal(other ref.Val) ref.Val {
+	o, ok := other.(IP)
+	if !ok {
+		return types.ValOrErr(other, "no such overload")
```
TristonianJones: CEL types should return false if the types aren't the same. This is a shift in behavior from how CEL originally started.

ext/network.go
```diff
@@ -0,0 +1,526 @@
...
+				cel.BinaryBinding(netCIDRContainsCIDRString)),
+		),
+		cel.Function(maskedFunc,
+			cel.MemberOverload("cidr_masked", []*cel.Type{CIDRType}, CIDRType,
+				cel.UnaryBinding(netCIDRMasked)),
+		),
+		cel.Function(prefixLengthFunc,
+			cel.MemberOverload("cidr_prefix_length", []*cel.Type{CIDRType}, cel.IntType,
+				cel.UnaryBinding(netCIDRPrefixLength)),
+		),
```
TristonianJones, line 286: Optional, but consider introducing static validators for ip and cidr conversions from string literals as this will save people a lot of trouble with runtime errors which could have been caught at compile time.

https://github.com/google/cel-go/blob/e9f15ea3c1841839cb0c1136e7dad95cde3e896b/cel/validator.go#L261-L279

TristonianJones (commented):
I've raised https://github.com/google/cel-spec/pull/507 to cel-spec, so shortly after both artifacts are checked in and a new cel-spec release cut, let's enable the conformance tests

ext/network.go
```diff
@@ -0,0 +1,526 @@
...
+		}
+
+		// Install the Adapter (Wrapping the existing one)
+		adapter := &networkAdapter{Adapter: e.CELTypeAdapter()}
+		return cel.CustomTypeAdapter(adapter)(e)
+	}
+}
+
+const (
+	// Function names matching Kubernetes implementation
```
jpbetz:
```suggestion
	// Function names matching the original Kubernetes implementation of this networking library
```

?

jpbetz:
@lalitc375

ext/network.go
```diff
@@ -0,0 +1,526 @@
+// Copyright 2025 Google LLC
```
jpbetz, line 1: The cost calculations for this library exist in https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apiserver/pkg/cel/library/cost.go in the Kuberhetes code.  Is it possible to also upstream the cost calculations?

TristonianJones: Yep, I'll tackle this after the initial implementation is in

ext/network.go
tdesrosi: Since the helper in cel/validator.go is not exported, I created a networkFormatValidator directly in network.go that mimics its behavior with AST tools. Is something like this what you had in mind?

jpbetz:
@TristonianJones https://github.com/kubernetes/kubernetes/issues/134224#issuecomment-4047131848 caught an issue with this library that I recommend addressing before releasing upstream.

tdesrosi:
Thanks @jpbetz, I caught myself up on https://github.com/kubernetes/kubernetes/issues/134224#issuecomment-4047131848, and I've added isStrictCIDR and isInterfaceAddress to try to address it. Let me know if this is the preferred way to handle this. Thanks!

jpbetz:
> Thanks @jpbetz, I caught myself up on [kubernetes/kubernetes#134224 (comment)](https://github.com/kubernetes/kubernetes/issues/134224#issuecomment-4047131848), and I've added isStrictCIDR and isInterfaceAddress to try to address it. Let me know if this is the preferred way to handle this. Thanks!

+ @danwinship CIDR specifics

ext/network.go
```diff
@@ -0,0 +1,635 @@
...
+//
+// The `isCIDR` function checks if a string is a valid CIDR notation. Note that
+// `isCIDR` is "loose" and allows CIDRs with non-zero host bits (e.g.,
+// '10.0.0.1/8'). For strict validation of subnets, use `isStrictCIDR`.
+//
+// The `isStrictCIDR` function checks if a string is a valid canonical CIDR
+// (no host bits).
+//
+// The `isInterfaceAddress` function is an alias for `isCIDR` that explicitly
+// signifies the intent to allow host bits.
```
danwinship: "loose"/"strict" is biased towards Kubernetes's use cases, where we assume the majority of APIs want a subnet/mask.

I think I'd suggest just `isCIDR()` as a top-level method, which notes that it accepts CIDR values with or without host bits, and then have a method `<cidr>.isMask() -> bool` which checks if the host bits are 0 (aka, if `<cidr> == <cidr>.masked()`).

danwinship: (We had talked about trying to deprecate `isCIDR` in k8s, but since we don't have any way to implement CEL function deprecation anyway, maybe a better approach is to just "lint" the CRD rules, and if we see `isCIDR(val)`, warn that they they should say either `isCIDR(val) && cidr(val).isMask()` or `isCIDR(val) && !cidr(val).isMask()`.)

tdesrosi: Removed isStrictCIDR and isInterfaceAddress, and added isMask() so users can use `isCIDR(val) && cidr(val).isMask()` to validate CIDR values with or without host bits. I've also adjusted the doc comment to reflect this.

TristonianJones (approved):

TristonianJones:
There appear to be a few negative tests which aren't exercised, but I'll add them in a follow up PR that enables the networking cel-spec tests:

1. Kubernetes evaluates expressions checking the type names, e.g. `type(ip("192.168.0.1")) == net.IP` and `type(cidr("192.168.0.0/24")) == net.CIDR`. Though such tests are absent here, they should be covered in the conformance tests.

2. Kubernetes verifies compile errors for type mismatches, such as passing a CIDR to isIP: `isIP(cidr("192.168.0.0/24"))` expecting found `no matching overload for 'isIP' applied to '(net.CIDR)'`.

ext/network.go
TristonianJones: That's a reasonable way to go, thanks!

Issue #974, opened 2024-07-04
[Policy compiler] Return explanation output policy


seirl:
The policy compiler now parses the "explanation" field of match blocks, but does not do anything with it.

It would be useful to provide an API that returns a copy of the policy, except the output would be replaced by the contents of the "explanation" block. This would effectively generate a separate expression that outputs the explanation instead of the result of the policy.

PR #1244, opened 2025-12-08
Refactor match compiling to accept user-defined logic.

catalinadurbala:
What I'd like to achieve here is having the match output compilation be customizable - I don't necessarily need the rest of CompileMatches, but I thought it might be simpler to reason about it as a whole. Please let me know if you'd prefer a different approach.

### DIFF
--- a/policy/compiler.go
+++ b/policy/compiler.go
@@ -119,6 +119,16 @@ type CompiledMatch struct {
 	nestedRule *CompiledRule
 }
 
+// NewCompiledMatch creates a CompiledMatch.
+func NewCompiledMatch(exprID int64, cond *cel.Ast, output *OutputValue, nestedRule *CompiledRule) *CompiledMatch {
+	return &CompiledMatch{
+		exprID:     exprID,
+		cond:       cond,
+		output:     output,
+		nestedRule: nestedRule,
+	}
+}
+
 // SourceID returns the source identifier associated with the compiled match.
 func (m *CompiledMatch) SourceID() int64 {
 	return m.exprID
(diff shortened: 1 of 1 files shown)


policy/compiler.go
```diff
@@ -248,22 +278,28 @@ func CompileRule(env *cel.Env, p *Policy, opts ...CompilerOption) (*CompiledRule
 			c.env = env
 		}
 	}
-	return c.compileRule(p.Rule(), c.env, iss)
+	return c.CompileRuleImpl(p.Rule(), p, c.env, iss)
 }
 
-type compiler struct {
+// Compiler holds the state of the policy compiler.
+type Compiler struct {
```
TristonianJones: Prefer using functional options to configure the fields you're looking for rather than exporting the struct.

TristonianJones:
Hi Catalina, I'm struggling to understand the usage pattern you're looking to enable. Could you help me understand the use case better? The API change is substantial, so I'd prefer not to make the change as-is without having a really compelling usage scenario. Perhaps there are simpler alternatives to consider?

catalinadurbala:
We would like to be able to compile an AST that outputs structured explanatory metadata instead of the actual output for a given policy. What we do now is compiling a policy once for the output, mutating the policy to swap the metadata (explanation) for its output, and then compiling this mutated policy again. While hacky, this works for the single string explanation field that we were using so far. However we now need more structure in our explanations (i.e. a proto that we parse from multiple match subtags) and we'd also like to keep line and column information for all of the components.

Ideally we would be able to use branch annotations and only do one policy compilation that yields both the actual output and explanations for it. As second best, we would like to leave the policy as it is and instead be able to customize the output compilation - https://github.com/google/cel-go/issues/974 is related, but we've realised that we would benefit from more structure for the explanations and therefore the current Explanation field would not suffice anyway.

Please let me know if there's a different approach that you prefer that would help us achieve this. TIA!

catalinadurbala:
Closing this in favour of https://github.com/google/cel-go/pull/1246.

PR #1246, opened 2025-12-12
Refactor match output compiling to accept user-defined logic.

catalinadurbala:
As follow-up to https://github.com/google/cel-go/pull/1244, I've narrowed down the changes to the minimum surface that we need to interact with in order to compile our explanations. Please let me know what you think!

### DIFF
--- a/policy/compiler.go
+++ b/policy/compiler.go
@@ -189,6 +189,49 @@ func MaxNestedExpressions(limit int) CompilerOption {
 	}
 }
 
+// MatchCompiler is an interface that provides the necessary functionality for compiling the output
+// of a match.
+type MatchCompiler interface {
+	// Env returns the rule environment for the match compiler.
+	Env() *cel.Env
+	// RelSource returns a RelativeSource for a given expression id.
+	RelSource(ValueString) *RelativeSource
+	// ReportErrorAtID reports an error at the given expression id.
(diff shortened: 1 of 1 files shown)


catalinadurbala:
I've added a layer of abstraction to the input parameters to hide as many of the internals as possible - unfortunately I realised that the rule Env is still needed, as we're planning to use the [optimizer](https://github.com/google/cel-go/blob/6c59bf580ce820689a3fcb0931ef86081fab1ec0/cel/optimizer.go#L49) with it. Do let me know if that's alright with you.

policy/compiler.go
```diff
@@ -189,6 +189,46 @@ func MaxNestedExpressions(limit int) CompilerOption {
...
+	// ReportErrorAtID reports an error at the given expression id.
+	ReportErrorAtID(int64, string, ...any)
+}
+
+type matchCompilerImpl struct {
+	env *cel.Env
+	c   *compiler
+	iss *cel.Issues
+}
+
```
TristonianJones, line 208: Please make sure to add comments for all exposed methods, even if it's just `// RelSource implements the MatchCompiler interface` or `Env returns the current environment available to the match expression`

TristonianJones (dismissed):

seirl (approved):
Approving this since it's almost identical to the changes @TristonianJones approved earlier.

PR #1250, opened 2025-12-29
REPL: Support reading and writing YAML env configs

jnthntatum:
REPL: Support reading and writing YAML env configs

Adds support for reading and wring YAML env configs from REPL.

Uses a custom trailer comment to apply settings not supported by the
environment (variable bindings, type loading options)


repl/evaluator.go
```diff
@@ -277,12 +281,26 @@ type Optioner interface {
...
 }
 
+type contextOption struct {
+	opt Optioner
+	// True if the option is captured by a yaml config.
+	yaml bool
+}
+
+type genericOption struct {
+	o cel.EnvOption
```
l46kok: nit: spell out o? (envOpt or otherwise)

repl/evaluator.go
```diff
@@ -637,14 +656,82 @@ func (e *Evaluator) DelLetFn(name string) error {
...
+
+	switch format {
+	case statusFormatYAML:
+		return e.parseYAMLConfig(src)
+	}
+
+	return "", fmt.Errorf("configure: not yet implemented %v %v %v ", format, fromFile, configRef)
+}
+
 // Status returns a stringified view of the current evaluator state.
```
l46kok:
```suggestion
// REPLConfig returns a stringified view of the current evaluator state.
```

repl/evaluator.go
```diff
@@ -588,9 +607,9 @@ func (e *Evaluator) AddDeclFn(name string, params []letFunctionParam, typeHint *
 // AddOption adds an option to the basic environment.
 // Options are applied before evaluating any of the let statements.
 // Returns an error if setting the option prevents planning any of the defined let expressions.
-func (e *Evaluator) AddOption(opt Optioner) error {
+func (e *Evaluator) AddOption(opt Optioner, isYAML bool) error {
```
l46kok: nit: consider splitting this into a named method, to avoid ambiguity at the callsite (ex: `AddConfigOption`, `AddYamlOption` etc?)

repl/evaluator.go
```diff
@@ -675,6 +762,162 @@ func (e *Evaluator) Status() string {
 	return status.String()
 }
 
+func (e *Evaluator) YAMLConfig() string {
+	var status strings.Builder
+	status.WriteString("# CEL environment YAML\n")
+	conf, err := e.getEffectiveEnv().ToConfig("repl-session")
+	if err != nil {
+		panic(fmt.Errorf("REPL env broken %w", err))
```
l46kok: Could we consider a more graceful error handling here, unless if you believe this is an unrecoverable situation? Largely thinking in terms of preventing app restarts on the web based REPL

repl/evaluator_test.go
```diff
@@ -979,6 +982,158 @@ func TestProcess(t *testing.T) {
...
+					src:        "2",
+					resultType: mustParseType(t, "int"),
+					params:     nil,
+				},
+				&simpleCmd{
+					cmd:  "status",
+					args: []string{"--yaml"},
+				},
+			},
+			wantText: `# CEL environment YAML
```
l46kok, line 1011: Could we add a roundtrip test for this?

repl/testdata/test_env.yaml
```diff
@@ -0,0 +1,37 @@
...
+    - name: pb3
+      type_name: cel.expr.conformance.proto3.TestAllTypes
+    - name: pb2
+      type_name: cel.expr.conformance.proto2.TestAllTypes
+features:
+    - name: cel.feature.macro_call_tracking
+      enabled: true
+
+
+# REPL Bindings:
```
l46kok, line 32: Just something to consider for the future: I wonder if a multi-document would be more appropriate here:

```yaml
# Document 1: CEL Config
name: repl-session
container: cel.expr.conformance
... existing config
---
# Document 2: REPL bindings
name: repl-bindings
commands:
  - %load_descriptors --pkg 'cel-spec-test-types'
  - %let pb3 = proto3.TestAllTypes{}
  - %let pb2 = proto2.TestAllTypes{}
```

It's likely the existing config YAML parsers in go/java would only process the first document by default, so in theory we should be safe compatibility wise.

repl/evaluator.go
jnthntatum: Done. I don't think this is realistic to trigger but propagating the error is probably cleaner.

repl/evaluator_test.go
jnthntatum: Added a basic case. The config ordering isn't strictly stable (e.g. functions and libraries can reorder based on hash map traversal) so exact round tripping isn't possible in general.

repl/testdata/test_env.yaml
jnthntatum: Noted.

l46kok (approved):

PR #1253, opened 2026-01-06
Preserve source information during CEL policy composition.

abenea:
The old implementation outputs the wrong source information for policy files derived from the dummy AST created by RuleComposer.Compose. This change fixes that by preserving the correct source and merging the offset ranges of the rule match expressions inserted by the composer optimizer into the final AST.

### DIFF
--- a/cel/optimizer.go
+++ b/cel/optimizer.go
@@ -15,6 +15,7 @@
 package cel
 
 import (
+	"fmt"
 	"sort"
 
 	"github.com/google/cel-go/common"
@@ -29,17 +30,43 @@ import (
 // passes to ensure that the final optimized output is a valid expression with metadata consistent
 // with what would have been generated from a parsed and checked expression.
 //
-// Note: source position information is best-effort and likely wrong, but optimized expressions
+// Note: source position information is best-effort and incomplete, but optimized expressions
 // should be suitable for calls to parser.Unparse.
 type StaticOptimizer struct {
 	optimizers []ASTOptimizer
+	// If set, Optimize() will use this Source instead of the one from the AST.
+	sourceOverride *Source
 }
+
+type OptimizerOption func(*StaticOptimizer) (*StaticOptimizer, error)
 
 // NewStaticOptimizer creates a StaticOptimizer with a sequence of ASTOptimizer's to be applied
 // to a checked expression.
-func NewStaticOptimizer(optimizers ...ASTOptimizer) *StaticOptimizer {
-	return &StaticOptimizer{
-		optimizers: optimizers,
+func NewStaticOptimizer(options ...any) (*StaticOptimizer, error) {
+	so := &StaticOptimizer{}
+	var err error
+	for _, opt := range options {
+		switch v := opt.(type) {
+		case ASTOptimizer:
+			so.optimizers = append(so.optimizers, v)
+		case OptimizerOption:
+			so, err = v(so)
+			if err != nil {
+				return nil, err
+			}
+		default:
+			return nil, fmt.Errorf("unsupported option: %v", v)
+		}
+	}
+	return so, nil
+}
+
+// OptimizeWithSource overrides the source used by the optimizer.
+// Note this will cause the source info from the AST passed to Optimize() to be discarded.
+func OptimizeWithSource(source Source) OptimizerOption {
+	return func(so *StaticOptimizer) (*StaticOptimizer, error) {
+		so.sourceOverride = &source
+		return so, nil
 	}
 }
 
@@ -49,15 +76,21 @@ func NewStaticOptimizer(optimizers ...ASTOptimizer) *StaticOptimizer {
 func (opt *StaticOptimizer) Optimize(env *Env, a *Ast) (*Ast, *Issues) {
 	// Make a copy of the AST to be optimized.
 	optimized := ast.Copy(a.NativeRep())
+	source := a.Source()
+	sourceInfo := optimized.SourceInfo()
+	if opt.sourceOverride != nil {
+		source = *opt.sourceOverride
+		sourceInfo = ast.NewSourceInfo(*opt.sourceOverride)
+	}
 	ids := newIDGenerator(ast.MaxID(a.NativeRep()))
 
 	// Create the optimizer context, could be pooled in the future.
-	issues := NewIssues(common.NewErrors(a.Source()))
+	issues := NewIssues(common.NewErrors(source))
 	baseFac := ast.NewExprFactory()
 	exprFac := &optimizerExprFactory{
 		idGenerator: ids,
 		fac:         baseFac,
-		sourceInfo:  optimized.SourceInfo(),
+		sourceInfo:  sourceInfo,
 	}
 	ctx := &OptimizerContext{
 		optimizerExprFactory: exprFac,
@@ -80,7 +113,7 @@ func (opt *StaticOptimizer) Optimize(env *Env, a *Ast) (*Ast, *Issues) {
 
 		// Recheck the updated expression for any possible type-agreement or validation errors.
 		parsed := &Ast{
-			source: a.Source(),
+			source: source,
 			impl:   ast.NewAST(expr, info)}
 		checked, iss := ctx.Check(parsed)
 		if iss.Err() != nil {
@@ -91,7 +124,7 @@ func (opt *StaticOptimizer) Optimize(env *Env, a *Ast) (*Ast, *Issues) {
 
 	// Return the optimized result.
 	return &Ast{
-		source: a.Source(),
+		source: source,
 		impl:   optimized,
 	}, nil
 }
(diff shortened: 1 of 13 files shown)


cel/optimizer.go
```diff
@@ -32,14 +32,27 @@ import (
...
+		optimizers:      optimizers,
+		mergeSourceInfo: false,
+	}
+}
+
+// NewStaticOptimizerWithSourceInfoMerging creates a StaticOptimizer with a sequence of
+// ASTOptimizer's to be applied to a checked expression. The source info of the optimized AST will
+// be merged with the source info of the input AST which is useful when merging multiple expressions
+// defined in the same file. Used only for policy composition.
+func NewStaticOptimizerWithSourceInfoMerging(optimizers ...ASTOptimizer) *StaticOptimizer {
```
TristonianJones: Prefer instead of making a new method to make source info merge a functional option by introducing the following changes:

```go
type OptimizerOption func(*StaticOptimizer) (*StaticOptimizer, error)

NewStaticOptimizer(options ...any) (*StaticOptimizer, error) {
   so := &StaticOptimizer{}
   var err error
   for _, opt := range options {
     switch v := opt.(type) {
       case ASTOptimizer:
          so.optimizers = append(so.optimizers, v)
       case OptimizerOption:
          so, err = v(so)
          if err != nil {
             return nil, err
          }
        default:
          return nil, fmt.Errorf("unsupported option: %v", v)
     }
   }
   return so
}
```

policy/compiler.go
```diff
@@ -411,16 +408,56 @@ func (c *compiler) checkUnreachableCode(rule *CompiledRule, iss *cel.Issues) {
...
+func (v *idVisitor) VisitExpr(e ast.Expr) {
+	v.ids[e.ID()] = true
+}
+
+func (v *idVisitor) VisitEntryExpr(e ast.EntryExpr) {
+	v.ids[e.ID()] = true
+}
+
+// getAstIds returns a set of AST node IDs
+func getAstIds(celAst *cel.Ast) map[int64]bool {
```
TristonianJones: Prefer to uppercase acronyms, e.g. `getASTIDs`. There's some debt around the old names, but we shouldn't propagate it.

TristonianJones: prefer operating on the `ast.AST` rather than the `cel.Ast`.

policy/composer.go
```diff
@@ -71,12 +71,18 @@ type RuleComposer struct {
 
 // Compose stitches together a set of expressions within a CompiledRule into a single CEL ast.
 func (c *RuleComposer) Compose(r *CompiledRule) (*cel.Ast, *cel.Issues) {
-	ruleRoot, _ := c.env.Compile("true")
+	// dummyExpr is a placeholder expression used as the root of the AST before optimization. Because
+	// StaticOptimizer copies the source info from it, we must set the correct source info here.
+	source := recoverOriginalSource(r)
+	sourceInfo := ast.NewSourceInfo(source)
+	dummyExpr := ast.NewExprFactory().NewLiteral(1, types.True)
+	ruleRoot := cel.NewAst(source, ast.NewAST(dummyExpr, sourceInfo))
```
TristonianJones: Rather than modify the existing `rootRule` approach, consider packaging up the original source as an option to the static optimizer since it appears there's already a need to introduce awareness of such capabilities from this change.

```go
source := recoverOriginalSource(r)
opt, err := cel.NewStaticOptimizer(composer, cel.OptimizeWithSource(source))
if err != nil {
    // add the error to the issues list and return
}
opt.Optimize(c.env, ruleRoot)
```

cel/env.go
```diff
@@ -48,6 +48,14 @@ type Ast struct {
 	impl   *celast.AST
 }
 
+// NewAst creates a new Ast value from a source and its native representation.
+func NewAst(source Source, impl *celast.AST) *Ast {
```
TristonianJones: Let's not use this method, but rather pass the desired source to the optimizer to use as the basis for source metadata merging since that seems to be the desired effect. More details below.

cel/optimizer.go
```diff
@@ -96,10 +110,30 @@ func (opt *StaticOptimizer) Optimize(env *Env, a *Ast) (*Ast, *Issues) {
 	}, nil
 }
 
+func updateOffsetRanges(idGen ast.IDGenerator, info *ast.SourceInfo) {
+	newRanges := make(map[int64]ast.OffsetRange)
+	sortedOldIDs := []int64{}
+	for oldID := range info.OffsetRanges() {
```
TristonianJones: Prefer `copy(dst, src)` to the for-loop

cel/optimizer.go
```diff
@@ -96,10 +110,30 @@ func (opt *StaticOptimizer) Optimize(env *Env, a *Ast) (*Ast, *Issues) {
...
+}
+
 // normalizeIDs ensures that the metadata present with an AST is reset in a manner such
 // that the ids within the expression correspond to the ids within macros.
-func normalizeIDs(idGen ast.IDGenerator, optimized ast.Expr, info *ast.SourceInfo) {
+func normalizeIDs(idGen ast.IDGenerator, optimized ast.Expr, info *ast.SourceInfo, mergeSourceInfo bool) {
 	optimized.RenumberIDs(idGen)
+	if mergeSourceInfo {
+		updateOffsetRanges(idGen, info)
+	}
```
TristonianJones: Consider moving `updateOffsetRanges` into the `ast.SourceInfo` object as `RenumerIDs(idGen)` for consistency with the other renumbering code. Should be easier to unit tests that a copy of an AST can be renumbered in a way that's consistent with your expectations.

policy/compiler.go
```diff
@@ -411,16 +408,56 @@ func (c *compiler) checkUnreachableCode(rule *CompiledRule, iss *cel.Issues) {
...
+	info := ast.NativeRep().SourceInfo()
+	var keepIDs map[int64]bool
+	if !found {
+    // Remove the newly created offset ranges if the ValueString is not associated with a source
+    // position. This currently happens because the parser creates a synthetic "true" condition if a
+    // match doesn't have one.
+		keepIDs = make(map[int64]bool)
+	} else {
+		// Remove offset ranges for ids without a corresponding AST node. This can happen because the
+		// checker deletes some nodes while rewriting the AST. For example the Select operand is deleted
```
TristonianJones: It seems like we should be able to do this cleanup within the parser and the checker rather than here. Could we explore that approach instead? Maybe as a separate PR so we can remove the cleanup here after that one is submitted?

policy/compiler_test.go
```diff
@@ -516,3 +520,81 @@ func normalize(s string) string {
...
+func verifySourceInfoCoverage(t testing.TB, policy *Policy, ast *cel.Ast) {
+	t.Helper()
+	info := ast.SourceInfo()
+	if info == nil {
+		return
+	}
+
+	exprLines := exprLinesFromPolicy(policy)
+	coveredLines := make(map[int]bool)
+	ids := getAstIds(ast)
```
TristonianJones: Perhaps we should just expose this function on the `ast.AST` instead ... no harm in doing it that way.

policy/compiler.go
abenea: Done in #1258.

cel/optimizer.go
abenea: According to https://pkg.go.dev/builtin#copy it  only works with slices, but in this case I am trying to copy the keys from a map.

parser/helper.go
```diff
@@ -159,7 +159,7 @@ func (p *parserHelper) id(ctx any) int64 {
 		offset.Start = p.sourceInfo.ComputeOffset(int32(c.GetLine()), int32(c.GetColumn()))
 		offset.Stop = offset.Start + int32(len(c.GetText()))
 	case common.Location:
-		offset.Start = p.sourceInfo.ComputeOffset(int32(c.Line()), int32(c.Column()))
+		offset.Start = p.sourceInfo.ComputeOffsetAbsolute(int32(c.Line()), int32(c.Column()))
```
TristonianJones, line 162: Can you help me understand why you need to compute the absolute offset here? Usually, RelSource would handle logic related to hiding such details, but maybe I missed something?

TristonianJones (approved):
In general, looks good to me, just one question related to offset computations as it could lead to surprising results in change-detector tests.

parser/helper.go
abenea: The locations returned by RelativeSource are all [absolute](https://github.com/google/cel-go/blob/52280ba990db5690c5ab8ea260c290c299a3c253/policy/source.go#L60). The code was previously shifting an already absolute location usually resulting in out of range line numbers which were turned into -1.

TristonianJones: Do you have an example where the offset was bad previously and is now fixed? it'd be great to have a test case to prevent future regressions. My impression was that the relative source would provide an absolute offset based on a relative snippet of source ... so I'm having a hard time understanding when this expectation breaks?

abenea: Yes, I added a [new check in parser_test.go](https://github.com/google/cel-go/pull/1253/changes#diff-6eaacb6a6fa7a8b6851fa3bc9917a3cfc6a7a3e191069b6d59ae5385b9beb6faR2279-R2300) that didn't pass before this change. You can see examples by reverting this line and running the test.

I was also surprised by the fact that RelativeSource produces relative offsets and absolute locations and I think that's a design that should be changed but I didn't want to add more big refactorings to this already large PR.

PR #1254, opened 2026-01-08
Add parse only evaluation to REPL

jcking:


repl/evaluator.go
```diff
@@ -1394,6 +1406,31 @@ func (e *Evaluator) Evaluate(expr string) (ref.Val, *exprpb.Type, error) {
...
+		return nil, nil, err
+	}
+
+	act, _ = env.PartialVars(act)
+	val, _, err := p.Eval(act)
+	// expression can be well-formed and result in an error
+	return val, ast.ResultType(), err
+}
+
+// Evaluate sets up a CEL evaluation using the current REPL context.
```
jnthntatum: nit: delete this

repl/evaluator.go
```diff
@@ -1394,6 +1406,31 @@ func (e *Evaluator) Evaluate(expr string) (ref.Val, *exprpb.Type, error) {
 	return val, ast.ResultType(), err
 }
 
+// Evaluate sets up a CEL evaluation using the current REPL context.
```
jnthntatum: EvaluateParseOnly runs the CEL expression against the current REPL context without type checking

repl/evaluator.go
jcking: Whoops, accidently left from copy pasting. Fixed.

repl/evaluator.go
jcking: Thanks. Fixed.

jnthntatum (approved):

PR #1256, opened 2026-01-13
Namespace resolution fix

TristonianJones:
Updates to the type resolution rules in accordance with https://github.com/google/cel-spec/pull/506


interpreter/attributes.go
```diff
@@ -304,15 +324,26 @@ func (a *absoluteAttribute) String() string {
 // a type, then the result is `nil`, `error` with the error indicating the name of the first
 // variable searched as missing.
 func (a *absoluteAttribute) Resolve(vars Activation) (any, error) {
-	for _, nm := range a.namespaceNames {
+	var rootVars Activation
+	if len(a.disambiguateNames) > 0 {
+		rootVars = vars
+		for rootVars.Parent() != nil {
+			rootVars = rootVars.Parent()
```
jnthntatum: Are there any cases where the library user might implement Parent? (something like user types baseActivation{} extendedActivation{parent: baseActivation*}?

or is it only ever implemented by the interpreter for locals?

jnthntatum (dismissed):

interpreter/attributes.go
TristonianJones: This is a good point. It's probably safer to detect local activations

TristonianJones: PTAL. I've updated to `Unwrap` activations where `Unwrap` is only supported on activations which contain local scope.

jnthntatum (approved):

PR #1258, opened 2026-01-16
Clean up unused source info after checker rewrites the AST.

abenea:


checker/checker.go
```diff
@@ -67,6 +67,15 @@ func Check(parsed *ast.AST, source common.Source, env *Env) (*ast.AST, *common.E
 	for id, t := range c.TypeMap() {
 		c.SetType(id, substitute(c.mappings, t, true))
 	}
+	// Remove source info for IDs without a corresponding AST node. This can happen because
+	// check() deletes some nodes while rewriting the AST. For example the Select operand is
+	// deleted when a variable reference is replaced with a Ident expression.
+	ids := c.AST.IDs()
```
TristonianJones: Since AST values are mutable, you may want to simply have a cleanup function 'ClearUnusedIDs()` and add the same call to the Parser.

abenea: I moved this code into ClearUnusedIDs() and added a check in parser_test.go but it appears the parser doesn't produced unused source info IDs.

common/ast/ast.go
```diff
@@ -160,6 +160,25 @@ func MaxID(a *AST) int64 {
...
+func (a *AST) IDs() map[int64]bool {
+	visitor := make(idVisitor)
+	PostOrderVisit(a.Expr(), visitor)
+	for _, call := range a.SourceInfo().MacroCalls() {
+		PostOrderVisit(call, visitor)
+	}
+	return visitor
+}
+
+func (a *AST) ClearUnusedIDs() {
```
TristonianJones, line 174: Please add a comment to the function.

TristonianJones (dismissed):

TristonianJones (approved):

PR #1270, opened 2026-02-10
Support mixing CEL values and native values in native structs

### LINKED ISSUE #1223: NativeTypes from ext doesn't honor ref.Val.Type
**Describe the bug**

When using `ext.NativeTypes` together with proto-mapped or manually mapped types as fields, it will try to map them as native types ignoring the fact they implement `ref.Val` and already exist in type provider.

**To Reproduce**

- [ ]  parser
- [ ]  checker
- [x]  interpreter

```cel
ip("127.0.0.1") == conn.addr
```

Test setup:
```go
package xfiles

import "k8s.io/apiserver/pkg/cel/library"

type Conn struct {
    Addr library.IP `cel:"addr"`
}

...

env, err := NewEnv(
    library.IP(),
    ext.NativeTypes(
        ext.ParseStructTag("cel"),
        reflect.TypeFor[*Conn](),
    ),
)
```

When you execute an expression, CEL will complain about missing overload for `library.IP`.

**Expected behavior**

`conn.addr` gets resolved into `net.IP` (i.e. `library.IPType`) and expression works as expected.

**Additional context**
~~Add any other context about the problem here.~~

TristonianJones:
Support a combination of Go native structs with CEL `ref.Val` types

Closes #1223

### DIFF
--- a/ext/native_test.go
+++ b/ext/native_test.go
@@ -474,7 +474,7 @@ func TestNativeTypesJsonSerialization(t *testing.T) {
 					NestedListVal: ["first", "second"],
 				},
 				StringVal: "string",
-        custom_name: "name",
+                custom_name: "name",
 			}`,
 			out: `{
 				"BoolVal":  true,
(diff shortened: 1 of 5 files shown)


ext/native_test.go
```diff
@@ -1098,6 +1141,10 @@ func mustParseTime(t *testing.T, timestamp string) time.Time {
 	return out
 }
 
+type Custom struct {
```
jnthntatum: nit: is this used anywhere?

jnthntatum (approved):

PR #1283, opened 2026-03-09
Add checker, ast, and type-provider support for JSON names

TristonianJones:
Initial implementation of JSON field name support in the checker, ast, and types sub-packages.

Top-level APIs and runtime support based on the presence of the AST extension will be introduced
in subsequent PRs.

#1214

### DIFF
--- a/checker/checker_test.go
+++ b/checker/checker_test.go
@@ -2441,6 +2441,14 @@ _&&_(_==_(list~type(list(dyn))^list,
             }~google.expr.proto2.test.TestAllTypes^google.expr.proto2.test.TestAllTypes.singleInt32~int`,
 			outType: types.IntType,
 		},
+		{
+			in:        `TestAllTypes{singleInt32: 1, single_bool: true}.singleInt32`,
+			container: "google.expr.proto2.test",
+			env:       testEnv{optionalSyntax: true, jsonFieldNames: true},
+			err: `ERROR: <input>:1:41: undefined field 'single_bool'
+             | TestAllTypes{singleInt32: 1, single_bool: true}.singleInt32
+             | ........................................^`,
+		},
 	}
 }
 
--- a/checker/checker_test.go
+++ b/checker/checker_test.go
@@ -2413,6 +2413,34 @@ _&&_(_==_(list~type(list(dyn))^list,
               @result~bool^@result)~bool`,
 			outType: types.BoolType,
 		},
+		{
+			in:        `TestAllTypes{?singleInt32: {}.?i}`,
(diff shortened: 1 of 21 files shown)


common/types/provider.go
```diff
@@ -93,11 +93,50 @@ type Registry struct {
...
 // message that proto depends upon in its FileDescriptor.
 func NewRegistry(types ...proto.Message) (*Registry, error) {
-	p := &Registry{
+	return NewProtoRegistry(ProtoTypes(types...))
+}
+
+// RegistryOption configures the behavior of the registry.
+type RegistryOption func(r *Registry) (*Registry, error)
+
+// JSONFieldNames configures JSON field name support within the protobuf types in the registry.
```
jnthntatum, line 105: fine to fix later when you add runtime support and the common EnvOption:

you might want to document that opting in picks the json name over the normal name -- protojson generally supports either of the json name or the normal field names on the parse side (though you can configure it to be strict iirc). similar config for export: (https://pkg.go.dev/google.golang.org/protobuf/encoding/protojson#MarshalOptions)

checker/checker_test.go
```diff
@@ -2413,6 +2413,34 @@ _&&_(_==_(list~type(list(dyn))^list,
...
+			outType: types.NewObjectType(
+				"google.expr.proto2.test.TestAllTypes",
+			),
+		},
+		{
+			in:        `TestAllTypes{?singleInt32: {'i': 20}.?i}.singleInt32`,
+			container: "google.expr.proto2.test",
+			env:       testEnv{optionalSyntax: true, jsonFieldNames: true},
+			out: `google.expr.proto2.test.TestAllTypes{
+              ?singleInt32:_?._(
```
jnthntatum, line 2435: Could you document the decision to use the json name in the AST instead of the actual field name somewhere? I think it's reasonable for consistency / unparseability, but I would have expected this to just work like an alias instead of using a different naming scheme. Ok for follow-up / when you add runtime support.

jnthntatum (dismissed):

checker/checker_test.go
TristonianJones: Ah, this could have been my misunderstanding of the way Sean had implemented JSON support. I effectively just use the json name if that was what was provided on input. Does the Java version normalize to the proto-based names? I can ask Sean when he gets back.

common/types/provider.go
TristonianJones: Sounds good. Thank you.

jnthntatum (approved):

PR #1284, opened 2026-03-11
Optionally include reachable fieldpaths in prompt

jnthntatum:
Update the AI prompt template to optionally include all of the reachable fields from the input variables along with any documentation.

The fields are described in terms of CEL types.

Will add a bazel rule for generating the file descriptor set with source info in a follow-up -- for testing included one built locally.


jnthntatum:
closed in favor of rolling into #1285

PR #1285, opened 2026-03-11
Optionally include reachable fieldpaths in prompt

jnthntatum:
- Add support for expanding the set of reachable fields from top level variables in the prompt template.
  Fields are described in terms of CEL types including documentation where available. 

- Add a bazel rule to generate a file descriptor set from a proto_library that preserves source info.
  This is used to extract comments for formatting CEL environment descriptions.


cel/fieldpaths.go
```diff
@@ -0,0 +1,152 @@
...
+// fieldPath represents a selection path to a field from a variable in a CEL environment.
+type fieldPath struct {
+	celType *Type
+	// path represents the selection path to the field.
+	path        string
+	description string
+	isLeaf      bool
+}
+
+func (f *fieldPath) Documentation() *common.Doc {
```
TristonianJones, line 21: nit: don't forget to add documentation on 'Documentation' :)

common/types/provider.go
```diff
@@ -206,9 +209,11 @@ func (p *Registry) FindStructFieldType(structType, fieldName string) (*FieldType
 		return nil, false
 	}
 	return &FieldType{
-		Type:    fieldDescToCELType(field),
-		IsSet:   field.IsSet,
-		GetFrom: field.GetFrom}, true
+		Type:        fieldDescToCELType(field),
+		IsSet:       field.IsSet,
+		GetFrom:     field.GetFrom,
+		Description: field.Documentation(),
```
TristonianJones: Do we want to make population of the field documentation optional? Perhaps we could have this as a separate method that we treat as a private interface?

```go
type documentationProvider interface {
   FindStructFieldDescription(structType, fieldName string) (string, bool)
}

if docs, ok := env.CELTypeProvider().(documentationProvider); ok {
  return docs.FindStructFieldDescription(...)
}
return ""
```

vendor/cel.dev/expr/conformance/proto3/BUILD.bazel
```diff
@@ -0,0 +1,39 @@
...
+    default_visibility = ["//visibility:public"],
+    licenses = ["notice"],  # Apache 2.0
+)
+
+go_library(
+    name = "go_default_library",
+    srcs = [
+        "test_all_types.pb.go",
+    ],
+    importpath = "cel.dev/expr/conformance/proto3",
```
TristonianJones, line 27: Do you know if this runs correctly with `go test`? I've had some troubles with getting the modules to resolve correctly against `cel.dev`. Alternatively, to avoid another type dependency we could use the `proto3pb` directory under `test` to accomplish a similar feat. Not a requirement, just curious to hear your thoughts.

cel/testdata/field_paths.prompt.txt
```diff
@@ -0,0 +1,403 @@
...
+Variables:
+
+* team is a cel.testdata.Team
+
+Field Paths:
+
+* team is a cel.testdata.Team
+* team.members is a list(cel.testdata.Mutant)
+
+      The members of the team.
```
TristonianJones: The docs for function / overloads are included next to the name with a `-` to help with delimiting the symbol from the description. It might be worth doing the same here:

`team` - type: `cel.testdata.Team`, description: The members of the team

Doesn't necessarily work with multi-line strings, so perhaps:

```md
* name: `team`
  type: `cel.testdata.Team`
  field paths: // skip 'self'
  * path: `team.members`
    type: `list(cel.testdata.Mutant)`
    description: The members of the team.
```

Alternatively, we could split between variables and attributes:

```md
Variables: 

* name: `type`
  type: ...

In addition to the Variables, attributes of the variables which may be used in expressions include the following field paths:

* attribute: `team.members`
  type: `list(cel.testdata.Mutant)`
  description: The members of the team
```

vendor/cel.dev/expr/conformance/proto3/BUILD.bazel
jnthntatum: `go test ./...` works for me locally (tm) -- is there anything else I should check?

cel/fieldpaths.go
jnthntatum: Done

common/types/provider.go
jnthntatum: done

cel/testdata/field_paths.prompt.txt
jnthntatum: I revised to follow the first option. let me know how you like the format.

vendor/cel.dev/expr/conformance/proto3/BUILD.bazel
TristonianJones: Should be good then :)

cel/testdata/field_paths.prompt.txt
TristonianJones: Looks very readable, thanks.

TristonianJones (dismissed):

TristonianJones (approved):

PR #1286, opened 2026-03-12
Json field names runtime support

TristonianJones:
Support for JSON or proto-based field name accesses when the `JSONFieldNames(true)` feature is enabled.

### DIFF
--- a/cel/cel_test.go
+++ b/cel/cel_test.go
@@ -2234,6 +2234,62 @@ func TestContextProto(t *testing.T) {
 	}
 }
 
+func TestContextProtoJSONFieldNames(t *testing.T) {
+	descriptor := new(proto3pb.TestAllTypes).ProtoReflect().Descriptor()
+	env := testEnv(t, JSONFieldNames(true), DeclareContextProto(descriptor))
+	expression := `
+	singleInt64 == 1
+	&& singleDouble == 1.0
+	&& singleBool == true
+	&& singleString == ''
+	&& singleNestedMessage == google.expr.proto3.test.TestAllTypes.NestedMessage{}
+	&& standaloneEnum == google.expr.proto3.test.TestAllTypes.NestedEnum.FOO
+	&& singleDuration == duration('5s')
+	&& singleTimestamp == timestamp(63154820)
+	&& singleAny == null
+	&& singleUint32Wrapper == null
+	&& singleUint64Wrapper == 0u
+	&& repeatedInt32 == [1,2]
+	&& mapStringString == {'': ''}
+	&& mapInt64NestedType == {0 : google.expr.proto3.test.NestedTestAllTypes{}}`
+	ast, iss := env.Compile(expression)
+	if iss.Err() != nil {
+		t.Fatalf("env.Compile(%s) failed: %s", expression, iss.Err())
+	}
+	prg, err := env.Program(ast)
+	if err != nil {
+		t.Fatalf("env.Program() failed: %v", err)
+	}
+	in := &proto3pb.TestAllTypes{
+		SingleInt64:  1,
+		SingleDouble: 1.0,
+		SingleBool:   true,
+		NestedType: &proto3pb.TestAllTypes_SingleNestedMessage{
+			SingleNestedMessage: &proto3pb.TestAllTypes_NestedMessage{},
+		},
+		StandaloneEnum: proto3pb.TestAllTypes_FOO,
+		SingleDuration: &durationpb.Duration{Seconds: 5},
+		SingleTimestamp: &timestamppb.Timestamp{
+			Seconds: 63154820,
+		},
+		SingleUint64Wrapper: wrapperspb.UInt64(0),
+		RepeatedInt32:       []int32{1, 2},
+		MapStringString:     map[string]string{"": ""},
+		MapInt64NestedType:  map[int64]*proto3pb.NestedTestAllTypes{0: {}},
+	}
+	vars, err := ContextProtoVars(in, types.JSONFieldNames(true))
+	if err != nil {
+		t.Fatalf("ContextProtoVars(%v) failed: %v", in, err)
+	}
+	out, _, err := prg.Eval(vars)
+	if err != nil {
+		t.Fatalf("prg.Eval() failed: %v", err)
+	}
+	if out.Equal(types.True) != types.True {
+		t.Errorf("prg.Eval() got %v, wanted true", out)
+	}
+}
+
 func TestRegexOptimizer(t *testing.T) {
 	var stringTests = []struct {
 		expr          string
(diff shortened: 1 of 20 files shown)


cel/cel_test.go
```diff
@@ -3607,6 +3663,109 @@ func TestAstProgramNilValue(t *testing.T) {
...
+			name: "proto simple field",
+			expr: `msg.single_int32 == 1`,
+		},
+		{
+			name: "proto map field",
+			expr: `msg.map_string_string['key'] == 'value'`,
+		},
+		{
+			name:           "json simple field",
+			expr:           `msg.singleInt32 == 1`,
```
jnthntatum, line 3682: could you add a test for the fallback behavior?

`dyn(msg).single_int32 == dyn(msg).singleInt32` ought to do it.

what happens with messages unpacked from Any? does the field lookup still happen through the type provider or does the field access happen through proto reflection directly?

common/ast/ast.go
```diff
@@ -438,6 +438,14 @@ func (s *SourceInfo) Extensions() []Extension {
 	return s.extensions
 }
 
+// HasExtension returns whether the source info contains the extension which satisfies the minimum version requirement.
+func (s *SourceInfo) HasExtension(id string, minVersion ExtensionVersion) bool {
```
jnthntatum, line 445: This doesn't matter yet since this is the first, but wouldn't you want the AST extension version to be no greater than what we support? (ext.major version <= 1)? Feel free to punt until we actually need to version an extension

jnthntatum (changes requested):
LGTM -- just suggest adding a test case.

cel/cel_test.go
TristonianJones: It should all happen through type-provider, but I'll add some tests to confirm.

common/ast/ast.go
TristonianJones: It's a good question. I'm not sure what the right choice is. Often there's an expection of version equivalence at the major version level. (so `ext.Version.Major == minVersion.Major`) with flexibility on the minor version `(ext.Version.Minor >= minVersion.Minor)`. How does that sound to you?

jnthntatum: Sounds good to me.

jnthntatum (approved):

Airblader:
@TristonianJones We've been needing this feature, and upon checking again, I was very happy to see it just so happened to have been implemented. However, unfortunately this seems like a hard switch, or am I missing something? We have stored CEL expressions in our system, and making it either-or means we'd have to migrate all expressions at the same time as rolling out enabling this feature.

Is there any way opting into JSON name support can be made smoother, e.g. by allowing a mixed mode where either JSON name or field name is supported?

TristonianJones:
@Airblader Old expressions should continue to work as-is since the JSON name support is only used at runtime if the expression was type-checked with JSON names enabled. Your existing stored and compiled expressions should continue to work. The test case below, reproduced from `cel_test.go` shows that at runtime the code will prefer JSON and fallback to proto-based names. 

```go
{
	name:           "json with proto fallback",
	expr:           `dyn(msg).single_int32 == dyn(msg).singleInt32`,
	jsonFieldNames: true,
},
```

If you have expressions that don't work with json names, you could always extend the existing environment via `env.Extend(cel.JSONFieldNames(true))` and compile the expression twice if you really needed to. Without more details though, it's hard for me to know what your use case is in practice. Care to share more examples?

Airblader:
@TristonianJones Thanks, I appreciate the reply.

> Without more details though, it's hard for me to know what your use case is in practice. Care to share more examples?

We basically do something akin to Kubernetes where we have CRDs that define CEL expressions for validation. We have existing CRDs using snake_case (= field names in the proto schema), but we'd like to move towards using JSON names as we use those for some other things, too. When I simply added in `cel.JSONNames(true)` into our CEL environment, tests started failing because snake_case names were no longer working.

Interestingly, using `registry.WithJSONFieldNames(true)` instead but not enabling it on the CEL environment seems to work. I do have to admit that I don't understand all the finer details around the CEL architecture here, and whether there's something bad about doing it this way.

*Edit:* I suppose the important detail implicitly mentioned here is that we have a custom registry.

Airblader:
I am running this a bit through Claude, and it suggests a gap in the implementation. By no means do I want to mindlessly repeat that evaluation, I trust you folks more than my half-informed Claude conversation, but perhaps you can help me clear up this misunderstanding somehow? 

```
The PR review comments on cel-go #1283 reveal that even the cel-go reviewer expected JSONFieldNames(true) to work as an alias (accepting both naming conventions), not as a strict replacement:

  ▎ jnthntatum (reviewer): "protojson generally supports either of the json name or the normal field names on the parse side [...] I would have expected this to just work like an alias instead of using a different naming scheme."

  And the author (TristonianJones) acknowledged this might not be correct:

  ▎ TristonianJones: "Ah, this could have been my misunderstanding of the way Sean had implemented JSON support. I effectively just use the json name if that was what was provided on input."

  So the checker's strict rejection of proto names when JSONFieldNames(true) is enabled was not an intentional design choice — it was the author's initial implementation, and the reviewer flagged that it should probably work as an alias. The "backwards compatible" second commit added the type-provider fallback, but the checker-level IsJSONField strictness
  was never relaxed to match.

  This means:
  1. The cel-go implementation has a design gap — the reviewer expected alias behavior, but the checker is stricter than intended
  2. Our approach (registry.WithJSONFieldNames(true) without the env option) is actually more aligned with the original reviewer's expectation — both names accepted as aliases
  3. We could upstream a fix to cel-go to make cel.JSONFieldNames(true) accept both as aliases, but that's out of scope for now

  The bottom line: cel.JSONFieldNames(true) was intended to be backward-compatible but the implementation fell short. Our workaround achieves what was originally intended.
```

jnthntatum:
Sure. Missing context is that this was ported from cel-java -- that implementation made the decision to only resolve JSON names at check time to avoid ambiguous expressions like `obj.jsonName == obj.json_name`. At runtime it is supposed to be more permissive (for parse-only expressions and cases where the struct is dynamic) hence the type provider change.

My comment that Claude picked up on was just a request to document the behavior, I think it was correct to behave the same as the Java implementation here. @l46kok fyi

l46kok:
While it's technically possible to make the type-checker permissive to allow resolution of both styles, we made an explicit design choice to require an all-or-nothing flag. Mixing naming conventions in a single expression causes both ambiguity as @jnthntatum mentioned, and degrades readability over time (e.g., `msg.user_id == 'foo' && msg.accountStatus == 'ACTIVE'`) . A core tenet of CEL's design philosophy is determinism and simplicity; we strongly prefer enforcing a single source of truth, especially for type-checking and canonical AST representation.

Another issue is collisions:

```
message UserEvent {
  string current_status = 1 [json_name = "previous_status"];
  string previous_status = 2; // Protobuf auto-generates the JSON name: "previousStatus"
}
```

expression: `event.previous_status`

With the current implementation, the type-checker unambiguously resolves to field 1 or 2 depending on the json field name option. If the type-checker were to allow resolution of both styles, we would either have to make such resolution an error state (undesirable from UX perspective), or define a strict, convoluted fallback priority. 

A fallback priority is dangerous for a policy language: if a developer explicitly types the native name of field 2 (previous_status), a JSON-first fallback would silently intercept and evaluate field 1 instead. We'd much rather force a loud compile-time error than allow silent logical shadowing. 

Hopefully that provides some clarity. If you really wanted to allow both, type-checking the expression twice with and without the option set is a workaround.

Airblader:
I appreciate your guys' time to reply to me. This all makes a ton of sense, and we definitely also want to get there (meaning using _only_ JSON names), but we need a migration path that allows either field names or JSON names in a given expression, though not being able to mix both in the same expression _is_ OK. In our specific case, collisions isn't something we are worried about, though of course a fair point.

Our schemas today use `json_name` options already, but CEL expressions are written entirely against field names (because cel-go previously didn't support anything else). So at least temporarily we need to allow expressions written with (exclusively) either field or JSON names. If I merely enable `cel.JSONNames(true)`, our tests break, however, because the field names (snake_case) used in existing CEL expressions no longer work. 

Is using `registry.WithJSONFieldNames(true)` instead an OK approach for this (it seems to work, at least), or is there a reason to prefer the path through `env.Extend(...)`?

TristonianJones:
My advice @Airblader, would be to try the following:

```go
env, err := cel.NewEnv(...options w/o JSON names...)
jsonEnv, err := env.Extend(cel.JSONFieldNames(true))

ast, iss := jsonEnv.Compile(expr)
if iss.Err() != nil {
  pbAst, pbIss := env.Compile(expr)
  if pbIss.Err() == nil {
    ast = pbAst
  }
}
``` 

You end up potentially compiling twice, but it also gives you the option to signal to your users (or internal processes) that an update to JSON names is available should they wish to translate to the new format (or have you translate for them)

Airblader:
Got it, again, really appreciate the discussion and support. Thanks all!

PR #1288, opened 2026-03-17
Add runtime cost tracking for string extension functions

qq96932100:
`stringLib.ProgramOptions()` returns an empty slice, so string extension calls (replace, split, join, indexOf, etc.) all get charged a flat runtime cost of 1 in `costCall()` regardless of input size. This makes `CostLimit` ineffective for these operations — chained `replace` calls can produce exponentially growing output while staying under the cost budget.

The lists and regex extensions already register proper cost trackers. This adds the same for string extension overloads, version-gated to match function availability.

### DIFF
--- a/ext/strings.go
+++ b/ext/strings.go
@@ -28,9 +28,11 @@ import (
 	"golang.org/x/text/language"
 
 	"github.com/google/cel-go/cel"
+	"github.com/google/cel-go/common"
 	"github.com/google/cel-go/common/types"
 	"github.com/google/cel-go/common/types/ref"
 	"github.com/google/cel-go/common/types/traits"
+	"github.com/google/cel-go/interpreter"
 )
 
 const (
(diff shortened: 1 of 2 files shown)


ext/strings.go
```diff
@@ -553,8 +555,41 @@ func (lib *stringLib) CompileOptions() []cel.EnvOption {
 }
 
 // ProgramOptions implements the Library interface method.
-func (*stringLib) ProgramOptions() []cel.ProgramOption {
-	return []cel.ProgramOption{}
+func (lib *stringLib) ProgramOptions() []cel.ProgramOption {
+	trackers := []interpreter.CostTrackerOption{
```
TristonianJones, line 559: Even though the functions are introduced at a specific version, changes in cost tracking also need to be versioned, so you'll have to make the version check `>= 4` for all cost changes.

ext/strings.go
```diff
@@ -553,8 +555,41 @@ func (lib *stringLib) CompileOptions() []cel.EnvOption {
 }
 
 // ProgramOptions implements the Library interface method.
-func (*stringLib) ProgramOptions() []cel.ProgramOption {
-	return []cel.ProgramOption{}
+func (lib *stringLib) ProgramOptions() []cel.ProgramOption {
```
TristonianJones, line 558: Note, it's important to introduce cost calculations for the checked expression as well, otherwise the runtime cost value may far exceed the expected cost

TristonianJones:
@qq96932100 thanks for the contributions!

PR #1289, opened 2026-03-20
Fuzzing: add FuzzMacros for comprehension macro coverage

SteppL10n-prog:
Add FuzzMacros fuzzer targeting comprehension macros (all, exists,
filter, map, exists_one) which are not covered by the existing
FuzzCompile or FuzzEval fuzzers.

The fuzzer wraps randomly generated expressions inside comprehension
contexts to increase coverage of macro expansion and type-checking
code paths in the CEL parser and checker.

This fuzzer is intended for use with OSS-Fuzz integration.

### DIFF
--- a/cel/fuzz_macros.go
+++ b/cel/fuzz_macros.go
@@ -1,0 +1,66 @@
+package cel
+
+// FuzzMacros targets comprehension macros (all, exists, filter, map)
+// and optional field selection which are not covered by FuzzCompile or FuzzEval.
+func FuzzMacros(data []byte) int {
+	if len(data) < 2 {
+		return 0
+	}
+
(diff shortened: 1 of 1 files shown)


TristonianJones:
Thank you for the contribution, but the fuzzers are actually maintained here: https://github.com/google/oss-fuzz/blob/master/projects/cel-go

The `fuzz_eval.go` does cover macros as well, though not explicitly. You're welcome to contribute there and we'd be happy to review and discuss contribution on that repo. 

Thanks!

SteppL10n-prog:
Thank you for clarifying Tristan!..I understand. I will close this PR.
I will open a new PR there shortly.

PR #1290, opened 2026-03-22
Preserve runtime error node IDs from Resolve

### LINKED ISSUE #1191: 6b5d14c3 breaks resolution of runtime error locations in macros
**Describe the bug**

We use the NodeID() value from runtime errors to be able to locate runtime errors in programs. The change in 6b5d14c3 (identified by git bisect) appears to have broken this for nodes that exist within the AST in macro expansions. After the change, errors are noted as existing at the site of the unexpanded macro, rather than at the site within the macro's body.

For example with the following program:
```
get(state.url).Body.decode_json().records.map(r,
	get(state.url+'/'+r.id).Body.decode_json()).as(events, {
//                    ^~~~ r.id not converted to string: can't add integer to string.
		"events": events,
})
```
note that in this context, `state.url` is a `dyn` that is not known until runtime.

(before):
```
failed eval: ERROR: <input>:3:26: no such overload
 |   get(state.url+'/'+r.id).Body.decode_json()).as(events, {
 | .........................^
```
(after):
```
failed eval: ERROR: <input>:5:14: no such overload
 |    "events": events,
 | .............^
```

Neither is perfect, but the one after the change is worse.

**To Reproduce**

- [ ]  parser
- [ ]  checker
- [x]  interpreter

Test setup:

Best done by reference to github.com/elastic/mito@48813c5bbc46326feb15c8aab98cb08bae2c1fa6 and the corresponding test adjustment in github.com/elastic/mito@6bd3ded032c5cd2bb44ca0aad50a14a80e736857

I can put together a smaller reproducer, but I'd like to get an idea of whether this is something that we just have to live with before I do that>

**Expected behavior**

Something more similar to the "before" case.

efd6:
# Pull Requests Guidelines

development branch.

## Commit Messages

```
Preserve runtime error node IDs from Resolve

When absoluteAttribute.Resolve() encounters a *types.Err variable value,
return it directly rather than calling Unwrap(). The Unwrap call strips
the Err wrapper and its node ID, causing callers to re-label the error
with the observation site's ID instead of the originating expression's ID.

Additionally, make WrapErr pass through existing *types.Err values rather
than double-wrapping them. This ensures that LabelErrNode sees the
original non-zero node ID and preserves it, regardless of the call site.

Fixes #1191
```

## Reviews

* Perform a self-review.

## Merging

  if applicable.

### DIFF
--- a/interpreter/attributes.go
+++ b/interpreter/attributes.go
@@ -349,7 +349,7 @@ func (a *absoluteAttribute) Resolve(vars Activation) (any, error) {
 		obj, found := v.ResolveName(nm)
 		if found {
 			if celErr, ok := obj.(*types.Err); ok {
-				return nil, celErr.Unwrap()
+				return nil, celErr
 			}
 			obj, isOpt, err := applyQualifiers(v, obj, a.qualifiers)
 			if err != nil {
--- a/common/types/err.go
+++ b/common/types/err.go
@@ -113,6 +113,9 @@ func ValOrErr(val ref.Val, format string, args ...any) ref.Val {
 
 // WrapErr wraps an existing Go error value into a CEL Err value.
 func WrapErr(err error) ref.Val {
+	if err, ok := err.(*Err); ok {
+		return err
+	}
 	return &Err{error: err}
 }
 
--- a/interpreter/attributes_test.go
+++ b/interpreter/attributes_test.go
@@ -1106,6 +1106,21 @@ func TestAttributeStateTracking(t *testing.T) {
 			),
 			out: types.NewUnknown(5, types.QualifyAttribute[string](types.NewAttributeTrail("a"), "b")),
 		},
+		{
+			expr: `['a', b.val, 'c'].filter(i, i + 'b' != 'ab')`,
+			vars: []*decls.VariableDecl{
+				decls.NewVariable("b", types.NewMapType(types.StringType, types.DynType)),
+			},
+			in: partialActivation(
+				map[string]any{
+					"b": map[string]any{
+						"val": 1,
+					},
+				},
+			),
+			// Error node 9 corresponds to the `i + 'b'` expression
+			out: types.LabelErrNode(9, types.NoSuchOverloadErr()),
+		},
 	}
 	for _, test := range tests {
 		tc := test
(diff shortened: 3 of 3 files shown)


interpreter/interpretable.go
```diff
@@ -167,6 +167,9 @@ func (test *evalTestOnly) Eval(ctx Activation) ref.Val {
 	val, err := test.Resolve(ctx)
 	// Return an error if the resolve step fails
 	if err != nil {
+		if celErr, ok := err.(*types.Err); ok {
```
TristonianJones: Can you wrap the conditional check into the `types.WrapErr` directly? That will cover more branches of execution.

interpreter/attributes.go
```diff
@@ -349,7 +349,7 @@ func (a *absoluteAttribute) Resolve(vars Activation) (any, error) {
 		obj, found := v.ResolveName(nm)
 		if found {
 			if celErr, ok := obj.(*types.Err); ok {
-				return nil, celErr.Unwrap()
+				return nil, celErr
```
TristonianJones, line 352: Can you add the following repro case into `attributes_test.go` L1106:


```go
               {
                       expr: `['a', b.val, 'c'].filter(i, i + 'b' != 'ab')`,
                       vars: []*decls.VariableDecl{
                               decls.NewVariable("b", types.NewMapType(types.StringType, types.DynType)),
                       },
                       in: partialActivation(
                               map[string]any{
                                       "b": map[string]any{
                                               "val": 1,
                                       },
                               },
                       ),
                       // Error node 9 corresponds to the `i + 'b'` expression
                       out: types.LabelErrNode(9, types.NoSuchOverloadErr()),
               },
```

You'll also have to update the test conditions to the following L1173:

```diff
                        } else if tc.out.Equal(out) != types.True {
                                t.Errorf("got %v, wanted %v", out, tc.out)
                        }
+                       if err, isErr := out.(*types.Err); isErr && err.NodeID() != 0 {
+                               wantErr := tc.out.(*types.Err)
+                               if err.NodeID() != wantErr.NodeID() {
+                                       t.Errorf("err.NodeID() got %d, wanted %d", err.NodeID(), wantErr.NodeID())
+                               }
+                       }
                        for id, val := range tc.state {

```

TristonianJones (changes requested):

efd6:
PTAL

TristonianJones (approved):

TristonianJones (approved):

PR #1291, opened 2026-03-23
fix: cache concatList.Size() to prevent O(N^2) evaluation time

Flo354:
`concatList.Size()` recomputes the total size by recursively calling `Size()` on `prevList` and `nextList`. After N concatenation operations, the resulting tree has depth N, and each `Size()` call traverses the full tree. List operations like `exists()`, `all()`, and `filter()` call `Size()` on each iteration, producing O(N^2) total time.

This means the CEL cost tracker (which uses `Size()` for cost estimation) reports O(N) cost while the actual evaluation time is O(N^2), bypassing `CostLimit`.

With N=1000 and `CostLimit` set: reported cost is 17011 (within budget), but actual evaluation takes 12s (706x the expected time).

The fix caches the computed size in a `sync.Once` field so repeated calls return in O(1).

**Files changed:** `common/types/list.go`

### DIFF
--- a/common/types/list.go
+++ b/common/types/list.go
@@ -126,12 +126,6 @@ func (l *baseList) Add(other ref.Val) ref.Val {
 	if !ok {
 		return MaybeNoSuchOverloadErr(other)
 	}
-	if l.Size() == IntZero {
-		return other
-	}
-	if otherList.Size() == IntZero {
-		return l
-	}
 	return newConcatList(l.Adapter, l, otherList)
 }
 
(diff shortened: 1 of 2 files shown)


common/types/list.go
```diff
@@ -477,7 +480,10 @@ func (l *concatList) Iterator() traits.Iterator {
 
 // Size implements the traits.Sizer interface method.
 func (l *concatList) Size() ref.Val {
-	return l.prevList.Size().(Int).Add(l.nextList.Size())
```
TristonianJones, line 480: I believe this is safe to do during object construction without a `sync` since the lists are immutable. Perhaps, create a `newConcatList` which abstracts the size checking logic currently adjacent to the two call sites where `&concatList` is created?

Flo354:
According to your last comment, I have added a newConcatList() constructor that precomputes the size, both Add() methods go through it now and removed the sync.

TristonianJones:
Thanks, @Flo354!

common/types/list.go
TristonianJones: The size checks here could be moved into the `newConcatList` method and then pruned from the other call site as well. Also, since both list sizes are checked, we can store the sizes into local variables and combine them into the `size` value set on the newly minted `concatList`

Flo354:
I did the changes, it's better indeed.
And from your comment on my other PR, I added a test case. It's on two angles. The first one verifies the size is correct, and the second one to check about the time required.

TristonianJones (approved):

PR #1292, opened 2026-03-23
fix: cap format string precision to prevent memory exhaustion

Flo354:
`parsePrecision()` in `ext/formatting.go` and `parsePrecisionV2()` in `ext/formatting_v2.go` accept unbounded precision values from format strings. An expression like `"%.9999999f".format([3.14])` causes `fmt.Sprintf` to allocate a string proportional to the precision value. At cost 501, a single expression can allocate 792MB.

The fix adds a `maxPrecision` constant (1000) and rejects precision values exceeding it. This is consistent with the practical limits of floating-point display (IEEE 754 double has ~15-17 significant digits).

**Files changed:** `ext/formatting.go`, `ext/formatting_v2.go`

### DIFF
--- a/ext/formatting.go
+++ b/ext/formatting.go
@@ -410,7 +410,9 @@ func (c *stringFormatter) Octal(arg ref.Val, locale string) (string, error) {
 
 // stringFormatValidator implements the cel.ASTValidator interface allowing for static validation
 // of string.format calls.
-type stringFormatValidator struct{}
+type stringFormatValidator struct {
+	maxPrecision int
+}
 
 // Name returns the name of the validator.
 func (stringFormatValidator) Name() string {
@@ -427,7 +429,7 @@ func (stringFormatValidator) Configure(config cel.MutableValidatorConfig) error 
 
 // Validate parses all literal format strings and type checks the format clause against the argument
 // at the corresponding ordinal within the list literal argument to the function, if one is specified.
-func (stringFormatValidator) Validate(env *cel.Env, _ cel.ValidatorConfig, a *ast.AST, iss *cel.Issues) {
+func (v stringFormatValidator) Validate(env *cel.Env, _ cel.ValidatorConfig, a *ast.AST, iss *cel.Issues) {
 	root := ast.NavigateAST(a)
 	formatCallExprs := ast.MatchDescendants(root, matchConstantFormatStringWithListLiteralArgs(a))
 	for _, e := range formatCallExprs {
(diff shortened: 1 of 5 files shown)


TristonianJones (approved):

TristonianJones (changes requested):
Would you mind adding test cases which trigger the failure condition?

ext/formatting.go
```diff
@@ -870,6 +870,8 @@ func parseFormattingClause(formatStr string, callback formatStringInterpolator)
 	}
 }
 
+const maxPrecision = 1000
```
TristonianJones: Would you mind making this limit a configurable option, e.g. `StringsOption` and by default set the limit to 100 if the string version library is greater than >= 5 (I believe we're at v4 currently)

Flo354:
I Made maxPrecision configurable with a new StringsMaxPrecision() option. 

As requested, default is 100 when version >= 5, no limit for earlier versions to keep backward compatibility. 
The limit is passed through the whole parsing chain down to parsePrecision(). 

As requested, I also added test cases.

ext/strings.go
```diff
@@ -544,9 +559,9 @@ func (lib *stringLib) CompileOptions() []cel.EnvOption {
 	}
 	if lib.version >= 1 {
 		if lib.version >= 4 {
-			opts = append(opts, cel.ASTValidators(stringFormatValidatorV2{}))
+			opts = append(opts, cel.ASTValidators(stringFormatValidatorV2{maxPrecision: maxPrecision}))
```
TristonianJones, line 565: I think you need to check whether the maxPrecision should be effectively unbounded at version 4 and set to the config value at version 5+ here and below.

Flo354: I want to make sure I understand correctly. Currently, maxPrecision defaults to 0 (unbounded) for version < 5 and 100 for version >= 5. It is computed once before the version-gated blocks. Do you want me to split the version >= 4 block so that version 4 explicitly gets 0 and only version >= 5 gets the configured limit? Or is there a different structure you have in mind?

For reference, at the beginning of CompileOptions, we have this code:
```
maxPrecision := lib.maxPrecision
if maxPrecision == 0 && lib.version >= 5 {
    maxPrecision = 100
}
```

TristonianJones: Hi @Flo354, I'm looking over the PR again, and it's probably just worth mentioning in a comment that only for version 5 or above would the default max precision be set. I didn't see the earlier defaulting.

Flo354:
Hello @TristonianJones, I moved the actual set of maxPrecision closer to its usage instead of letting it at the beginning of CompileOptions. And I added a comment.
Tests are running, please, test also on your end to avoid any regression.

TristonianJones (approved):

PR #1300, opened 2026-04-06
fix: deleted `IntToDuration` overload

alexsnaps:
The conversion isn't implemented, but mostly isn't part of the language specification. The behavior remains mostly unchanged, the resulting error changes tho, still an error:

 - pre: type conversion error from 'int' to 'google.protobuf.Duration'
 - post: found no matching overload for 'duration' applied to '(int)'

The error would now be caught by the checker, instead of the evaluation time one previously.

### DIFF
--- a/cel/testdata/field_paths.prompt.txt
+++ b/cel/testdata/field_paths.prompt.txt
@@ -256,7 +256,6 @@ Functions:
 * duration - convert a value to a google.protobuf.Duration
 
       duration(duration('1s')) // duration('1s')
-      duration(int) -> google.protobuf.Duration
       duration('1h2m3s') // duration('3723s')
 
 * dyn - indicate that the type is dynamic for type-checking purposes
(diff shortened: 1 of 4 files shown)


alexsnaps:
> Thanks for your pull request! It looks like this may be your first contribution to a Google open source project. 

It's not, changed the email to the one you have on file for the CLA.

TristonianJones (dismissed):

TristonianJones:
@alexsnaps could you update [testdata/standard_env.prompt.txt](https://github.com/google/cel-go/tree/master/cel/testdata/standard_env.prompt.txt) and remove line 222:

https://github.com/google/cel-go/blob/6b8f6d621e464429f39c01140d7f60a467eef5ae/cel/testdata/standard_env.prompt.txt#L222

alexsnaps:
Fixed

alexsnaps:
If I get this failure right, it complains about the line being gone now?
```
 - 	      duration(int) -> google.protobuf.Duration
```

Is that right? I really have no where the "authority" for that is sourced from... I have no experience with this test suite tbh

TristonianJones:
@alexsnaps it looks like there's one more file where the line `duration(int) -> google.protobuf.Duration` needs to be removed: `cel/testdata/field_paths.prompt.txt`

alexsnaps:
:crossed_fingers:

TristonianJones (approved):

PR #1301, opened 2026-04-10
Support shorthand types in env yaml and REPL 

jnthntatum:
Upstream parser for shorthand type specifier `map<int, string>`. This is used for describing environments.

Add support for parsing yaml environment configs using the string type specifier and update the REPL to use the new format.

Also rework repl to track types in terms of the environment config
TypeDesc. This removes many of the usages of deprecated APIs in the
REPL.


common/env/io.go
```diff
@@ -0,0 +1,270 @@
...
+	return nil
+}
+
+type typeDescParser struct {
+	text   string
+	pos    int
+	length int
+}
+
+func ParseTypeDesc(text string) (*TypeDesc, error) {
```
TristonianJones, line 117: Please add a comment to the exported method.

repl/parser/Commands.g4
```diff
@@ -59,16 +59,20 @@ qualId: leadingDot='.'? rid=IDENTIFIER ('.' qualifiers+=IDENTIFIER)*;
...
 typeParamList:
-    '(' ( types+=type (',' types+=type)* )? ')' ;
+    '<' ( types+=type (',' types+=type)* )? '>' ;
 
 // lexer rules:
 COMMAND: '%' IDENTIFIER;
 FLAG: '-' ('-' IDENTIFIER)+;
 ARROW: '->';
 EQUAL_ASSIGN: '=';
+PARAM_SPECIFIER: '~' (LETTER | '_' ) ( LETTER | DIGIT | '_')*;
```
TristonianJones: I think the code only supports a single uppercase letter here.

repl/parser/Commands.g4
```diff
@@ -59,16 +59,20 @@ qualId: leadingDot='.'? rid=IDENTIFIER ('.' qualifiers+=IDENTIFIER)*;
 startType : t=type EOF;
 
 type :
-    id=typeId params=typeParamList? ;
+    (typeId params=typeParamList?) | paramId ;
```
TristonianJones, line 62: This needs to move up because of function return types, right? I hadn't really considered that case, more just that I didn't want it to be something you could specify on the input. We'd just have to restrict the usage of type param types to only container types and return types, not variables.

common/env/testdata/extended_env.yaml
```diff
@@ -49,6 +54,22 @@ functions:
...
+        examples:
+          - "''.isEmptyAlt() // true"
+        target: "google.protobuf.StringValue"
+        return: "bool"
+      - id: "list_isEmpty"
+        examples:
+          - "[].isEmptyAlt() // true"
+          - "[1].isEmptyAlt() // false"
+        target: "list<~T>"
+        return: "bool"
```
TristonianJones, line 72: Can you add a function that returns a parameterized type, e.g. `list<~T>.first() -> ~T`

common/env/io.go
jnthntatum: Done.

repl/parser/Commands.g4
jnthntatum: Done. I think we use "elem" and "key" internally in a couple of places, but probably fine to be more restricted in the environment config.

repl/parser/Commands.g4
jnthntatum: Done. I added a check in the config.Validate call. Let me know if I should check somewhere else.

common/env/testdata/extended_env.yaml
jnthntatum: Done.

common/env/io.go
```diff
@@ -0,0 +1,271 @@
...
+			IsTypeParam: buf.IsTypeParam,
+		}
+	} else if buf.Type != nil {
+		v.TypeDesc = buf.Type
+	}
+	return nil
+}
+
+// MarshalYAML implements yaml.Marshaler
+func (v *Variable) MarshalYAML() (any, error) {
```
TristonianJones, line 66: I think usually you would try to keep all functionality in the same file, but since Variable is in the mega-env.go file, this is probably a reasonable separation. Do you need to support the same for Functions and Overloads or no?

common/env/testdata/extended_env.yaml
TristonianJones, line 1: This can be done in a follow-up, but let's try and move all examples to the new type format. We can leave one for 'legacy_types_env.yaml' that exercises the older codepaths

TristonianJones (approved):

common/env/io.go
jnthntatum: No -- I think it's just an odd behavior in how the go yaml library handles embedding by pointer so only need the extra logic for variable.

common/env/testdata/extended_env.yaml
jnthntatum: Will send a follow-up.

PR #1304, opened 2026-04-15
ext/strings: add runtime cost tracking for string extension functions

MindflareX:
## Summary

The string extension library (`ext/strings.go`) does not register any `OverloadCostTracker` entries in its `ProgramOptions()` method, causing all string extension function calls to fall through to the default case in `costCall()` (`interpreter/runtimecost.go:337`) which charges a cost of 1 regardless of input size or output size.

This is inconsistent with `ext/lists.go`, `ext/sets.go`, and `ext/regex.go` which all register proper cost trackers.

## Impact

The missing cost tracking allows resource exhaustion when `CostLimit()` is used as a defense mechanism:

- **`indexOf()`/`lastIndexOf()`**: Uses an O(N×M) naive nested-loop search implementation (`ext/strings.go:608-619`). A 500K string searched against a 200K almost-matching pattern burns **24 seconds of CPU** at `CostLimit(100)` with no error.

- **`join()`**: Concatenates N strings into a single output string. Joining 10,000 strings of 10,000 chars each allocates **587 MB** at `CostLimit(100)` with no error.

- **`replace()`**: Output can be significantly larger than input (e.g., replacing each character with a long string). `replace("a", "<1000 chars>")` on 100K input produces 95 MB with no cost error.

- **`lowerAscii()`/`upperAscii()`/`reverse()`/`split()`/`trim()`**: All perform O(N) work on the input string but are tracked as cost=1.

## Fix

This change adds `OverloadCostTracker` registrations for all string extension functions. The cost formulas follow the patterns already used by built-in string operations in `interpreter/runtimecost.go`:

| Function | Cost Formula |
|---|---|
| `indexOf`/`lastIndexOf` | `str_size * substr_size * StringTraversalCostFactor` |
| `join`/`replace` | `result_size * StringTraversalCostFactor` |
| `lowerAscii`/`upperAscii`/`reverse`/`trim` | `input_size * StringTraversalCostFactor` |
| `split` | `result_size * StringTraversalCostFactor + ListCreateBaseCost` |

## Test

Added `TestStringsCostTracking` which verifies:
- Large `indexOf`/`lastIndexOf`/`join` operations correctly exceed cost limits
- Small operations remain within cost limits

All existing tests pass.

### DIFF
--- a/ext/strings.go
+++ b/ext/strings.go
@@ -28,9 +28,11 @@ import (
 	"golang.org/x/text/language"
 
 	"github.com/google/cel-go/cel"
+	"github.com/google/cel-go/common"
 	"github.com/google/cel-go/common/types"
 	"github.com/google/cel-go/common/types/ref"
 	"github.com/google/cel-go/common/types/traits"
+	"github.com/google/cel-go/interpreter"
 )
 
 const (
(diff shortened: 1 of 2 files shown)


TristonianJones:
There's already an open PR for this which hasn't yet responded to feedback

PR #1305, opened 2026-04-15
Use shorthand type specifiers in env yaml files

jnthntatum:
Also add a test that JSON as yaml works.

Marshaling will still use the structured / yaml-map types, but we could default to the string form after the other implementations all support it.

### DIFF
--- a/common/env/env.go
+++ b/common/env/env.go
@@ -258,6 +258,8 @@ type Variable struct {
 
 	// Type represents the type declaration for the variable.
 	//
+	// When serialized, 'type' is used for shorthand specifier string.
+	//
 	// Deprecated: use the embedded *TypeDesc fields directly.
 	Type *TypeDesc `yaml:"type,omitempty"`
 
--- a/common/env/env.go
+++ b/common/env/env.go
@@ -260,7 +260,7 @@ type Variable struct {
 	//
 	// When serialized, 'type' is used for shorthand specifier string.
 	//
-	// Deprecated: use the embedded *TypeDesc fields directly.
+	// Use GetType() for getting the effective type.
 	Type *TypeDesc `yaml:"type,omitempty"`
 
 	// TypeDesc is an embedded set of fields allowing for the specification of the Variable type.
--- a/common/env/testdata/context_env.yaml
+++ b/common/env/testdata/context_env.yaml
@@ -43,23 +43,20 @@ functions:
           - "coalesce(null, 1) // 1"
           - "coalesce(2, 1) // 2"
         args:
-          - type_name: "google.protobuf.Int64Value"
-          - type_name: "int"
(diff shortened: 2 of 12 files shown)


common/env/testdata/context_env.yaml
```diff
@@ -43,23 +43,20 @@ functions:
           - "coalesce(null, 1) // 1"
           - "coalesce(2, 1) // 2"
         args:
-          - type_name: "google.protobuf.Int64Value"
-          - type_name: "int"
-        return:
-          type_name: "int"
+          - "google.protobuf.Int64Value"
```
TristonianJones, line 46: Elsewhere we use `int_wrapper`, should we make the proto type name and short-name of `{type}_wrapper` both work?

common/env/testdata/json_env.json
```diff
@@ -0,0 +1,132 @@
...
+  ],
+  "variables": [
+    {
+      "name": "msg",
+      "type_name": "google.expr.proto3.test.TestAllTypes",
+      "description": "msg represents all possible type permutation which CEL understands from a proto perspective"
+    },
+    {
+      "name": "opt_msg",
+      "type": "optional_type<google.expr.proto3.test.TestAllTypes>",
```
TristonianJones, line 22: Should we normalize to `type` in this file?

TristonianJones (dismissed):

common/env/testdata/context_env.yaml
jnthntatum: I added a test in /cel/env_test.go.

Right now the _wrapper names get mapped to the internal types.Type in the AsCELType call (just a manual check, before we look at the type provider), but the env.TypeDesc  just stores whatever was written in config. On export I think we normalize to the protobuf wrapper names. Does that all sound reasonable to you?

common/env/testdata/json_env.json
jnthntatum: This was just a 1-1 translation of the yaml version of the same. I renamed and trimmed this file down.

common/env/testdata/context_env.yaml
TristonianJones: So long as the alias and the fully qualified name are treated the same way internally, I'm happy

common/env/env.go
```diff
@@ -258,6 +258,8 @@ type Variable struct {
 
 	// Type represents the type declaration for the variable.
 	//
+	// When serialized, 'type' is used for shorthand specifier string.
+	//
 	// Deprecated: use the embedded *TypeDesc fields directly.
```
TristonianJones: Let's remove the deprecation notice

TristonianJones (dismissed):

TristonianJones (approved):

PR #1310, opened 2026-04-30
ext/lists: add max size check to genRange() to prevent OOM

### LINKED ISSUE #1309: ext/lists: lists.range() unbounded memory allocation allows OOM DoS
`lists.range(N)` in `ext/lists.go` allocates N elements with no upper bound check. A large value causes OOM and crashes the process.

This is exploitable in Kubernetes ValidatingAdmissionPolicy where user-controlled fields feed into CEL expressions. For example:

```yaml
validations:
  - expression: "lists.range(object.spec.replicas).all(i, i >= 0)"
```

An attacker sets `replicas: 2147483647` and the API server allocates ~16 GB, gets OOM-killed.

The root cause is `genRange()` at line 406:

```go
func genRange(n types.Int) (ref.Val, error) {
    var newList []ref.Val
    for i := types.Int(0); i < n; i++ {
        newList = append(newList, i)
    }
    return types.DefaultTypeAdapter.NativeToValue(newList), nil
}
```

The cost estimator doesn't help at Lists v1-v2 because cost estimators for `lists.range` are only registered at v3+ (line 354). `CostLimit` is also opt-in.

Confirmed: `lists.range(10000000)` allocates 865 MB in ~5 seconds.

I'll submit a PR with the fix — adding a `maxRangeSize` check and rejecting negative values.

marwan9696:
lists.range(N) allocates N elements with no upper bound. A large
value like 2147483647 allocates ~16 GB and crashes the process.

This is dangerous in Kubernetes ValidatingAdmissionPolicy where
user-controlled input feeds into CEL expressions.

Add maxRangeSize (10M) check and reject negative values.

Fixes #1309

### DIFF
--- a/ext/lists.go
+++ b/ext/lists.go
@@ -160,7 +160,7 @@ func Lists(options ...ListsOption) cel.EnvOption {
 	return cel.Lib(l)
 }
 
-const defaultMaxRangeSize = 10_000_000
+const defaultMaxRangeSize = 1_000_000
 
 type listsLib struct {
 	version      uint32
--- a/ext/lists.go
+++ b/ext/lists.go
@@ -403,8 +403,17 @@ func (lib *listsLib) ProgramOptions() []cel.ProgramOption {
 	return opts
 }
 
+// maxRangeSize is the maximum number of elements lists.range() will allocate.
+const maxRangeSize = 10_000_000
+
 func genRange(n types.Int) (ref.Val, error) {
-	var newList []ref.Val
+	if n < 0 {
+		return nil, fmt.Errorf("lists.range: size must be non-negative, got %d", n)
+	}
+	if n > maxRangeSize {
+		return nil, fmt.Errorf("lists.range: size %d exceeds maximum allowed (%d)", n, maxRangeSize)
+	}
+	newList := make([]ref.Val, 0, n)
 	for i := types.Int(0); i < n; i++ {
 		newList = append(newList, i)
 	}
--- a/ext/lists.go
+++ b/ext/lists.go
@@ -153,15 +153,18 @@ var comparableTypes = []*cel.Type{
 //	].sortBy(e, e.score).map(e, e.name)
 //	== ["bar", "foo", "baz"]
 func Lists(options ...ListsOption) cel.EnvOption {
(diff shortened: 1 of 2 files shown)


marwan9696:
CLA signed. Please re-check.

ext/lists.go
```diff
@@ -403,8 +403,17 @@ func (lib *listsLib) ProgramOptions() []cel.ProgramOption {
 	return opts
 }
 
+// maxRangeSize is the maximum number of elements lists.range() will allocate.
+const maxRangeSize = 10_000_000
```
TristonianJones: Could you make this a functional option for the `Lists` library instead?

marwan9696:
Done — moved it to a `ListsMaxRangeSize(n)` functional option. Default is still 10M, setting to zero disables the check. Let me know if you'd like any other changes.

ext/lists_test.go
```diff
@@ -521,3 +521,35 @@ func testListsEnv(t *testing.T, opts ...cel.EnvOption) *cel.Env {
...
+	val, err := genRange(10, defaultMaxRangeSize)
+	if err != nil {
+		t.Fatalf("genRange(10) failed: %v", err)
+	}
+	if val == nil {
+		t.Fatal("genRange(10) returned nil")
+	}
+
+	// Over the limit should fail, not allocate.
+	_, err = genRange(defaultMaxRangeSize+1, defaultMaxRangeSize)
```
TristonianJones, line 542: Would you mind testing both the default range and a smaller range?

ext/lists.go
```diff
@@ -153,15 +153,18 @@ var comparableTypes = []*cel.Type{
...
 func Lists(options ...ListsOption) cel.EnvOption {
-	l := &listsLib{version: math.MaxUint32}
+	l := &listsLib{version: math.MaxUint32, maxRangeSize: defaultMaxRangeSize}
 	for _, o := range options {
 		l = o(l)
 	}
 	return cel.Lib(l)
 }
 
+const defaultMaxRangeSize = 10_000_000
```
TristonianJones: We can make this 1M. Seems reasonably high that no one should ever hit it on accident.

TristonianJones (approved):

PR #1313, opened 2026-05-08
Fix type variable leakage in block variables list (issue #1312)

### LINKED ISSUE #1312: Compiler type-checking failure with empty list variables
## Description
We have encountered a subtle and relatively specific type-unification bug in the CEL compiler (specifically when compiling CEL policies). 

When a list of strings is unwrapped from a list of optional strings (`list(optional(string))`) using standard comprehensions (`filter`/`map`), and then concatenated with an untyped empty list `[]` defined as a variable, the CEL type checker incorrectly propagates the `list(optional(string))` constraint down the addition chain. 

This causes the compiler to incorrectly infer the empty list variable as `list(optional(string))` instead of `list(string)`, resulting in a compilation failure:

```
found no matching overload for '_+_' applied to '(list(string), list(optional_type(string)))'
```

This bug is highly sensitive to variable boundaries and ONLY triggers when all steps are declared as separate variables in the policy. Inlining any of these steps acts as a workaround and allows the compiler to succeed.

**To Reproduce**

- [ ]  parser
- [x]  checker
- [ ]  interpreter

---

## Minimal Reproducible Example (MRE)

```yaml
name: "mre"
rule:
  variables:
    # type: list(dyn) - Initially untyped empty list
    - name: "empty_list"
      expression: |
        []

    # type: list(optional(string)) - Source list of optional strings
    - name: "list_optional_string"
      expression: |
        [optional.of("foo")]

    # type: list(string) - Unwrapped string list using standard comprehensions
    - name: "string_list"
      expression: |
        variables.list_optional_string.filter(u, u.hasValue()).map(u, u.value())

    # Expected type: list(string)
    # ERROR TRIGGERS HERE:
    # "found no matching overload for '_+_' applied to '(list(string), list(optional_type(string)))'"
    - name: "concatenated_list"
      expression: |
        variables.string_list +
        variables.empty_list
  match:
    - output: |
        variables.concatenated_list
```

When compiling the above MRE, we get the following error message:

```
Compile(mre_policy.celpolicy) failed: 
ERROR: mre_policy.celpolicy:24:31: found no matching overload for '_+_' applied to (list(string), list(optional_type(string)))
 |         variables.string_list +
 | ..............................^
```

---

This bug seems strictly tied to how the compiler manages type constraints across variable boundaries. It only triggers when the compilation steps are split across separate variables:

1.  **Empty list must be a variable:** 
    The empty list `[]` **must** be defined as a variable (`empty_list = []`). 
    *   Inlining it directly into the addition chain (e.g., `variables.string_list + []`) compiles successfully.
2.  **Source optional list must be a variable:** 
    The source optional list `list_optional_string` **must** be defined as a variable.
    *   Inlining it directly into the comprehension expression (e.g., `[optional.of("foo")].filter(...)`) compiles successfully.
3.  **Unwrapped list must be a variable:** 
    The mapped string list `string_list` **must** be defined as a variable.
    *   Inlining it directly into the final addition (e.g., `variables.list_optional_string.filter(...).map(...) + variables.empty_list`) compiles successfully.

If you inline **any** of these three variables, the type checker correctly unifies the types. Only when they are declared as separate variables does the compiler incorrectly "leak" the type of the original optional list (`list_optional_string`) to the empty list variable in the addition step, thinking we are trying to add `list(string)` and `list(optional(string))`.

seirl:
Under the hood, when you compile a CEL policy with local variables, the compiler bundles them into a single internal function call called `cel.@block`.

The first argument of this call is a list containing all the variable initializers, and the second is the body of the expression: `cel.@block([var1_init, var2_init, ...], body)`

For example, if you define two variables:

    empty_list = []  // (inferred as list(dyn))
    string_list = ["foo"]  // (inferred as list(string))

They get bundled into: `cel.@block([[], ["foo"]], ...)`

When CEL type-checks this composed expression, it looks at the first argument `[[], ["foo"]]` and sees a standard list literal. CEL lists are designed to be homogeneous. To enforce this, the type checker eagerly unifies the types of all elements in a list literal.

However, when it unifies `[]` (which has an unbound type parameter `list(_var0)`) with `["foo"]` (`list(string)`), it decides they are compatible by binding the type parameter `_var0` to `string`.

Even if the list as a whole later defaults to `list(dyn)` because of other incompatible elements, it's already too late: the type checker has already recorded that the independent `empty_list` is of type `list(string)` and will keep this information for the rest of the type checking step.

If you later try to concatenate `empty_list + [1]`, the type checker returns an error:

    found no matching overload for '_+_' applied to '(list(string), list(int))'

Basically, the type of `string_list` bled into the completely independent `empty_list` simply because they were temporarily bundled in the same initializer list.

The problem is that the variables "list" passed to `cel.@block` is semantically not a normal homogeneous list. It is more like a heterogeneous tuple of independent expressions.

To fix this, we special-case the internal `cel.@block` function in the core type checker.

* When we encounter `cel.@block`, we type-check each variable initializer in the list independently.
* We do not run the unification logic on them, completely preventing any type bleeding.
* We set the type of the variables list directly to `list(dyn)` (matching the `cel.@block` signature) and then proceed to type-check the body.

This allows local variables to maintain their independent types, resolving the compilation failures while preserving the performance benefits of `cel.@block`.

Fix #1312

### DIFF
--- a/checker/checker.go
+++ b/checker/checker.go
@@ -266,6 +266,10 @@ func (c *checker) checkCall(e ast.Expr) {
 		c.checkOptSelect(e)
 		return
 	}
+	if fnName == "cel.@block" {
+		c.checkCelBlock(e)
+		return
+	}
 
 	args := call.Args()
 	// Traverse arguments.
@@ -771,3 +775,29 @@ var (
 		types.UintKind:      "google.protobuf.UInt64Value",
 	}
 )
+
+func (c *checker) checkCelBlock(e ast.Expr) {
+	call := e.AsCall()
+	if len(call.Args()) != 2 {
+		c.errors.unexpectedASTType(e.ID(), c.location(e), "cel.@block", "incorrect argument count")
+		c.setType(e, types.ErrorType)
+		return
+	}
+
+	varsList := call.Args()[0]
+	if varsList.Kind() != ast.ListKind {
+		c.check(varsList)
+	} else {
+		create := varsList.AsList()
+		for _, elem := range create.Elements() {
+			c.check(elem)
+		}
+		c.setType(varsList, types.NewListType(types.DynType))
+	}
+
+	body := call.Args()[1]
+	c.check(body)
+	c.setType(e, c.getType(body))
+
+	c.setReference(e, ast.NewFunctionReference("cel_block_list"))
+}
--- a/policy/helper_test.go
+++ b/policy/helper_test.go
@@ -210,6 +210,15 @@ var (
 			: optional.none())))
 	  : optional.of(@index3.format([@index0, @index2])))`,
 		},
+		{
+			name: "empty_list_unification_issue1312",
+			expr: `
+	cel.@block([
+	  [],
+	  ["foo"]],
+	  @index0 + [1])
+			`,
+		},
 	}
 
 	composerUnnestTests = []struct {
--- a/policy/testdata/empty_list_unification_issue1312/config.yaml
+++ b/policy/testdata/empty_list_unification_issue1312/config.yaml
@@ -1,0 +1,1 @@
+name: "empty_list_unification_issue1312"
--- a/policy/testdata/empty_list_unification_issue1312/policy.yaml
+++ b/policy/testdata/empty_list_unification_issue1312/policy.yaml
@@ -1,0 +1,9 @@
+name: "empty_list_unification_issue1312"
+rule:
+  variables:
+    - name: "empty_list"
+      expression: "[]"
+    - name: "string_list"
+      expression: "['foo']"
+  match:
+    - output: "variables.empty_list + [1]"
--- a/policy/testdata/empty_list_unification_issue1312/tests.yaml
+++ b/policy/testdata/empty_list_unification_issue1312/tests.yaml
@@ -1,0 +1,8 @@
+description: Issue 1312 simple test
+section:
+  - name: "basic"
+    tests:
+      - name: "test"
+        input: {}
+        output:
+          expr: "[1]"


TristonianJones:
@seirl The individual expressions placed into the cel.@block are type-checked ahead of time, and then assembled into the block format as an optimization. Can you help me understand what the issue is in practice?

Is this mostly that the individual slots need to have their type information preserved during the optimization for the final type-check to provide a sensible output?

seirl:
@TristonianJones have you checked the minimal reproducible example in #1312 ? The problem is that compiling:

```
name: "mre"
rule:
  variables:
    - name: "empty_list"
      expression: "[]"
    - name: "string_list"
      expression: "['foo']"
  match:
    - output: "variables.empty_list + [1]"
```

fails with:

```
found no matching overload for '_+_' applied to '(list(string), list(int))'
```

TristonianJones:
That's actually a known bug in CEL type checking where the type parameter assignment gets confused by empty list sometimes. The expression `[] + [1]` (or some variation) should be all you need to repro.

As for CEL block, the individual entries are type checked separately, and the type assignment propagated to the @index variables. If you fix the main bug, I believe you'll get more predictable results.

seirl:
I can't reproduce the bug with the expression `[] + [1]` or any other similar variation that I tried. Actually there's a bunch of tests in `checker_test.go` that check exactly that, and it seems to work fine AFAICT. The only thing where I saw this behavior is for `cel.@block`.

Do you have examples for the "main bug" that should be fixed alongside this one, or is there an existing bug for it?

TristonianJones:
Hi @seirl, I believe the issue that's causing troubles here is the same or similar to the one which fails for type promotion within a list `wrapper_int` and `int` types:  

https://github.com/google/cel-spec/blob/cb51b4176013ad19bd00df94be273c322916a620/tests/simple/testdata/type_deduction.textproto#L520

In the past, there have been a few other cases of type resolution challenges due to the erasure of type parameters to `dyn`, which may also be causing some trouble here, but I'm not sure.

There should be a carve-out for `cel.@block` already to be heterogeneously typed and the appropriate @index variables have a strong type; however this would still reduce to a type-promotion failure in the type-checker that looks similar to the test in the cel-spec that cel-go is currently failing (it's expressly disabled in the `conformance` tests)

-Tristan

seirl:
OK, I think I understand where you see the connection. The type unification is greedy for lists, so if you have a type unification for `[wrapper(int), int]`, it thinks the list is heterogeneous, even though theoretically you could unify this as `[int, int]`.

However, I think in my case, fixing this behavior still wouldn't be enough. Imagine you had a non-greedy type unification, and you waited for the end of your list literal to assign a type to your list elements. If you had `[[], [3]]`  I think you would *still* want to unify this as `list(list(int))`, because an empty list shouldn't make an otherwise homogeneous list a dyn list.

And I think this behavior is correct for regular lists, but not for `cel.@block`, because you don't care about the list being homogeneous, conceptually you just want to type-check it as a tuple. So I think we still need to special-case it in some way. Does that make sense?

jnthntatum:
Poking a bit, it looks like the mis identified type happens when the composer does the second pass to unnest rules (in the basic compose indexes are manually mapped from the inferred type of the independent expressions, in the unnester it reads the inferred type from checking the init list type inferences). https://github.com/google/cel-go/blob/a82c68b770ac0cb67f7b4f76166827c14b145eb8/policy/composer.go#L275. I think cel-go handles this in conformance tests by just marking all of the index variables as dyn for the purposes of checking a hand-rolled AST with the block construct.

To get the conformance tests for cel.block working in C++ I added a similar change to this PR -- https://github.com/google/cel-cpp/pull/1968. It still has some odd behaviors around free type parameters that we might need to address. Block makes it extra tricky since the variable is intended to be referenced in multiple places.

 - variables.empty_list + ['foo'] -> list(dyn) 
 - variables.empty_list + [1] -> list(dyn)
 - variables.empty_optional.orValue('foo') -> dyn

seirl:
@jnthntatum Cool! So just so I understand fully, you had to implement something similar to what I do in this pull request to make the C++ checker pass the conformance tests? Does that mean that the cel-go code *doesn't* pass the conformance tests?

Is the C++ implementation able to compile the example that I gave correctly, with the empty list variable?

TristonianJones:
Both Java and Go have some issues in type resolution with empty container types. Often the result is dyn and doesn't have a meaningful impact, but it would be good to fix. I just don't know if the solution proposed is more of a bandaid over the core issue or a fix for an issue that makes the underlying cause worse

l46kok:
A potential alternative: we could simply use the declared type on the CompiledRule itself as the source of truth rather than trying to use the inferred type from the checker:

```go
-                       celType := a.GetType(v.ID()) 
+                       celType := opt.rule.Variables()[i].Declaration().Type() // Illustrative example. May have to collect all variables in this scope. 
```

This is what we do in CEL-Java's composer implementation to bypass the unification pollution problem. I agree though that supporting this in checker is probably more ideal:

https://github.com/google/cel-java/blob/14d4c2e39151f2e99e36f9818a9118b01c1d9ed3/policy/src/main/java/dev/cel/policy/RuleComposer.java#L66

PR #1314, opened 2026-05-12
fix: respect variable scoping when folding comprehensions

### LINKED ISSUE #1296: Constant folding fails for comprehensions containing a shadowed variable
**Describe the bug**
Attempting to constant fold the following expression: `[(x - 1 > 3) ? (x - 1) : 5].exists(x, x - 1 > 3)`

Produces the following error:

`Optimize() generated an invalid AST: ERROR: <input>:-1:0: constant-folding evaluation failed: no such attribute(s): x`

**Expected behavior**
Though there's nothing to fold, it shouldn't cause an error

SAY-5:
The constant folding optimizer treated a comprehension's iteration variable as in-scope while inspecting the iteration range and accumulator initializer, so a free variable in the iteration range that happened to share the iteration variable's name was misjudged foldable and the fold failed at evaluation time. The matcher now walks comprehensions with proper scoping. Closes #1296.

### DIFF
--- a/cel/folding.go
+++ b/cel/folding.go
@@ -506,26 +506,84 @@ func (opt *constantFoldingOptimizer) constantExprMatcher(ctx *OptimizerContext, 
 		if isNestedComprehension(e) {
 			return false
 		}
-		vars := map[string]bool{}
-		constantExprs := true
-		visitor := ast.NewExprVisitor(func(e ast.Expr) {
-			if e.Kind() == ast.ComprehensionKind {
-				nested := e.AsComprehension()
-				vars[nested.AccuVar()] = true
-				vars[nested.IterVar()] = true
-			}
-			if e.Kind() == ast.IdentKind && !vars[e.AsIdent()] {
-				constantExprs = false
-			}
-			// Late-bound function calls cannot be folded.
-			if e.Kind() == ast.CallKind && isLateBoundFunctionCall(ctx, a, e) {
-				constantExprs = false
-			}
-		})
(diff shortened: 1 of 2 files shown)


cel/folding.go
```diff
@@ -506,26 +506,84 @@ func (opt *constantFoldingOptimizer) constantExprMatcher(ctx *OptimizerContext,
...
+	}
+	return opt.constantScopedExpr(ctx, a, compre.LoopCondition(), inner) &&
+		opt.constantScopedExpr(ctx, a, compre.LoopStep(), inner) &&
+		opt.constantScopedExpr(ctx, a, compre.Result(), inner)
+}
+
+// constantScopedExpr reports whether an expression references only the given in-scope variables
+// and contains no late-bound function calls, recursing into nested comprehensions with their
+// own iteration and accumulator variables.
+func (opt *constantFoldingOptimizer) constantScopedExpr(ctx *OptimizerContext, a *ast.AST, e ast.Expr, vars map[string]bool) bool {
```
TristonianJones, line 541: I think there's probably a simpler way to achieve this check which the cel-java implementation uses:

https://github.com/google/cel-java/blob/14d4c2e39151f2e99e36f9818a9118b01c1d9ed3/optimizer/src/main/java/dev/cel/optimizer/optimizers/ConstantFoldingOptimizer.java#L224

Thanks for looking into this! Consider using the `NavigableExpr` methods to mirror how Java checks for this case.

PR #1316, opened 2026-05-14
Managed execution frame with async call foundations

TristonianJones:
There are a lot of concepts present in the `ExecutionFrame`, but it overall introduces the following concepts:

* `ExecutionFrame` is a custom `context.Context` value whose lifecycle is owned entirely by CEL
* `functions.AsyncOp` and `functions.BlockingAsyncOp` are the signature expected for async call management. 

The next PRs will introduce the following changes:

## Interpretation updates

* Integration of `ExecutionFrame` via an `InterpretableV2` interface
* Cleanup of the state tracing and cost tracking logic by hoisting it into `ExecutionFrame`
* Unified `ExecutionFrame` and `Activation` pooling

## Async calls

* Iterative eval with different drain strategies
* Async call caching support

## Checkpoint evaluation

* Determine incrementally re-evaluate from a checkpoint to assist with minimizing expensive computations in response to new async results.

### DIFF
--- a/common/functions/functions.go
+++ b/common/functions/functions.go
@@ -68,6 +68,11 @@ type FunctionOp func(...ref.Val) ref.Val
 
 // AsyncOp is a function that accepts zero or more arguments and produces
 // a value or error asynchronously via a channel.
+//
+// AsyncOp is an internal interface intended for use by CEL to manage goroutines and
+// channels associated with async calls. For public API usage, use BlockingAsyncOp.
+// Implementers should listen for context cancellation on the provided context for
+// resource cleanup.
 type AsyncOp func(context.Context, ...ref.Val) <-chan ref.Val
 
 // BlockingAsyncOp is a function that accepts zero or more arguments and blocks until
--- a/common/functions/functions.go
+++ b/common/functions/functions.go
@@ -15,7 +15,11 @@
 // Package functions defines the standard builtin functions supported by the interpreter
 package functions
 
-import "github.com/google/cel-go/common/types/ref"
+import (
+	"context"
+
+	"github.com/google/cel-go/common/types/ref"
+)
 
 // Overload defines a named overload of a function, indicating an operand trait
 // which must be present on the first argument to the overload as well as one
@@ -41,9 +45,11 @@ type Overload struct {
 	// Binary defines the overload with a BinaryOp implementation. May be nil.
 	Binary BinaryOp
 
-	// Function defines the overload with a FunctionOp implementation. May be
-	// nil.
+	// Function defines the overload with a FunctionOp implementation. May be nil.
 	Function FunctionOp
+
+	// Async defines the overload with an AsyncOp implementation. May be nil.
+	Async AsyncOp
 
 	// NonStrict specifies whether the Overload will tolerate arguments that
 	// are types.Err or types.Unknown.
@@ -51,11 +57,20 @@ type Overload struct {
 }
 
 // UnaryOp is a function that takes a single value and produces an output.
-type UnaryOp func(value ref.Val) ref.Val
+type UnaryOp func(ref.Val) ref.Val
 
 // BinaryOp is a function that takes two values and produces an output.
-type BinaryOp func(lhs ref.Val, rhs ref.Val) ref.Val
+type BinaryOp func(ref.Val, ref.Val) ref.Val
 
 // FunctionOp is a function with accepts zero or more arguments and produces
 // a value or error as a result.
-type FunctionOp func(values ...ref.Val) ref.Val
+type FunctionOp func(...ref.Val) ref.Val
+
+// AsyncOp is a function that accepts zero or more arguments and produces
+// a value or error asynchronously via a channel.
+type AsyncOp func(context.Context, ...ref.Val) <-chan ref.Val
+
+// BlockingAsyncOp is a function that accepts zero or more arguments and blocks until
+// the result is available. When used with AsyncBinding, the framework runs the function
+// in its own goroutine and manages channel lifecycle internally.
+type BlockingAsyncOp func(context.Context, ...ref.Val) ref.Val
--- a/interpreter/frame.go
+++ b/interpreter/frame.go
@@ -18,19 +18,7 @@ import (
 	"context"
 	"sync"
 	"sync/atomic"
-
-	"github.com/google/cel-go/common/functions"
-	"github.com/google/cel-go/common/types"
-	"github.com/google/cel-go/common/types/ref"
 )
-
-// AsyncObserver provides callbacks for monitoring the lifecycle of asynchronous function calls.
-type AsyncObserver interface {
-	// OnCallStarted is called when an asynchronous function is first launched.
-	OnCallStarted(callID int64, function, overload string, args []ref.Val)
-	// OnCallFinished is called when an asynchronous function completes.
-	OnCallFinished(callID int64, function, overload string, res ref.Val)
-}
 
 // evalContext contains the stateful information needed for a single evaluation.
 //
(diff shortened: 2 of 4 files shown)


interpreter/frame.go
```diff
@@ -0,0 +1,509 @@
...
+
+func (acs *asyncCallState) equals(other *asyncCallState) bool {
+	if acs == nil || other == nil {
+		return false
+	}
+	if acs.function != other.function || acs.overload != other.overload {
+		return false
+	}
+	for i, v := range acs.argVals {
+		if types.Equal(v, other.argVals[i]) != types.True {
```
l46kok: Can we add a length check at the start?

```go
if len(acs.argVals) != len(other.argVals) {
	return false
}
```

interpreter/frame.go
```diff
@@ -0,0 +1,509 @@
...
+
+func (t *asyncCallStateTracker) getOrCreate(id int64, function, overload string, argVals []ref.Val, impl functions.AsyncOp, completions chan<- int64) *asyncCallState {
+	t.mu.RLock()
+	acs, ok := t.calls[id]
+	t.mu.RUnlock()
+
+	potentialState := newAsyncCallState(id, function, overload, argVals, impl)
+	if ok && acs.equals(potentialState) {
+		return acs
+	}
```
l46kok: I wonder if we need to store the list of actual states within the calls map here, like a multimap.

For repeated async calls with different arguments, this would overwrite the `calls` map here.
 ex: `[1,2].map(x, asyncFunc(x))`. 

I think it would help me to look at the wiring to better understand the problem, but on top of unnecessarily spawning a new goroutine, there seems to be a potential livelock scenario here:

```
x = 1, t.calls state is recorded, Goroutine 1 is launched, Unknown(1) is returned
x = 2, t.calls state is overridden (acs.equals check failure), Goroutine 2 is launched, Unknown(2) is returned
Both goroutine finishes and signals completion
Advance evaluation
x = 1, t.calls state is overridden again (fetched acs state with x = 2, acs.equals check failure), Goroutine 3 is launched, Unknown(3) is returned
x = 2, t.calls state is overridden again, Goroutine 4 is launched and so forth
```

interpreter/frame.go
```diff
@@ -0,0 +1,509 @@
...
+					return
+				}
+			}
+			ch := acs.impl(ctx, acs.argVals...)
+			select {
+			case res := <-ch:
+				acs.mu.Lock()
+				acs.result = res
+				acs.mu.Unlock()
+				if observer != nil {
```
l46kok: It's worth checking if the context was cancelled before executing callbacks here as well.

Testing this might be a bit annoying (I'm not familiar with how go normally does concurrency tests, but perhaps something like):

```go
for i 0 to 100 (or a comfortable enough number):

  ComputeResult("1, veryFastAsyncFunc, ...)

  // Immediately cancel evaluation (timeout or short circuit)
  cancel()

  time.Sleep(time.Millisecond)

  if len(completionCh) > 0 {
     // Stale Call ID, this is a race condition
  }

```

common/functions/functions.go
```diff
@@ -41,21 +45,32 @@ type Overload struct {
...
+type BinaryOp func(ref.Val, ref.Val) ref.Val
 
 // FunctionOp is a function with accepts zero or more arguments and produces
 // a value or error as a result.
-type FunctionOp func(values ...ref.Val) ref.Val
+type FunctionOp func(...ref.Val) ref.Val
+
+// AsyncOp is a function that accepts zero or more arguments and produces
+// a value or error asynchronously via a channel.
+type AsyncOp func(context.Context, ...ref.Val) <-chan ref.Val
```
l46kok, line 76: wdyt about explicitly documenting some requirements as an API contract? Something like:  implementations must return a buffered channel of capacity at least 1, or ensure background goroutines select on ctx.Done() when sending to prevent leaks when evaluation gets cancelled.

l46kok (commented):
First pass, I need to re-look at this a couple of more times, but I've mostly focused on the async side of things first

interpreter/frame.go
TristonianJones: Right, this is more of a multimap case. I'm not sure why that I didn't consider this earlier, kind of a silly mistake. In a different CL, I actually computed unique call trace ids, but I haven't incorporated them here yet. Let me udpate this to a multi-map

interpreter/frame.go
TristonianJones: I believe context cancellation should result in a ctx.Done() and a return from the call. Some of the context cancellation tests _should_ exercise this, but I'll have to double check.

interpreter/frame.go
TristonianJones: Since they're the same function and overload id, they should always have the same arg counts, but I can add a check.

TristonianJones:
@l46kok I've removed all notions of the async call for now and will focus first on the plumbing to introduce `ExecutionFrame` through the rest of the code base. PTAL

interpreter/frame.go
TristonianJones: Resolving for now, this change has been moved to another PR

interpreter/frame.go
TristonianJones: Moved to a later PR, will revisit down the road.

interpreter/frame.go
TristonianJones: Will revisit this a bit later. Removing for now.

interpreter/frame.go
TristonianJones: Will revisit this later. Removed for now.

interpreter/frame.go
```diff
@@ -0,0 +1,221 @@
...
+		}
+		f.ctx.ctx = nil
+		f.ctx.interrupt = nil
+		f.ctx.state = nil
+		f.ctx.costs = nil
+		f.ctx.interrupted.Store(false)
+		f.ctx.interruptCheckCount.Store(0)
+		f.ctx.interruptCheckFrequency = 0
+		evalContextPool.Put(f.ctx)
+		f.ctx = nil
```
l46kok, line 102: Is the context shared between parent and child frames? Should we check for the root frame before closing here?

```go
if f.parent != nil {
    // Defensive no-op or record the error somewhere
}
```

interpreter/frame.go
```diff
@@ -0,0 +1,221 @@
...
+
+// NewExecutionFrame creates a new execution frame from the pool.
+func NewExecutionFrame(vars Activation) *ExecutionFrame {
+	f := frameStack.Get().(*ExecutionFrame)
+	f.Activation = vars
+	return f
+}
+
+// SetContext sets the context for the execution frame.
+func (f *ExecutionFrame) SetContext(ctx context.Context, interruptCheckFrequency uint) {
```
l46kok, line 76: Similar question here, perhaps we should also check whether we're on the root frame:

```go
if f.parent != nil {
   // Accessed from child frame
}

if f.ctx != nil {
   // Unintended double initialization?
}
```

interpreter/frame.go
```diff
@@ -0,0 +1,221 @@
...
+
+func (pool *activationStackPool) create(parent, child Activation) Activation {
+	h := pool.Get().(*hierarchicalActivation)
+	h.child = child
+	h.parent = parent
+	return h
+}
+
+func (pool *activationStackPool) release(activation Activation) {
+	h, ok := activation.(*hierarchicalActivation)
```
l46kok, line 200: Thoughts on defining a private activation here to only capture activations the frame here created (e.g: frameActivation)? In case if users construct their own hierarchicalActivation that live outside the boundaries of this frame.

l46kok (commented):
LGTM overall, mostly nits

interpreter/frame.go
TristonianJones: Sounds good, I'll return an error from this function.

interpreter/frame.go
TristonianJones: I actually marked the activation as `poolAllocated` and the top-level entry point to `NewHierarchicalActivation` ensures this value is `false`. So, this case should be covered.

interpreter/frame.go
TristonianJones: Yep, that would have been bad. I've fixed it so these call paths only happen on the root.

l46kok (approved):

TristonianJones:
@l46kok my apologies, I addressed the nits in another branch. Will send the review shortly.

PR #1321, opened 2026-05-16
Policy variable scoping fixes

jnthntatum:
- fix variable resolution when a variable in a nested rule shadows an outer one
 - propagate the variable type from composition to unnesting instead of
   reading the post optimize inferred type (should address #1312)

### DIFF
--- a/policy/composer.go
+++ b/policy/composer.go
@@ -146,12 +146,12 @@ func (opt *ruleComposerImpl) lookupLocal(name string) (int, bool) {
 	return -1, false
 }
 
-func (opt * ruleComposerImpl) enterScope() {
+func (opt *ruleComposerImpl) enterScope() {
 	opt.scopes = append(opt.scopes, localScope{})
 }
 
-func (r *ruleComposerImpl) exitScope() {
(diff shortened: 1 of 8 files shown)


TristonianJones (dismissed):

policy/composer.go
```diff
@@ -256,25 +282,28 @@ func (opt *ruleUnnesterImpl) Optimize(ctx *cel.OptimizerContext, a *ast.AST) *as
...
+		// definitions in the optimizer. Extract the value expressions and types to set up the
+		// checker environment.
 		block := ruleExpr.AsCall()
 		ruleExpr = block.Args()[1].(ast.NavigableExpr)
 
 		// Collect the list of variables associated with the block
 		blockList := block.Args()[0].(ast.NavigableExpr)
 		vars := blockList.AsList()
+		if vars.Size() != len(opt.varIndices) {
+			ctx.ReportErrorAtID(ruleExpr.ID(), "ast var list size and recorded one do not match")
```
l46kok: nit: it's probably safer to early return for this case

l46kok (dismissed):

l46kok (approved):

PR #1323, opened 2026-05-19
Update documentation for nested first-match rules

jnthntatum:
First match rules have an implicit unwrap behavior if they are unconditional. This wasn't documented and behaved inconsistently depending on where the rule was in the policy tree.

### DIFF
--- a/policy/README.md
+++ b/policy/README.md
@@ -194,6 +194,10 @@ of the parent rule (otherwise, it would leave subsequent matches unreachable).
 Example:
 
 ```
+# Policy is a string (not wrapped)
+# input -> output
+# 3 -> 'b'
+# 2 -> 'c'
 rule:
   match:
     - rule:
@@ -201,9 +205,9 @@ rule:
         - name: "foo"
           value: 1 + 2
       match:
-        - condition: "input.foo > variables.foo"
+        - condition: "input > variables.foo"
           output: '"a"'
-        - condition: "input.foo == variables.foo"
+        - condition: "input == variables.foo"
           output: '"b"'
     - output: '"c"'
 
--- a/policy/README.md
+++ b/policy/README.md
@@ -201,14 +201,14 @@ Example:
 rule:
   match:
     - rule:
-      variables:
-        - name: "foo"
-          value: 1 + 2
-      match:
-        - condition: "input > variables.foo"
-          output: '"a"'
-        - condition: "input == variables.foo"
-          output: '"b"'
+        variables:
+          - name: "foo"
+            expression: "3"
+        match:
+          - condition: "input > variables.foo"
+            output: '"a"'
+          - condition: "input == variables.foo"
+            output: '"b"'
     - output: '"c"'
 
 ```
--- a/policy/README.md
+++ b/policy/README.md
@@ -184,6 +184,32 @@ rule:
     - output: "'outer_default'"
 ```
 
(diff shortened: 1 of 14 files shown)


policy/README.md
```diff
@@ -184,6 +184,36 @@ rule:
     - output: "'outer_default'"
 ```
 
+Nested first-match rules may be unconditional (i.e. the condition is omited or
```
l46kok:
```suggestion
Nested first-match rules may be unconditional (i.e. the condition is omitted or
```

policy/helper_test.go
```diff
@@ -210,6 +210,27 @@ var (
...
+	cel.@block([3],
+	((x > @index0) ? optional.of("a") : ((x == @index0) ? optional.of("b") : optional.none()))
+	  .orValue("c"))`,
+		},
+		{
+			name: "nested_rules_unconditional_chaining_optional",
+			expr: `
+	cel.@block([3],
+	((x > @index0) ? optional.of("a") : ((x == @index0) ? optional.of("b") : optional.none()))
+	  .or((x == 1) ? optional.of("c") : optional.none()))`,
```
l46kok, line 225: Not related to this pr, but it's weird that it pulls out constant literals into block indices. I would imagine this doesn't buy us anything performance wise. Actually, might be worse because now we have to do a scoped identifier lookup. optional.none() otoh, is likely worth extracting. Maybe worth filing a bug?

policy/README.md
l46kok: Up to you, but I personally found this to be a bit too condensed (maybe it's my brain not working). wdyt about breaking this down a bit?


#### Using Unconditional Nested Rules

In a first-match policy, you can create a nested rule without an explicit condition (or by simply setting the condition to true). 

This behavior is useful for two reasons:

1. Scoping Variables
You can use an unconditional rule to declare variables (like foo in the example below) that only apply to a specific group of matches, rather than cluttering the global scope.

2. Fallback Behavior (Chaining)
The inner matches inside your unconditional rule do not need to cover every possible scenario. If an input runs through the inner conditions and doesn't find a match, the engine doesn't fail. Instead, it simply steps back out to the parent rule and continues down the list.

Important: if your unconditional nested rule does cover every possible scenario (meaning it is exhaustive and will always generate an output), it must be the very last item in your parent rule's list. Placing it anywhere else will create dead code, as the engine will never reach the rules written beneath it.

policy/README.md
```diff
@@ -184,6 +184,36 @@ rule:
...
+is the literal `true`). Unconditional rules can be used as a scoping or
+grouping mechanism. The matches of the parent rule "chain" after an
+unconditional rule. This allows for unwrapping optionals from an inexhaustive
+sub rule. An uncondtional nested rule that is exhaustive must be the last match
```
l46kok:
```suggestion
sub rule. An unconditional nested rule that is exhaustive must be the last match
```

l46kok (dismissed):

policy/helper_test.go
jnthntatum: xref: http://b/515150714. Go composer doesn't inspect the variable definition at all, just always maps it to block index.

policy/README.md
jnthntatum: This reads better, thanks. I included this (reworded to match tone of the rest of the doc).

l46kok (approved):

PR #1324, opened 2026-05-20
Fix has() unknown propagation during partial evaluation

0xASTRA:
# Fix has() unknown propagation during partial evaluation

## Summary

This patch fixes stale unknown-type handling in `testOnlyQualifier.Qualify`.

`types.Unknown` is represented as `*types.Unknown`, but this code path still checks for the old value type:

```go
out.(types.Unknown)
```

As a result, when `QualifyIfPresent` returns a `*types.Unknown`, the assertion fails and `has()` can incorrectly return a definite boolean instead of propagating `Unknown`.

## Change

Update the type assertion from:

```go
if unk, isUnk := out.(types.Unknown); isUnk {
```

to:

```go
if unk, isUnk := out.(*types.Unknown); isUnk {
```

This allows `Unknown` values to be propagated correctly during partial evaluation.

## Impact

This restores the expected behavior for expressions such as:

```cel
has(request.auth.claims.admin)
```

when an intermediate qualifier resolves to `*types.Unknown`.

Instead of converting the result into a definite boolean, the evaluator now preserves the unknown value so the caller can defer evaluation as intended.

## Notes

This is a minimal compatibility fix for a stale type assertion left behind after the migration of `types.Unknown` to pointer-based usage.

### DIFF
--- a/interpreter/interpretable.go
+++ b/interpreter/interpretable.go
@@ -194,7 +194,7 @@ func (q *testOnlyQualifier) Qualify(vars Activation, obj any) (any, error) {
 	if err != nil {
 		return nil, err
 	}
-	if unk, isUnk := out.(types.Unknown); isUnk {
+	if unk, isUnk := out.(*types.Unknown); isUnk {
 		return unk, nil
 	}
 	return present, nil
--- a/interpreter/interpreter_test.go
+++ b/interpreter/interpreter_test.go
@@ -1542,6 +1542,16 @@ func testData(t testing.TB) []testCase {
 			out: types.NewUnknown(2, types.QualifyAttribute[int64](types.NewAttributeTrail("a"), 0)),
 		},
 		{
+			name: "macro_has_map_key_unknown_propagates",
+			expr: `has(a.b)`,
+			vars: []*decls.VariableDecl{
+				decls.NewVariable("a", types.NewMapType(types.StringType, types.BoolType)),
+			},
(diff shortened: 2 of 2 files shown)


interpreter/interpretable.go
```diff
@@ -194,7 +194,7 @@ func (q *testOnlyQualifier) Qualify(vars Activation, obj any) (any, error) {
 	if err != nil {
 		return nil, err
 	}
-	if unk, isUnk := out.(types.Unknown); isUnk {
+	if unk, isUnk := out.(*types.Unknown); isUnk {
```
TristonianJones, line 197: Would you mind adding a test case for this so we don't have a regression in the future? Thanks!

0xASTRA: Thanks for the review! Added macro_has_map_key_unknown_propagates in interpreter/interpreter_test.go (commit e97846f). It marks a as unknown via partial activation and asserts has(a.b) returns *types.unknown rather than a definite false.

TristonianJones (approved):

TristonianJones (approved):

0xASTRA:
Hi TristonianJones, thanks for the approval. I noticed the PR is still waiting on the “GitHub Actions ScanExpected” check. Is there anything needed from my side, or is this just waiting on internal automation?

jnthntatum:
It looks to be stuck -- it's advisory so I can merge.

PR #1327, opened 2026-05-22
fix(containers): allow aliases for simple names

### LINKED ISSUE #1306: Allow aliases to support names without field selections
**Describe the bug**
This constraint was relaxed in Java for usability reasons. We should match the behavior here.

The following should be allowed:

```
alias: "short", target: "really_long_package_name"
```

The following should be disallowed:

```
// Nothing to abbreviate
abbreviations: really_long_package_name

**To Reproduce**

- [ ]  parser
- [ x ]  checker
- [ x ]  interpreter

**Additional context**
b/450617503

rohan-patnaik:
Fixes #1306

## Summary
This allows `containers.Alias` to point at a simple name, not only a dotted qualified name.

## Why
Before this change, `Alias("really_long_package_name", "short")` failed validation even though explicit aliases should be able to abbreviate a name without a field selection. `Abbrevs` still requires a dotted qualified name, so the stricter abbreviation behavior is preserved.

## Test plan
- `go test ./common/containers`
- `go test ./common/containers ./cel ./checker ./interpreter`

### DIFF
--- a/common/containers/container.go
+++ b/common/containers/container.go
@@ -227,7 +227,7 @@ func Abbrevs(qualifiedNames ...string) ContainerOption {
 			}
 			alias := qn[ind+1:]
 			var err error
-			c, err = aliasAs("abbreviation", qn, alias)(c)
+			c, err = aliasAs("abbreviation", qn, alias, true)(c)
 			if err != nil {
 				return nil, err
 			}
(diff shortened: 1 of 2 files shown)


TristonianJones (approved):
@jnthntatum and @l46kok we should probably make a similar change for C++ and Java alias support since it's a different use case from abbreviations.

l46kok:
This change is already present in Java

Issue #1317, opened 2026-05-14
[Security / Logic] Precision loss in compareIntDouble() causes incorrect equality evaluation for integers above 2^53 - potential authorization bypass


herdiyana256:
I was investigating the behavior of cross-type numeric comparisons in `cel-go`  and found that `compareDoubleInt()` in `common/types/compare.go` performs an  implicit `Int → float64` cast before comparison. This causes silent precision  loss for integer values above `2^53`, making two mathematically distinct numbers  evaluate as equal — a spec violation with real authorization bypass implications.

---

## Root Cause
`common/types/compare.go`

```go
func compareDoubleInt(d Double, i Int) Int {
    if d < math.MinInt64 {
        return IntNegOne
    }
    if d > math.MaxInt64 {
        return IntOne
    }
    return compareDouble(d, Double(i))  // ← implicit Int → float64 cast HERE
}

func compareIntDouble(i Int, d Double) Int {
    return -compareDoubleInt(d, i)
}
```

The line `compareDouble(d, Double(i))` silently converts `i` (int64) to  `float64` before comparison. IEEE 754 double-precision floating-point can only  represent integers exactly up to `2^53`. Any integer beyond that boundary loses  precision when cast to float64.

For example:
- `int64(9007199254740993)` → cast to float64 → becomes `9007199254740992.0`
- Result: `9007199254740993 == 9007199254740992.0` evaluates to `true`

---

## Proof of Concept
Tested on: `github.com/google/cel-go v0.22.0`, Go 1.23.0, linux/amd64

```go
package main

import (
    "fmt"
    "github.com/google/cel-go/cel"
)

func main() {
    env, _ := cel.NewEnv(
        cel.Variable("val_int", cel.DynType),
        cel.Variable("val_double", cel.DoubleType),
    )

    ast, _ := env.Compile("val_int == val_double")
    prg, _ := env.Program(ast)

    valInt    := int64(9007199254740993)       // 2^53 + 1
    valDouble := float64(9007199254740992.0)   // 2^53

    out, _, _ := prg.Eval(map[string]any{
        "val_int":    valInt,
        "val_double": valDouble,
    })

    fmt.Printf("9007199254740993 == 9007199254740992.0 → %v\n", out.Value())
    // Output: true   ← INCORRECT, should be false
}
```

**Expected:** `false` - these are two mathematically distinct numbers  
**Actual:** `true` — precision lost during implicit cast

---

## CEL Spec Violation
[...]

PR #1328, opened 2026-05-26
Fix precision loss in compareDoubleInt/compareDoubleUint for values outside float64 safe integer range

XananasX7:
## Summary

Fix precision loss in `compareDoubleInt` and `compareDoubleUint` when comparing `double` with `int`/`uint` values outside the safe float64 integer range (> 2^53).

## Problem

The current implementation casts `Int` (int64) or `Uint` (uint64) to `float64` before comparison:

```go
func compareDoubleInt(d Double, i Int) Int {
    // ...bounds checks...
    return compareDouble(d, Double(i))  // Double(i) = float64(i): loses precision!
}
```

IEEE 754 `float64` has a 53-bit mantissa. Any integer with absolute value greater than `2^53 = 9007199254740992` cannot be represented exactly as `float64`. This means two distinct `int64` values can map to the same `float64`, causing the comparison to incorrectly return equal.

### Demonstrated bug

```go
// int(9007199254740993) == double(9007199254740992.0) evaluates to TRUE
// but these are different numbers — off by 1
result := compareDoubleInt(Double(9007199254740992), Int(9007199254740993))
// result == 0 (equal) — WRONG
```

Other affected values:
- `int(100000000000000001) == double(1e17)` → `true` (should be `false`)
- `int(1000000000000000001) == double(1e18)` → `true` (should be `false`)

## Impact

CEL is used as the expression engine in GCP IAM Conditions, Firebase Security Rules, and Kubernetes admission webhooks. A policy rule that compares an integer field to a double literal near these boundary values can produce incorrect authorization decisions.

## Fix

Use `math/big.Float` for exact comparison, eliminating the int→float64 precision loss:

```go
func compareDoubleInt(d Double, i Int) Int {
    if d < math.MinInt64 { return IntNegOne }
    if d > math.MaxInt64 { return IntOne }
    bf := new(big.Float).SetFloat64(float64(d))
    bi := new(big.Float).SetInt64(int64(i))
    return Int(bf.Cmp(bi))
}
```

The same fix is applied to `compareDoubleUint`.

The fast-path bounds checks (`< MinInt64`, `> MaxInt64`) remain unchanged and avoid the `big.Float` allocation for values clearly out of range. The `big.Float` path is only reached for values in the representable int64 range where precision matters.

## Testing

All existing comparison tests continue to pass. The previously incorrect cases now return the correct result:

```
int(9007199254740993) == double(9007199254740992.0) → false ✅
int(100000000000000001) == double(1e17)             → false ✅
int(1000000000000000001) == double(1e18)            → false ✅
```

### DIFF
--- a/common/types/compare.go
+++ b/common/types/compare.go
@@ -16,10 +16,16 @@ package types
 
 import (
 	"math"
+	"math/big"
 
 	"github.com/google/cel-go/common/types/ref"
 )
 
+// compareDoubleInt compares a CEL double and int value for ordering purposes.
+//
+// Direct conversion of Int to float64 loses precision for integer values
+// outside the safe integer range of float64 (i.e., abs(i) > 2^53). Using
+// math/big.Float ensures an exact comparison without rounding errors.
 func compareDoubleInt(d Double, i Int) Int {
 	if d < math.MinInt64 {
 		return IntNegOne
@@ -27,13 +33,20 @@ func compareDoubleInt(d Double, i Int) Int {
 	if d > math.MaxInt64 {
 		return IntOne
 	}
-	return compareDouble(d, Double(i))
+	bf := new(big.Float).SetFloat64(float64(d))
+	bi := new(big.Float).SetInt64(int64(i))
+	return Int(bf.Cmp(bi))
 }
 
 func compareIntDouble(i Int, d Double) Int {
 	return -compareDoubleInt(d, i)
 }
 
+// compareDoubleUint compares a CEL double and uint value for ordering purposes.
+//
+// Direct conversion of Uint to float64 loses precision for values outside
+// the safe integer range of float64 (i.e., u > 2^53). Using math/big.Float
+// ensures an exact comparison without rounding errors.
 func compareDoubleUint(d Double, u Uint) Int {
 	if d < 0 {
 		return IntNegOne
@@ -41,7 +54,9 @@ func compareDoubleUint(d Double, u Uint) Int {
 	if d > math.MaxUint64 {
 		return IntOne
 	}
-	return compareDouble(d, Double(u))
+	bf := new(big.Float).SetFloat64(float64(d))
+	bu := new(big.Float).SetUint64(uint64(u))
+	return Int(bf.Cmp(bu))
 }
 
 func compareUintDouble(u Uint, d Double) Int {
--- a/common/types/compare_test.go
+++ b/common/types/compare_test.go
@@ -1,0 +1,75 @@
+// Copyright 2026 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package types
+
+import "testing"
+
+// TestCompareDoubleIntPrecision verifies that compareDoubleInt and
+// compareDoubleUint correctly distinguish integer values that are adjacent
+// to float64 representable boundaries (> 2^53), where naive float64 casting
+// loses precision and can incorrectly equate distinct integers.
+func TestCompareDoubleIntPrecision(t *testing.T) {
+	tests := []struct {
+		d    Double
+		i    Int
+		want Int
+		desc string
+	}{
+		// Values within safe float64 range — must still work correctly.
+		{Double(1.0), Int(1), IntZero, "1.0 == 1"},
+		{Double(1.5), Int(1), IntOne, "1.5 > 1"},
+		{Double(0.5), Int(1), IntNegOne, "0.5 < 1"},
+		{Double(9007199254740992), Int(9007199254740992), IntZero, "2^53 == 2^53"},
+
+		// Precision-loss zone (> 2^53): distinct integers must NOT compare equal.
+		// Bug: Double(i) cast loses low bits, collapsing adjacent ints to same float64.
+		{Double(9007199254740992), Int(9007199254740993), IntNegOne, "2^53 < 2^53+1"},
+		{Double(9007199254740992), Int(9007199254740994), IntNegOne, "2^53 < 2^53+2"},
+		{Double(1e17), Int(100000000000000001), IntNegOne, "1e17 < 1e17+1"},
+		{Double(1e18), Int(1000000000000000001), IntNegOne, "1e18 < 1e18+1"},
+
+		// Symmetric: int smaller than double.
+		{Double(9007199254740993), Int(9007199254740992), IntOne, "2^53+1 > 2^53 (as Double)"},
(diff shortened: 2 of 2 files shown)


TristonianJones:
@jnthntatum @l46kok @jcking What are your thoughts on switching to `BigInt`, `BigFloat` under the covers for comparisons like this? If we were to make such a change we'd need it to be consistent across all stacks.

jcking:
Isn't this the same problem as https://github.com/google/cel-go/issues/1317?

I'd much rather not switch to BigFloat/BigInt, its a pretty heavy weight dep in other stacks and typically performs memory allocation. I am pretty sure you can implement this correctly without using those.

TristonianJones:
@jcking it's mostly the same, but with a wider type representation. The way in which doubles are rounded depends on compiler flags and platform settings, so I'm not sure if switching to a wider representation above int53 precision would avoid other incongruities relating to floating point. That's my main thought.

jcking:
Rounding is controlled by IEEE 754 behavior. We should just be able to brute force it with:

```
func compare(a int64, b float64) Value {
  if math.IsInf(b) {
    if b < 0.0 {
      return 1;
    }
    return -1;
  }
  if math.IsNaN(b) {
    // Um...error?
  }
  // Round toward zero to nearest integer.
  c := math.Trunc(b)
  d := int64(c)
  // Compare non-fractional parts. If they are diffferent the fractional part doesn't matter.
  if a < d {
    return -1
  }
  if a > d {
    return 1
  }
  // Check the fractional part by removing the non-fractional part via subtraction.
  b -= c
  if c < 0 {
    return -1
  }
  if c > 0 {
    return 1
  }
  return 0
}
```

That should get pretty close. we would have to check the minimum representable value before zero and infinity and see if they hold. But that can be peeked at as well by force casting to int and back and comparing.

XananasX7:
Thanks @jcking — that's a clean and correct approach. The brute-force comparison using `math.Trunc` avoids the BigInt/BigFloat dependency entirely while still handling the edge cases correctly.\n\nI'll implement that algorithm for the int64/float64 comparison. A few edge cases to verify:\n- NaN comparison should return an error (as you noted with the comment)\n- `math.MinInt64` and `math.MaxInt64` boundaries — need to check whether their float representations are exact or rounded\n- Negative infinity / positive infinity handled by the existing check\n\nI'll push an updated implementation using your suggested approach. Thanks for working through this.

xhon-pelushi (commented):
The bug is real and worth fixing. Since the two commits here (2026-05-26) predate @jcking's `math.Trunc` suggestion and @XananasX7's "I'll implement that algorithm" reply, I've focused on the three edge cases listed in that reply, which as far as I can tell nobody has answered in the two and a half months since. All three have concrete answers, and one of them is a trap the `Trunc` implementation will hit if it reuses the existing bounds checks unchanged.

## The three open edge cases

### 1. "NaN comparison should return an error"

Already handled, above these helpers — no new logic needed. Every entry point guards NaN before calling in:

| caller | guard |
| --- | --- |
| `Double.Compare` (`double.go:54`, `:59`) | `NewErr("NaN values cannot be ordered")` |
| `Double.Equal` (`double.go:159`, `:164`) | `False` |
| `Int.Compare` (`int.go:68`) | `NewErr("NaN values cannot be ordered")` |
| `Int.Equal` (`int.go:218`) | `False` |
| `Uint.Compare` (`uint.go:62`) | `NewErr("NaN values cannot be ordered")` |
| `Uint.Equal` (`uint.go:187`) | `False` |

and `compare.go`, `double.go`, `int.go`, `uint.go` are the only files referencing the four `compare{Double,Int,Uint}{Int,Uint,Double}` helpers, so there is no unguarded path in.

**But that precondition becomes load-bearing in a way it isn't today**, and it's worth a guard or at least a comment. Today `compareDouble(NaN, x)` falls through both branches and returns `IntZero` — wrong, but harmless. With `big.Float` it becomes a panic:

```go
// math/big/float.go
// SetFloat64 panics with [ErrNaN] if x is a NaN.
func (z *Float) SetFloat64(x float64) *Float {
    ...
    if math.IsNaN(x) {
        panic(ErrNaN{"Float.SetFloat64(NaN)"})
    }
```

and with `math.Trunc` + `int64(...)` it becomes an implementation-defined value. I could not construct a reachable panic — the six guards above hold — so this is latent, not a live bug. Still, "a NaN reaching this function panics the evaluator" is a sharp enough edge that a `math.IsNaN` early return, or a doc comment stating the precondition, seems cheap insurance for a library that ends up in admission controllers.

### 2. "MinInt64 and MaxInt64 boundaries — exact or rounded?"

**`MinInt64` is exact; `MaxInt64` rounds up.** Measured:

```
(double)INT64_MIN  = -9223372036854775808.0   exact  (= -2^63)
(double)INT64_MAX  =  9223372036854775808.0   rounds UP to 2^63
largest float64 strictly below 2^63 = 9223372036854774784.0
```

So the existing guard `if d > math.MaxInt64` is really `if d > 2^63`, and **a `d` of exactly `2^63` passes it** and reaches the comparison body. The same holds on the uint side:

```
(double)UINT64_MAX = 18446744073709551616.0   rounds UP to 2^64
d == 2^64 passes `d > math.MaxUint64`:  yes
largest float64 strictly below 2^64 = 18446744073709549568.0
```

**This is the trap for the `Trunc` version.** With `big.Float` it's harmless: `2^63` compares exactly against any `int64` and yields `1`, which is correct. But jcking's sketch does

```go
c := math.Trunc(b)
d := int64(c)
```

and `int64(9223372036854775808.0)` is out of `int64` range. Per the Go spec, "in all non-constant conversions involving floating-point values, if the result type cannot represent the value the conversion succeeds but the result value is implementation-dependent." So the `Trunc` implementation needs the guards tightened to reject at the boundary rather than past it — `d >= 2^63` for the int case and `d >= 2^64` for the uint case — or the conversion has to be done only after proving the value is in range. Reusing the current `>` guards would move the precision bug to a much rarer input rather than removing it.

### 3. "Negative inf"

Both infinities are already excluded before the body: `+Inf > 2^63` returns `IntOne` and `-Inf < -2^63` returns `IntNegOne`, and the same on the uint side. So neither implementation ever sees an infinity in the comparison core. (For completeness, `big.Float.SetFloat64` handles them fine anyway — it sets `form = inf` rather than panicking; it's only NaN that panics.)

## On `big.Float` vs `Trunc`

Nothing to argue here since the author has already agreed to switch, but for the record the objection is easy to quantify: `new(big.Float).SetFloat64(...)` plus `new(big.Float).SetInt64(...)` is **two heap allocations per comparison**, on a path that currently allocates nothing, and numeric comparison is about as hot as CEL evaluation gets. That matches @jcking's "typically performs memory allocation" concern.

## The bug itself reproduces

For anyone wanting the one-line confirmation that this is worth fixing at all:

```
2^53                              = 9007199254740992
int64 9007199254740993 -> double  = 9007199254740992.0
```

so `int(9007199254740993) == double(9007199254740992.0)` is `true` today, off by one, exactly as described.

## What I could not test

There's no Go toolchain on this machine, so **I did not compile or run any of this** — no `go test`, and the new `compare_test.go` is unverified. Everything above is source reading against `master` (cel-go and Go's `math/big`) plus the IEEE-754 boundary values computed in C, which has the same `binary64` semantics but is not a substitute for running the Go code. Worth re-confirming the `2^63` / `2^64` guard behaviour with an actual Go test before relying on it.

PR #1332, opened 2026-05-27
Remove duplicate policy tests covered by policy conformance tests

l46kok:


policy/BUILD.bazel
```diff
@@ -75,152 +75,157 @@ go_test(
...
 
 proto_descriptor_set(
     name = "test_all_types_fds",
     deps = [
-        "//test/proto3pb:test_all_types_proto",
+        "@dev_cel_expr//proto/cel/expr/conformance/proto3:test_all_types_proto",
     ],
 )
 
 cel_go_test(
```
jnthntatum: These are effectively the same as the conformance tests, right? probably just remove if it is genuinely identical.

If it's a useful regression test for the test runner library, maybe just keep separate (and remove ones that are just for demonstrating policy behaviors covered by the conformance tests).

policy/helper_test.go
```diff
@@ -48,7 +51,7 @@ var (
 	  resource.labels.?environment.orValue("prod"),
 	  resource.labels.?break_glass.orValue("false") == "true"],
 	  !(@index1 || resource.containers.all(c, c.startsWith(@index0 + ".")))
-	    ? optional.of("only %s containers are allowed in namespace %s".format([@index0, resource.namespace]))
+	    ? optional.of("only " + @index0 + " containers are allowed in namespace " + resource.namespace)
```
jnthntatum: I don't think this is a great idea -- even if we end up writing similar policies for the sake of conformance testing and the unit tests, it's probably better to have the unit test input collocated with the test instead in a separate repo.

It may be reasonable to trim these down if we're getting coverage from the conformance test runner.

policy/BUILD.bazel
l46kok: Yeah I think we can remove all of the policy tests here. We have sufficient coverage for cel_go_test macro itself elsewhere imo.

policy/helper_test.go
l46kok: Fair point, these exercise the expected structure of composed ASTs, so the coverage is definitely needed. Let me see if I can trim these down to a more representative set. We may end up just keeping the duplicate policies for now.

TristonianJones (approved):
I wonder if we should move the `policy` folder into its own repo under cel-expr? Maybe cel-expr/cel-go-policy? Seems like it might be a good idea at some point in the future.

l46kok:
Yeah I think once we're comfortable in doing so, we can pursue that. Probably when we add a few more representative semantics .

PR #1334, opened 2026-06-04
Add depth validation for ASTs loaded through ParsedExprToAst / Checke...

### LINKED ISSUE #1333: Add depth validation for ASTs loaded through ParsedExprToAst / CheckedExprToAst
## Summary

`cel-go` currently applies recursion/depth limits during string parsing, but the same kind of depth validation does not appear to be enforced when expressions are loaded from pre-built AST protos through APIs such as `ParsedExprToAst` / `CheckedExprToAst`.

This means callers that store, deserialize, or otherwise ingest CEL ASTs directly can bypass the parser recursion guard and pass a deeply nested AST into later phases such as checking or program construction.

In extreme cases, this can lead to unbounded recursive traversal and a Go runtime stack overflow rather than a normal returned error.

## Affected path

The issue is not about normal CEL source parsing. The parser already has recursion limiting.

The affected scenario is non-parser AST ingestion, for example:

```go
cel.ParsedExprToAst(...)
cel.CheckedExprToAst(...)
env.Check(...)
env.Program(...)
```

A caller may reasonably use these APIs to load previously parsed or checked expressions from storage. If the stored AST is deeply nested, later recursive traversal in the checker / planner can reach excessive depth.

## Expected behavior

ASTs loaded through proto/native conversion APIs should be subject to a bounded expression-depth validation step before recursive checking or planning.

If an AST exceeds the supported maximum depth, `cel-go` should return an ordinary error, similar in spirit to parser depth-limit failures.

## Actual behavior

The parser depth guard is bypassed when the input is already an AST/proto. Later recursive processing can traverse the AST without an equivalent depth limit.

For sufficiently nested inputs, this can crash the host process with:

```text
fatal error: stack overflow
```

instead of returning a recoverable error.

## Suggested fix

Add depth validation for ASTs entering through non-parser ingestion paths, ideally reusing the same limit as the parser default.

Possible approaches:

1. Add a reusable AST depth validator in the common AST layer.
2. Call it from `ParsedExprToAst` / `CheckedExprToAst`, or from the public `Check()` / `Program()` construction path.
3. Ensure checker and planner recursive traversals cannot exceed the configured maximum expression depth.
4. Add a regression test using a moderately deep synthetic AST that exceeds the configured limit but does not attempt to exhaust the process stack.

The default parser recursion limit appears to be 250, so using the same default for loaded ASTs would keep behavior consistent.

## Security / reliability impact

Any embedder that accepts stored or serialized CEL ASTs from untrusted or semi-trusted storage can currently receive an unrecoverable process crash instead of a validation error.

This is especially relevant for systems that treat stored CEL ASTs as policy objects and compile or load them in a long-running service.

## Notes

I originally reported this through Google Bug Hunters, and they confirmed that since this is not being tracked for reward, it is acceptable to proceed publicly through GitHub.

wilyan09007:
## Summary

`cel-go` enforces a recursion/expression-depth limit while parsing CEL source, but ASTs ingested directly — through `ParsedExprToAst` / `CheckedExprToAst` — bypass that guard. A deeply nested loaded AST then flows into the recursive checker (`Env.Check`) and planner (`Env.Program` / `PlanProgram`), which can exhaust the Go stack and abort the process with `fatal error: stack overflow` instead of returning a recoverable error. Embedders that load stored or serialized CEL ASTs as policy objects are the most exposed.

This validates expression nesting depth before those recursive phases:

- Adds `ast.ExceedsMaxDepth(expr, maxDepth)` in `common/ast` — a **bounded** traversal that never recurses past `maxDepth+1` levels, so it stays safe on the same adversarially deep inputs it guards against. The default `ast.MaxNestingDepth = 250` mirrors the parser's default `maxRecursionDepth`, keeping loaded ASTs consistent with parsed ones.
- `Env.Check` and `Env.PlanProgram` now run that check up front and return an ordinary error/issue — `input exceeds maximum expression nesting depth: 250` — rather than recursing into a stack overflow.
- Adds a regression test that loads a synthetic over-deep AST and asserts `Check` and `Program` return an error (not a crash), while normal expressions remain unaffected.

Closes #1333


common/ast/depth.go
```diff
@@ -0,0 +1,101 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package ast
+
+// MaxNestingDepth is the default maximum expression nesting depth applied to
+// ASTs that enter through non-parser ingestion paths. It mirrors the parser's
+// default maxRecursionDepth so loaded ASTs are validated against the same
+// bound as ASTs produced from CEL source.
+const MaxNestingDepth = 250
```
TristonianJones: It's preferred to create functional options with sensible defaults for user-configurable behavior. CEL doesn't currently validate the AST size on load since the parser _should_ be the only entry point for a user-authored AST. Defense in depth is great, but could you please hoist this out of a publicly mutable variable and add it to the `cel.NewEnv` as a functional option that gets checked during `Program` calls?

cel/options.go
TristonianJones, line 124: Please add the `limitMaxASTDepth` to the list of `limitIDsToNames`

common/ast/depth.go
```diff
@@ -0,0 +1,101 @@
...
+
+// ExceedsMaxDepth reports whether the given expression nests deeper than
+// maxDepth. The traversal itself is bounded: it never recurses past
+// maxDepth+1 levels, so it is safe to call on adversarially deep inputs that
+// would otherwise blow the Go stack during later checking or planning.
+//
+// A non-positive maxDepth disables the check and always returns false.
+func ExceedsMaxDepth(e Expr, maxDepth int) bool {
```
TristonianJones: Prefer instead to use a visitor that observes the expression depth to the `navigable.go` file.

Existing methods:
https://github.com/google/cel-go/blob/f456a6ec60db68b418851535ac572c51703caa48/common/ast/navigable.go#L174-L183

Proposed impl:
```go
// ExceedsDepth determines whether the AST contains expressions nested beyond a specified maxDepth.
func ExceedsDepth(a *AST, depth int) bool {
  exceedsDepth := false
  visitor := NewExprVisitor(func(e Expr) {
      nav := e.(NavigableExpr)
      if nav.Depth() >= depth {
        exceedsDepth = true
      }
  })
  root := NavigateAST(a)
  PostOrderVisit(root, visitor)
  return exceedsDepth
}
```

cel/env.go
```diff
@@ -693,6 +693,19 @@ func (e *Env) Program(ast *Ast, opts ...ProgramOption) (Program, error) {
...
+	// evaluation would otherwise risk a Go stack overflow on adversarially
+	// deep inputs. The limit is configurable via ExpressionNestingDepthLimit;
+	// a zero value falls back to the parser-matching default and a negative
+	// value disables the check.
+	maxDepth := e.limits[limitMaxASTDepth]
+	if maxDepth == 0 {
+		maxDepth = defaultMaxASTDepth
+	}
+	if a != nil && celast.ExceedsMaxDepth(a.Expr(), maxDepth) {
+		return nil, fmt.Errorf("input exceeds maximum expression nesting depth: %d", maxDepth)
```
TristonianJones: Maybe if the maxDepth is 0 just skip the check. The other alternative is to shift this check to the `io.go` file instead which converts from the proto representation to the AST. In fact, let's maybe do that instead since a hand-rolled golang AST isn't possible without full control of the code.

cel/io_test.go
```diff
@@ -317,31 +317,64 @@ func TestCheckedExprToAstMissingInfo(t *testing.T) {
...
+			Id: int64(i + 2),
+			ExprKind: &exprpb.Expr_CallExpr{
+				CallExpr: &exprpb.Expr_Call{
+					Function: operators.LogicalNot,
+					Args:     []*exprpb.Expr{expr},
+				},
+			},
+		}
+	}
+	deepAst := ParsedExprToAst(&exprpb.ParsedExpr{Expr: expr})
```
TristonianJones: Let's instead add the depth check in the `ParsedExprToAst` and `CheckedExprToAst` since the program construction path is the hot path and the defense in depth check should be something users can disable if it doesn't apply since it's added compute without added security in a number of cases.

TristonianJones (dismissed):

common/ast/navigable.go
```diff
@@ -181,6 +181,29 @@ func PreOrderVisit(expr Expr, visitor Visitor) {
...
 }
 
+// ExceedsDepth determines whether the AST contains expressions nested deeper than the specified
+// maxDepth. The root expression has depth 0, so a maxDepth of 250 permits expressions nested up
+// to and including 250 levels deep.
+//
+// The traversal is bounded: it descends at most maxDepth+1 levels, so it remains safe to call on
+// adversarially deep inputs that could otherwise exhaust the Go stack during later checking or
+// planning. A non-positive maxDepth disables the check and returns false.
+func ExceedsDepth(a *AST, maxDepth int) bool {
```
TristonianJones, line 191: Would you mind adding unit tests within the `navigable_test.go`?

TristonianJones (dismissed):

common/ast/navigable.go
wilyan09007: Sure, no problem.

TristonianJones (approved):

PR #1335, opened 2026-06-06
validate offset on empty-string path in indexOf and lastIndexOf

alhudz:
`indexOfOffset` and `lastIndexOfOffset` validate the start offset only after the empty-search-string short circuit, so an empty needle skips the check. `'tacocat'.indexOf('', -1)` returns `-1` and `'tacocat'.indexOf('', 30)` returns `30`, whereas `'tacocat'.indexOf('a', -1)` errors with `index out of range` and `'tacocat'.indexOf('a', 30)` reports `-1`. `lastIndexOf` behaves the same way.

Move the offset bounds check ahead of the empty-string branch in both functions so a negative offset errors and an offset past the end reports `-1`, while a valid offset (including the end position) still returns the offset. Keeping the check in the callee means every `indexOf`/`lastIndexOf` overload shares one rule rather than depending on the search string being non-empty.

### DIFF
--- a/ext/strings.go
+++ b/ext/strings.go
@@ -669,9 +669,9 @@ func indexOfOffset(str, substr string, offset int64) (int64, error) {
 	}
 	runes := []rune(str)
 	if substr == "" {
-		// The empty string matches at the search offset, which must fall within the string.
+		// The empty string matches at the search offset, clamped to the end of the string.
 		if off > len(runes) {
-			return -1, nil
+			return int64(len(runes)), nil
 		}
 		return offset, nil
 	}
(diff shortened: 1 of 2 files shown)


ext/strings.go
```diff
@@ -663,15 +663,19 @@ func indexOf(str, substr string) (int64, error) {
...
-	runes := []rune(str)
-	subrunes := []rune(substr)
 	if off < 0 {
 		return -1, fmt.Errorf("index out of range: %d", off)
 	}
+	runes := []rune(str)
+	if substr == "" {
+		// The empty string matches at the search offset, which must fall within the string.
+		if off > len(runes) {
+			return -1, nil
```
TristonianJones: Java seems to treat this case `len(runes)`

```java
System.out.println("".indexOf("", 30)); // 0
System.out.println("x + y".indexOf("", 30)); // 5
```

alhudz: Good catch. Both paths now clamp to `len(runes)` for an out-of-range offset on the empty needle, so `'tacocat'.indexOf('', 30)` reports `7` instead of `-1`, matching Java. That also lines up with `lastIndexOf('')` already returning `len(runes)`. Updated the regression cases to match.

TristonianJones (approved):

PR #1337, opened 2026-06-09
guard int32/uint32 map key narrowing in qualifyInternal

alhudz:
Repro: indexing a native `map[int32]any` with a key outside the `int32` range, e.g. `m[4294967296]` where `m == {0: "x"}`, returns `"x"` instead of a no-such-key error. Same for `map[uint32]any`.
Cause: `intQualifier.qualifyInternal` and `uintQualifier.qualifyInternal` cast the `int64`/`uint64` index to `int32`/`uint32` before the lookup, so `4294967296` wraps to `0` and collides with the entry stored at key `0`.
Fix: only perform the narrowed lookup when the index round-trips through the key type (`int64(int32(i)) == i`), otherwise fall through to the existing missing-key path.

### DIFF
--- a/interpreter/attributes_test.go
+++ b/interpreter/attributes_test.go
@@ -399,30 +399,48 @@ func TestAttributesConditionalAttrFalseBranch(t *testing.T) {
 func TestAttributesNarrowMapKeyQualifier(t *testing.T) {
 	reg := newTestRegistry(t)
 	attrs := NewAttributeFactory(containers.DefaultContainer, reg, reg)
-	vars, _ := NewActivation(map[string]any{
+	vars, err := NewActivation(map[string]any{
 		"i32": map[int32]any{0: "zero"},
 		"u32": map[uint32]any{0: "zero"},
 	})
-
+	if err != nil {
+		t.Fatalf("NewActivation() failed: %v", err)
+	}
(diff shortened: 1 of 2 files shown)


TristonianJones (dismissed):

interpreter/attributes_test.go
```diff
@@ -396,6 +396,36 @@ func TestAttributesConditionalAttrFalseBranch(t *testing.T) {
...
 }
 
+func TestAttributesNarrowMapKeyQualifier(t *testing.T) {
+	reg := newTestRegistry(t)
+	attrs := NewAttributeFactory(containers.DefaultContainer, reg, reg)
+	vars, _ := NewActivation(map[string]any{
+		"i32": map[int32]any{0: "zero"},
+		"u32": map[uint32]any{0: "zero"},
+	})
+
```
TristonianJones: Optional, but you could make this test parameterized with expected inputs and outputs.

alhudz: Done, rewrote it as a table-driven test with expected outputs and exact `no such key` errors, same shape as `TestAttributesOptional`. Also covers the in-range `u32` lookup now.

TristonianJones (approved):

PR #1338, opened 2026-06-10
reject non-RFC3339 timestamp strings in timestamp() conversion

### LINKED ISSUE #1108: `cel-go` inherits RFC 3339 parsing bugs from Golang's `time.Parse()`
**Describe the bug**
`cel-go` inherits RFC 3339 parsing bugs from Golang's `time.Parse()`

**To Reproduce**

- [ ]  parser
- [ ]  checker
- [x]  interpreter

```cel
(
  timestamp("2025-01-17T01:00:00.001Z") ==
  timestamp("2025-01-17T01:00:00,001Z") // comma-delimited fractional seconds
) &&
(
  timestamp("2025-01-17T01:00:00Z") ==
  timestamp("2025-01-17T1:00:00Z") // single-digit hour
) &&
(
  timestamp("2025-01-17T01:01:01.001+00:01") ==
  timestamp("2025-01-18T01:01:01.001+24:01") // offset > 24:00
) &&
(
  timestamp("2025-01-17T01:01:01.001+01:00") ==
  timestamp("2025-01-17T01:01:01.001+00:60") // offset minute > 59
)
```

Test setup:
```go
package main

import (
  "fmt"
)

func main() {
  env, _ := cel.NewEnv()
  ast, _ := env.Compile(`
    (
      timestamp("2025-01-17T01:00:00.001Z") ==
      timestamp("2025-01-17T01:00:00,001Z") // comma-delimited fractional seconds
    ) &&
    (
      timestamp("2025-01-17T01:00:00Z") ==
      timestamp("2025-01-17T1:00:00Z") // single-digit hour
    ) &&
    (
      timestamp("2025-01-17T01:01:01.001+00:01") ==
      timestamp("2025-01-18T01:01:01.001+24:01") // offset > 24:00
    ) &&
    (
      timestamp("2025-01-17T01:01:01.001+01:00") ==
      timestamp("2025-01-17T01:01:01.001+00:60") // offset minute > 59
    )
  `)
  prg, _ := env.Program(ast)
  out, _, _ := prg.Eval(map[string]any{})

  fmt.Println(out)
}
```

**Expected behavior**
Each of the latter examples should result in an error from the `timestamp()` function

**Additional context**
If compatibility is needed, I recommend we add an option in `cel-go` to provide compatibility rather than forcing other implementations to replicate Golang's bugs.

alhudz:
Repro: `timestamp("2025-01-17T01:00:00,001Z")`, `timestamp("2025-01-17T1:00:00Z")`, `timestamp("2025-01-18T01:01:01.001+24:01")` and `timestamp("2025-01-17T01:01:01.001+00:60")` all evaluate to a value instead of erroring.
Cause: the `string`-to-timestamp conversion parses with `time.Parse(time.RFC3339, ...)`, which accepts inputs RFC 3339 forbids: a `,` fractional separator, single-digit time fields, and offset hours past `23` or minutes past `59`. These silently shift the parsed instant or get wrongly accepted.
Fix: gate the conversion on a strict RFC 3339 pattern in the callee before `time.Parse` runs, so the listed forms report the usual conversion error while `time.Parse` keeps doing the calendar validation.

Closes #1108


common/types/string.go
```diff
@@ -122,11 +122,14 @@ func (s String) ConvertToType(typeVal ref.Type) ref.Val {
 			return durationOf(d)
 		}
 	case TimestampType:
-		if t, err := time.Parse(time.RFC3339, s.Value().(string)); err == nil {
-			if t.Unix() < minUnixTime || t.Unix() > maxUnixTime {
-				return celErrTimestampOverflow
+		str := s.Value().(string)
+		if strictRFC3339Pattern.MatchString(str) {
```
TristonianJones: Prefer early return on an error for an invalid format as this will provide a more helpful message to the user (or agent).

alhudz: Done. The pattern check now returns early with `invalid RFC 3339 timestamp "<value>"` rather than falling through to the generic conversion error, and the regression test asserts the exact message for each rejected form. `go test ./...` still passes.

TristonianJones: Would you mind adding a little benchmark to see the perf impact of the regex?

alhudz: Added `BenchmarkStringConvertToTimestamp` over the conversion path. On an M4, converting a valid `2025-01-01T12:34:56.123456789Z` runs around 600 ns/op at 1 alloc/op. The regex `MatchString` is the bulk of that (~500 ns/op, 0 allocs) against ~90 ns/op for the bare `time.Parse`, so the gate adds no allocations but does roughly 6x the parse cost. Happy to swap the regex for a small hand-rolled scan if that overhead matters on the hot path.

TristonianJones: If you wouldn't mind a small hand-rolled scan. I looked into it a bit and it's doable, but roughly 100-ish lines ... seems easy to conformance test against the regex though, so maybe keep the regex for testing purposes.

alhudz: Done. The conversion path now goes through a hand-rolled `isStrictRFC3339` scan instead of the regex. Kept `strictRFC3339Pattern` as the reference and added `TestIsStrictRFC3339MatchesPattern` which asserts the scan and the regex agree across the field boundaries (hour/minute/second/offset edges, leap second, comma separator, single-digit fields, missing/empty zone).

The scan is allocation-free and roughly an order of magnitude cheaper than the regex `MatchString` on the same machine, so the gate no longer dominates the convert. `go test ./...` passes.

alhudz:
any update?

TristonianJones:
Hi @alhudz, I was OOO for a couple weeks. I'll take a look this week.

common/types/timestamp.go
```diff
@@ -52,6 +53,89 @@ const (
...
+// isStrictRFC3339 reports whether s matches strictRFC3339Pattern, hand-rolled to
+// keep the conversion path off the regexp engine and its per-call cost.
+func isStrictRFC3339(s string) bool {
+	// Shortest accepted form is "2006-01-02T15:04:05Z" (20 bytes): a 19-byte
+	// fixed-width date-time followed by at least a 'Z'/'z' zone.
+	if len(s) < 20 {
+		return false
+	}
+	// date: \d{4}-\d{2}-\d{2}
+	if !isDigit(s[0]) || !isDigit(s[1]) || !isDigit(s[2]) || !isDigit(s[3]) || s[4] != '-' ||
```
TristonianJones: You might want to consider passing slices and using `strconv.ParseUint`:

```go
!isYear(s[0:4]) || !isChar(s[4], '-') || !isMonth(s[5:7]) || !isChar(s[7], '-') || !isDay(s[8:10]) {
   return false
}
```

```go
func isYear(s string) bool {
   u, err := strconv.ParseUint(v, 10, 64)
   return err == nil && u <= 9999
}

func isChar(got, want byte) bool {
  g, w := rune(got), rune(want)
  return g == w || unicode.ToLower(g) == w
}
```

alhudz: Done. The scan now slices each field into `strconv.ParseUint` helpers (`isYear`/`isMonth`/`isDay`/`isHour`/`isMinute`/`isSecond` over a shared `inRange`, plus `isChar` for the separators). Since `isMonth`/`isDay` now check the grammar ranges 01-12/01-31, I tightened the reference regex to match and added the month/day boundary cases to the conformance test; day-of-month vs month and leap years still fall to `time.Parse`.

On perf, `ParseUint` costs a bit over the raw byte compares: the convert benchmark goes from ~103 ns/op to ~197 ns/op on the same M4, still allocation-free in the gate and roughly 3x cheaper than the regex version. `go test ./...` passes.

TristonianJones: Perfect, thanks for the benchmark. It seems like a reasonable perf / safety / maintenance tradeoff.

TristonianJones (approved):

PR #1339, opened 2026-06-11
feat: expose issue error offsets

### LINKED ISSUE #189: Make Errors more useful
**Feature request checklist**

- [ ] There are no issues that match the desired change

**Change**

Have an Error method to obtain the raw offset (or range?) for the error, in addition to the line and column, to make it more useful when the caller embeds the CEL expression in a larger document and would like to translate errors back to the original context.

Furthermore, Error.ToDisplayString takes a Source argument, but in the case where Env.Parse() fails, we don't necessarily have an Ast to obtain the Source from, so you currently need to go outside the 'cel' package API in order to translate Issues into nice display strings.

rohan-patnaik:
## What changed
- Added `Issues.Source()` for the source attached to parse/check issues.
- Added `Issues.ErrorOffset()` and `Issues.ErrorOffsetRange()` for raw error positions.

## Why
Callers that embed CEL in a larger document can map CEL errors back to their own source ranges without parsing display strings.

Fixes #189.

## Testing
- go test ./common ./cel -run 'TestIssues|TestErrors'

### DIFF
--- a/cel/env.go
+++ b/cel/env.go
@@ -1009,6 +1009,41 @@ func (i *Issues) Errors() []*Error {
 	return i.errs.GetErrors()
 }
 
+// Source returns the source associated with the issues.
+func (i *Issues) Source() Source {
+	if i == nil {
+		return nil
+	}
+	return i.errs.Source()
+}
+
+// ErrorOffset returns the start character offset for the error, if available.
+func (i *Issues) ErrorOffset(err *Error) (int32, bool) {
+	offsetRange, found := i.ErrorOffsetRange(err)
+	if found {
+		return offsetRange.Start, true
+	}
+	return 0, false
+}
+
+// ErrorOffsetRange returns the character offset range for the error, if available.
+func (i *Issues) ErrorOffsetRange(err *Error) (celast.OffsetRange, bool) {
+	if i == nil || err == nil {
+		return celast.OffsetRange{}, false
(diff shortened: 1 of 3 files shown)


cel/env.go
```diff
@@ -1009,6 +1009,41 @@ func (i *Issues) Errors() []*Error {
 	return i.errs.GetErrors()
 }
 
+// Source returns the source associated with the issues.
+func (i *Issues) Source() Source {
```
TristonianJones, line 1013: Usually the best way to take care of this is to preserve the `Source` by calling `cel.CompileSource` and then performing these operations in terms of the `Issues` and `Source` values. This is usually more guaranteed to be correct than poking into the internal `Issues` state.

rohan-patnaik: Ok ,makes sense. Seems the preferred pattern is for callers to create/preserve a `Source`, call `CompileSource` or `ParseSource`, and then map each issue’s `Error.Location` through `Source.LocationOffset`, rather than exposing more of the internal `Issues` state.

Would you prefer that I close this PR, convert it into docs/example coverage for that usage pattern, or narrow it to a smaller convenience API if there is still a gap worth covering?

TristonianJones: If you'd like to add an example into `examples`, I think that's useful. Thanks!

PR #1341, opened 2026-06-11
fix: track unknowns through comprehensions

### LINKED ISSUE #321: Attribute Tracking Through Comprehensions
**Feature request checklist**

- [] If this is a bug, please file a [Bug Report](./ISSUE_TEMPLATE).

**Change**
Add support for tracking attributes through comprehensions.

**Example**
`var[2].field` is declared unknown
for expression:
`var.exists(x, x.field == 'test')`
I would expect that `var[2].field` might be returned as an unknown value. 

Adding tracking here to keep feature parity with cel-cpp. I'm not sure what the current level of support is in cel-go. Let me know if I can add more detail.

rohan-patnaik:
## What changed
- Let comprehensions iterate when only a child of the range is unknown.
- Mark matching range elements as unknown during iteration.
- Preserve the original attribute trail, e.g. `items[2].field`.

## Why
Partial evaluation currently stops at the range variable for descendant unknowns. This lets expressions like `items.exists(x, x.field == 'test')` report the specific unknown child attribute.

Fixes #321.

## Testing
- go test ./interpreter -run 'TestInterpreter_ComprehensionTracksUnknownAttribute|TestInterpreter$'
- go test ./interpreter

### DIFF
--- a/interpreter/interpretable.go
+++ b/interpreter/interpretable.go
@@ -799,7 +799,7 @@ func (fold *evalFold) Eval(ctx Activation) ref.Val {
 	f := newFolder(fold, ctx)
 	defer releaseFolder(f)
 
-	foldRange := fold.iterRange.Eval(ctx)
+	foldRange := fold.iterRange.Eval(f.iterRangeActivation())
 	if types.IsUnknownOrError(foldRange) {
 		return foldRange
 	}
@@ -1309,6 +1309,7 @@ type folder struct {
 	accuVal     ref.Val
 	iterVar1Val any
 	iterVar2Val any
+	iterIndex   int64
 
 	// bookkeeping flags to modify Activation and fold behaviors.
 	initialized   bool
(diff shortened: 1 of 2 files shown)


TristonianJones:
@rohan-patnaik I'm holding off on making changes here since I've been looking into operation "fusing" that might make this tricky to support (tradeoff is 10x performance boost). On the other hand, the unknown tracing becomes more important with async dispatch... Let me see how far I get on "fusing" ... Lots of tradeoffs to weigh

rohan-patnaik:
Agreed that preserving unknown tracing will become more important with async dispatch. 
Happy to adapt this PR once the direction/tradeoffs are clearer 👍

rohan-patnaik:
@TristonianJones, has the fusing direction settled enough for me to refresh and adapt this PR? Thanks!

rohan-patnaik:
@TristonianJones, has the fusing direction settled enough for me to refresh and adapt this PR? Happy to make the changes needed—thanks!

PR #1342, opened 2026-06-15
refactor: replace interface{} with any (Go 1.18+)

MD-Mushfiqur123:
Replace `interface{}` with `any` (Go 1.18+ type alias).

11 files changed via code search.

Fixes type literal usage only (not interface definitions).


TristonianJones:
Hi @MD-Mushfiqur123, can you revert the `gen/` and `repl/parser` related files. They're code-generated and so we don't control the content.

PR #1344, opened 2026-06-15
Execution frame integration with updated IntepretableV2

TristonianJones:
Plugs the `ExecutionFrame` into the CEL interpreter and program APIs and introduces
a new `InterpretableV2` for calling program steps with the `ExecutionFrame` rather than
the `Activation`.


interpreter/interpreter.go
```diff
@@ -44,15 +44,15 @@ type EvalObserver func(vars Activation, id int64, programStep any, value ref.Val
 // StatefulObserver observes evaluation while tracking or utilizing stateful behavior.
 type StatefulObserver interface {
 	// InitState configures stateful metadata on the activation.
```
l46kok: nit: `s/activation/ExecutionFrame`

interpreter/runtimecost.go
```diff
@@ -62,76 +62,39 @@ func CostObserver(opts ...costTrackPlanOption) PlannerOption {
...
+	return frame.ctx.costs
 }
 
 // Observe computes the incremental cost of each step and records it into the CostTracker associated
 // with the evaluation.
 func (ct *costTrackerFactory) Observe(vars Activation, id int64, programStep any, val ref.Val) {
-	tracker, found := asCostTracker(vars)
-	if !found {
+	frame := AsFrame(vars)
+	tracker := ct.GetState(frame).(*CostTracker)
```
l46kok: Consider checking for `nil` from `ct.getState` call first before the cast

interpreter/interpretable.go
```diff
@@ -112,44 +145,81 @@ type InterpretableConstructor interface {
...
-	return oi.ObserveEval(vars, func(any) {})
+	return oi.ObserveExec(AsFrame(vars), func(any) {})
 }
 
-// ObserveEval evaluates an interpretable and performs per-evaluation state-tracking.
+// ObserveExec evaluates an interpretable and performs per-evaluation state-tracking.
 //
 // This method is concurrency safe and the expectation is that the observer function will use
 // a switch statement to determine the type of the state which has been reported back from the call.
-func (oi *ObservableInterpretable) ObserveEval(vars Activation, observer func(any)) ref.Val {
```
l46kok, line 133: Would the rename here be a breaking change?

interpreter/interpreter.go
```diff
@@ -44,15 +44,15 @@ type EvalObserver func(vars Activation, id int64, programStep any, value ref.Val
 // StatefulObserver observes evaluation while tracking or utilizing stateful behavior.
 type StatefulObserver interface {
```
l46kok, line 45: This is technically a breaking change.. but not sure if it matters at all. Can external authors author a custom cost tracker against this interface and inject it into the evaluator?

ext/bindings.go
```diff
@@ -223,14 +223,18 @@ func (b *dynamicBlock) ID() int64 {
...
 }
 
-// Eval implements the Interpretable interface method.
-func (b *dynamicBlock) Eval(activation cel.Activation) ref.Val {
+func (b *dynamicBlock) Exec(frame *interpreter.ExecutionFrame) ref.Val {
 	sa := b.slotActivationPool.Get().(*dynamicSlotActivation)
-	sa.Activation = activation
+	sa.Activation = frame
 	defer b.clearSlots(sa)
 	return b.expr.Eval(sa)
```
l46kok: Should we push the slot into a child frame? `sa` no longer takes in an Activation, so this would force allocate a new frame on the heap per bind right?

interpreter/interpreter.go
TristonianJones: It's a breaking change if you use `interpreter` directly but I couldn't find any references in public GitHub repos of doing this. Importantly, the `EvalObserver` and it's signature on the `StatefulObserver` are preserved which is usually what people use instead from the top-level `cel` package. Also, since the cost trackers factory is a private struct, I believe the change is safe.

ext/bindings.go
TristonianJones: You're quite right. I looked at memory profiles for `interpreter`, but forgot to check them here. I've exposed `Push` and `Pop` on the execution frame and added more benchmarks.

interpreter/interpretable.go
TristonianJones: I wasn't able to find any references of this in the wild, but I reintroduced the `ObserveEval` call to keep this signature. The implementation just proxies to the `ObserveExec` in a manner consistent with the `InterpretableV2` updates.

l46kok (approved):

PR #1345, opened 2026-06-16
Cost library for ext/network

TristonianJones:
Adds cost computations and cost tests from K8s:

https://github.com/kubernetes/kubernetes/blob/808a553dc88f0343a895e4bbf9917107500b6488/staging/src/k8s.io/apiserver/pkg/cel/library/cost_test.go#L314

There is one small nit in that the equality costing in CEL is slightly 
richer than the one used in K8s and the cost estimates have a range
that's slightly different from the one computed by K8s which doesn't
look at the arguments to the function, but rather returns a unit cost
of one.

### DIFF
--- a/ext/network.go
+++ b/ext/network.go
@@ -16,13 +16,16 @@ package ext
 
 import (
 	"fmt"
+	"math"
 	"net/netip"
 	"reflect"
 
 	"github.com/google/cel-go/cel"
+	"github.com/google/cel-go/checker"
 	"github.com/google/cel-go/common/ast"
 	"github.com/google/cel-go/common/types"
 	"github.com/google/cel-go/common/types/ref"
+	"github.com/google/cel-go/interpreter"
 )
 
 const (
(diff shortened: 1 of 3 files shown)


TristonianJones:
FYI @jpbetz

jpbetz:
@lalitc375 would you check for compatibility with our downstream costs?

impartialstate:
@lalitc375 the tests are ported from K8s. The only difference I noticed is that the cost for equality is overridden in K8s and will produce a slightly smaller cost

lalitc375:
K8s overrides _==_ for IP/CIDR to a flat {1,1} estimate and runtime 1, regardless of v4 or v6. This PR doesn't override _==_, so it falls through to cel-go's default Equals path: estimate {1,  ceil(min(lhsMax, rhsMax) * 0.1)} and runtime ceil(min(lhsSize, rhsSize) * 0.1), using the new Size() method that returns 4 for v4 and 16 for v6. 

Net effect: estimate Max goes from 1 to 2 for all IP/CIDR equality; runtime stays 1 for v4 but becomes 2 for v6.

TristonianJones:
If it's at all possible, I would prefer not to change the way equality estimates cost. If it's a breaking change, we can override it within the networking lib and mark it as something we'll remove in future libraries ... I'm not sure how best you want to approach this, to be honest.

jpbetz:
I think we should just downstream the breaking change into k8s.  TBH, I don't think anyone will notice given that it's only for equality for these relatively new types.

dmitriplotnikov (dismissed):

dmitriplotnikov (approved):

PR #1350, opened 2026-06-22
Ensure receiver and global matches cost estimates agree

TristonianJones:
Ensure receiver and global matches cost estimates agree, i.e. `cost(matches(input, regex)) == cost(input.matches(regex))`

### DIFF
--- a/checker/cost.go
+++ b/checker/cost.go
@@ -791,18 +791,26 @@ func (c *coster) functionCost(e ast.Expr, function, overloadID string, target *A
 			return CallEstimate{CostEstimate: c.sizeOrUnknown(args[1]).MultiplyByCostFactor(1).Add(argCostSum())}
 		}
 	// O(nm) functions
-	case overloads.MatchesString:
+	case overloads.Matches, overloads.MatchesString:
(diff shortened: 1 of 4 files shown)


checker/cost.go
```diff
@@ -791,18 +791,26 @@ func (c *coster) functionCost(e ast.Expr, function, overloadID string, target *A
...
 	// O(nm) functions
-	case overloads.MatchesString:
+	case overloads.Matches, overloads.MatchesString:
 		// https://swtch.com/~rsc/regexp/regexp1.html applies to RE2 implementation supported by CEL
-		if target != nil && len(args) == 1 {
+		var strNode, regexNode AstNode
+		if overloadID == overloads.MatchesString && target != nil && len(args) == 1 {
+			strNode = *target
+			regexNode = args[0]
+		} else if overloadID == overloads.Matches && len(args) == 2 {
```
jcking: Do we need `target == nil` here?

TristonianJones: Fair point ... added.

jcking (approved):

PR #1352, opened 2026-06-22
Update lists library cost computations for improved accuracy

TristonianJones:
Provide better estimates for distinct / sort / flatten calls using the item hint structure expected elsewhere within the cost
estimation library. 

These changes are introduced in v4 of the `ext/lists.go` library.

### DIFF
--- a/ext/lists.go
+++ b/ext/lists.go
@@ -349,13 +349,15 @@ func (lib listsLib) CompileOptions() []cel.EnvOption {
 	if lib.version >= 3 {
 		estimators := []checker.CostOption{
 			checker.OverloadCostEstimate("list_slice", estimateListSlice),
-			checker.OverloadCostEstimate("list_flatten", estimateListFlatten),
-			checker.OverloadCostEstimate("list_flatten_int", estimateListFlatten),
 			checker.OverloadCostEstimate("lists_range", estimateListsRange),
 			checker.OverloadCostEstimate("list_reverse", estimateListReverse),
 		}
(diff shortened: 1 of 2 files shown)


dmitriplotnikov (approved):

ext/lists.go
```diff
@@ -698,37 +709,53 @@ func estimateListSortBy(u *types.Type) checker.FunctionEstimator {
...
+	sz := estimateSize(estimator, keysNode)
+	itemSize := estimateItemSize(estimator, target)
+	elemCost := estimateElementEqualityCost(estimator, elemType, itemSize)
+
+	costSize := sz.Multiply(sz)
+	cost := costSize.MultiplyByCost(elemCost).MultiplyByCostFactor(2.0)
+	return estimateListCallWithDirectCost(cost, sz, true)
+}
+
 // estimateListSortCost estimates an O(n^2) sort operation with a cost factor of 2 for the equality
```
dmitriplotnikov, line 727: Out of curiosity: how come sort is O(n^2) rather than O(n*log(n))?

dmitriplotnikov (commented):
Terrific!

ext/lists.go
TristonianJones: I think I estimated a naive sort cost of O(n^2). It looks like `sort.Slice` has a O(n log n) cost, but with a high constant factor since it uses reflection heavily. If I update to `slices.Sort`, then I can probably lower the cost estimate

PR #1354, opened 2026-06-22
Add cost tracking for encoders

TristonianJones:
Introduce cost tracking for `base64.encode` and `base64.decode`

The cost tracking support is introduced at v1 of the library.


ext/encoders_test.go
```diff
@@ -93,3 +94,128 @@ func TestEncodersVersion(t *testing.T) {
...
+		cel.EnableMacroCallTracking(),
+	}
+	env, err := cel.NewEnv(append(baseOpts, opts...)...)
+	if err != nil {
+		t.Fatalf("cel.NewEnv(Encoders()) failed: %v", err)
+	}
+	return env
+}
+
+func TestEncodersCosts(t *testing.T) {
```
maskri17, line 128: should there be an empty str case: "base64.encode(x) == b''"

maskri17 (dismissed):

ext/encoders_test.go
TristonianJones: I added a few more tests

maskri17 (dismissed):

maskri17 (approved):

PR #1355, opened 2026-06-22
Async function execution internals

TristonianJones:
Internal support for async function evaluation.

* Declaration support with guard functions just like sync functions
* The framework manages async function launches according to configurable concurrency limits
* Async functions are provided the `context.Context` object and expected to block
  until a result is available and return it. The framework manages go routines

The implementation supports minimal call deduplication logic which users
will be able to extend with functionality introduced into later PRs.

### DIFF
--- a/interpreter/async.go
+++ b/interpreter/async.go
@@ -102,6 +102,8 @@ func (fn *evalAsyncFunc) Exec(frame *ExecutionFrame) ref.Val {
 	// Early return if any argument to the function is unknown or error.
 	for i, arg := range fn.args {
 		argVals[i] = arg.Exec(frame)
+		// TODO: early return only on errors, aggregate unknowns and validate
+		// whether any argument is unknown before proceeding with the call.
 		if types.IsUnknownOrError(argVals[i]) {
 			return argVals[i]
 		}
@@ -113,18 +115,11 @@ func (fn *evalAsyncFunc) Exec(frame *ExecutionFrame) ref.Val {
 // asyncCallStateTracker manages async call states across re-evaluations of a single program.
 type asyncCallStateTracker struct {
 	mu sync.RWMutex
-	// calls buckets call states by a composite hash of (node id, overload, string/bool args).
+	// calls buckets call states by a composite hash of (node id, overload, string/int/double/uint/bool args).
 	// A single AST node id may host many concurrently-live calls when it is evaluated inside a
 	// comprehension (once per element with different arguments), so each bucket may hold more
 	// than one state. The exact match within a bucket is resolved via asyncCallState.equals,
 	// which applies CEL's full equality semantics to the arguments.
-	//
-	// Caveat for future implementers: dedup relies on argument equality being reflexive. CEL
-	// values whose equality is not reflexive — notably NaN doubles, where NaN != NaN — will never
-	// match a previously registered state. Such a call is therefore relaunched on every
-	// re-evaluation pass and never converges (the evaluation terminates only via context
-	// cancellation/deadline). This is inherent to the CEL equality model and is not special-cased
-	// here; if it ever needs handling, give equals an identity-based fast path before equality.
 	calls        map[uint64][]*asyncCallState
 	callsByID    map[int64]*asyncCallState
 	nextCallID   atomic.Int64
@@ -149,10 +144,10 @@ var (
 
 // hashCall computes the composite bucket key for an async call.
 //
-// Only string and bool argument values contribute to the hash. Numeric and more complex types
-// rely on a richer notion of equivalence (e.g. cross-type numeric equality, unordered maps) that
-// a byte-level hash cannot capture safely, so they are deliberately excluded from the key and are
-// instead disambiguated within the bucket by asyncCallState.equals.
+// Only string, int, double, uint, and bool argument values contribute to the hash. More complex types
+// rely on a richer notion of equivalence (e.g. unordered maps, proto equality, custom types)
+// that a byte-level hash cannot capture safely, so they are intentionally excluded from the key
+// and are instead disambiguated within the bucket by asyncCallState.equals.
 func hashCall(id int64, overload string, args []ref.Val) uint64 {
 	h := fnv.New64a()
 	var idBuf [8]byte
@@ -183,6 +178,15 @@ func hashCall(id int64, overload string, args []ref.Val) uint64 {
 			h.Write(buf[:])
 		case types.Double:
 			h.Write(hashNumberMarker)
+			if math.IsNaN(float64(v)) {
+				h.Write([]byte("NaN"))
+				continue
+			}
+			// Normalize -0.0 to 0.0. Go will treat -0.0 as 0.0 at compile time,
+			// but the function math.Copysign(0.0, -1.0) can be used to test the -0.0 case.
+			if v == types.Double(0.0) && math.Signbit(float64(v)) {
+				v = types.Double(0.0)
+			}
 			var buf [8]byte
 			binary.LittleEndian.PutUint64(buf[:], math.Float64bits(float64(v)))
 			h.Write(buf[:])
@@ -350,6 +354,21 @@ func (t *asyncCallStateTracker) launch(ctx context.Context, acs *asyncCallState,
 			// Release the concurrency slot when this call finishes or is cancelled.
 			defer func() { <-semaphore }()
 		}
+		if acs.completions != nil {
+			defer func() {
+				// Prioritize context cancellation to prevent racy completion signals.
+				if ctx.Err() != nil {
+					return
+				}
+				// Signal completion to the evaluator. Use a select with ctx.Done() to prevent
+				// the goroutine from leaking if the evaluator has already timed out or aborted.
+				select {
+				case acs.completions <- acs.callID:
+				case <-ctx.Done():
+				}
+			}()
+		}
+
 		ch := acs.impl(ctx, acs.argVals...)
 		// Early terminate with a CEL error when an implementation returns an empty channel.
 		if ch == nil {
@@ -359,7 +378,6 @@ func (t *asyncCallStateTracker) launch(ctx context.Context, acs *asyncCallState,
 				fmt.Sprintf("function %s returned an empty channel", acs.function))
 			return
 		}
-
 		// Wait for the async computation to finish or for the context to be cancelled.
 		select {
 		case r := <-ch:
@@ -368,14 +386,6 @@ func (t *asyncCallStateTracker) launch(ctx context.Context, acs *asyncCallState,
 			acs.mu.Unlock()
 			if observer != nil {
 				observer.OnCallFinished(acs.callID, acs.function, acs.overload, r)
-			}
-			if acs.completions != nil {
-				// Signal completion to the evaluator. Use a select with ctx.Done() to prevent
-				// the goroutine from leaking if the evaluator has already timed out or aborted.
-				select {
-				case acs.completions <- acs.callID:
-				case <-ctx.Done():
-				}
 			}
 		case <-ctx.Done():
 			// Evaluation context cancelled before the async operation completed.
--- a/interpreter/async.go
+++ b/interpreter/async.go
@@ -1,0 +1,428 @@
+// Copyright 2026 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package interpreter
+
+import (
+	"context"
+	"encoding/binary"
+	"hash/fnv"
+	"sync"
+	"sync/atomic"
+
+	"github.com/google/cel-go/common/functions"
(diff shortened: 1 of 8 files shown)


interpreter/async.go
```diff
@@ -0,0 +1,428 @@
...
+	}
+	go func() {
+		if semaphore != nil {
+			// Release the concurrency slot when this call finishes or is cancelled.
+			defer func() { <-semaphore }()
+		}
+		ch := acs.impl(ctx, acs.argVals...)
+		// Wait for the async computation to finish or for the context to be cancelled.
+		select {
+		case r := <-ch:
```
l46kok, line 349: We should probably check for the channel being nil, or the channel being closed without a value (or the channel itself sends a nil).

interpreter/async.go
```diff
@@ -0,0 +1,428 @@
...
+	hashZeroMarker      = []byte{0}
+	hashStringMarker    = []byte{'s'}
+	hashBoolTrueMarker  = []byte{'b', 1}
+	hashBoolFalseMarker = []byte{'b', 0}
+	hashDefaultMarker   = []byte{'x'}
+)
+
+// hashCall computes the composite bucket key for an async call.
+//
+// Only string and bool argument values contribute to the hash. Numeric and more complex types
```
l46kok: Instead of excluding numbers, wdyt about normalizing to`float64`? 

The concern I have is for an expression like `[1, 2, ..., 1000].map(x, asyncFn(x))`, everything gets hashed into the same bucket, and it's a O(N^2) pass given that `findInBucket` is a linear scan. 

Ideally, we should handle things like numbers beyond 53 bit mantissa, lists/maps too, but those are rare enough that we could likely defer.

interpreter/async_test.go
```diff
@@ -0,0 +1,900 @@
...
+		}
+	}
+}
+
+func TestAsyncSetupWithoutContextErrors(t *testing.T) {
+	frame, err := NewExecutionFrame(EmptyActivation())
+	if err != nil {
+		t.Fatalf("NewExecutionFrame() failed: %v", err)
+	}
+	defer frame.Close()
```
l46kok, line 1007: nit: this is unlikely, but perhaps consider adding a simple guard for double close in ExecutionFrame's Close() function (something like `if.Activation == nil { return }` should make it idempotent)

interpreter/async.go
```diff
@@ -0,0 +1,428 @@
...
+	Overload() string
+}
+
+// evalAsyncFunc is the planned Interpretable for an asynchronous function call.
+type evalAsyncFunc struct {
+	id       int64
+	function string
+	overload string
+	args     []InterpretableV2
+	impl     functions.AsyncOp
```
l46kok, line 71: Question: what happens if the user provided implementation panics? Do we ordinarily recover gracefully?

TristonianJones: That's a good test case. Currently, I think the whole expression goes into a `recover` just like the sync functions, but it would be an interesting refinement to consider for all extensions in the future. I'll open an issue to track

interpreter/async.go
TristonianJones: I totally hear you. I was wondering about that as well, but wanted to start with a case that was unambiguous ... normalizing to `float64` is very reasonable though since it's just a hash. I'll add it.

interpreter/async.go
TristonianJones: Since the `impl` is always a wrapper over what the customer provides, CEL allocates the channel internally. I can add a check, but it should never be `nil`

interpreter/async.go
TristonianJones: Done, and the hashing looks good when normalized, number-line equivalent values get hashed the same.

interpreter/async.go
TristonianJones: Tracked in: https://github.com/cel-expr/cel-go/issues/1356

TristonianJones:
PTAL, thanks for the detailed review. I also added a NaN case to the async function call equality check.

interpreter/async.go
```diff
@@ -0,0 +1,463 @@
...
+			defer func() { <-semaphore }()
+		}
+		ch := acs.impl(ctx, acs.argVals...)
+		// Early terminate with a CEL error when an implementation returns an empty channel.
+		if ch == nil {
+			acs.mu.Lock()
+			defer acs.mu.Unlock()
+			acs.result = types.NewErrFromString(
+				fmt.Sprintf("function %s returned an empty channel", acs.function))
+			return
```
l46kok, line 345: Do we also need to signal the completions channel here? Otherwise, `pendingCalls` will never be decremented

interpreter/async.go
```diff
@@ -0,0 +1,463 @@
...
+
+// Exec implements the InterpretableV2 interface method.
+func (fn *evalAsyncFunc) Exec(frame *ExecutionFrame) ref.Val {
+	argVals := make([]ref.Val, len(fn.args))
+	// Early return if any argument to the function is unknown or error.
+	for i, arg := range fn.args {
+		argVals[i] = arg.Exec(frame)
+		if types.IsUnknownOrError(argVals[i]) {
+			return argVals[i]
+		}
```
l46kok, line 109: Just a food for thought (I don't think we should make this change now), but do you envision a future enhancement where we add a configuration to accumulate unknowns for maximum concurrency? 

I think this only matters for `strict_func(asyncA(), asyncB())`, so we can definitely defer until we see a concrete pattern or a request on this.

interpreter/async.go
```diff
@@ -0,0 +1,463 @@
...
+	acs.mu.Unlock()
+	t.pendingCalls.Add(1)
+
+	if observer != nil {
+		observer.OnCallStarted(acs.callID, acs.function, acs.overload, acs.argVals)
+	}
+	go func() {
+		if semaphore != nil {
+			// Release the concurrency slot when this call finishes or is cancelled.
+			defer func() { <-semaphore }()
```
l46kok: Do we need to also release the semaphore right before sending to the completion channel? defer runs at the very end of the goroutine right? If the semaphore is at capacity, there might be a small window where the next async call is rejected because defer hasn't ran.

interpreter/async.go
TristonianJones: That's a fair point. We tend to assume that functions are strict and thus shouldn't accept unknown (definitely the assumption I'm making for all async functions).

I think here you'd be doing something like `asyncCall(asyncFirst(), asyncSecond())` Generally, there's some expectation that we can't ever issue `asyncCall` if any of the arguments is unknown ... I can see a case for accumulating and just early returning on error though. I've added a TODO to come back to it.

interpreter/async.go
```diff
@@ -0,0 +1,463 @@
...
+			h.Write(buf[:])
+		case types.Uint:
+			h.Write(hashNumberMarker)
+			var buf [8]byte
+			binary.LittleEndian.PutUint64(buf[:], math.Float64bits(float64(v)))
+			h.Write(buf[:])
+		case types.Double:
+			h.Write(hashNumberMarker)
+			var buf [8]byte
+			binary.LittleEndian.PutUint64(buf[:], math.Float64bits(float64(v)))
```
l46kok, line 191: For doubles -- I think we'll want to normalize `0.0` and `-0.0` explicitly, just so that they aren't hashed to different buckets. And probably `NaN` as well, in case if we have NaN values with differing bit representations

interpreter/async.go
```diff
@@ -0,0 +1,463 @@
...
+type asyncCallStateTracker struct {
+	mu sync.RWMutex
+	// calls buckets call states by a composite hash of (node id, overload, string/bool args).
+	// A single AST node id may host many concurrently-live calls when it is evaluated inside a
+	// comprehension (once per element with different arguments), so each bucket may hold more
+	// than one state. The exact match within a bucket is resolved via asyncCallState.equals,
+	// which applies CEL's full equality semantics to the arguments.
+	//
+	// Caveat for future implementers: dedup relies on argument equality being reflexive. CEL
+	// values whose equality is not reflexive — notably NaN doubles, where NaN != NaN — will never
```
l46kok: Now that we're handling the NaNs in the equality check, we should update this comment.

interpreter/async.go
l46kok: So fun fact -- legacy Java async interpreter does indeed evaluate all arguments in parallel by default (it of course, does not inspect the value being unknown or whatsoever before issuing the call), so at least there's one precedent for it. Imo, we shouldn't follow that model blindly for CEL semantics adherence, but at least something to reference. 

(definitely meant non_strict_func on above, my mistake)

TristonianJones: It's a larger issue if we're talking about non-strict, but the lesson still applies. I'll have to update a few different blocks of code to aggregate unknowns and permit that behavior -- I think it mostly applies to the strict functions, but non-strict could benefit too.

interpreter/async.go
TristonianJones: Yes, I'm going to move the completions up to a deferred call after the semaphore release.

interpreter/async.go
TristonianJones: I don't think we do. The completions signal indicates when a result is ready for re-evaluation (this will be more obvious in later PRs). The idea is that a batch of results are being accumulated and some logic is trying to decide whether to wait for all pending completions, a certain number of results, or a certain amount of time. The semaphore release at the end is, I think, the right thing to do since it should block new calls until the decision about whether re-evaluate is made.

TristonianJones:
PTAL and let me know what you think regarding the last open comment about the semaphore

interpreter/async.go
l46kok: With the semaphore release at the end, here's the scenario I'm wondering about, let me know if this sounds realistic to you. Suppose we have `asyncA() + asyncB()`, with semaphore capacity of 1. Rough timeline:

First evaluation pass:
- asyncA() starts, acquires the semaphore. 
- asyncB() fails to acquire it then backs off. Evaluator now waits for completions channel. 
- asyncA() finishes. completions <- stateA is notified. Evaluator gets woken up. Note that semaphore isn't released at this time.

Second evaluation pass: 
- asyncA() is now cached so it's skipped, asyncB() starts, attempts to acquire the semaphore
- asyncA()'s goroutine hasn't executed the `defer <- semaphore` yet. asyncB() fails to acquire the slot.
- asyncA() gets done, asyncB() was rejected. The pass yields an unknown. At this point, `frame.PendingAsyncCalls == 0`. The evaluator prematurely aborts the evaluation.

TristonianJones: Actually, the behavior was even worse than you thought-- turns out it would evaluate twice and hang. Yikes.

I made a mistake in how I thought about the flow. I introduced an `asyncGate` to `async.go` that groups the concepts together - acquire the semaphore, increment a count. complete the task, release the semaphore, decrement a count, notify on `completions` with the call id. Seems much cleaner now, but I may still have something off.

interpreter/async.go
```diff
@@ -0,0 +1,512 @@
...
+		select {
+		case r := <-ch:
+			acs.mu.Lock()
+			acs.result = r
+			acs.mu.Unlock()
+			if observer != nil {
+				observer.OnCallFinished(acs.callID, acs.function, acs.overload, r)
+			}
+		case <-ctx.Done():
+			// Evaluation context cancelled before the async operation completed.
```
l46kok, line 354: I think you may want to also clean up and notify the observer for the cancellation case here (might be easiest to consolidate the cleanup logic into a `defer`)

interpreter/async.go
```diff
@@ -0,0 +1,512 @@
...
+		case types.Uint:
+			h.Write(hashNumberMarker)
+			var buf [8]byte
+			binary.LittleEndian.PutUint64(buf[:], math.Float64bits(float64(v)))
+			h.Write(buf[:])
+		case types.Double:
+			h.Write(hashNumberMarker)
+			if math.IsNaN(float64(v)) {
+				h.Write([]byte("NaN"))
+				continue
```
l46kok, line 183: Did you mean to break here? We'll be skipping the `hashZeroMarker` write below.

interpreter/async.go
l46kok: This is now tracked in asyncGate, so we should be able to remove this

interpreter/async.go
```diff
@@ -0,0 +1,512 @@
...
+		h.Write(hashZeroMarker)
+	}
+	return h.Sum64()
+}
+
+// findInBucket returns the call state in the bucket matching the same node id and call identity,
+// or nil if no match is present.
+func findInBucket(bucket []*asyncCallState, id int64, want *asyncCallState) *asyncCallState {
+	for _, acs := range bucket {
+		if acs.id == id && acs.equals(want) {
```
l46kok: optional: you could consider just taking in the raw variables for the signature match here (e.g: `func findInBucket(bucket []*asyncCallState, id int64, function, overload string, argVals []ref.Val)`. For comprehensions, allocating `asyncCallState` struct on every single loop iteration just for equality check seems a bit expensive.

Do feel free to just ack if this feels like a premature optimization, this does pollute the arguments a bit. Readability wise, I prefer what's currently here.

interpreter/async.go
TristonianJones: Quite right. Fixed.

interpreter/async.go
TristonianJones: Removed.

interpreter/async.go
TristonianJones: The readability isn't substantially degraded with the change. So I've updated to your suggestion to avoid added allocations.

interpreter/async.go
TristonianJones: Usually context cancellation should just halt execution and release resources (assuming all parties are listening for cancellation correctly), but I'll take a look. Are you saying we should add a cancelled observation method, perhaps?

l46kok (approved):

PR #1357, opened 2026-06-26
avoid repeated construction of cost tracker

pohly:
Most fields in the CostTracker struct are only set once during construction. This takes a surprising amount of allocations, in particular for the overload map. Constructing the tracker once and cloning it for each evaluation avoids doing that work repeatedly.

In a Kubernetes benchmark exercising (among other things) CEL expression evaluation, the cumulative CPU time and allocs for Eval dropped from 3.87s to 1.99s resp. from 3.85G to 2.1G.

### DIFF
--- a/cel/cel_test.go
+++ b/cel/cel_test.go
@@ -1969,6 +1969,48 @@ func TestCostLimit(t *testing.T) {
 	}
 }
 
+func TestCostTrackingConsistentAcrossEvals(t *testing.T) {
+	env := testEnv(t,
+		Variable("val1", IntType),
+		Variable("val2", IntType),
+	)
+	ast, iss := env.Compile(`val1 + val2`)
+	if iss.Err() != nil {
+		t.Fatalf("env.Compile() failed: %v", iss.Err())
+	}
+	checkedAst, iss := env.Check(ast)
+	if iss.Err() != nil {
+		t.Fatalf("env.Check() failed: %v", iss.Err())
+	}
+	program, err := env.Program(checkedAst, CostTracking(nil))
+	if err != nil {
+		t.Fatalf("env.Program() failed: %v", err)
+	}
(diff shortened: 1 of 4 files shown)


TristonianJones (dismissed):

TristonianJones:
@pohly Thanks for this change. Would you mind adding a test case that initiates initiates the cost tracking against a single compiled expression in a short loop to help prevent future regressions?

pohly:
How can such a test verify that Eval doesn't call NewCostTracker again? Just calling Eval multiple times will not fail, with or without this patch.

TristonianJones:
It's purely defensive ... more that if something changes in the cost tracker in the future to make sure we don't miss the updates to `Clone`. Currently the cost tests exercise single evaluation which probably won't hit the `Clone` case.

pohly:
A single Eval always uses NewCostTracker (during program construction) followed by Clone (when the factory function is called). But it doesn't hurt to do Eval more than once to cover the scenario that Clone is called more than once - will add something.

pohly:
@TristonianJones: see https://github.com/cel-expr/cel-go/compare/c0b1db8e505326cc8463b8fed381c116a5ff9de1..164e53d3c0baad4da72f0df5400ef1526ffcc6a7 for unit tests.

`computeCost` in `interpreter` now exercises the `NewCostTracker -> Clone` code path, to detect early if Clone breaks. `TestCostTrackingConsistentAcrossEvals` in `cel` checks that repeated evaluation using the same program yields consistent costs.

pohly:
@TristonianJones: gentle ping. Is there a chance to get this into a tagged release soonish? Code freeze deadline for Kubernetes is July 22nd, it would be great to bump cel-go there well in advance.

TristonianJones (approved):

PR #1364, opened 2026-07-15
Introduce async DrainStrategy, DrainAction, Observer, and Call types

TristonianJones:
Create `cel/async` package with interfaces used by concurrent eval

* `async.Call` is a read-only view of a pending or completed async operation
* `async.Observer` can be used to receive start / stop notification for an `async.Call`
* `async.DrainStrategy` controls how and when to consume async call results in an evaluation
* `async.DrainAction` indicates how the concurrent evaluation should treat the batch of results

`ConcurrentEval` will initiate a CEL expression evaluation that triggers one or more async
calls. As calls complete, the expression evaluation will be performed again using the results
from the call to advance the expression evaluation to a conclusive state.

### DIFF
--- a/cel/async/BUILD.bazel
+++ b/cel/async/BUILD.bazel
@@ -1,0 +1,27 @@
+load("@io_bazel_rules_go//go:def.bzl", "go_library", "go_test")
+
+package(
+    licenses = ["notice"],  # Apache 2.0
+)
+
+go_library(
+    name = "go_default_library",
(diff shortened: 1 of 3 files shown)


TristonianJones:
Test issue is unrelated, see: https://github.com/cel-expr/cel-go/pull/1365 for the fix.

jnthntatum (approved):

PR #1369, opened 2026-07-17
Concurrent Eval with context.Context bound functions

### LINKED ISSUE #557: Support context propagation on overloads

TristonianJones:
The `ConcurrentEval` function for CEL enables you to connect asynchronous extensions
that respect context cancellation and deadlines.

**Execution Frame as Context**

The features is built upon the CEL-managed `ExecutionFrame` which serves the role of a
custom `context.Context` object for the duration of a single `ConcurrentEval` execution. 
The `ExecutionFrame` must never be stored or used outside of CEL: it's purely internal

**Async Blocking**

Async extensions are expected to block with CEL managing channel creation / concurrency
limits which can be configured through top-level options. Over time, CEL will add more utilities
for customizing extension timeouts, retries, and caching, e.g. #1367 

**Async Pruning**

This implementation uses CEL's unknown feature to prune async dispatches before they're
executed if they're not relevant to the result.

**Fixes for Unknown Merging**

All code paths now properly early return on error, but aggregate unknowns, allowing for better
constant folding and expression pruning in addition to better identification of the batches of
calls which are relevant to result computation.

Closes #557


interpreter/async.go
```diff
@@ -330,31 +348,25 @@ func (t *asyncCallStateTracker) launch(ctx context.Context, acs *asyncCallState,
...
-			defer acs.mu.Unlock()
-			acs.result = types.NewErrFromString(
-				fmt.Sprintf("function %s returned an empty channel", acs.function))
+			acs.SetResult(types.NewErrFromString(
+				fmt.Sprintf("function %s returned an empty channel", acs.function)))
 			return
 		}
 		// Wait for the async computation to finish or for the context to be cancelled.
 		select {
 		case r := <-ch:
```
l46kok: Should we check channel being empty? Currently, wrapAsyncOp guarantees creation of a buffered channel, so I realize it's impossible to exercise, but the following seems like a pretty cheap way to future proof. Up to you.

```go
case r, ok := <-ch:
   if !ok // ... empty channel
```

interpreter/async.go
```diff
@@ -330,31 +348,25 @@ func (t *asyncCallStateTracker) launch(ctx context.Context, acs *asyncCallState,
 	go func() {
 		defer func() {
 			if observer != nil {
-				observer.OnCallFinished(acs.callID, acs.function, acs.overload, acs.result)
+				observer.OnCallFinished(acs.callID, acs.function, acs.overload, acs.ResultOrUnknown())
 			}
 			gate.Complete(ctx, acs.callID)
 		}()
 
 		ch := acs.impl(ctx, acs.argVals...)
```
l46kok, line 356: Given that we are gracefully recovering from panics for [sync evals](https://github.com/cel-expr/cel-go/blob/master/cel/program.go#L313), would you want to handle it the same here? If anything for consistency reasons (unless if there are anything that makes it structurally unsafe to continue with CEL evaluation, but potentially crashing the whole process seems undesirable).

interpreter/prune_test.go
```diff
@@ -367,17 +367,17 @@ var testCases = []testInfo{
 	{
 		in:   unknownActivation(),
 		expr: `test == {'a': 1, 'field': 2}.field`,
-		out:  `test == 2`,
```
l46kok, line 370: Did we lose pruning capability on some of these test cases? It feels like a regression.

cel/folding.go
```diff
@@ -159,12 +160,14 @@ func (opt *constantFoldingOptimizer) tryFold(ctx *OptimizerContext, a *ast.AST,
 	if activation == nil {
 		activation = NoVars()
 	}
-	out, _, err := prg.Eval(activation)
-	if err != nil {
-		return err
+	resCh := prg.ConcurrentEval(context.Background(), activation)
```
l46kok: This is a great enhancement -- Can we add a configuration for timeouts? (For things like async functions that can hang indefinitely)

cel/program.go
```diff
@@ -383,3 +434,183 @@ func (p *prog) newExecutionFrame(input any) (*interpreter.ExecutionFrame, error)
...
+	}
+	frame.SetAsyncObserver(p.asyncObserver)
+	frame.SetAsyncMaxConcurrency(resolveAsyncMaxConcurrency(p.asyncMaxConcurrency))
+	return frame, nil
+}
+
+// defaultAsyncMaxConcurrency bounds the number of concurrently launched async calls when the
+// program does not configure AsyncMaxConcurrency. It exists so that a wide fan-out (e.g. an async
+// call inside a comprehension over a large list) cannot spawn an unbounded number of goroutines.
+const defaultAsyncMaxConcurrency = 1000
```
l46kok: Question: What happens if we exhaust the concurrency limit, and more items than the specific limit is writing to `g.completions`? It looks like the new `DrainStrategy` forces re-evaluation, and looks at the same `g.completions` channel. Are we safely dropping the overflowing items (via non-blocking send?)

It might make sense to stress test this in some way, ex: `items.map(x, async_fn(x))` over 3k items or so. Just to ensure we won't leak goroutines anywhere

interpreter/prune_test.go
TristonianJones: I updated the way that errors are handled so errors will result in an early-return. In this case `test` is never actually declared, so it should error and not attempt any changes. I can add a comment to the effect if that makes it easier to reason about, and have the equivalent with an unknown variable show that the folds still work. Thoughts?

cel/folding.go
TristonianJones: I actually mark all Async functions as late-bindings so they never get called; however, the presence of an async function will early terminate anything that isn't `ConcurrentEval`. I can add a comment to this effect.

cel/program.go
TristonianJones: I believe the overflowing items never send now since they have to be initiated via the `DispatchPendingAsyncCalls` method that happens post-evaluation (I've switched from eager to lazy dispatch). I'll add a test to stress this case though, limit of 10 concurrent calls, 1k calls that wait 1ms each. Does that sound like a good match to what you'd like to see?

interpreter/prune_test.go
l46kok: Ah makes perfect sense. Yeah I also agree that this shouldn't have been pruning in the first place, and the fix is trivial at the callsite. Documenting the effect SGTM

cel/folding.go
l46kok: Interesting, could you elaborate on the purpose of this change? I feel like just using `Eval` is effectively faster to avoid the extra goroutine overhead, unless if you want to introduce compile time async folding later on.

cel/program.go
l46kok: Yeah that SGTM. Effectively, what I'm worried about is getting stuck here: https://github.com/TristonianJones/cel-go/blob/c92a515d897e8ad72f15160dee86a42288126c98/interpreter/async.go#L512-L516. I'm wondering if there is a scenario where if we release all slots, the completions channel is not notified before the next round of eval from ConcurrentEval launches the next 1k goroutine.

cel/folding.go
TristonianJones: It's more that if someone adds a reference to an async function, that `Eval` won't attempt to run. It's a perf hit, but hopefully minor.

cel/program.go
TristonianJones: The goroutines aren't started until after the `ConcurrentEval` yields, so I don't think it's possible to have an empty completions channel since the completions would buffer until they're consumed by `ConcurrentEval`. Does that help answer your question?

interpreter/async.go
TristonianJones: Done.

l46kok (dismissed):

l46kok (approved):

PR #1371, opened 2026-07-18
Fix constant folding and local variable resolution

### LINKED ISSUE #1296: Constant folding fails for comprehensions containing a shadowed variable

TristonianJones:
Align constant folding and attribute resolution behavior with cel-java, fixing local variable pattern matching during partial evaluation.

* Updated `matchesUnknownPatterns` to check whether a variable is locally bound (e.g. comprehension iterator/accumulator variables `i`, `accuVar`) so local variables are not incorrectly flagged as unknown global attributes during partial evaluation.
* Updated folding to skip attempting computations over local variables within a comprehension.
* Added support for partial evaluation over unknown values to fold properly with Unknown rather than error.
* Added `in` optimizations to cover element or identity equality even when the variable is unknown.

Closes #1296

### DIFF
--- a/cel/folding.go
+++ b/cel/folding.go
@@ -17,7 +17,6 @@ package cel
 import (
 	"errors"
 	"fmt"
-	"strings"
 
 	"github.com/google/cel-go/common/ast"
 	"github.com/google/cel-go/common/operators"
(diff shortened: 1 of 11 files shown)


cel/folding.go
l46kok, line 536: Do we also need to check for IterVar2 here?

TristonianJones: Updated. Thanks for flagging this. Added more tests to capture two-var comprehension edge.

l46kok (approved):

PR #1374, opened 2026-07-21
Fix unsound constant folding of `x in [x]` for NaN values

pranit0808:
**Constant folding rewrites `x in [x]` to `true`**

Matching the needle against a list element by identifier name assumes the value is equal to itself, which does not hold for a NaN double: with `x` bound to NaN, `x in [1, 2, x]` evaluates to `false` before optimization and `true` after. The identifier match is now limited to the scalar types that cannot carry a NaN, so dyn, double, and aggregate needles keep their runtime behavior.

### DIFF
--- a/cel/folding.go
+++ b/cel/folding.go
@@ -624,8 +624,9 @@ func constantCallMatcher(e ast.NavigableExpr) bool {
 //
 // Matching an identifier against a list element by name only proves list membership when the
 // value the name resolves to is self-equal. A double may be NaN, which is not equal to itself,
-// and the dyn, aggregate, and abstract types may all hold a NaN at runtime, so the check is
-// limited to the scalar types which cannot.
+// and dyn, abstract, and struct types may all hold a NaN at runtime, so the check is limited
+// to the scalar types which cannot, and to the aggregate types whose type parameters are
+// themselves self-equal.
 func isSelfEqualIdent(e ast.Expr) bool {
 	if e.Kind() != ast.IdentKind {
 		return false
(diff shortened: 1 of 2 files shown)


cel/folding.go
```diff
@@ -619,6 +619,34 @@ func constantCallMatcher(e ast.NavigableExpr) bool {
...
+	nav, ok := e.(ast.NavigableExpr)
+	if !ok {
+		return false
+	}
+	t := nav.Type()
+	if t == nil {
+		return false
+	}
+	switch t.Kind() {
+	case types.BoolKind, types.BytesKind, types.DurationKind, types.IntKind,
```
TristonianJones, line 647: Would you mind adding `types.TypeType` as well?

cel/folding_test.go
```diff
@@ -378,6 +379,10 @@ func TestConstantFoldingOptimizer(t *testing.T) {
 		},
 		{
 			expr:   `x in [1, 2, x]`,
+			folded: `x in [1, 2, x]`,
```
TristonianJones, line 382: Please add some additional tests for the other types in the list. You might want to carve this out into its own table-driven test to demonstrate why the behavior is structured like it is. You should also be able to tell whether aggregate types like `list` and `map` have consistently comparable type parameters. It's a lesser use case, for sure, but it's worth considering. 

Structs are possible to consider, but require more work than it may be worth.

cel/folding.go
pranit0808: Added, along with list and map: those compare element-wise, so they're self-equal exactly when their type parameters are. list(int) and map(string, int) fold now, list(dyn)/list(double)/map(string, double) don't. Left structs and abstract types out as you suggested.

cel/folding_test.go
pranit0808: Carved it out into TestConstantFoldingInListIdent with a row per type: the self-equal scalars, the aggregate cases on both sides of the line, double/dyn, and struct/abstract needles. The rows that don't fold also evaluate the optimized program with a NaN binding so it's clear what the fold would have changed. Kept a couple of literal-needle rows too since those go through real CEL equality and are unaffected.

cel/folding.go
```diff
@@ -619,6 +619,49 @@ func constantCallMatcher(e ast.NavigableExpr) bool {
...
 
+// isSelfEqualIdent indicates whether the expression is an identifier whose static type
+// guarantees that its runtime value is equal to itself.
+//
+// Matching an identifier against a list element by name only proves list membership when the
+// value the name resolves to is self-equal. A double may be NaN, which is not equal to itself,
+// and dyn, abstract, and struct types may all hold a NaN at runtime, so the check is limited
+// to the scalar types which cannot, and to the aggregate types whose type parameters are
+// themselves self-equal.
+func isSelfEqualIdent(e ast.Expr) bool {
```
TristonianJones, line 630: I would just pass e ast.NavigableExpr here rather than validate and return false later ... The input is guaranteed to be navigable based on the call path.

TristonianJones (approved):

PR #1383, opened 2026-07-23
Regex program plan size controls

### LINKED ISSUE #1142: Support option to limit RE2 program size
**Feature request checklist**

### Change
To make CEL environment setup consistent across CEL implementations, I propose to add a program option to set maximum RE2 program. This option should work similar to `CelOptions#maxRegexProgramSize(int)` in CEL-Java and `InterpreterOptions.regex_max_program_size` in CEL-Cpp (see nuances below).

#### Implementation details
1. To support this option,  cel-go will have to switch its regexp implementation from the [regex](https://pkg.go.dev/regexp) package to the [regexp/syntax](https://pkg.go.dev/regexp/syntax) (both in standard go library). In latter, the instruction count is available as the length of [regexp/syntax\#Prog.Inst](https://pkg.go.dev/regexp/syntax#Prog.Inst).
2. RE2 program size should be verified when the CEL program is created from the AST (AFAIK this is [how cel-cpp works](https://github.com/google/cel-cpp/blob/e6ba30e29a51a2fec3e176e24789e6113e1cd2af/runtime/standard/regex_functions.cc)). If this is not possible, [java-style implementation](https://github.com/google/cel-java/pull/550) where the error is raised during program execution is acceptable as well.

#### Nuances
The program size represents a very approximate measure of a regexp's "cost". There are no guarantees on the implementation details or claims about the properties of the program size (except "larger numbers are more expensive than smaller numbers"). 

Currently the program size is the same as the number of instructions of the regex program. However, the number of instructions in a regex depends on the concrete RE2 implementation. The implication of using the number of instructions as the program size is that:

> [!IMPORTANT]
> There's no guarantee that RE2 program size has the exact same value in CPP, Go and Java.
> We should communicate this in the docs.

For example:
```py
["", "a", "^", "^$", "a+b", "a+b?", "(a+b)", "a+b.*", "(a+b?)"]  # pattern
[4,   5,   2,   2,     7,     8,       9,      15,       10]     # program size: cpp re2-2024-07-02
[3,   3,   3,   4,     5,     6,       7,       7,        8]     # program size: java re2j v1.8 
[3,   3,   3,   4,     5,     6,       7,       7,        8]     # program size: go v1.24.1
# (go v1.24.1 is identical to re2j v1.8)
```

### Example
```go
// Actual option names up to the implementer.
prg, err := env.Program(ast, cel.MaxRegexProgramSize(100))
```

### Related
* cel-java feature request: google/cel-java#545
* cel-java implementation: google/cel-java#550, cl/715048086
* Internal ref: go/grpc-cel-integration

TristonianJones:
This implementation mirrors the logic used in RE2 within cel-java and cel-cpp. Regex implementations across platforms do not guarantee equivalent program plan sizes, but this at least provides a means to control the plan size per-platform.

Closes https://github.com/cel-expr/cel-go/issues/1142


jnthntatum (dismissed):

TristonianJones:
@jnthntatum PTAL, I needed to resolve a merge conflict.

jnthntatum (approved):

PR #1384, opened 2026-07-24
Respect composed adapter for unregistered structs

### LINKED ISSUE #1343: NativeTypes doesn't use baseAdapter when it should
**Describe the bug**

**To Reproduce**

- [ ]  parser
- [ ]  checker
- [x]  interpreter

```cel
x == "abc"
```

Test setup:
```go

//
// setup [ref.Val] wrapper
//

type bogus struct {
    bogus string
}

var BogusType = cel.OpaqueType("bogus.Wrapper")

type wrapper bogus

var _ ref.Val = wrapper{}

func (w wrapper) ConvertToNative(reflect.Type) (any, error)
func (w wrapper) ConvertToType(ref.Type) ref.Val
func (w wrapper) Equal(ref.Val) ref.Val

func (w wrapper) Type() ref.Type { return BogusType }
func (w wrapper) Value() any { return bogus(w) }

//
// setup [types.Adapter]
//

type bogusAdapter struct {
    baseAdapter types.Adapter
}

func (ba *bogusAdapter) NativeToValue(value any) ref.Val {
    switch v := value.(type) {
    case wrapper: return v
    case bogus: return wrapper(v)
    }

    return ba.baseAdapter.NativeToValue(value)
}

func bogusAdapterOption(env *cel.Env) (*cel.Env, error) {
    adapter := bogusAdapter{baseAdapter: env.CELTypeAdapter()},
    return env.Extend(cel.CustomTypeAdapter(&adapter))
}

//
// setup CEL environment
//

env := cel.Env(
    bogusAdapterOption,
    ext.NativeTypes(...),
)
```

**Expected behavior**

I am not sure what'd be the solution though. This very well may be the intended behaviour but took me a while to get to the bottom of it. But it doesn't seem like it. `ext.NativeTypes` accepts `reflect.Type` and proto descriptors which are getting registered explicitly to `nativeTypeProvider`. So my assumption was that only registered types will be converted through `NativeToValue`

**Additional context**

`nativeTypeProvider` treats _all_ structures as native types and uses `newNativeObject`. Thus `bogusAdapter.NativeTypes` not being called at all.

If you define type as `type wrapper string` then it should (I think) fall to `baseAdapter` which makes this even more confusing.

When you swap `EnvOption` order, it does work.

lopster568:
`nativeTypeProvider.NativeToValue` wrapped every struct-kind value as a native object regardless of whether the type was registered through `ext.NativeTypes`. Every other method on the provider checks the `nativeTypes` registry first and delegates unregistered types to the composed base provider or adapter; `NativeToValue` was the exception, so a `cel.CustomTypeAdapter` composed alongside `ext.NativeTypes` never saw struct values it should convert itself.

This mirrors the registry check from `NewValue`: registered struct types are still wrapped as native objects, and unregistered ones now reach the base adapter. This is the fix proposed in the issue and changes behavior for unregistered structs.

Fixes #1343

### DIFF
--- a/ext/native.go
+++ b/ext/native.go
@@ -395,7 +395,16 @@ func (tp *nativeTypeProvider) NativeToValue(val any) ref.Val {
 			time.Time:
 			return tp.baseAdapter.NativeToValue(val)
 		default:
-			return tp.newNativeObject(val, rawVal)
+			// Only claim struct values whose type was registered via
+			// ext.NativeTypes. Every other method on nativeTypeProvider checks
+			// the registry before handling a type; without the same check here
+			// a composed base adapter never sees unregistered structs it wants
+			// to convert itself.
+			typeName := fmt.Sprintf("%s.%s", simplePkgAlias(refVal.Type().PkgPath()), refVal.Type().Name())
+			if _, found := tp.nativeTypes[typeName]; found {
+				return tp.newNativeObject(val, rawVal)
+			}
+			return tp.baseAdapter.NativeToValue(val)
 		}
 	default:
 		return tp.baseAdapter.NativeToValue(val)
--- a/ext/native_test.go
+++ b/ext/native_test.go
@@ -1277,3 +1277,61 @@ type TestRefValFieldType struct {
 	IntVal       types.Int
 	CELTime      types.Timestamp `cel:"time"`
 }
+
+// registeredNativeStruct is registered with NativeTypes in the delegation test.
+type registeredNativeStruct struct {
+	Name string
+}
+
+// unregisteredNativeStruct is not registered, so NativeToValue should hand it to
+// the composed base adapter rather than wrapping it as a native object.
+type unregisteredNativeStruct struct {
+	Name string
+}
+
+// recordingAdapter converts unregisteredNativeStruct into a sentinel string and
+// records that it was asked to, so the test can confirm nativeTypeProvider
+// delegated the value. Everything else falls through to the base adapter.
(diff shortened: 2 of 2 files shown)


ext/native.go
```diff
@@ -395,7 +395,16 @@ func (tp *nativeTypeProvider) NativeToValue(val any) ref.Val {
 			time.Time:
 			return tp.baseAdapter.NativeToValue(val)
 		default:
-			return tp.newNativeObject(val, rawVal)
+			// Only claim struct values whose type was registered via
+			// ext.NativeTypes. Every other method on nativeTypeProvider checks
+			// the registry before handling a type; without the same check here
+			// a composed base adapter never sees unregistered structs it wants
+			// to convert itself.
+			typeName := fmt.Sprintf("%s.%s", simplePkgAlias(refVal.Type().PkgPath()), refVal.Type().Name())
```
TristonianJones, line 403: Let's make this a helper function to create a typeName from a value (including reflect.Type) since it appears a few places in this file.

TristonianJones (changes requested):
Thanks for the fix! I just have one minor code maintenance request, but otherwise LGTM

ext/native.go
lopster568: Done, pulled it into a nativeTypeName helper used at all three sites

TristonianJones (approved):

TristonianJones:
@lopster568 I'm just waiting on the CLA check before I can accept the contribution. Thanks for the fix!

PR #1385, opened 2026-07-24
Updates to type unification with parameterized types

### LINKED ISSUE #1312: Compiler type-checking failure with empty list variables

TristonianJones:
Addresses type unification issues identified in conformance tests.

Closes #1312


checker/checker_test.go
```diff
@@ -2452,6 +2452,157 @@ _&&_(_==_(list~type(list(dyn))^list,
...
+			in: `[null_int, null, 1]`,
+			env: testEnv{
+				idents: []*decls.VariableDecl{
+					decls.NewVariable("null_int", types.NewNullableType(types.IntType)),
+				},
+			},
+			outType: types.NewListType(types.NewNullableType(types.IntType)),
+		},
+		{
+			in:      `[1, null]`,
```
jnthntatum, line 2489: These are a bit interesting, do you think we should automatically promote to a union type based on component elements, or require that one of the elements already inferred to be a union? (maybe less interesting for just nullables right now, but becomes more interesting if we add a JSON value type union or something like any of from JSON schema and have to decide if we keep a set of type union candidates or allow arbitrary ones).

checker/checker_test.go
```diff
@@ -2452,6 +2452,157 @@ _&&_(_==_(list~type(list(dyn))^list,
...
+			in: `[null, 1, null_int]`,
+			env: testEnv{
+				idents: []*decls.VariableDecl{
+					decls.NewVariable("null_int", types.NewNullableType(types.IntType)),
+				},
+			},
+			outType: types.NewListType(types.NewNullableType(types.IntType)),
+		},
+		{
+			in: `[null_int, null, 1]`,
```
jnthntatum, line 2480: May want an explicit test to check that different nullables don't coalesce [m.string_wrapper, null, m.int64_wrapper] -> list(dyn)

jnthntatum (approved):

PR #1393, opened 2026-07-29
Simplify support for native object types

TristonianJones:
Shift most of the logic from the NativeTypeProvider over to the NativeObject wrappers. 

This change is a stepping stone to consolidating type providers into the common/types
package and making it less painful to manage custom type adaptation.

### DIFF
--- a/ext/native.go
+++ b/ext/native.go
@@ -720,12 +720,14 @@ func (t *nativeType) FindFieldType(fieldName string) (*types.FieldType, bool) {
 	return &types.FieldType{
 		Type: celType,
 		IsSet: func(obj any) bool {
+			// TODO: determine what to do if refVal is Invalid()
 			refVal := reflect.Indirect(reflect.ValueOf(obj))
 			// Check if field path exists and is set
 			refFieldVal := safeGetFieldByIndex(refVal, refField.Index)
 			return refFieldVal.IsValid() && !refFieldVal.IsZero()
 		},
 		GetFrom: func(obj any) (any, error) {
+			// TODO: determine what to do if refVal is Invalid()
 			refVal := reflect.Indirect(reflect.ValueOf(obj))
 			// Check if field path exists and is set
 			refFieldVal := safeGetFieldByIndex(refVal, refField.Index)
--- a/ext/native.go
+++ b/ext/native.go
@@ -581,8 +581,11 @@ func newNativeTypes(fieldNameHandler NativeTypesFieldNameHandler, rawType reflec
 		}
 		result = append(result, nt)
 
-		for idx := 0; idx < t.NumField(); idx++ {
-			iterateStructMembers(t.Field(idx).Type)
+		for _, field := range reflect.VisibleFields(t) {
+			if !field.IsExported() || !isSupportedType(field.Type) {
+				continue
+			}
+			iterateStructMembers(field.Type)
 		}
 	}
 	iterateStructMembers(rawType)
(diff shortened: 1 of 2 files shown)


ext/native.go
```diff
@@ -756,34 +686,67 @@ func (t *nativeType) Value() any {
...
-	return reflect.StructField{}, false
+	celType, ok := convertToCelType(refField.Type)
+	if !ok {
+		return nil, false
+	}
+	return &types.FieldType{
+		Type: celType,
+		IsSet: func(obj any) bool {
+			refVal := reflect.Indirect(reflect.ValueOf(obj))
+			refField := refVal.FieldByIndex(refField.Index)
```
l46kok: nit: existing, but consider guarding the access here with `refVal.isValid()`

ext/native.go
```diff
@@ -756,34 +686,67 @@ func (t *nativeType) Value() any {
...
+func (t *nativeType) hasField(fieldName string) (reflect.StructField, bool) {
+	f, found := t.fieldsByName[fieldName]
+	if !found {
 		return reflect.StructField{}, false
 	}
+	return f, true
+}
 
-	if t.fieldNameHandler == nil {
-		return t.refType.FieldByName(fieldName)
```
l46kok, line 767: It's possible we may have just lost the ability to handle promoted fields in embedded structs. Could we add a test for this? (I think we're only iterating on top-level fields now).

jnthntatum (dismissed):

ext/native.go
TristonianJones: I did shift it so we wouldn't traverse through unexported fields. This was checked in one place, but not in others. It's kind of odd to be able to access an unexported field (I think it only works for executions in the same package). Is that what you're referring to?

l46kok: I think it's an issue with exported fields within an embedded struct. Something like this:

```go
type TestPromotedInner struct {
       PromotedField string
}

type TestPromotedOuter struct {
       TestPromotedInner
}

envOpts := []cel.EnvOption{
               NativeTypes(
                       reflect.TypeFor[*TestPromotedOuter](),
               ),
               cel.Variable("test", cel.ObjectType("ext.TestPromotedOuter")),
       }

// This doesn't compile
ast, iss := env.Compile("test.PromotedField == 'hello'")
```

ext/native.go
```diff
@@ -680,35 +610,35 @@ func newNativeType(fieldNameHandler NativeTypesFieldNameHandler, rawType reflect
...
-			}
-			if _, found := fieldNames[fieldName]; found {
-				return nil, fmt.Errorf("invalid field name `%s` in struct `%s`: %w", fieldName, refType.Name(), errDuplicatedFieldName)
-			} else {
-				fieldNames[fieldName] = struct{}{}
-			}
+	// Collect the set of visible / exported fields, ensuring that unsupported types and sentinel
+	// 'skip' tags such as `-` are filtered out.
+	fieldsByName := make(map[string]reflect.StructField)
+	for idx := 0; idx < refType.NumField(); idx++ {
```
l46kok: Looks like go provides a handy utility for this that also looks at fields in embedded structs, we could probably use it instead:
https://cs.opensource.google/go/go/+/go1.26.5:src/reflect/visiblefields.go;l=16

ext/native.go
TristonianJones: I've added a TODO since I have to look into what it would look like to support zero-values on access at greater depth.

ext/native.go
TristonianJones: Ah, well, it looks like we didn't really have it before, but I've added support and added checks to make field getting and field setting safe for promoted fields in an embedded, uninitialized, struct pointer.

ext/native.go
TristonianJones: Thanks! I've updated the code.

l46kok (approved):

PR #1395, opened 2026-07-30
Support self-describing, self-adapting struct types

TristonianJones:
The `StructTypeDescriptor` interface allows structs to self-describe
their own fields, field types, value creation, as well as how they should
be adapted when their `reflect.Type` is encountered during data reads

### DIFF
--- a/common/types/provider.go
+++ b/common/types/provider.go
@@ -476,11 +476,24 @@ func (p *Registry) RegisterMessage(message proto.Message) error {
 // to CEL, even when they're not based on protobuf types.
 func (p *Registry) RegisterType(types ...ref.Type) error {
 	for _, t := range types {
-		if st, ok := t.(StructTypeDescriptor); ok {
-			typeName := t.TypeName()
-			if celType, isCelType := t.(*Type); isCelType && celType.Kind() == TypeKind && len(celType.parameters) > 0 {
-				typeName = celType.parameters[0].TypeName()
+		existing, found := p.revTypeMap[t.TypeName()]
+		celType := maybeForeignType(t)
+		if found {
+			if !existing.IsEquivalentType(celType) {
+				return fmt.Errorf("type registration conflict. found: %v, input: %v", existing, celType)
 			}
+			if existing.traitMask != celType.traitMask {
+				return fmt.Errorf(
+					"type registered with conflicting traits: %v with traits %v, input: %v",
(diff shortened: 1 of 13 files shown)


common/types/provider_test.go
```diff
@@ -107,60 +138,80 @@ func TestRegistryFindStructType(t *testing.T) {
...
-	_, found = reg.FindType(msgTypeName + "Undefined")
-	if found {
-		t.Fatalf("FindType() found: %q", msgTypeName+"Undefined")
+
+	tests := []struct {
+		typeName  string
+		wantFound bool
+	}{
+		{
+			typeName:  ".google.expr.proto3.test.TestAllTypes",
```
l46kok, line 147: Consider adding a similar test case like this (leading dot with a FQN custom type)

common/types/provider.go
```diff
@@ -382,6 +476,21 @@ func (p *Registry) RegisterMessage(message proto.Message) error {
 // to CEL, even when they're not based on protobuf types.
 func (p *Registry) RegisterType(types ...ref.Type) error {
 	for _, t := range types {
+		if st, ok := t.(StructTypeDescriptor); ok {
+			typeName := t.TypeName()
+			if celType, isCelType := t.(*Type); isCelType && celType.Kind() == TypeKind && len(celType.parameters) > 0 {
+				typeName = celType.parameters[0].TypeName()
+			}
+			p.structTypes[typeName] = st
```
l46kok, line 497: Do we also need to sanitize the struct type name here?

l46kok, line 497: Do you want to also do a collision check here against revTypeMap, similar to what we're doing just below?

TristonianJones: The `TypeName()` reference should always be sanitized. When given a string form of `typeName` like in `NewValue()`, it could be coming from the parsed expression and have a leading '.'.

TristonianJones: I think I should just early exit on the revTypeMap ... this could leave it in a strange state if somehow you have two different identically named types.

common/types/provider_test.go
TristonianJones: I actually had this lower down (huge amount of test changes -- sorry!), but added more test assertions to make sure the type is uniform.

common/types/provider.go
TristonianJones: Done.

l46kok (approved):

PR #1396, opened 2026-07-31
Move native type support into the Core CEL library

TristonianJones:
This change shifts all of the `ext.NativeTypes` functionality into core CEL

The motivation for this change is that composing type providers is tricky and
error prone. Additionally, making it simple to package Go structs instead of
proto for type representations associated with extensions is highly desirable
from a velocity and performance standpoint.


common/types/native.go
```diff
@@ -0,0 +1,578 @@
...
+		jsonStruct, err := o.ConvertToNative(jsonStructType)
+		if err != nil {
+			return nil, err
+		}
+		return structpb.NewStructValue(jsonStruct.(*structpb.Struct)), nil
+	case jsonStructType:
+		refVal := reflect.Indirect(o.refValue)
+		fields := make(map[string]*structpb.Value, refVal.NumField())
+		for fieldName, fieldType := range o.valType.fieldsByName {
+			fieldValue := refVal.FieldByIndex(fieldType.Index)
```
l46kok, line 272: nit: For consistency sake, could we use `safeGetFieldByIndex` here too?

l46kok (approved):

common/types/native.go
TristonianJones: I've picked up this chain on the next PR, thanks!

Issue #1397, opened 2026-08-03
Program construction rebuilds a dispatcher containing every function binding of the environment


jpbetz:
Kubernetes uses CEL heavily for CRD validation. This results in a large number of CEL
environments where the only difference is the schema of the `value` and `oldValue` variables.
All function bindings and all other settings are identical across the environments.

`newProgram` creates a fresh `interpreter.Dispatcher` per program and copies every
function-overload binding of the environment into it (`disp.Add(e.functionBindings...)`).
For a stdlib + extensions environment that is hundreds of map inserts per program, and
the dispatcher contents are byte-identical across every program of the environment.
Additionally, `funcBindOnce` memoizes binding materialization per *environment*, so
consumers that extend a base environment per compilation unit (e.g. Kubernetes CRD
validation, one extended env per rule-bearing schema node) re-materialize the full
binding set once per extended env even though the function declarations are shared.

Programs are frequently retained (Kubernetes retains them for the lifetime of each CRD),
so this is resident heap, not just construction-time garbage. Heap profiling of
Kubernetes API servers with large numbers of CRDs attributes ~18KB resident per retained
program to dispatcher/binding duplication.

## Measurements

Benchmarks: `Program()` on an already-compiled AST — marginal cost on a warm env
(`ProgramK8sShape`) and first-program cost per extended env
(`ProgramFirstPerEnvK8sShape`). Harness is `cel/env_sharing_bench_test.go` on the branch
below. go1.24.1, linux/amd64, Intel Ultra 7 165U, `-count=10`, benchstat. Base for this
comparison is the extend-sharing branch (see companion issue); the improvement relative
to master is the same, since extend-sharing does not change these benchmarks.

```
                              │ extend-sharing │        + dispatcher-sharing           │
                              │     sec/op     │    sec/op     vs base                 │
ProgramK8sShape-14              15147.0n ±  4%   887.1n ± 15%  -94.14% (p=0.000 n=10)
ProgramFirstPerEnvK8sShape-14    38.496µ ± 10%   1.867µ ± 30%  -95.15% (p=0.000 n=10)

                              │ extend-sharing │        + dispatcher-sharing           │
                              │      B/op      │     B/op      vs base                 │
ProgramK8sShape-14              14.266Ki ± 0%    1.070Ki ± 0%  -92.50% (p=0.000 n=10)
[...]

PR #1399, opened 2026-08-03
Program plan optimizations

TristonianJones:
A significant amount of overhead comes from the instantiation of the dispatcher
with an immutable set of functions. Instead of recreating the dispatcher, create
a shared one across program instances.

Additionally, check whether optimizations are needed for an expression before
applying them as decorators.

| Benchmark Case | Before (ns/op) | After (ns/op) | Δ Time | Before (B/op) | After (B/op) | Δ Memory | Before (allocs) | After (allocs) | Δ Allocs | 
| :--- | :---: | :---: | :---: | :---: | :---: | :---: | :---: | :---: | :---: | 
| BenchmarkProgramPlan/Default | 8,153 | 930 | **-88.6%** | 8,320 | 1,416 | **-83.0%** | 36 | 26 | **-27.8%** |
| BenchmarkProgramPlan/OptimizeUnneeded | 7,344 | 1,150 | **-84.3%** | 8,784 | 1,512 | **-82.8%** | 50 | 32 | **-36.0%** | 
| BenchmarkProgramPlan/OptimizeNeeded | 8,370 | 2,164 | **-74.1%** | 10,224 | 2,976 | **-70.9%** | 67 | 52 | **-22.4%** |

Addresses the dispatcher-sharing proposed in #1397, which is the first half of the optimizations suggested to minimize startup costs.

### DIFF
--- a/cel/cel_test.go
+++ b/cel/cel_test.go
@@ -4002,6 +4002,58 @@ func BenchmarkDynamicDispatch(b *testing.B) {
 	})
 }
 
+func BenchmarkProgramPlan(b *testing.B) {
+	env, err := NewEnv(
+		Variable("ai", IntType),
+		Variable("ar", MapType(StringType, StringType)),
+	)
+	if err != nil {
+		b.Fatalf("NewEnv() failed: %v", err)
+	}
+	astSimple, iss := env.Compile("ai == 20 || ar['foo'] == 'bar'")
+	if iss.Err() != nil {
+		b.Fatalf("env.Compile() failed: %v", iss.Err())
+	}
+	astOpt, iss := env.Compile("ai in [10, 20, 30] || 'foo' in ar")
+	if iss.Err() != nil {
+		b.Fatalf("env.Compile() failed: %v", iss.Err())
+	}
+
+	b.Run("Default", func(b *testing.B) {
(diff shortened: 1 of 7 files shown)


TristonianJones:
@jpbetz FYI

jpbetz (commented):
LGTM, this looks like a cleaner/simpler approach than I had considered. Thanks!

TristonianJones:
@jpbetz I went ahead and folded in more of the changes from your proposed PR. The numbers I see are close to what you observed:

```
pkg: github.com/google/cel-go/cel
cpu: Apple M1 Pro
BenchmarkEnvExtendK8sShape-10             	   30596	     38461 ns/op	   23332 B/op	     278 allocs/op
BenchmarkProgramK8sShape-10               	 1579915	       755.6 ns/op	    1144 B/op	      20 allocs/op
BenchmarkProgramFirstPerEnvK8sShape-10    	 1000000	      1440 ns/op	    1144 B/op	      20 allocs/op
```

There's two more changes to make to reduce the `EnvExtend` cost further, but then I think we'll have captured everything.

jpbetz:
Awesome. This LGTM

jnthntatum (approved):

PR #1402, opened 2026-08-05
NativeToValue Go-based JSON type support

TristonianJones:
NativeToValue type conversion support for `json.Number` and `json.RawMessage`


ext/jwt/jwt.go
```diff
@@ -0,0 +1,835 @@
...
+
+	lowerIss := strings.ToLower(issuer)
+	if strings.HasPrefix(lowerIss, "http://") {
+		if !strings.HasPrefix(lowerIss, "http://localhost") && !strings.HasPrefix(lowerIss, "http://127.0.0.1") && !strings.HasPrefix(lowerIss, "http://[::1]") {
+			return "", errors.New("cannot fetch JWKS key over insecure HTTP: issuer URL must use HTTPS")
+		}
+	}
+
+	url := issuer
+	if !strings.HasPrefix(url, "http://") && !strings.HasPrefix(url, "https://") {
```
TristonianJones: This check is possibly redundant with the above check. Also, not the lack of lower-casing in comparisons.

common/types/provider.go
```diff
@@ -663,6 +664,18 @@ func (p *Registry) NativeToValue(value any) ref.Val {
 		if v != nil {
 			return Uint(*v)
 		}
+	case json.Number:
```
jnthntatum, line 667: Question for my benefit: Does this handle stringified numbers for ints >~ 1e53 (and I guess NaN +/-inf) ?  Or does it just permit the Int conversion if it's losslessly convertible?

jnthntatum (approved):

common/types/provider.go
TristonianJones: Apparently the Golang json encoders / decoders, from which this type alias to a string is generated, don't handle NaN and Infinity. Also, the promises of Int vs. Float are absent as well. Effectively, if it looks 'int-like' attempt to convert it to int vis `strconv` otherwise convert to a float. I'm not totally sure how I feel about that, but it would seem to capture most of the cases we care about since the decoder captures the number type as a string and doesn't automatically convert to a floating point rep first and rejects the edge cases.

PR #1404, opened 2026-08-06
Support aggregate size computations over list, maps, and structs

TristonianJones:
To support best cost tracking, introduce version-aware `AggregateSize` methods 
on list, map, and object types. 

The implementations currently memoize aggregate size computations
for immutable objects. The aggregate size is capped at `uint32` to minimize
the allocation overhead associated with the book-keeping fields on these
objects.

As far as possible, the allocation cost associated computing the size 
list, maps, and structs has been reduced to a minimum. The purpose of 
having different, or versioned, `AggregateSize` implementations is to allow 
users to tailor which kinds of costs they're willing to consider before giving
up and assigning a maximum size.


common/types/overflow.go
```diff
@@ -427,3 +427,27 @@ func uint64ToInt64Lossless(v uint64) (int64, bool) {
...
+	if n < 0 {
+		return 0
+	}
+	if uint64(n) > math.MaxUint32 {
+		return math.MaxUint32
+	}
+	return uint32(n)
+}
+
+func safeUint32FromBoxedInt(v Int) uint32 {
```
l46kok, line 445: nit: do you want to clamp the negatives? (-1 will return MaxUint32)

common/types/util.go
```diff
@@ -46,3 +54,270 @@ func Equal(lhs ref.Val, rhs ref.Val) ref.Val {
...
+		for it.HasNext() == True {
+			total = safeAddUint32(total, s.AggregateSize(it.Next()))
+		}
+		return total
+	case String:
+		return safeUint32FromInt(utf8.RuneCountInString(string(v)))
+	case Bytes:
+		return safeUint32FromInt(len(v))
+	case traits.Sizer:
+		return safeUint32FromBoxedInt(v.Size().(Int))
```
l46kok: Do you want to check for int type first? (Sizer's Size() returns ref.Val).

common/types/list.go
```diff
@@ -480,6 +490,11 @@ func (l *concatList) Size() ref.Val {
 	return l.cachedSize
 }
 
+// AggregateSize implements the AggregateSizeVisitor interface method.
+func (l *concatList) AggregateSize(sizer AggregateSizer) uint32 {
+	return safeAddUint32(sizer.AggregateSize(l.prevList), sizer.AggregateSize(l.nextList))
```
l46kok, line 495: Question: Should this be adding 1 to the cost? In relation to `concat_list` test case in list_test.go:

- l1 (3,4): base list cost of 1, 2 elements = 3
- l2 (7,8): base list cost of 1, 2 elements = 3

Total: 6

But we might be missing the cost of final concatenated list allocation itself (I think AggregateSize implementation for baseList above starts at base cost 1 to account for this).

common/types/util.go
```diff
@@ -46,3 +54,270 @@ func Equal(lhs ref.Val, rhs ref.Val) ref.Val {
...
+	total := uint32(1)
+	m.Range(func(k protoreflect.MapKey, v protoreflect.Value) bool {
+		total = safeAddUint32(total, getProtoValueAggregateSize(s, k.Value()))
+		total = safeAddUint32(total, getProtoValueAggregateSize(s, v))
+		return true
+	})
+	return total
+}
+
+func getReflectValueAggregateSize(s AggregateSizer, fieldVal reflect.Value) uint32 {
```
l46kok: wdyt about adding a guard for cyclic structs? (In case if the host application passes one in)

common/types/list.go
TristonianJones: So, the concat list being a set of pointers, I just kept it at the cost of the sublists since you get 1 for each child container (base cost) and the size of the list. I figured it sort of evened out or slightly over-counted. Namely since this ends up being a different answer if you do `([a] + [b]).filter(i, i)` -- you get one less due to the newly formed list. That, and concat as a view is pretty cheap since we don't reallocate lists like we do for `filter` (for example)

common/types/util.go
TristonianJones: I think adding a budget and depth make a lot of sense. I have this drafted, to a degree, in a memory limit PR, but let me move it over here.

common/types/util.go
TristonianJones: `Size` is always an `Int`. At some point I was worried about `Unknown` and `Err` types, but the function guards prevent these values from ever being extracted from objects now -- they didn't in the early days and by then it was too late to change the interface.

common/types/overflow.go
TristonianJones: I figure if you pass in a negative it's not intentional, and I should flip it to max as there's probably a bug. Thoughts?

common/types/list.go
l46kok: Makes perfect sense to me, thanks for elaborating

common/types/util.go
l46kok: I was actually wondering about custom structs implementing Sizer and downstreams accidentally returning `Uint` (this is pretty contrived, feel free to just ack)

common/types/overflow.go
l46kok: I think we should at least make this consistent one way or the other (`safeUint32FromInt` above clamps negatives). 

As a general utility -- I'd think clamping to 0 is the common behavior but returning max uint32 is just fine to signal fail-closed for incorrect cost computation.

TristonianJones: Oh, I'm so sorry, I didn't notice the incongruent clamping behavior. I'll fix it.

common/types/util.go
TristonianJones: There's unfortunately a lot of other code that would break if `Size` were not an `Int`, so I'm going to take this one as a given until someone files a bug for it.

common/types/util.go
TristonianJones: I've added a depth and traversal limit to the calculator which will result in a saturated uint32 once either is hit.

common/types/map.go
```diff
@@ -303,6 +303,23 @@ func (m *baseMap) Size() ref.Val {
...
 
+// AggregateSize implements the AggregateSizeVisitor interface method.
+func (m *baseMap) AggregateSize(sizer AggregateSizer) uint32 {
+	if m.aggSize != 0 {
+		return m.aggSize
+	}
+	var total uint32 = 1
+	it := m.Iterator()
+	for it.HasNext() == True {
+		key := it.Next()
```
l46kok: wdyt about just calling m.Fold() to avoid the iterator reallocation and the double lookup (m.Find(key)) cost? I think you can move the total cost computation to a small private struct and implement FoldEntry there. Something like:

```go
type aggSizeFolder struct {
	sizer AggregateSizer
	total uint32
}

func (f *aggSizeFolder) FoldEntry(k, v any) bool {
	f.total = safeAddUint32(f.total, f.sizer.AggregateSize(k))
	f.total = safeAddUint32(f.total, f.sizer.AggregateSize(v))
	return true
}
```

l46kok (dismissed):

common/types/map.go
TristonianJones: I moved created a common folder. I'm not sure how much it changes, but it at least moves the logic to a central place and pattern. I also added more tests and cleaned up some other duplicated code paths in the traversal.

l46kok (approved):

PR #1408, opened 2026-08-07
Aggregate semantics in Policy Compiler

TristonianJones:
Add `aggregate` semantics into Policy Compiler.

The CEL policy compiler initially started as `first-match` policy compilation, ensuring at
most one result would be produced from a policy expression evaluation. With `aggregate`
semantics, multiple outputs can be produced on a single pass. For those familiar with K8s
validating admission policies, the effect is similar.

Ported from cel-expr/cel-java/pull/1052

====

Aggregate walks through all matching rules (including nested ones) and appends them into a list:

```yaml
rule:
  aggregate:
    - condition: "true"
      output: "'FOO'"
    - condition: "true"
      output: "'BAR'"

# Output: ['FOO', 'BAR']
```

Few noteworthy design decisions below. All examples assume all conditions matched:

1. For usability reasons, subrules under an aggregate ancestor will always have their **lists flattened**:

```YAML
name: aggregate_flat_flattening_example
rule:
  aggregate:
    - rule:
        match:
          - condition: "resource.is_admin == true"
            output: "['GDPR_STANDARD', 'EU_B2C_NOTICE']"
    - condition: "true"
      output: "'FALLBACK'"
# Output: ['GDPR_STANDARD', 'EU_B2C_NOTICE', 'FALLBACK']
```

2. Base case of an aggregate rule is an empty list. Nested conditional rules within an aggregate rule which outputs `optional.none()` are pruned (except in cases where policy output explicitly emits an `optional.none()`):

```YAML
name: optional_pruning_example
rule:
  aggregate:
    - rule:
        match:
          - condition: "1 == 2"
            output: "'EU_NOTICE'"
    - condition: "true"
      output "'ALWAYS'"

# Output: ['ALWAYS']
```

```YAML
name: explicit_optional_none_example
rule:
  aggregate:
    - rule:
        match:
          - condition: "resource.is_b2c == true"
            output: "optional.none()"   # Explicitly authored by user

    - condition: "true"
      output: "optional.of('ALWAYS')"

# Output: [optional.none(), optional.of('ALWAYS')]
```

Note: nesting `aggregate` clauses is currently not allowed, and will result in a compilation error.


policy/composer.go
```diff
@@ -189,43 +189,68 @@ func (opt *ruleComposerImpl) optimizeRule(ctx *cel.OptimizerContext, r *Compiled
...
-			output = step.combine(output)
-			continue
+		if isAggregate {
+			output = opt.combineAggregate(ctx, currentStep, output)
+		} else {
+			output = currentStep.combine(output)
+		}
+	}
+
+	if output == nil {
```
TristonianJones: @l46kok  - This is a bit different from cel-java because output isn't initialized to a base step in order to avoid superfluous `+ []` appearing in the output expression. I believe you handle this with the constant folding and CSE in Java, but constant-folding isn't turned on by default for policy composition in order to preserve the round-trip to source metadata for coverage purposes.

policy/compiler_test.go
```diff
@@ -46,52 +46,6 @@ func TestCompile(t *testing.T) {
 	}
 }
 
-func TestRuleComposerError(t *testing.T) {
```
TristonianJones, line 49: These moved into composer_test.go

policy/composer.go
l46kok: Const folding in CEL-Java doesn't actually strip away `+ []` at the moment (requires all subtrees to contain constant literals), so the superfluous list concatenation is actually retained at the end: https://github.com/cel-expr/cel-java/blob/main/policy/src/test/java/dev/cel/policy/CelPolicyCompilerImplTest.java#L265. But this is  unintended -- I'll update Java to strip the redundant concatenation away.

policy/composer.go
```diff
@@ -189,43 +189,69 @@ func (opt *ruleComposerImpl) optimizeRule(ctx *cel.OptimizerContext, r *Compiled
...
 	matchCount := len(matches)
 	var output compositionStep = nil
-	// If the rule has an optional output, the last result in the ternary should return
+	// If the rule is non-aggregate and has an optional output, the last result in the ternary should return
 	// `optional.none`. This output is implicit and created here to reflect the desired
 	// last possible output of this type of rule.
-	if r.HasOptionalOutput() {
+	if !returnList && r.HasOptionalOutput() {
 		output = newOptionalCompositionStep(ctx, ctx.NewLiteral(types.True), ctx.NewCall("optional.none"))
 	}
```
l46kok: You'll want to handle the composition step for non-optional cases here too (handled [here in Java](https://github.com/cel-expr/cel-java/blob/main/policy/src/main/java/dev/cel/policy/RuleComposer.java#L150-L153))

I've added more conformance test cases around this -- https://github.com/cel-expr/cel-policy/pull/19. aggregate_nested_single_conditional_match and aggregate_nested_all_rules_fall_through_to_empty should trigger the failure here

policy/composer.go
```diff
@@ -494,12 +537,20 @@ func (s nonOptionalCompositionStep) combine(step compositionStep) compositionSte
 		// Likely a candidate for dead-code warnings.
 		return s
 	}
+	if !s.isConditional() {
+		return s
+	}
+	stepExpr := step.expr()
```
l46kok: I think with the base step fix mentioned above -- this will effectively become dead code (condition will never be dropped)

policy/composer.go
TristonianJones: I added a fix into the concatenate step that checks for `+` with an empty list arg, and then corrected the base step to match cel-java.

policy/composer.go
TristonianJones: Good call. Removed.

policy/composer.go
TristonianJones: Fixed, and tested against cel-expr/cel-policy#19, came back clean.

l46kok (approved):

Issue #1240, opened 2025-11-24
Feature Request: Add deep merge support to CEL maps library


kahirokunn:
**Feature request checklist**

* [x] There are no issues that match the desired change

---

**Change**
Add deep merge support for nested map structures in the CEL `maps` library.

Currently, `map.merge` only performs a shallow merge: when two maps share the same key whose value is also a map, the value from the second map completely replaces the value from the first map. This makes it hard to work with configuration-like structures that naturally contain nested maps and lists (for example, Kubernetes-style objects).

The requested change is to provide a deep merge operation, either as:

* A new function (e.g. `mergeDeep`, `deepMerge`, or similar), or
* An additional mode/parameter on the existing `merge` function,

with the following behavior:

* **Non-map values**: values from the second map overwrite values from the first map (same behavior as current shallow `merge`).
* **Map values**: if both maps contain map values for the same key, recursively merge those nested maps instead of replacing them entirely.
* **List values containing maps**: support merging list items that represent configuration objects, by matching on an identifying key (e.g. `name` for Kubernetes containers).

  * When two list elements share the same identifier, deep merge those elements.
  * When identifiers differ, keep all elements (concatenate the lists).
* **Lists of primitive values**: behavior could be configurable (concatenate vs. replace) or follow a simple, documented default.

This feature would significantly improve ergonomics when expressing configuration overlays and patches directly in CEL.

---

**Example**

Nested maps:

```cel
// Current behavior (shallow merge)
{'a': {'x': 1, 'y': 2}}.merge({'a': {'y': 3, 'z': 4}})
// => {'a': {'y': 3, 'z': 4}}

// Desired deep merge behavior
{'a': {'x': 1, 'y': 2}}.mergeDeep({'a': {'y': 3, 'z': 4}})
// => {'a': {'x': 1, 'y': 3, 'z': 4}}
```

Lists of maps (e.g. Kubernetes containers):

```cel
// Current behavior (shallow merge)
{'containers': [{'name': 'app', 'image': 'v1'}]}.merge({
  'containers': [
    {
      'name': 'app',
      'env': [{'name': 'DEBUG', 'value': 'true'}],
    },
  ],
})
// => {'containers': [{'name': 'app', 'env': [{'name': 'DEBUG', 'value': 'true'}]}]}

// Desired deep merge behavior
{'containers': [{'name': 'app', 'image': 'v1'}]}.mergeDeep({
  'containers': [
    {
      'name': 'app',
[...]

PR #1409, opened 2026-08-08
Add a maps extension with maps.merge

lopster568:
## Description

CEL has no way to combine two maps. The `+` operator concatenates strings, bytes, and lists, but has no map overload, and there is no `maps` extension.

This adds one, with a single function:

```
maps.merge(map(K, V), map(K, V)) -> map(K, V)
```

It returns a new map holding the entries of both arguments, with the second argument's values winning on conflicting keys. Neither input is modified. The merge is shallow, so a value that is itself a map is replaced rather than merged recursively.

```
maps.merge({}, {})                              // {}
maps.merge({'a': 1}, {'b': 2})                  // {'a': 1, 'b': 2}
maps.merge({'a': 1}, {'a': 2})                  // {'a': 2}
maps.merge({'a': {'x': 1}}, {'a': {'y': 2}})    // {'a': {'y': 2}}
```

## Scope

This is deliberately only the `replace` step from #1240. @TristonianJones suggested starting there:

> I think `replace` semantics is where I'd start -- last write wins, followed by set-if-absent as the deep merge is the trickiest case

So set-if-absent and recursive merge are left for follow-ups rather than guessed at here.

On the API shape, I went with a bare `maps.merge(a, b)` in a new `maps` namespace rather than a single entry point taking a strategy argument, because that is what the existing libraries do: `sets.contains` / `sets.equivalent` / `sets.intersects`, `math.ceil` / `math.floor` / `math.abs`. No function in `ext` takes a mode or strategy parameter today, so a later `maps.mergeIfAbsent` would fit that pattern without changing this signature. Happy to reshape it if you would rather have something else.

## Implementation notes

* Cost estimation and tracking scale with the combined size of both inputs plus map allocation, following `ext/sets.go`. Without them the checker would price this as a fixed cost of 1 despite the O(n+m) copy.
* `mapAllocCost` is added to `ext/costs.go` alongside the existing `listAllocCost`.
* Registered in `ext/extension_option_factory.go` (factory and alias) and in `repl/evaluator.go`.
* `ext/README.md` gains a `Maps` section in the style of the surrounding entries.

## Testing

`ext/maps_test.go` covers empty inputs, disjoint keys, conflicting keys, nested maps being replaced rather than merged, non-string keys, merging a variable, associativity over disjoint keys, compile-time rejection of non-map arguments, non-map arguments at runtime, and that the cost estimate scales with input size and brackets the tracked cost.

`go build ./...`, `go test ./...`, `go vet ./ext/...` and `gofmt` are clean.


ext/maps.go
```diff
@@ -0,0 +1,161 @@
...
+// LibraryName implements the SingletonLibrary interface method.
+func (mapsLib) LibraryName() string {
+	return "cel.lib.ext.maps"
+}
+
+// CompileOptions implements the Library interface method.
+func (mapsLib) CompileOptions() []cel.EnvOption {
+	mapType := cel.MapType(cel.TypeParamType("K"), cel.TypeParamType("V"))
+	return []cel.EnvOption{
+		cel.Function("maps.merge",
```
TristonianJones: Normally, I'd say this is a good idea, but let's just make it a member function instead since sometimes it's hard to remember the direction of the merge when it's a global function.

lopster568: Done. It's `a.merge(b)` now, same semantics as before: on conflicting keys the argument
overrides the receiver.

ext/maps.go
```diff
@@ -0,0 +1,161 @@
...
+// LibraryName implements the SingletonLibrary interface method.
+func (mapsLib) LibraryName() string {
+	return "cel.lib.ext.maps"
+}
+
+// CompileOptions implements the Library interface method.
+func (mapsLib) CompileOptions() []cel.EnvOption {
+	mapType := cel.MapType(cel.TypeParamType("K"), cel.TypeParamType("V"))
+	return []cel.EnvOption{
+		cel.Function("merge",
```
TristonianJones: Let's maybe call this something other than `merge`. 

A couple of good examples might be `putAll` (Java convention) or `extend` (Rust) since `merge` might imply semantics like combining the values of a colliding key. 

I'm inclined toward `putAll` since it's easy to document what it does, and you could have other methods like `<map(K,V)>.put(K, V) -> map(K,V)` if you need to update a specific key (and you might want to add `<map(K,V)>.remove(K) -> map(K,V)`, `<map(K,V)>.removeAll(<list(K)>) -> map(K,V)`, `map(K,V).keys()`, `map(K,V).values()` while you're at it)

ext/maps.go
```diff
@@ -0,0 +1,161 @@
...
+	cost := entries.MultiplyByCostFactor(1).Add(mapAllocCost).Add(callCostEstimate)
+	// The result holds at least as many entries as the larger input, when every
+	// key collides, and at most the sum of both, when none do.
+	resultSize := rangedSizeEstimate(max(lhsSize.Min, rhsSize.Min), entries.Max)
+	return callEstimate(cost, &resultSize)
+}
+
+// trackMapsMergeCost mirrors estimateMapsMergeCost against the actual inputs.
+func trackMapsMergeCost(args []ref.Val, _ ref.Val) *uint64 {
+	entries := safeAdd(actualSize(args[0]), actualSize(args[1]))
```
TristonianJones, line 121: I'm in the process of overhauling the cost framework, so this is fine for now, but please add a `TODO` to update this method once the new model is available.

ext/maps.go
```diff
@@ -0,0 +1,161 @@
...
+	}
+	return types.NewRefValMap(types.DefaultTypeAdapter, merged)
+}
+
+// copyEntries writes every entry of m into dst, overwriting entries whose keys
+// are already present. It returns a non-nil ref.Val only when the map yields an
+// error or unknown value.
+func copyEntries(m traits.Mapper, dst map[ref.Val]ref.Val) ref.Val {
+	it := m.Iterator()
+	for it.HasNext() == types.True {
```
TristonianJones, line 152: Consider using the `folder` interfaces to manage setting entries from `traits.Foldable`

ext/maps.go
lopster568: Done, renamed to putAll

ext/maps.go
lopster568: Done

ext/maps.go
lopster568: I kept `Find` for now. `Fold` passes native Go values for reflect/string-backed maps, and the only adapter reachable in the binding is `DefaultTypeAdapter`, so `NativeTypes`/custom-adapter entries would error on re-adaptation. `Find` returns already-adapted values, so it's correct for every backend.

TristonianJones (approved):

PR #1414, opened 2026-08-14
Parsing helper for working with different types and formats of timestamp

TristonianJones:
Simplify working with timestamps that could come from JSON or other sources.

### DIFF
--- a/common/types/timestamp_test.go
+++ b/common/types/timestamp_test.go
@@ -617,9 +617,39 @@ func TestParseTimestamp(t *testing.T) {
 			want: epochFloatTime,
 		},
 		{
+			name: "float64 negative",
+			val:  float64(-1700000000.5),
+			want: time.Unix(-1700000000, -500000000).UTC(),
+		},
+		{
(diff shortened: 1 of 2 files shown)


common/types/timestamp_test.go
```diff
@@ -551,3 +552,148 @@ func TestIsStrictRFC3339MatchesPattern(t *testing.T) {
...
+		{
+			name: "int64",
+			val:  int64(epoch),
+			want: epochTime,
+		},
+		{
+			name: "float64",
+			val:  float64(1700000000.5),
+			want: epochFloatTime,
+		},
```
jnthntatum, line 618: Up to you: it may be worth a test case for converting negative numbers / before unix epoch. (I think the math looks right, just suggesting for completeness)

jnthntatum (dismissed):

common/types/timestamp_test.go
TristonianJones: Thanks, I added some overflow checks as well.

jnthntatum (approved):

PR #1418, opened 2026-08-16
mem-track: Add memory tracker, scale-down string and byte sizes

TristonianJones:
Simple peak memory tracker and scale down the string & bytes costs.

Future changes will hook up the tracker to a memory limiter, and make aggregate
computations atomic on a given object. The caveat for proper use is that the
`SizeCalculator` should not change between environment configurations.

### DIFF
--- a/common/types/memory_tracker.go
+++ b/common/types/memory_tracker.go
@@ -1,0 +1,156 @@
+// Copyright 2026 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package types
+
+const (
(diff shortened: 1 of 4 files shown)


common/types/memory_tracker.go
```diff
@@ -0,0 +1,156 @@
...
+	}
+	return t
+}
+
+// Version returns the tracking version.
+func (t *MemoryTracker) Version() int {
+	return t.version
+}
+
+// Track observes a set of coexistent values as a single watermark, returning their combined
```
jnthntatum, line 103: For something like strcat or proto construction, wouldn't the high watermark need to be the sum of args plus output?

jnthntatum (dismissed):

common/types/memory_tracker.go
TristonianJones: The observers pass in all argument and output values, and since the aggregate size call computes the size of the proto and all it's fields, the high watermark will be exactly as you suggest (at least for proto since it's usually dealing with nested messages). The strcat case is a bit more painful to catch, and on second though, the proto + string fields case is too.

Since peak comes after construction, it's possible to carefully select inputs such that the threshold would only kick in at `2*threshold - 1`. In general, I think this is okay, but let me know if you think we should approach it differently.

TristonianJones: Per offline convo, will document the limitations of this approach regarding peak as a function of all observed arguments and the result size which may not be directly correlated.

jnthntatum (dismissed):

jnthntatum (approved):

PR #1419, opened 2026-08-17
Report every evaluation step to every observer

TristonianJones:
Cost tracking silently reported a cost of zero whenever state tracking was also enabled:

	cel.CostTracking(nil)                                 -> 5
	cel.CostTracking(nil) + cel.EvalOptions(OptTrackState)   -> 0
	cel.CostTracking(nil) + cel.EvalOptions(OptExhaustiveEval) -> 0

Each observer installed its own decorator, and decObserveEval returns a node which is already wrapped in a watcher untouched. Since the planner applies decorators in order, the state observer's decorator wrapped each node first and the cost observer's decorator then found a watcher and left it alone, so the cost observer's per-node callback was never installed. ObservableInterpretable still ran the tracker's InitState and GetState, so evaluation produced a CostTracker reporting a cost of zero rather than an error or a nil result -- including for programs configured with a cost limit, which then could not be exceeded.

Observers now register only as observers, and the planner installs a single decorator which reports each observation to all of them.

### DIFF
--- a/interpreter/planner.go
+++ b/interpreter/planner.go
@@ -23,6 +23,7 @@ import (
 	"github.com/google/cel-go/common/functions"
 	"github.com/google/cel-go/common/operators"
 	"github.com/google/cel-go/common/types"
+	"github.com/google/cel-go/common/types/ref"
 )
 
 // newPlanner creates an interpretablePlanner which references a Dispatcher, TypeProvider,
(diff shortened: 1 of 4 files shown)


interpreter/planner.go
```diff
@@ -73,6 +74,12 @@ type planBuilder struct {
 // such as state-tracking, expression re-write, and possibly efficient thread-safe memoization of
 // repeated expressions.
 func (p *planner) Plan(expr ast.Expr) (InterpretableV2, error) {
+	if len(p.observers) != 0 {
+		// A single decorator reports to every observer. One decorator per observer would not
+		// work, since the second decorator would find the node already wrapped by the first and
+		// leave it alone, silently dropping the second observer's observations.
+		p.decorators = append(p.decorators, decObserveEval(observeAll(p.observers)))
```
jnthntatum, line 81: Would this still have the same problem if there's a custom decorator applied before the observers? Or are there not many practical uses for custom decorators for client code?

jnthntatum (approved):

interpreter/planner.go
TristonianJones: The observer decorator wraps each plan step into a private `evalWatch` object which tries to prevent nested watchers to avoid unnecessary complexity in resolving the actual call to invoke. It's not possible to create these instances as part of the public API, and while it is possible to create new plan nodes, they have to go through the public APIs like `NewCall` or `NewConstVal` which create the usual plan nodes that the observers wrap -- observers should always be planned at the end.

PR #1427, opened 2026-08-20
[Pratt parser] Port Pratt parser from cel/cpp

dmitriplotnikov:
This PR ports the Pratt parser and hand-written lexer from [github.com/cel-expr/cel-cpp/tree/master/parser/internal](https://github.com/cel-expr/cel-cpp/tree/master/parser/internal).

### Highlights:
- **`PrattParser`**: Adds a recursive-descent top-down operator precedence (Pratt) parser implementation.
- **Fast Lexer**: Adds a dedicated lexer (`parser/lexer.go`) replacing ANTLR lexing with single-pass rune scanning.
- **Performance**: Pratt parser demonstrates significant performance improvements over ANTLR in `BenchmarkParsers`:
  - **8.4x – 14.8x faster** execution time
  - **6.0x – 12.5x less memory** consumed
  - **4.2x – 6.6x fewer memory allocations**

*This parser is still WIP and should not yet be used by clients of cel-go*


parser/lexer.go
```diff
@@ -0,0 +1,727 @@
...
+
+	// Operators
+	tokDot              // .
+	tokComma            // ,
+	tokMinus            // -
+	tokPlus             // +
+	tokAsterisk         // *
+	tokSlash            // /
+	tokPercent          // %
+	tokQuestion         // ?
```
TristonianJones, line 64: From ANTLR, I recall that we had to add special lexemes for character combinations like `[?` and `?.` to remove ambiguity from the ternary. I'll keep reading to see how this was addressed in the parser, but was wondering about this for potential future features as well.

parser/lexer_test.go
```diff
@@ -0,0 +1,590 @@
...
+			expected: []expectedToken{
+				{kind: tokInt, text: "0"},
+				{kind: tokWhitespace, text: " "},
+				{kind: tokUint, text: "0u"},
+				{kind: tokWhitespace, text: " "},
+				{kind: tokInt, text: "0x0"},
+			},
+		},
+		{
+			name: "StringsAndBytes",
```
TristonianJones, line 163: I believe the conformance tests also capture these cases ... maybe we should pull out some `parser` conformance tests?

parser/lexer.go
```diff
@@ -0,0 +1,727 @@
...
+		return l.setError(start, l.pos, "unexpected single '&', expected '&&'")
+	case '|':
+		l.advance(1)
+		if l.consume('|') {
+			return l.makeToken(tokLogicalOr, start, l.pos)
+		}
+		return l.setError(start, l.pos, "unexpected single '|', expected '||'")
+	case '_':
+		return l.consumeIdent()
+	case '`':
```
TristonianJones, line 708: Nice to see this in the default.

parser/pratt_parser.go
```diff
@@ -0,0 +1,1012 @@
...
+	}
+	return p, nil
+}
+
+// Parse parses the expression represented by source using the Pratt parser and returns the result.
+func (p *PrattParser) Parse(source common.Source) (*ast.AST, *common.Errors) {
+	errs := common.NewErrors(source)
+	pratt := p.newWorker(source, errs)
+	var out ast.Expr
+	if pratt.length > int32(p.expressionSizeCodePointLimit) {
```
TristonianJones: Do we end up doing a sizable allocation before this check? Maybe we should move this up to a comparison against common.Source, though I can't recall what we did with the other parser.

parser/pratt_parser.go
```diff
@@ -0,0 +1,1012 @@
...
+	return p.parseSelectorChainTail(lhs)
+}
+
+func (p *prattParser) parseSelectorChainTail(lhs ast.Expr) ast.Expr {
+	for {
+		switch p.peekTok.kind {
+		case tokDot:
+			dotTok := p.nextToken()
+			optional := false
+			if p.peekTok.kind == tokQuestion {
```
TristonianJones, line 502: Nice, this makes sense ... very tiny lookahead, definitely a win.

parser/pratt_parser.go
```diff
@@ -0,0 +1,1012 @@
...
+			ops = ops[:len(ops)-1]
+			operand = p.parseNegativeDoubleLiteral(lastOp.id)
+		default:
+			operand = p.parseSelectorChain()
+		}
+	} else {
+		operand = p.parseSelectorChain()
+	}
+
+	for i := len(ops) - 1; i >= 0; i-- {
```
TristonianJones, line 684: Why are the logical not and negate surfacing in the same loop? This feels like an odd combination

dmitriplotnikov: From the syntax perspective, they are both unary operation, so processing them separately would lead to code duplication.  In practice, we would never encounter things like `-!!--!!-42`, but the parser still needs to be able to handle them.

parser/lexer.go
dmitriplotnikov: The optional syntax is handled explicitly in the parser. There are tests for optional syntax in pratt_parser_test.go, e.g.

	{
		I: `a[?0]`,
		P: `_[?_](
			a^#1:*expr.Expr_IdentExpr#,
			0^#3:*expr.Constant_Int64Value#
		)^#2:*expr.Expr_CallExpr#`,
		Opts: []Option{EnableOptionalSyntax(true)},
	},

I wonder if it would actually be more efficient to have these lexemes.  Let me experiment separately, starting with the C++ parser, which I would like to treat as the SoT for the moment.

parser/lexer.go
dmitriplotnikov: We check the `enableIdentEscapeSyntax` option in the parser, so this is safe to recognize in the lexer unconditionally.

parser/lexer_test.go
dmitriplotnikov: For the moment, I prefer to keep the pratt parser and its tests self-contained for convenience. In a follow-up PR I will add an option to use this parser implementation.  At that point, we will ensure that *all* conformance tests pass. Once that path is available, we can consider removing all redundant tests.

parser/pratt_parser.go
dmitriplotnikov: Done

parser/pratt_parser.go
dmitriplotnikov: As mentioned above, I will experiment with defining a lexeme for this making even this tiny lookahead unnecessary.

parser/pratt_parser.go
dmitriplotnikov: Do you think we should disallow chains like `-!!--!!-42` since they will lead to typechecking or runtime errors anyway?

TristonianJones: I think it's something we should clean up as this is will certainly error at check or runtime, but it can wait for a follow-up pass.

parser/pratt_parser.go
```diff
@@ -0,0 +1,1006 @@
...
+	if len(text) == 0 {
+		return ""
+	}
+	if text[0] == '`' {
+		if !allowQuoted {
+			p.reportError(tok, "unexpected quoted identifier")
+			return ""
+		}
+		if !p.enableIdentEscapeSyntax {
+			p.reportError(tok, "unsupported syntax '`'")
```
TristonianJones, line 392: Did you want to return from here or continue?

parser/pratt_parser.go
```diff
@@ -0,0 +1,1006 @@
...
+	opID := p.nextID(qTok)
+	trueExpr := p.parseBinaryAndTernary(1)
+	if !p.expect(tokColon, "expected ':' in conditional expression") {
+		return lhs
+	}
+	falseExpr := p.parseBinaryAndTernary(0)
+	return p.helper.newGlobalCall(opID, operators.Conditional, lhs, trueExpr, falseExpr)
+}
+
+func (p *prattParser) parseBalancedLogicalChain(lhs ast.Expr, opInfo binaryOpInfo) ast.Expr {
```
TristonianJones: nit: maybe rename to `parseLogicalChain` since the output might or might not be balanced.

TristonianJones (dismissed):

parser/lexer.go
dmitriplotnikov: I tried this and discovered that having [? and ?. as lexemes actually complicates matters because we allow whitespace/comment between these characters, so now parser has to deal with tokBracketQuestion and *also* with tokBracket tokWhitespace tokQuestion.

parser/pratt_parser.go
dmitriplotnikov: Yes, it is safe to continue here.  I have extended a test to acertain:

```
	{
		I: "msg.`ident` + msg.`other`",
		E: "ERROR: <input>:1:5: unsupported syntax '`'\n" +
			" | msg.`ident` + msg.`other`\n" +
			" | ....^\n" +
			"ERROR: <input>:1:19: unsupported syntax '`'\n" +
			" | msg.`ident` + msg.`other`\n" +
			" | ..................^",
		Opts: []Option{EnableIdentEscapeSyntax(false)},
	},

```

parser/pratt_parser.go
dmitriplotnikov: Done

TristonianJones (approved):

PR #1430, opened 2026-08-21
Switch to direct observation of costs on a per-operation basis

TristonianJones:
Updates to the way costs are observed to reduce allocations an inefficiencies.

Benchmark | Master HEAD | Direct Push (Observerless) | Delta
-- | -- | -- | --
simple_comparison | 300.0 ns / 550 B / 10 allocs | 205.1 ns / 224 B / 4 allocs | -31.6% latency, -59.3% memory, -60.0% allocs
function_calls | 549.3 ns / 798 B / 14 allocs | 355.6 ns / 288 B / 7 allocs | -35.3% latency, -63.9% memory, -50.0% allocs
comprehension | 11,829 ns / 7,748 B / 148 allocs | 7,583 ns / 5,188 B / 131 allocs | -35.9% latency, -33.0% memory, -11.5% allocs
nested_comprehensions | 14,484 ns / 4,956 B / 250 allocs | 7,357 ns / 3,340 B / 106 allocs | -49.2% latency (~2x faster), -32.6% memory, -57.6% allocs

### DIFF
--- a/ext/encoders_test.go
+++ b/ext/encoders_test.go
@@ -249,7 +249,7 @@ func TestEncodersCosts(t *testing.T) {
 				"x": 100,
 			},
 			estimatedCost: checker.CostEstimate{Min: 2, Max: math.MaxUint64},
-			actualCost:    1,
+			actualCost:    math.MaxUint64,
 			version:       1,
 		},
 	}
--- a/cel/program.go
+++ b/cel/program.go
@@ -341,7 +341,7 @@ func newProgram(e *Env, a *ast.AST, opts []ProgramOption) (Program, error) {
 			observers = append(observers, interpreter.EvalStateObserver())
 		}
 		if p.evalOpts&OptTrackCost == OptTrackCost {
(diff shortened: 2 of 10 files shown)


ext/encoders_test.go
```diff
@@ -249,7 +249,7 @@ func TestEncodersCosts(t *testing.T) {
 				"x": 100,
 			},
 			estimatedCost: checker.CostEstimate{Min: 2, Max: math.MaxUint64},
-			actualCost:    1,
+			actualCost:    math.MaxUint64,
```
TristonianJones, line 252: The fact that this wasn't 1 was actually a bug.

dmitriplotnikov: Is there a test that now covers the MaxUint64 case?

dmitriplotnikov (approved):

ext/encoders_test.go
TristonianJones: Ah, this is the test ... the test assertion wasn't correct, it needed to emit MaxUint64 rather than 1

PR #1438, opened 2026-08-26
Fix panic when sizing counted regex repetitions

rolinh:
An environment configured with `RegexProgramSizeLimit` cannot handle an expression whose regex contains a counted repetition such as `[0-9]{3}`. `Env.Compile` panics with `"regexp: unhandled case in compile"` when the pattern is a literal, and evaluation fails with `"internal error: regexp: unhandled case in compile"` when the pattern comes from a variable.

The size is computed by parsing the pattern and compiling it with `regexp/syntax`. The standard library's `regexp.Compile` does the same, except that it calls `Simplify` on the parsed tree in between, which rewrites a counted repetition into its expanded form so that `a{3}` becomes `aaa`. That step is not optional: `syntax.Compile` has no case for the `OpRepeat` node the parser produces for `{n}` and panics on it.

Simplify the parsed pattern before compiling it, so the size is measured on the same program `regexp.Compile` builds.

### DIFF
--- a/cel/cel_test.go
+++ b/cel/cel_test.go
@@ -2681,6 +2681,17 @@ func TestRegexProgramSizeLimit(t *testing.T) {
 			vars: map[string]any{"pattern": "[0-9]+"},
 			want: types.True,
 		},
+		{
+			name:       "constant_counted_repetition_exceeds_limit_ast_validation",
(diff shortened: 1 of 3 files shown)


TristonianJones (approved):

TristonianJones:
@rolinh Thanks!

PR #1441, opened 2026-08-27
costs: Unified cost model and strategy

TristonianJones:
Introduce a unified cost model framework

This allows for the specification of cost semantics as a relationship between function arguments
- Helpers for common case computations, such as O(n) or O(n^2)
- Supports size tracking of recursive object structures
- Self-consistent generated trackers and estimators 
- Flexible sizing strategies for runtime and check-time


common/cost/model.go
```diff
@@ -0,0 +1,924 @@
...
+	// Target returns the size estimate for the receiver/target object, if present.
+	Target() (SizeEstimate, bool)
+
+	// Result returns the size estimate of the evaluated result, if known.
+	Result() (SizeEstimate, bool)
+
+	// Estimator returns the user-provided Estimator, if configured.
+	Estimator() Estimator
+
+	// Size returns the estimated size of the node (delegating to SizingStrategy if present).
```
dmitriplotnikov: I am not entirely clear on how a custom SizingStrategy is provided.  Or is this supposed to be "delegating to Estimator if present"?

dmitriplotnikov (dismissed):

common/cost/model.go
TristonianJones: Ah, yes, the `SizingStrategy` gets configured in the `Estimator` and this becomes more obvious in later PRs that update the top-level `cel` package ... but it should always be configured, so this comment should mention that the `SizingStrategy` determines the size always.

dmitriplotnikov (dismissed):

dmitriplotnikov (approved):

dmitriplotnikov (approved):

PR #1459, opened 2026-08-31
costs: Migrate ext/network to the common cost APIs

TristonianJones:
Migrate ext/network to the common cost APIs

### DIFF
--- a/ext/network_test.go
+++ b/ext/network_test.go
@@ -20,7 +20,7 @@ import (
 	"testing"
 
 	"cel.dev/cel-go/cel"
-	"cel.dev/cel-go/checker"
+	"cel.dev/cel-go/common/cost"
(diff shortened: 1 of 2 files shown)


ext/network_test.go
```diff
@@ -738,151 +738,151 @@ func TestNetworkCost(t *testing.T) {
 	tests := []struct {
 		name          string
 		expr          string
-		estimatedCost checker.CostEstimate
+		estimatedCost cost.CostEstimate
```
maskri17, line 741: for readability purposes, should we just name it: cost.Estimate

maskri17 (approved):

ext/network_test.go
TristonianJones: I hear you, but it wound up conflicting with a public method for performing estimation, so I couldn't do that.

PR #1467, opened 2026-09-03
Structural equivalence utilities for CEL ASTs

### LINKED ISSUE #1368: Provide formatting-insensitive structural equality for parsed CEL expressions
**Feature request checklist**

**Change**

Provide, or document an existing supported mechanism for, comparing two
parsed CEL expressions by structure while ignoring source formatting and
parse metadata.

The intended comparison would:

- ignore whitespace, redundant parentheses, source positions, and expression
  node IDs;
- treat equivalent literal spellings as equal when they parse to the same
  literal value;
- work for macro-backed expressions such as `exists`, `all`, and `map`;
- compare expression structure rather than rendered text; and
- avoid logical or algebraic equivalence. For example, `a && b` and `b && a`
  would remain different.

A concrete downstream use case is configuration reconciliation. A Terraform
provider may receive a CEL expression from configuration, while the backing
API returns the same expression with different whitespace or fully
parenthesized formatting. Treating those values as different produces a
perpetual plan diff even though the parsed expression has not changed.

One example of this use case and its current workaround is:

https://github.com/wandb/terraform-provider-orca/pull/32

The current cel-go APIs provide the underlying pieces, but not an explicit
equality contract:

- `cel.AstToString` produces stable, semantically equivalent text when it can
  unparse the expression.
- `cel.EnableMacroCallTracking` allows macro calls to be reconstructed during
  unparsing.
- Comparing the expression protobufs with `proto.Equal` works for the cases
  tested, but it is unclear whether equality across independently parsed ASTs,
  including the treatment of expression IDs, is a supported guarantee.
- `(*cel.Ast).Expr()` is deprecated in favor of `cel.AstToParsedExpr`, but
  neither API documents structural comparison as a supported use case.

An explicit helper would let embedders use this behavior without depending on
unparser behavior or undocumented protobuf details.

**Example**

A possible API might look like this; the exact name and package are open for
discussion:

```go
env, _ := cel.NewEnv()

left, _ := env.Parse(`
    resource.kind == "service"
      && resource.name == "api"
`)
right, _ := env.Parse(`
    (resource.kind == "service") &&
    (resource.name == "api")
`)

same := cel.AstStructurallyEqual(left, right)
// same == true

macroLeft, _ := env.Parse(`[1,2].exists(x,x>1)`)
macroRight, _ := env.Parse(`([1, 2].exists(x, x > 1))`)

sameMacro := cel.AstStructurallyEqual(macroLeft, macroRight)
// sameMacro == true

different, _ := env.Parse(`resource.name == "worker"`)

sameExpression := cel.AstStructurallyEqual(left, different)
// sameExpression == false
```

The implementation could instead live in `common/ast`, for example:

```go
func ExprEqual(left, right ast.Expr) bool
```

The important part is a documented contract that ignores IDs and source
metadata while comparing the parsed expression structure.

**Alternatives considered**

1. **Compare `cel.AstToString` output**

   This works for many expressions when macro-call tracking is enabled, but
   equality becomes dependent on an unparser that can return errors. Unparser
   behavior also evolves independently; for example, #1325 required #1326 to
   fix invalid output for doubles in scientific notation.

2. **Compare expression protobufs with `proto.Equal`**

   Comparing only the `Expr` fields from `cel.AstToParsedExpr` works for the
   formatting, parenthesization, and macro cases tested. However, it is not
   documented whether expression IDs are guaranteed to be stable across
   independent parses or whether this is a supported equality mechanism.

   If this is already the recommended approach, documenting that guarantee
   would satisfy this request without adding another API.

3. **Implement an application-specific recursive AST visitor**

   This avoids unparsing but duplicates cel-go's handling of expression kinds,
   literals, comprehensions, optional syntax, and future AST additions in each
   embedding application.

4. **Continue comparing raw source strings**

   This cannot distinguish formatting-only changes from actual expression
   changes and causes persistent configuration diffs.

TristonianJones:
Format insensitive equivalent testing for CEL expressions and ASTs

Closes #1368

### DIFF
--- a/common/ast/equiv.go
+++ b/common/ast/equiv.go
@@ -1,0 +1,352 @@
+// Copyright 2024 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package ast
+
+import (
+	"math"
+
+	"cel.dev/cel-go/common/types"
+	"cel.dev/cel-go/common/types/ref"
+)
+
+// EquivOption configures the structural equivalence comparison between ASTs or expressions.
+type EquivOption func(*equivOptions)
+
+// EquivTypes configures the type maps to use for validating type equivalence between two expressions or ASTs.
+func EquivTypes(aTypes, bTypes map[int64]*types.Type) EquivOption {
+	return func(opts *equivOptions) {
+		opts.aTypes = aTypes
+		opts.bTypes = bTypes
+		opts.checkTypes = aTypes != nil || bTypes != nil
+	}
+}
+
+// EquivReferences configures the reference maps to use for validating reference equivalence between two expressions or ASTs.
+func EquivReferences(aRefs, bRefs map[int64]*ReferenceInfo) EquivOption {
+	return func(opts *equivOptions) {
+		opts.aRefs = aRefs
+		opts.bRefs = bRefs
+		opts.checkRefs = aRefs != nil || bRefs != nil
+	}
+}
+
+// EquivIgnoreNames configures whether identifier and field names should be ignored during equivalence comparison.
+func EquivIgnoreNames(enabled ...bool) EquivOption {
+	return func(opts *equivOptions) {
+		if len(enabled) == 0 {
+			opts.ignoreNames = true
+			return
+		}
+		opts.ignoreNames = enabled[0]
+	}
+}
+
+type equivOptions struct {
+	aTypes      map[int64]*types.Type
+	bTypes      map[int64]*types.Type
+	aRefs       map[int64]*ReferenceInfo
+	bRefs       map[int64]*ReferenceInfo
+	checkTypes  bool
+	checkRefs   bool
+	ignoreNames bool
+}
+
+// EquivAST determines whether two ASTs are structurally equivalent.
+//
+// The comparison ignores numeric IDs except for resolving references and types.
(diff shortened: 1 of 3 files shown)


common/ast/equiv_test.go
```diff
@@ -0,0 +1,879 @@
...
+
+	if !ast.EquivExpr(e1, e2, ast.EquivReferences(fnRef1, fnRef2)) {
+		t.Errorf("ast.EquivExpr with reordered overloads = false, want true")
+	}
+	if ast.EquivExpr(e1, e2, ast.EquivReferences(fnRef1, fnRefDiff)) {
+		t.Errorf("ast.EquivExpr with different overloads = true, want false")
+	}
+}
+
+func TestEquivIgnoreIdentifiers(t *testing.T) {
```
jnthntatum, line 292: I think I'd expect you need to keep a mapping to make sure the identifiers are consistent (but tolerate them being renamed).

Would something like the following trip this up?

```
[2, 3, 4].all(outer, [0, 1, 2].all(inner, outer > inner))

[2, 3, 4].all(inner, [0, 1, 2].all(outer, outer > inner))
```

common/ast/equiv.go
```diff
@@ -0,0 +1,405 @@
...
+	return func(opts *equivOptions) {
+		if len(enabled) == 0 {
+			opts.ignoreIdentifiers = true
+			return
+		}
+		opts.ignoreIdentifiers = enabled[0]
+	}
+}
+
+type equivOptions struct {
```
jnthntatum, line 79: Do you want to add an option to support checking macro calls?

A couple translation use cases rely on reversing macros (though maybe it's special case enough to ignore until asked for).

TristonianJones: I was thinking of ignoring it for now if the structural outputs were the same, but it does make sense when we have to change to macro definition for some reason.

common/ast/equiv_test.go
TristonianJones: Yeah, I'll fix this to be tighter than it is now. The logic would currently miss the semantic shift of switching around identifiers

TristonianJones:
@jnthntatum - Updated with the fix and added feature for macro call-based equivalence. PTAL

jnthntatum (approved):

PR #1473, opened 2026-09-10
Document the Strings `format` function in ext/README.md

### LINKED ISSUE #1116: `string.format` missing from cel-go/ext/README.md
**Describe the bug**

`<string>.format` is supported in the Strings extension (see https://github.com/google/cel-go/blob/master/ext/strings.go#L57-L102)

But it looks like it is accidentally omitted from the README: https://github.com/google/cel-go/blob/master/ext/README.md

Sanjays2402:
Fixes #1116

The Strings extension supports `<string>.format(<list>)` since version 1
(ext/strings.go), but the function was accidentally omitted from the
documented Strings functions in ext/README.md. This adds the missing
documentation entry, matching the implementation's documented behavior.

Test evidence:
- No code changed, so no unit tests were added. Instead, every example in
  the new README section was evaluated against the interpreter
  (`cel.NewEnv(ext.Strings())`) and the documented outputs verified:
  - `"this is a string: %s and an integer: %d".format(["str", 42])` -> `"this is a string: str and an integer: 42"`
  - `"%f".format([3.14])` -> `"3.140000"`
  - `"5 in binary: %b".format([5])` -> `"5 in binary: 101"`
  - `"26 in hex: %x".format([26])` -> `"26 in hex: 1a"`
  - `"26 in hex (uppercase): %X".format([26])` -> `"26 in hex (uppercase): 1A"`
  - `"30 in octal: %o".format([30])` -> `"30 in octal: 36"`
  - `"duration: %s".format([duration("1h45m47s")])` -> `"duration: 6347s"`


ext/README.md
```diff
@@ -942,6 +942,46 @@ Examples:
     'hello'.charAt(5)  // return ''
     'hello'.charAt(-1) // error
 
+### Format
+
+**Introduced in version 1**
```
TristonianJones, line 947: Even though this has been present since version 1, would you mind indicating the v4 revision that removes locale support and fixes formatting? It's a much better implementation and how the other stacks are implemented.

Sanjays2402: Good point — added a note to the Format section: revised in version 4, where formatting was updated to match the CEL spec and locale support was removed (`StringsLocale` is ignored at version 4+).

TristonianJones (approved):

TristonianJones:
Thanks @Sanjays2402!

PR #1479, opened 2026-09-10
Fix type unification for type parameters

l46kok:
Port of 
https://github.com/cel-expr/cel-java/pull/1211
https://github.com/cel-expr/cel-cpp/pull/2260


checker/types.go
```diff
@@ -143,7 +162,24 @@ func internalIsAssignable(m *mapping, t1, t2 *types.Type) bool {
...
 		return t2.IsAssignableType(t1)
 	case types.TypeKind:
-		return kind2 == types.TypeKind
+		if kind2 != types.TypeKind {
+			return false
+		}
+		if len(t1.Parameters()) == 0 || len(t2.Parameters()) == 0 {
+			return len(t2.Parameters()) == 0
+		}
+		fromType := t1.Parameters()[0]
```
TristonianJones, line 179: This seems to assume that type only contains a single parameter, so we should check that this is the case  and return false if there's a mismatch.

checker/types.go
```diff
@@ -143,7 +162,24 @@ func internalIsAssignable(m *mapping, t1, t2 *types.Type) bool {
...
+		toType := t2.Parameters()[0]
+		// If either type contains a type parameter (e.g., type(T) in foo(data, type(T)) -> T),
+		// delegate to inner type unification to bind or validate type parameter substitutions.
+		// Returns true if the inner types structurally match, unify with an unbound type param,
+		// or conform to an existing binding in 'm'. Returns false on structural/kind mismatches
+		// (e.g., int vs list(T)), occurs-check cycles, or conflicting type param bindings.
+		if hasTypeParam(fromType) || hasTypeParam(toType) {
+			return internalIsAssignable(m, fromType, toType)
```
TristonianJones, line 187: I think there might be some issue with the `occurs` check with inputs like:

```
type(map(string, T)), type(T)
```

l46kok: We should be ok, the type substitution check propagates for map parameters. I've added some tests around this just to be sure.

TristonianJones (approved):

PR #1483, opened 2026-09-11
Support for late-bound functions resolved from the Activation

### LINKED ISSUE #356: Late Function Bindings in Go Cel Evaluation

TristonianJones:
Support for late bound functions provided as variables on the `Activation`

Closes #356


cel/program.go
```diff
@@ -115,6 +116,55 @@ func PartialVars(vars any,
...
+//
+// The `vars` value may either be an Activation or any valid input to the NewActivation call.
+//
+// Late-bound functions apply to a whole evaluation rather than to a lexical scope, so only one
+// activation may supply them. An error is returned when `vars` already does, whether directly or
+// through a partial or hierarchical activation which wraps one; supply the complete set of
+// bindings in a single call instead. The same restriction applies to the combination of Globals
+// and the activation passed to Eval, which is reported by Eval.
+//
+// Since the bindings are supplied for a single evaluation alongside the variables of that
```
jnthntatum: This is reads a bit strange to me, I might suggest something more direct like:

```

One of the motivations for late-bound functions is to support cases where the function behavior depends
on the context of a particular evaluation. This can be accomplished by closing over some state that is not
otherwise provided to CEL in the function definition added to the activation.

example:
...

CEL expects that within a given evaluation a function behaves the same given the same arguments, even 
if late-bound. For example, a function comparing against a validity window `valid_until(t)` could be implemented as a late bound function closing over a fixed timestamp value.

```


More of a preference up to you if you want to revise.

jnthntatum (dismissed):

cel/program.go
TristonianJones: Updated to make more concise and add clarity around expected behaviors.

jnthntatum (dismissed):

TristonianJones:
PTAL, the merge conflicts and docs should be in good shape now.

jnthntatum (approved):

PR #1485, opened 2026-09-11
AST Matcher documentation

TristonianJones:
Create documentation for the AST Matcher utilities.

### DIFF
--- a/docs/ast-matchers.md
+++ b/docs/ast-matchers.md
@@ -1,0 +1,349 @@
+# AST Pattern Matcher Design
+
+## 1. Overview
+
+CEL tooling frequently needs to recognize the *shape* of an expression:
+linters detect anti-patterns such as `_1 == _1`, optimizers extract
+comprehension loop steps, and validators reject disallowed sub-expressions.
+
+`ast.EquivAST` and `ast.EquivExpr` in
+[`common/ast/equiv.go`](../common/ast/equiv.go) answer whether two ASTs are
+structurally equivalent. The matcher generalizes that comparison: one side
+becomes a *pattern* containing **slots** that match and capture arbitrary
+sub-expressions.
+
+Patterns are written in ordinary CEL. Sentinel identifiers (`_`, `_0`..`_9`)
+and receiver macros supply the pattern vocabulary, so no grammar or lexer
+changes are required.
+
+> [!NOTE]
+> Matching is structural. Proof of semantic equivalence is a separate concern,
+> available from the CEL-Java verifier tool.
+
+---
+
+## 2. Pattern Syntax
+
+### 2.1 Slots
+
+| Concept | Syntax | Matches |
+| :--- | :--- | :--- |
+| Anonymous wildcard | `_` | Any sub-expression, captures nothing. |
+| Positional slot | `_0` .. `_9` | Any sub-expression, captured by index. |
+| Kind constraint | `_1.exprKind(ident)` | Nodes of a given `ast.ExprKind`. |
+| Type constraint | `_1.type(int)` | Nodes of a given checked type. |
+| Collections | `[_1.star()]`, `{_1.star(): _2.star()}` | List elements and map entries. |
+
+```mermaid
+graph TD
+    A["Pattern Source String"] --> B["Pratt Parser + Pattern Macros"]
+    B --> C["Pattern AST"]
+    C --> D["CompileSubtree & Slot Extraction"]
+    D --> E["Compiled Pattern"]
+    E --> E1["Slots Table: _0.._9"]
+    E --> E2["Pre-Indexed slotByID Array"]
+    E --> E3["Macro Calls Cache"]
+```
+
+Slot indices are limited to `_0`..`_9`, which lets captures live in fixed-size
+arrays rather than maps. `_10` and beyond are rejected at compile time.
+
+#### Kind Constraints
+
+`exprKind` accepts `ident`, `call`, `select`, `list`, `map`, `struct`,
+`literal`, and `comprehension`. Struct matching may be narrowed to a single
+message type:
+
+```cel
+_1.exprKind(struct, google.protobuf.Duration)
+```
+
+#### Type Constraints
+
+`type` accepts any name the configured `types.Provider` can resolve, such as
+`_1.type(int)` or `_1.type(google.protobuf.Duration)`. Unresolvable names fail
+at compile time, and matching requires a checked AST.
+
+#### Validation
+
+Repeating `exprKind` or `type` on one slot, or combining contradictory bounds
+(`.star().plus()`, `.atLeast(5).atMost(2)`), is a compile error.
+
+---
+
+### 2.2 Quantifiers
+
+Within call arguments, list elements, and map entries, a slot may match a
+contiguous run of nodes.
+
+```mermaid
+classDiagram
+    class SlotQuantifier {
+        +int minOccurs
+        +int maxOccurs
+        +bool hasMinOccurs
+        +bool hasMaxOccurs
+    }
+    class Macros {
+        +optional() -> atMost(1)
+        +star() -> atLeast(0)
+        +plus() -> atLeast(1)
+        +repeated(n) -> atLeast(n).atMost(n)
+    }
+    Macros --> SlotQuantifier : expands during parse
+```
+
+`atLeast(m)` and `atMost(n)` set the bounds directly. The receiver macros
+expand into those calls during parsing:
+
+| Macro | Expansion | Cardinality |
+| :--- | :--- | :--- |
+| `_1.optional()` | `_1.atMost(1)` | 0 or 1 |
+| `_1.star()` | `_1.atLeast(0)` | 0 or more |
+| `_1.plus()` | `_1.atLeast(1)` | 1 or more |
+| `_1.repeated(n)` | `_1.atLeast(n).atMost(n)` | exactly n |
+
+Variable quantifiers are only valid in the final position of a sequence.
+Fixed-cardinality slots such as `_1` and `_1.repeated(2)` may appear anywhere,
+so `[_1.repeated(2), _2.star()]` compiles while `[_1.star(), _2]` does not.
+
+---
+
+### 2.3 Matching Rules
+
+| Rule | Example |
+| :--- | :--- |
+| Map entries match in order | `{'a': _1, 'b': _2}` matches `{'a': 1, 'b': 2}`, not `{'b': 2, 'a': 1}` |
+| Entry quantifiers bind key and value | `{_1.star(): _2.star()}` |
+| Struct fields match in order, exhaustively | `pkg.Msg{a: _1, b: _2}` matches only two-field messages |
+| Optionality must agree | `{?'k': _1}` matches `{?'k': v}`, not `{'k': v}` |
+| Qualified calls are folded | `pkg.fn(_1)` matches `pkg.fn(x)` |
+| Macros match expanded or unexpanded targets | `_1.filter(x, true)` matches with or without macro call tracking |
+
+---
+
+## 3. Using the Matcher
+
+### 3.1 Compiling
+
+| Constructor | Input |
+| :--- | :--- |
+| `Compile` / `MustCompile` | Pattern source |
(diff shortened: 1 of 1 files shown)


docs/ast-matchers.md
```diff
@@ -0,0 +1,349 @@
...
+Fixed-cardinality slots such as `_1` and `_1.repeated(2)` may appear anywhere,
+so `[_1.repeated(2), _2.star()]` compiles while `[_1.star(), _2]` does not.
+
+---
+
+### 2.3 Matching Rules
+
+| Rule | Example |
+| :--- | :--- |
+| Map entries match in order | `{'a': _1, 'b': _2}` matches `{'a': 1, 'b': 2}`, not `{'b': 2, 'a': 1}` |
```
l46kok, line 116: Is the ordering requirement there to simplify implementation or is there something structural that prevents us from an unordered match? Same question for struct fields

docs/ast-matchers.md
```diff
@@ -0,0 +1,349 @@
...
+becomes a *pattern* containing **slots** that match and capture arbitrary
+sub-expressions.
+
+Patterns are written in ordinary CEL. Sentinel identifiers (`_`, `_0`..`_9`)
+and receiver macros supply the pattern vocabulary, so no grammar or lexer
+changes are required.
+
+> [!NOTE]
+> Matching is structural. Proof of semantic equivalence is a separate concern,
+> available from the CEL-Java verifier tool.
```
l46kok, line 21: nit: consider linking in documentation

docs/ast-matchers.md
```diff
@@ -0,0 +1,349 @@
...
+    }
+    Macros --> SlotQuantifier : expands during parse
+```
+
+`atLeast(m)` and `atMost(n)` set the bounds directly. The receiver macros
+expand into those calls during parsing:
+
+| Macro | Expansion | Cardinality |
+| :--- | :--- | :--- |
+| `_1.optional()` | `_1.atMost(1)` | 0 or 1 |
```
l46kok, line 101: Should we use something other than `optional` (or maybe just exclude it)? To avoid confusion and collision with our first class optional type.

docs/ast-matchers.md
```diff
@@ -0,0 +1,349 @@
...
+   options override the defaults.
+2. **Slot consistency**: a slot repeated in a pattern must bind equivalent
+   expressions, so `_1 == _1` matches `claims.sub == claims.sub` but not
+   `claims.sub == claims.iss`.
+3. **Macro matching**: macro calls recorded in `SourceInfo.MacroCalls()` are
+   compared directly, and comprehension variables are unified, so
+   `_1.all(x, x > 0)` matches `items.all(it, it > 0)`.
+4. **Map sequences**: entries match in order and each consumes one key and one
+   value, so `{_1.star(): _2.star()}` matches any map while `{_1: _2.star()}`
+   matches a single entry.
```
l46kok, line 266: I'm having a hard time grokking this section, maybe an explanation might help.

Why does the following need a star? What's the difference between the following two?

```
// We say this matches a single entry
{_1: _2.star()}

// What would this match differently versus the above?  
{_1: _2}
```

Also, `.star()` is supposed to match "0 or more". Does this mean this will also match an empty map? (since a map entry cannot have an empty value).

docs/ast-matchers.md
```diff
@@ -0,0 +1,349 @@
...
+
+Repeating `_2` in both branches ties the accumulator to a single expression.
+
+---
+
+## 7. Future Work
+
+### 7.1 Declarative AST Optimizers & Term Rewriters
+
+Authoring optimizers still requires procedural transformation logic. A
```
l46kok, line 320: Food for thought -- the pattern matching currently is agnostic of their types. 

My question: should it? There are two reasons why I bring it up for consideration

1. Things like maps or field selections are type constrained (map keys must be comparable, field selections must be done on strings, with a few exceptions like indexor or backtick accessors). An untyped matching in either case might be too permissive
2. Thinking ahead for optimization cases -- there is a class of mutations that's unsafe type wise (dyn_x in list, dyn_x && true, associativity rewrite that can lead to precision errors or overflows etc).

docs/ast-matchers.md
TristonianJones: It's a good question. The order matches what appears within the `ast.Equiv`. Currently, it's pretty strict and requires matching in the exact order of declaration. If we loosen this requirement for maps and structs, then `ast.Equiv` becomes more powerful and the matcher _should_ inherit the same leniency in match order within these types.

docs/ast-matchers.md
TristonianJones: I have it here mostly for equivalence with regex `?`, but I'm okay with just saying `atMost`.

docs/ast-matchers.md
TristonianJones: Since `_1` always "exactly one", the `_2.star()` could only ever match a single entry for a valid map-literal. In your example, the behavior would be identical to `{_1: _2}`. Whereas `{_1.star(): _2.star()}` would match any shape of map (empty, 1, N). Does that explanation make sense?

docs/ast-matchers.md
TristonianJones: It's possible to use `type(<type-name>)` to constrain the type of match, e.g. `_1.type(string) == null`; however, the matcher should always be applied to a type-checked expression so the expectation is that you're matching against something where these constraints already hold. Since the match is only intended to figure out which expr nodes in the expression to extract, there may be cases where the content needs to be inspected post-match. 

I think `dyn` would use the `matches everything` behavior, but we could make matchers be much more specific: 'only match dyn typed nodes', or `_1.notType(dyn) && _2` as the case may be. What do you think?

docs/ast-matchers.md
l46kok: Given that `{_1: _2}` effectively behaves the same as `{_1: _2.star()}` (and `{_1.star(): _2}`, though this case is interesting type-constraint wise on the key), would it be better to stay opinionated and just treat those as an error?. It would connote better that a map must have an atomic entry quantifier (having a star invites cardinality into consideration on either key/value).

I honestly think `{_1.star(): _2.star()}` is pretty confusing too, but maybe this one's ok if there's no other way of doing a partial map match (empty, 1, N case). Or would `_.exprKind(map)` work?

l46kok: `{_1.star(): _2.star()}` is probably ok, for doing things like "matching that a specific key exists, followed by any number of optional entries (ex: `{'env': 'prod', _1.star(): _2.star()}`)

TristonianJones: `_1.exprKind(map)` will catch the entire map, `{_.star(): _1.star()}` will catch all values, `{_1.star(): _.star()}` will catch all keys. Expressions like `{_1.atMost(3): _2.atMost(3)}` is interesting, `{_1.type(string).atMost(3): _2.type(string).atMost(3)}` are also interesting. It's not perfect, but not bad considering the behavior is well-defined.

docs/ast-matchers.md
l46kok: We can defer, but just from an expected behavior standpoint, I'd prefer seeing unordered match supported at some point (since maps and struct fields are unordered in CEL). In relation to any key/value matching, it would prevent surprises like:

matcher: `{'env': 'prod', _1.star(): _2.star()}`

matches: `{'env': 'prod', 'region': 'us'}`
does not match: `{'region': 'us', 'env': 'prod'}`

docs/ast-matchers.md
l46kok: Interesting, I think that would imply the following about the cardinality:

```
{_1.atMost(2): _2.atMost(5)} -> matches at most 2 entries?
{_1.plus(): _2.star()} -> matches at least 1 entry
{_1: _2.star()} -> `matches exactly 1 entry`
```

I would think at least the first case might be better served as a compilation error (since there's no real benefit in allowing users in writing a disjoint bound instead of a uniform one).

docs/ast-matchers.md
l46kok: `notType` does sound useful (`dyns` are generally where the myriads of edge cases come from wrt/ rewriting). I'd be in favor of it .

docs/ast-matchers.md
```diff
@@ -0,0 +1,349 @@
...
+## 2. Pattern Syntax
+
+### 2.1 Slots
+
+| Concept | Syntax | Matches |
+| :--- | :--- | :--- |
+| Anonymous wildcard | `_` | Any sub-expression, captures nothing. |
+| Positional slot | `_0` .. `_9` | Any sub-expression, captured by index. |
+| Kind constraint | `_1.exprKind(ident)` | Nodes of a given `ast.ExprKind`. |
+| Type constraint | `_1.type(int)` | Nodes of a given checked type. |
```
jnthntatum: You might consider expanding this a bit (although could be a future addition).

For cases like list.sort / sortBy, would be nice to have a concise way to express type constraints like
`_<_` is defined for `~T`.

docs/ast-matchers.md
```diff
@@ -0,0 +1,349 @@
...
+| `_1.filter(x, true)` | `users.filter(u, true)` | `_1 = users` |
+| `_1.filter(x, _2).filter(x, _3)` | `users.filter(u, u.active).filter(u, u.age >= 18)` | `_1 = users`, `_2 = u.active`, `_3 = u.age >= 18` |
+
+### 6.3 Sequences & Collections
+
+| Pattern | Expression | Captures |
+| :--- | :--- | :--- |
+| `concat(_1, _2.star())` | `concat("prefix", a, b, "suffix")` | `_1 = "prefix"`, `_2 = [a, b, "suffix"]` |
+| `slice(_1, _2, _3.optional())` | `slice(items, 0)` | `_1 = items`, `_2 = 0` |
+| `[_1.repeated(2), _2.star()]` | `[10, 20, 30, 40]` | `_1 = [10, 20]`, `_2 = [30, 40]` |
```
jnthntatum: Would you mind working a few examples mixing constraints and quantifiers?

top of mind example:
```
cel.@attribute(
  _1,  // operand
  [
    _2.exprKind(list).plus()  // select path
   ]
)
```

docs/ast-matchers.md
jnthntatum: I also find this hard to follow. As an alternative you could alter the match syntax for maps to put the quantifier on an entry match?

This diverges more from normal CEL syntax but it's a little easier to read for me.

```
match_map({_1: _2}.star())
match_map({'prefix': _1.type(int)}, {?_2: _3}.star())
```

docs/ast-matchers.md
```diff
@@ -0,0 +1,349 @@
...
+### 3.2 Matching
+
+| Method | Target | Notes |
+| :--- | :--- | :--- |
+| `Match` | `*ast.AST` | Matches the root; required for `type()` constraints. |
+| `MatchExpr` | `ast.Expr` | Matches any sub-expression, such as a loop step. |
+| `MatchAll` / `MatchAllExpr` | either | Invokes a handler per match; return `false` to stop. |
+| `FindAll` / `FindAllExpr` | either | Collects matches into a slice. |
+
+Traversal is in-order and matches may overlap. `MatchDisjoint()` skips matched
```
jnthntatum: You mean this is always top-down? Are there any use cases that might prefer leaf to root ordering?

docs/ast-matchers.md
```diff
@@ -0,0 +1,349 @@
...
+arrays rather than maps. `_10` and beyond are rejected at compile time.
+
+#### Kind Constraints
+
+`exprKind` accepts `ident`, `call`, `select`, `list`, `map`, `struct`,
+`literal`, and `comprehension`. Struct matching may be narrowed to a single
+message type:
+
+```cel
+_1.exprKind(struct, google.protobuf.Duration)
```
jnthntatum: Would you mind adding a few examples for member calls and field selections e.g. `_1.matches(_2)` ? I think the current set of constraints and quantifiers are pretty unambiguous, but if you expect to expand pattern matching over time we might want to think about how to distinguish between the unique pattern matching concepts vs normal CEL in the pattern expression?

PR #1487, opened 2026-09-11
Make the optMap / optFlatMap fixes versioned and backwards compatible

### LINKED ISSUE #1476: Non backward compatible cost changes in release 0.32.0
**Describe the bug**
Recent updates in cel-go introduce changes to macro shapes and baseline costs that result in unexpected cost increases for existing CEL expressions.

While cost reductions are easily adoptable, unexpected cost increases break backward compatibility for downstream consumers like Kubernetes, causing valid existing policies to fail validation checks.

The details are in https://github.com/kubernetes/kubernetes/pull/141830/
Cost changes observed due to the version bump:

| | Expression | Before → After |
|---|---|---|
| runtime | `self.listMap.exists(m, m.v2 == 'z')` | 24 → 22 |
| runtime | `!self.listMap.all(m, m.v2 != 'z')` | 22 → 20 |
| runtime | `optional.of('a').optMap(v, v == 'value').hasValue()` | 8 → **9** |
| estimated | `optional.of('a').optMap(v, v == 'value').hasValue()` | 18 → **29** |
| estimated | `optional.of('abcdefgabcdefg').optMap(v, v.size()).hasValue()` | 18 → **29** |
| estimated | `optional.of([1, 2, 3, 4, 5]).optMap(v, v.size()).hasValue()` | 38 → **39** |
| estimated | `self.l[?0].optMap(v, v == 'a').hasValue()` | 22 → **31** |

**To Reproduce**

- [ ]  parser
- [ ]  checker
- [ ]  interpreter

```cel
// sample expression string
```

Test setup:
```go
// test case in go
```

**Expected behavior**
Provide a flag-guarded option or maintain the previous behavior for the affected cost changes in release 0.32 to allow downstream users to adopt cost model adjustments gradually.

**Additional context**

TristonianJones:
Fixes #1476

### DIFF
--- a/cel/library.go
+++ b/cel/library.go
@@ -388,6 +388,14 @@ func (*stdLibrary) ProgramOptions() []ProgramOption {
 // optional.of(optional.of(42)).hasValue(optional.of(42)) // true
 // optional.of(42).hasValue(dyn(optional.of(42)))         // false
 // optional.of(42).hasValue(dyn(42))                      // true
+//
+// # OptMap and OptFlatMap efficiency improvements
+//
+// Introduced in version: 4
+//
+// In version 4 and later, optMap and optFlatMap avoid evaluating complex target
+// expressions more than once by binding the target to a variable in an enclosing
+// comprehension when the target is not a simple identifier.
 func OptionalTypes(opts ...OptionalTypesOption) EnvOption {
 	lib := &optionalLib{version: math.MaxUint32}
 	for _, opt := range opts {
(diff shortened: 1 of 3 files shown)


dmitriplotnikov (approved):

cel/library.go
```diff
@@ -388,6 +388,14 @@ func (*stdLibrary) ProgramOptions() []ProgramOption {
 // optional.of(optional.of(42)).hasValue(optional.of(42)) // true
 // optional.of(42).hasValue(dyn(optional.of(42)))         // false
 // optional.of(42).hasValue(dyn(42))                      // true
+//
+// # OptMap and OptFlatMap efficiency improvements
+//
+// Introduced in version: 4
+//
+// In version 4 and later, optMap and optFlatMap avoid evaluating complex target
```
dmitriplotnikov: nit: Perhaps add a brief explanation of why this cannot be done unconditionally?

dmitriplotnikov (dismissed):

cel/library.go
TristonianJones: Updated.

dmitriplotnikov (approved):

PR #1490, opened 2026-09-15
Remove hot-paths in unknown and error checking

TristonianJones:
Currently the unknown and error checking are consuming about 3ns of evaluation
time per call, which adds up quickly within comprehensions and more complex
expressions.


interpreter/interpretable.go
```diff
@@ -353,10 +353,14 @@ func (or *evalOr) Exec(frame *ExecutionFrame) ref.Val {
 			return types.True
 		}
 		if !ok {
-			isUnk := false
-			unk, isUnk = types.MaybeMergeUnknowns(val, unk)
-			if !isUnk && err == nil {
-				if types.IsError(val) {
+			if unkVal, isUnk := val.(*types.Unknown); isUnk {
+				if unk == nil {
```
dmitriplotnikov, line 357: is `isUnk` guaranteed to be true here?

interpreter/interpretable.go
```diff
@@ -401,10 +405,14 @@ func (and *evalAnd) Exec(frame *ExecutionFrame) ref.Val {
 			return types.False
 		}
 		if !ok {
-			isUnk := false
-			unk, isUnk = types.MaybeMergeUnknowns(val, unk)
-			if !isUnk && err == nil {
-				if types.IsError(val) {
+			if unkVal, isUnk := val.(*types.Unknown); isUnk {
+				if unk == nil {
+					unk = unkVal
```
dmitriplotnikov, line 410: Ditto

interpreter/interpretable.go
TristonianJones: When you type cast, you can add an optional second argument that returns `true` if the cast was successful. Since the `isUnk` appears in the condition at the end, it means that `isUnk == true` and thus `unkVal` is an unknown value.

dmitriplotnikov (approved):

interpreter/interpretable.go
TristonianJones: Yep, guaranteed :)

interpreter/interpretable.go
```diff
@@ -1354,10 +1470,14 @@ func (or *evalExhaustiveOr) Exec(frame *ExecutionFrame) ref.Val {
 			isTrue = true
 		}
 		if !ok && !isTrue {
-			isUnk := false
-			unk, isUnk = types.MaybeMergeUnknowns(val, unk)
-			if !isUnk && err == nil {
-				if types.IsError(val) {
+			if unkVal, isUnk := val.(*types.Unknown); isUnk {
+				if unk == nil {
```
dmitriplotnikov, line 1474: ditto

interpreter/interpretable.go
```diff
@@ -1406,10 +1526,14 @@ func (and *evalExhaustiveAnd) Exec(frame *ExecutionFrame) ref.Val {
 			isFalse = true
 		}
 		if !ok && !isFalse {
-			isUnk := false
-			unk, isUnk = types.MaybeMergeUnknowns(val, unk)
-			if !isUnk && err == nil {
-				if types.IsError(val) {
+			if unkVal, isUnk := val.(*types.Unknown); isUnk {
+				if unk == nil {
```
dmitriplotnikov, line 1530: Ditto

interpreter/interpretable.go
dmitriplotnikov: Didn't notice the `;isUnk` part of the condition.

PR #1491, opened 2026-09-15
Support fast-path identifier resolution for unambiguous identifiers

TristonianJones:
Improve variable resolution by adding fast-path identifier resolution
when the identifier is unambiguous.


interpreter/attributes.go
```diff
@@ -617,14 +732,27 @@ func (a *relativeAttribute) QualifyIfPresent(vars Activation, obj any, presenceO
...
+	frame, ok := vars.(*ExecutionFrame)
+	if !ok {
+		var err error
+		frame, err = NewExecutionFrame(vars)
+		if err != nil {
+			return nil, err
+		}
+		defer frame.Close()
+		vars = frame
+	}
```
l46kok, line 744: Since the interpretable v2 introduction, I think we've been using `AsFrame` to promote activations. Do we want to follow the same pattern here? It would persist cost tracker and eval state appropriately too I think.

TristonianJones: In #1489 I moved all core Activation logic up into `ExecutionFrame` and no longer treat them as separate values. As a result, I removed all references to `AsFrame`. This PR was split from those ones, but preserved this small change as a by-product of the split. In practice the `if !ok { ... }` block should never occur; however, there is a small possibility that someone could use a decorator to produce a new local `Activation` and attempt to evaluate against it. It's super unlikely, but this is purely defensive while retaining the appropriate lifecycle for pooled frames.

l46kok (dismissed):

TristonianJones:
@l46kok PTAL, I fixed the merge conflict

l46kok (approved):

PR #1495, opened 2026-09-18
Fix undefined cost helpers in late-bound eval

adrianozp:
Fixes missed cost-tracking helpers from PR #1493

#1493 inlined the `track*` cost-tracking helpers but missed the five call sites in `evalLateBoundFunc.Exec`, leaving the interpreter package uncompilable. Restore the inline nil-guarded `CostTracker` pattern used throughout the file.

Semantics are unchanged.


TristonianJones (approved):

TristonianJones:
Thanks @adrianozp, not sure why the presubmit checks didn't catch this before, but I'm happy for the fix.

PR #1498, opened 2026-09-19
Generic slice list for improved type guarantees and speed

TristonianJones:
Introduce list generics for CEL to improve iteration performance.

Optimizations also applied to `ext/lists.go` with `slice` and `reverse` showing 50% gains in perf.

Subsequent PRs extend the approach to maps and native types.

https://github.com/cel-expr/cel-go/issues/1293


ext/lists.go
```diff
@@ -596,7 +599,11 @@ func slice(list traits.Lister, start, end types.Int) (ref.Val, error) {
 		return nil, fmt.Errorf("cannot slice(%d, %d), list is length %d", start, end, listLength)
 	}
 
-	var newList []ref.Val
+	if sliced, ok := types.MaybeSliceList(types.DefaultTypeAdapter, list, int(start), int(end)); ok {
```
l46kok: Could we add a test case that uses a custom adapter (Proto messages and/or NativeTypes)? I think ordinarily, we want the list to be propagating the same adapter configured from the top-level APIs

(Same ask for reverse)

common/types/list.go
```diff
@@ -90,19 +155,87 @@ func NewJSONList(adapter Adapter, l *structpb.ListValue) traits.Lister {
...
+	return NewDynamicList(adapter, sub), true
+}
+
+// MaybeReverseList attempts to reverse a traits.Lister using its underlying native slice representation,
+// preserving the original element type. If the underlying representation is not a slice, it returns (nil, false).
+func MaybeReverseList(adapter Adapter, list traits.Lister) (traits.Lister, bool) {
+	if list == nil {
+		return nil, false
+	}
+	val := list.Value()
```
l46kok, line 199: Could you add a few test cases around ensuring that we don't eagerly unwrap the values upon list adaptation here? 

```
([optional.of(1), optional.none()]).reverse()[0] == optional.none()

// Something similar for `concatList`
([optional.of(1)] + [optional.none()]).reverse()[0] == optional.none()
```

TristonianJones: The `.Value()` function in this case would only ever return the underlying Go value, and only if it's feasible to do so. The fallback for `MaybeReverseList` is to iterate over the entries and create a new []ref.Val list. Optional wouldn't factor into this call path.

ext/lists.go
TristonianJones: I sort of wondered about this one too. The code at the end uses the default adapter because it assumes that all values have been converted via `Get`; however, that wouldn't necessarily be the case here. I'll add an `adapt` capture and pass it through to the string extensions (along with tests))

TristonianJones: Done.

TristonianJones:
Updated, PTAL.

l46kok (dismissed):

l46kok (approved):

PR #1500, opened 2026-09-21
Type-check error correction suggestions

TristonianJones:
Add suggestions for using different type, function, or variable names based on the environment config.

Also adds entries for `OptionalTypes` since more than a couple of extensions depend on them.

Minor change to default checker error messages, removing end clause `(container is '')` when it's empty.


checker/checker.go
```diff
@@ -322,6 +330,37 @@ func (c *checker) checkCall(e ast.Expr) {
...
+	}
+	return false
+}
+
+// checkCatalogSymbol checks whether an unresolved qualified name exists in the environment catalog
+// (such as a function, type, or macro in an un-enabled library) and records an undeclared reference error
+// with helpful library suggestions if found.
+//
+// Returns true if a matching catalog symbol was found and an error was recorded.
+func (c *checker) checkCatalogSymbol(e ast.Expr, qualifierPrefixes []string, qualifiedName string) bool {
```
jnthntatum: Very minor: it looks like this always looks up the recommended candidates twice here. I can't imagine this would be a hot path, but consider adding a different version of 'undeclaredReference' that accepts the recommendations.

jnthntatum (approved):

checker/checker.go
TristonianJones: You mean add another overload that includes the possible prefixes? I think I can do that and simplify the error checks and skip the internal `Find`. I'd be happy to do that. And if you mean something else, I'm happy to do that too :)

jnthntatum: Right -- something like `errors.undeclaredReferenceWithSuggestedReplacement` (but I assume your naming would be better than mine 😄 ).

common/env/catalog.go
```diff
@@ -0,0 +1,273 @@
...
+}
+
+// EditDistance calculates the Damerau-Levenshtein distance between two strings,
+// taking into account insertions, deletions, substitutions, and adjacent transpositions.
+//
+// When evaluating string similarity for suggestions, the following distance thresholds are recommended:
+//   - length <= 3: max distance 1
+//   - 4 <= length <= 8: max distance 2
+//   - length > 8: max distance length / 3
+func EditDistance(a, b string) int {
```
l46kok: consider capping the input length for suggestions and max distance to look for, just to avoid any unknown pathological cases on the error path

checker/checker.go
```diff
@@ -322,6 +330,37 @@ func (c *checker) checkCall(e ast.Expr) {
...
+	if c.env.Catalog() == nil || c.hasDeclaredPrefix(qualifierPrefixes) {
```
l46kok: nit: could we flip the ordering here such that we don't end up checking the catalog for happy path

checker/env.go
```diff
@@ -75,6 +77,9 @@ type Env struct {
 	aggLitElemType      aggregateLiteralElementType
 	filteredOverloadIDs map[string]struct{}
 	jsonFieldNames      bool
+	catalogSupplier     CatalogSupplier
+	catalogOnce         sync.Once
+	catalog             *env.Catalog
```
l46kok: Should we preserve the catalog when entering/exiting scopes?

https://github.com/cel-expr/cel-go/blob/f51e122dce0a4213027a29eb6a642ce849e2869b/checker/env.go#L324-L332

common/env/catalog.go
```diff
@@ -0,0 +1,273 @@
...
+	}
+	var editMatches []match
+	minDist := math.MaxInt32
+	seen := make(map[string]bool)
+
+	for _, sym := range c.symbols {
+		if seen[sym.Name] {
+			continue
+		}
+		if strings.Contains(sym.Name, ".") != hasDot {
```
l46kok: nit: this feels a bit overbroad fuzzy matching wise. Does it catch cases like:

variable decl: `my_namespace.user_name`
typed: `my_namespace.usr_name.size()` (two dots)

Maybe this one's ok, it looks like it will print out the whole expression `my_namespace.usr_name.size` instead of `my_namespace.usr_name`

